PR #4, opened 2019-06-24
Protecting Exports with Muta

kumavis:
currently breaks with `class` syntax


kumavis:
replacing with https://github.com/MetaMask/sesify/pull/5

PR #5, opened 2019-07-18
Exports defense with Kowtow

kumavis:
https://github.com/kumavis/kowtow


kumavis:
This is not perfect, but kowtow is a decent protective measure for now

PR #11, opened 2019-11-06
ExportsDefense Strategies

kumavis:
This refactor introduces an `exportsDefenseStrategy` interface for securing + caching `moduleExports`.

includes two `exportsDefense` strategies that can be specified in config:
- `magicCopy` (default, mutable copy, but retains closure state)
- `harden` (immutable, cached once per bundle, very low overhead)


kumavis:
i had made some refactors and removed a lot of half-implemented / unused features in preperation for the audit. I think it improved readability but it broke (completely removed) `magicCopy`'s cache, making `magicCopy` fairly meaningless as you would just get a fresh module source eval every time.

Here I fixed that and added a new strategy `harden`. I don't currently have any config autogeneration that would suggest `harden`. I'm not entirely sure how safely I could suggest harden just by looking at the source. maybe if they only export a class. This suggests there may be a need for per-module config as well.

the diff looks big mostly because I moved a large block of code into an if statement

PR #15, opened 2019-11-11
Prelude - remove unnecessary template string

EtDu:


src/preludeTemplate.js
```diff
@@ -152,15 +152,15 @@
             prepareRealmGlobalFromConfig(moduleRealm.global, globalsConfig, endowments, globalStore)
             // execute in module realm with modified realm global
             try {
-              moduleInitializer = moduleRealm.evaluate(`${moduleSource}`)
+              moduleInitializer = moduleRealm.evaluate(moduleSource)
```
dckc, line 155: This idiom can be used to ensure that the result is a string. It's not always a noop.

kumavis:
closed for now

PR #23, opened 2019-12-29
ExportsDefense with cytoplasm Membrane

kumavis:
TODOs:
- [ ] readable-stream patch merged for browserify otherwise this will break all streams usage https://github.com/nodejs/readable-stream/pull/423/files (or workaround)
- [x] remove all "strategies" e.g. harden, as they have been removed, including tests
- [x] need to ensure top level module cache is happening 
- [x] put endowments in membrane space


kumavis:
:rocket:

PR #27, opened 2020-01-01
Module exports tests refactoring

EtDu:


test/moduleExports.js
```diff
@@ -5,341 +5,121 @@ const {
...
+  function defineTwo() {
+    module.exports = {}
+  }
+
+  const one = await runSimpleOneTwo({ defineOne, defineTwo })
+  const result = { one: one, xyz: one.xyz }
 
-  const result = await evalModulesArray(t, { files, pluginOpts: { lavamoatConfig: config } })
-  t.equal(result, 42)
+  t.deepEqual(result, { one: {}, xyz: undefined })
```
kumavis: Creating the obj and then testing it is a little confusing, especially with it being circular. `deepEqual(one, {})` would be better. But it looks like the original test was for result to equal 42? This test seems to originally be about ensuring two modules from the same non-entry package can decorate each other's exports. We shouldn't change that. Instead create another helper like runSimpleOneTwo that tests a pair of modules in the same package

kumavis (changes requested):
see comment below. Dont change the expected test result when refactoring tests.
- [x] create a new test utility like runSimpleOneTwo where the two modules are from the same package

test/moduleExports.js
```diff
@@ -2,344 +2,122 @@ const test = require('tape-promise').default(require('tape'))
...
-
-  const config = {
-    "resources": {
-      "<root>": {
-        "packages": {
-          "test": true,
-        }
-      }
-    }
+  result = {
```
kumavis: missing `const`

kumavis (changes requested):

test/moduleExports.js
kumavis:
```suggestion
  const result = {
```

kumavis (approved):

PR #32, opened 2020-01-03
Default config path 

EtDu:


kumavis (changes requested):
should provide a default config path even if writeAutoConfig is not specified

kumavis (changes requested):
- [x] should not write file without `writeAutoConfig` specified
- [x] document `dryRun` (if you keep it)
- [x] if no config specified AND no file at default path, it should use an empty config
- [x] remove opts.lavamoatConfig ( such functionality should just happen via options.config )
- [x] pluginOpts should not be modified
- [x] in the plugin (index.js) inspect and interpret the pluginOpts, and create an object with the intended configuration (e.g. a function for loading the config be it from file or the object returned in) and a consistent api surface

src/index.js
```diff
@@ -20,70 +21,93 @@ module.exports.args = {
 }
 
 function plugin (browserify, pluginOpts) {
+  //pluginOpts.config is config path
+  const configuration = {
+    writeAutoConfig: pluginOpts.writeAutoConfig,
+    config: function getConfig() {
```
kumavis: lets call it `configuration.getConfig` or `getManifest` or something.
might as well use an arrow fn here.

src/index.js
```diff
@@ -20,70 +21,93 @@ module.exports.args = {
...
+    
     // inject package name into module data
     browserify.pipeline.splice('emit-deps', 0, createPackageDataStream())
-
     // helper to dump autoconfig to a file
-    if (pluginOpts.writeAutoConfig) {
-      const filename = pluginOpts.config
-      if (typeof filename !== 'string') {
+    if (configuration.writeAutoConfig) {
+      if (typeof configuration.autoConfigPath !== 'string') {
```
kumavis: Lets add the feature to this PR where we write the autoconfig to the default path if not specified

src/index.js
```diff
@@ -20,70 +21,93 @@ module.exports.args = {
...
         throw new Error('LavaMoat - "writeAutoConfig" was specified but "config" is not a string')
       }
-      pluginOpts.autoConfig = function writeAutoConfig (config) {
-        fs.writeFileSync(filename, config)
-        console.warn(`LavaMoat Autoconfig - wrote to "${filename}"`)
+
+      if (!fs.existsSync(configuration.autoConfigPath)) {
+        fs.mkdirSync("./lavamoat")
```
kumavis: this should use the `autoConfigPath` and not a hardcoded value. you'll want to use `path` to get the parent dir and you can use package `mkdirp` to create the parent path. to may also want to look at `graceful-fs`

src/index.js
```diff
@@ -20,70 +21,93 @@ module.exports.args = {
...
+      }
+
+      configuration.autoConfig = function writeAutoConfig (autoConfig) {
```
kumavis: this is a function so lets name it as a verb. `writeAutoConfig` or `writeManifest` sounds good

src/index.js
```diff
@@ -20,70 +21,93 @@ module.exports.args = {
...
     }
+    
   }
 }
 
 module.exports.generatePrelude = generatePrelude
 module.exports.createLavamoatPacker = createLavamoatPacker
 
+function getConfigFrom(pluginOpts) {
+  if (!pluginOpts.config && !pluginOpts.dryRun) {
```
kumavis: are you still using `dryRun`?

kumavis (changes requested):

kumavis:
getting there :muscle:

src/index.js
EtDu: writeAutoConfig is already being used by `configuration`. I will rename this to `generateAutoConfig`

src/index.js
EtDu: Not anymore!!

src/index.js
EtDu: As it stands, it writes the autoconfig to the default path if a path is not given.

test/config.js
```diff
@@ -91,3 +95,24 @@ test('config - dunder proto not allowed in globals path', async (t) => {
     t.ok(err.message.includes('"__proto__"'))
   }
 })
+
+test('config - default config path is generated with autoconfig if path is not specified', async (t) => {
+  const tmpObj = tmp.dirSync({ keep: true });
```
kumavis: can we drop the `keep: true` ?

kumavis (commented):
is this parameter still supported?
https://github.com/LavaMoat/lavamoat-browserify/blob/85a27988a748084aff062d9ef643d2846174a324/test/util.js#L85

kumavis:
we should add something that throws an error if an unrecognized parameter is present

kumavis:
```js
const allowedKeys = new Set(['config', ...]
const invalidKeys = Reflect.ownKeys(opts).filter(key => !allowedKeys.has(key)))
if (invalidKeys.length) throw new Error(`found invalid keys ${invalidKeys}`) 
```

kumavis:
```js
> z={};Object.defineProperty(z,'a', { value:1, enumerable: false }); Object.keys(z)
[]
> z={};Object.defineProperty(z,'a', { value:1, enumerable: false }); Reflect.ownKeys(z)
[ 'a' ]
```

src/index.js
```diff
@@ -20,70 +22,95 @@ module.exports.args = {
...
-      pluginOpts.autoConfig = function writeAutoConfig (config) {
-        fs.writeFileSync(filename, config)
-        console.warn(`LavaMoat Autoconfig - wrote to "${filename}"`)
+
+      if (!fs.existsSync(configuration.autoConfigPath)) {
+        const configDirectory = path.dirname(configuration.autoConfigPath)
+        mkdirp.sync(configDirectory)
+      }
+
+      configuration.generateAutoConfig = function writeAutoConfig (autoConfig) {
```
kumavis: do not modify the configuration object during plugin runtime
this should be moved above runPlugin

kumavis (approved):

kumavis (approved):

kumavis:
this is a breaking change

PR #36, opened 2020-01-26
rebundles on config change with watchify + tests

EtDu:
Creates a new bundle on override config change when watchify is enabled. Overwrites the original config file with the corresponding change. index.js tweak --> now returns the default config path for `configPath` instead of null when config is an object (or anything other than a string).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -119,8 +119,8 @@ function getConfigurationFromPluginOpts(pluginOpts) {
           const configOverride = JSON.parse(configOverrideSource)
           const mergedConfig = mergeDeep(primaryConfig, configOverride)
           //Overwrite source config file 
-          const configPath = path.resolve(configuration.configPath)
-          fs.writeFileSync(configPath, JSON.stringify(mergedConfig, null, 3))
+          const configPath = configuration.configPath
+          fs.writeFileSync(configPath, JSON.stringify(mergedConfig, null, 2))
           return mergedConfig
         }
         return primaryConfig
@@ -157,7 +157,7 @@ function getConfigurationFromPluginOpts(pluginOpts) {
             }
           }
         }
-        fs.writeFileSync(overrideConfigPath, JSON.stringify(basicConfig, null, 3))
+        fs.writeFileSync(overrideConfigPath, JSON.stringify(basicConfig, null, 2))
         console.warn(`LavaMoat Override Config - wrote to "${overrideConfigPath}"`)
       }
       console.warn(`LavaMoat Config - wrote to "${configPath}"`)
--- a/src/index.js
+++ b/src/index.js
@@ -31,7 +31,7 @@ function plugin (browserify, pluginOpts) {
   function setupPlugin () {
     
     applySesTransforms(browserify)
-    
+
     const customPack = createLavamoatPacker(configuration)
     // replace the standard browser-pack with our custom packer
     browserify.pipeline.splice('pack', 1, customPack)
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -154,7 +157,7 @@ function getConfigurationFromPluginOpts(pluginOpts) {
             }
           }
         }
-        fs.writeFileSync(overrideConfigPath, JSON.stringify(basicConfig))
+        fs.writeFileSync(overrideConfigPath, JSON.stringify(basicConfig, null, 3))
```
kumavis: use 2 as the whitespace value

src/index.js
```diff
@@ -118,6 +118,9 @@ function getConfigurationFromPluginOpts(pluginOpts) {
           const configOverrideSource = fs.readFileSync(resolvedPath, 'utf-8')
           const configOverride = JSON.parse(configOverrideSource)
           const mergedConfig = mergeDeep(primaryConfig, configOverride)
+          //Overwrite source config file 
+          const configPath = path.resolve(configuration.configPath)
```
kumavis: Is the config path already defined+resolved somewhere? If so let's use that value or elevate it so it can be used here. If that doesn't sound like a good idea, explain why

test/util.js
```diff
@@ -100,6 +101,19 @@ function fnToCodeBlock (fn) {
   return fn.toString().split('\n').slice(1,-1).join('\n')
 }
 
+async function createWatchifyBundle (pluginOpts) {
+  const bundler = browserify([], {
```
kumavis, line 105: 👌

test/config.js
```diff
@@ -252,3 +254,47 @@ test("Config override is applied if not specified and already exists at default
...
+  bundler.emit('file', './lavamoat/lavamoat-config-override.json')
+
+  const configFile = fs.readFileSync(configPath)
+  const configFileString = configFile.toString()
+  
+  const overrideString = JSON.stringify(configDefault)
+  fs.writeFileSync(overridePath, overrideString)
+  
+
+  bundler.on('update', async() => {
```
kumavis: Tests won't wait for this to complete
Use pify or await new promise here for bundle to complete and then run subsequent code

kumavis (changes requested):

src/index.js
EtDu: I got rid of `path.resolve`. configuration.configPath is where it's defined

test/config.js
EtDu: Solved, after a while of head banging. In some cases the 'update' event was firing, others not. (even after re-running tests without changing any code)Turned out to be an underlying filesystem watch issue. Set `poll: true` in watchify plugin opts. Does the trick for testing purposes.

src/index.js
EtDu: done

test/config.js
kumavis: Weird!

test/util.js
```diff
@@ -107,7 +107,7 @@ async function createWatchifyBundle (pluginOpts) {
     packageCache: {},
     plugin: [
       [sesifyPlugin, pluginOpts],
-      [watchify]
+      [watchify, {poll: true}]
```
kumavis, line 111: Add a comment why this is important

test/config.js
```diff
@@ -279,14 +278,13 @@ test('Config edits trigger re-bundle if using watchify', async (t) => {
...
 
   const configFile = fs.readFileSync(configPath)
   const configFileString = configFile.toString()
-
-  const overrideString = JSON.stringify(configDefault)
-  fs.writeFileSync(overridePath, overrideString)
-
-  await new Promise((resolve) => {
+  
+  await new Promise(async(resolve) => {
```
kumavis: - [ ] Have this promise resolve when update is called, so the rest of those code is not indented
- [ ] use `.once` instead of `.on` since we only care about the first event

kumavis (changes requested):

kumavis (approved):

kumavis:
Nice work

PR #37, opened 2020-02-01
autoconfig now generated the same as config override

EtDu:
Quick fix: when generating the autoconfig, make it the same as the override config

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -149,15 +149,7 @@ function getConfigurationFromPluginOpts(pluginOpts) {
       fs.writeFileSync(configPath, configString)
       //Write default override config to file if it doesn't already exist
       if (!fs.existsSync(overrideConfigPath)) {
-        const basicConfig = {
-          "resources": {
-            "<root>": {
-              "packages": {
-              }
-            }
-          }
-        }
-        fs.writeFileSync(overrideConfigPath, JSON.stringify(basicConfig, null, 2))
+        fs.writeFileSync(overrideConfigPath, configString)
         console.warn(`LavaMoat Override Config - wrote to "${overrideConfigPath}"`)
       }
       console.warn(`LavaMoat Config - wrote to "${configPath}"`)


kumavis:
Does this mean the autoconfig contains all of the content of the normal config?
So far I've intended the autoconfig to only contain the explicit overrides from default. This makes it easier to see what's manually been modified to have behavior different than the default. Can you elaborate on your motivations for your change? I may be misunderstanding it

EtDu:
It would contain the same content at first when they are both generated. After testing with the examples repo, it seemed cumbersome to have a barebones override config such as this:
```
"resources": {	
            "<root>": {	
              "packages": {	
              }	
            }	
          }
```
then, have to fill in the desired changes manually. With this new way, effort from the user's standpoint is diminished because everything is already there, they would just need to change simple values like `true` and `false`, or destroy entries completely. 

It's the same thing but the user doesn't need to add `metamask-logo` again since it already exists in the dep graph

EtDu:
But you're right, it makes for a less clear understanding of what changes from the default

kumavis:
This is a significant change to workflow, which is not well mapped out and battle tested. I think having the original config in the override diminishes the purpose of the override. My intuition is that the override should be minimal and we'll commented (an issue for json). But I'm happy to have those challenged against use in practice.

While the override changed values would be visible in version control, i think it helps auditiblity to see just the manual overrides in the override file.

Tell me more about working with the examples

EtDu:
This idea spawned purely from a user experience standpoint. Was testing automatic re-bundling with watchify and felt that it's extra work to fill in information (deps) in `lavamoat-config-override.json` that's already present in  `lavamoat-config.json`

EtDu:
That being said, it would be less of a consideration if we could add comments in the config to better guide the user. Changing to a different comment-supported filetype such as YAML could be a viable solution.

kumavis:
here is part of my logic:
- if the two files are the same(-ish), there should just be one
- if there is just one, re-running autogen will squash user adjustments
- so we should keep both and keep `overrides` minimal

in order to keep the devx good, we should strive to make it work well out of the box and make overrides rarely necessary 

i appreciate the exploration of possible devx improvements :man_scientist:

PR #38, opened 2020-02-18
Set root package's membrane space to `endowments`

EtDu:
This minor change sets the root package's membrane space to the same as endowments, so that root receives unwrapped refs


kumavis (approved):
nice nice

Issue #39, opened 2020-02-21
Respond to LeastAuthority audit


kumavis:
- [x] trivial fixes
  - [X] Issue G: Prevent access to `__proto__` from `deepGet`
https://github.com/LavaMoat/lavamoat-browserify/commit/f903354ca64b7aa46511d6c83d13921253df6ab1
  - [x] Issue I: Code Injection via Label in `wrapWithReturnCjsExports`
https://github.com/LavaMoat/lavamoat-browserify/pull/41/commits/13014b15944bb57799ad0f3c1065fabaa860653c
https://github.com/LavaMoat/lavamoat-browserify/pull/40/commits/dc2497f8040ee43f3e24c256a7e703c0faf36c88
  - [x] Suggestion 3: Detect Additional Primordials
  - [x] Suggestion 4: Hide stack traces
https://github.com/LavaMoat/lavamoat-browserify/pull/44/commits/9176838271f09dbbcea37ecb735206433ff9b86d

- [x] moduleExports defense
  - Issue H: Exported Factory Function can return Shared Object
  - Suggestion 5: Stronger Magic Copy
  - Issue J: Child Dependencies can Access a Parent Module’s Exports before Harden is Applied
When modules have circular dependencies, child modules can access parent while being initialized.

kumavis:
response sent :mailbox_with_mail:

PR #41, opened 2020-02-21
Module content wrap - disallow filenames containing newlines 

kumavis:
Addresses Audit "issue I" #39


EtDu:
Would not work for multiline comments!

PR #47, opened 2020-02-25
Parse config override

### LINKED ISSUE #42: Feature: Strict config parsing
we should strictly parse the config to catch things like typos or unsupported options

EtDu:
Ensure that the override config was typed correctly.
Fixes #42


kumavis (changes requested):
- break up the individual test into many test (can be in its own file iff you want)
- create a test helper fn so you can just call
- move example configs out of util and into where the test is actually happening (fine to just define them inline)
```js
testConfigParser({ ... }, 'should catch typo ...')
```

src/index.js
```diff
@@ -226,4 +228,34 @@ function applySesTransforms(browserify) {
 
   browserify.transform(removeHtmlComment, { global: true })
   browserify.transform(changeImportString, { global: true })
+}
+
+function parseOverrideConfig(configOverride) {
```
kumavis: rename to `validateConfig` not "parse"

src/index.js
```diff
@@ -226,4 +228,34 @@ function applySesTransforms(browserify) {
...
+  if (typeof configOverride !== 'object') {
+    throw new Error("LavaMoat - Expected config override to be an object")
+  }
+  
+  if (Object.keys(configOverride)[0] !== 'resources') {
```
kumavis: the order is not important here, but we should check that 'resources' exists

src/index.js
```diff
@@ -226,4 +228,34 @@ function applySesTransforms(browserify) {
...
+    throw new Error("LavaMoat - Expected label 'resources' for configuration key")
+  }
+
+  Object.entries(configOverride['resources']).forEach((resource, index) => {
```
kumavis: you can destructure here
`forEach(([packageName, packageOptions], index) => {`

kumavis:
nice, good stuff :+1: 
just some factoring and cleanup and its good to go

src/index.js
```diff
@@ -226,4 +228,34 @@ function applySesTransforms(browserify) {
...
   browserify.transform(removeHtmlComment, { global: true })
   browserify.transform(changeImportString, { global: true })
+}
+
+function validateConfig(configOverride) {
+  if (typeof configOverride !== 'object') {
+    throw new Error("LavaMoat - Expected config override to be an object")
+  }
+  
+  if (!'resources' in Object.keys(configOverride)) {
```
kumavis: `'resources' in configOverride` or `Object.keys(configOverride).includes('resources')`

src/index.js
```diff
@@ -226,4 +228,34 @@ function applySesTransforms(browserify) {
...
+    throw new Error("LavaMoat - Expected label 'resources' for configuration key")
+  }
+
+  Object.entries(configOverride['resources']).forEach(([packageName, packageOpts], index) => {
+    if (index === 0 && packageName !== '<root>') {
```
kumavis: i dont think we need to validate the presence of `<root>`

src/index.js
```diff
@@ -226,4 +228,34 @@ function applySesTransforms(browserify) {
...
+      throw new Error("LavaMoat - Expected package '<root>' as first resource")
+    }
+    const packageOptions = Object.keys(packageOpts)
+    const packageEntries = Object.values(packageOpts)
+    if (!packageOptions.every(packageOpt => packageOpt === 'globals' || packageOpt === 'packages')) {
+      throw new Error("LavaMoat - Unrecognized package options. Expected 'globals' or 'packages'")
+    }
```
kumavis, line 251:
```js
optionsWhitelist = [...]
packageOptions.every(packageOpt => optionsWhitelist.includes(packageOpt))
```

src/index.js
```diff
@@ -226,4 +228,34 @@ function applySesTransforms(browserify) {
...
+
+    packageEntries.forEach((entry) => {
+      Object.values(entry).forEach((value) => {
+        if (value !== true) {
```
kumavis: `'write'` is a valid value for globals. it allows packages to write to `global[key]`

test/config.js
```diff
@@ -297,4 +297,248 @@ test('Config edits trigger re-bundle if using watchify', async (t) => {
 
   t.notOk(configFileString.includes('"three": 12345678'))
   t.ok(updatedConfigFileString.includes('"three": 12345678'))
-})
\ No newline at end of file
+})
+
+test('Bundle fails with first configOverride', async (t) => {
```
kumavis: improve test titles
- call out what the mistake is in the test title, eg `'invalid "resources" key'`
- "config validation fails"

test/config.js
```diff
@@ -297,4 +297,248 @@ test('Config edits trigger re-bundle if using watchify', async (t) => {
...
+  testConfigValidator(configOverride, config, true, t)
+})
+
+
+
+async function testConfigValidator(configOverride, config, shouldBeValid, t) {
+  try {
+    await createBundleFromRequiresArray([], {
+      config,
+      configOverride: configOverride
```
kumavis:
```suggestion
      configOverride,
```

kumavis (changes requested):
Nice, looking clean. Couple more nits

EtDu (commented):
Solid

kumavis (approved):

kumavis:
i suspect there will be a lint error

PR #48, opened 2020-02-27
Ci lint

### LINKED ISSUE #31: lint is not configured correctly
it currently ignores all files

EtDu:
Configure standard lint properly and fix errors 
Fixes #31


kumavis:
@EtDu why did you make this change? `fix utf-8 in filepath`

EtDu:
> @EtDu why did you make this change? `fix utf-8 in filepath`

Lint yelled for not using `path.join` instead of `+` so I changed it, then it gave me an error because `utf-8` was showing up in the file path, then I changed it back and lint ignored it

kumavis:
removing the `utf8` argument results in the fs call returning a buffer instead of a string, so that would have caused the breakage. lint is correct about `path.join` (for platform agnosticism), i think the issue is that you accidentally added `'utf8'` to the path argument, not the last encoding argument

kumavis (approved):

kumavis:
noice

PR #49, opened 2020-03-04
gernerateConfig - breakout createModuleInspector

kumavis:
now you can use `createModuleInspector` to generate config

```js
const inspector = createModuleInspector()
// for each module
inspector.inspectModule(moduleData)
// when done
const config = inspector.generateConfig()
```


kumavis:
`inspector.inspectModule(moduleData)` expects the following shape for `moduleData`
```js
{ package, id, source}
```

PR #51, opened 2020-03-04
Feature: Static Prelude, "loadBundle" api, bundling without prelude

kumavis:
- Config has been moved out of the prelude, it is now static (takes no options, not app specific)
- Bundles are now loaded into the prelude via `LavaMoat.loadBundle(...)`
- Prelude can be optionally not included in bundle via `{ includePrelude: false }` pluginOpts
- (untested) Using the 3 features above, apps can be run by loading separate bundle files


EtDu:
YEEEEEEEEE

PR #54, opened 2020-03-05
Feature: bundle factoring

kumavis:
- Adds test ensuring support of bundle factoring via `bify-package-factor`
- Removes requirement for package to be in `<root>`s whitelist. this is because each bundle can contain `<root>` modules, and we only allow setting the config for each package once.

TODO:
- config factoring


kumavis:
etienne sez: lit :fire:

PR #63, opened 2020-03-23
Add deps-dump feature to Lavamoat

EtDu:
- depsDump plugin option
- pass config option to createConfigSpy
- format module data with ID as the key
- if depsDump enabled, write module data to disk
- test added


kumavis (changes requested):
rename the option to `writeAutoConfigDebug` and allow it to be true with a default path, or a string that is used as a path

src/index.js
```diff
@@ -186,6 +190,12 @@ function getConfigurationFromPluginOpts (pluginOpts) {
     throw new Error('LavaMoat - Unrecognized value for writeAutoConfig')
   }
 
+  if (typeof pluginOpts.writeAutoConfigDebug === 'string') {
+    configuration.writeAutoConfigDebug = pluginOpts.writeAutoConfigDebug
+  } else if (pluginOpts.writeAutoConfigDebug !== false) {
```
kumavis:
```suggestion
  } else if (pluginOpts.writeAutoConfigDebug) {
```

src/generateConfig.js
```diff
@@ -130,16 +133,27 @@ function aggregateDeps ({ packageModules, moduleIdToPackageName }) {
...
     cb(null, data)
   }, (cb) => {
     // call flush observer
     onEnd()
-    // end normally
+    // if writeAutoConfigDebug option enabled, dump allDeps to a file
+    if (writeAutoConfigDebug) {
+      const serialized = jsonStringify(allDeps, { space: 2 })
+      console.warn(`deps-dump - writing to ${writeAutoConfigDebug}`)
+      fs.writeFileSync(writeAutoConfigDebug, serialized)
```
kumavis: this part
```js
    // if writeAutoConfigDebug option enabled, dump allDeps to a file
    if (writeAutoConfigDebug) {
      const serialized = jsonStringify(allDeps, { space: 2 })
      console.warn(`deps-dump - writing to ${writeAutoConfigDebug}`)
      fs.writeFileSync(writeAutoConfigDebug, serialized)
    }
```

should appear in index.js or its own file. there should be a through or a spy added at the `'debug'`  pipeline phase

kumavis (changes requested):

src/generateConfig.js
EtDu: doooone

src/index.js
```diff
@@ -1,8 +1,11 @@
 const fs = require('fs')
 const path = require('path')
+const clone = require('clone')
 const mkdirp = require('mkdirp')
+const through = require('through2')
```
kumavis:
```suggestion
const through = require('through2').obj
```

kumavis (changes requested):

PR #90, opened 2020-08-20
Session "Proboscus"

kumavis:
random bugs found when running `eth-faucet` in the new version


kumavis:
dude looks good
thanks

PR #91, opened 2020-08-24
Session: Mic Check

kumavis:
last set of changes before initial release


kumavis:
config major change for core and node (doesnt affect browserify as native modules are not supported there)

kumavis:
compat warnings looking scrumptious. maybe a little too alarming but a huge improvement
image
mad love for `@babel/code-frame` here :heart:

PR #92, opened 2020-09-10
fix default config paths

EtDu:


kumavis (changes requested):
in test/project1 plz delete the old lavamoat dir

kumavis (approved):

PR #101, opened 2020-10-15
improvements - node resolution, compatibility logging

kumavis:
fixes from running lavamoat against the mm build system

### DIFF
--- a/packages/node/src/parseForConfig.js
+++ b/packages/node/src/parseForConfig.js
@@ -21,7 +21,22 @@ const createRequire = (url) => {
   return {
     resolve: (requestedName) => {
       const basedir = path.dirname(url.pathname)
-      return resolve.sync(requestedName, { basedir })
+      const result = resolve.sync(requestedName, { basedir })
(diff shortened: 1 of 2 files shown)


packages/node/src/parseForConfig.js
```diff
@@ -86,12 +101,12 @@ function makeImportHook ({
...
     const packageData = packageDataForModule({ id: specifier, file: filename }, rootPackageName)
 
-    if (extension === '.node') {
-      return makeNativeModuleRecord(specifier, filename, packageData)
-    }
     if (['.js', '.cjs'].includes(extension)) {
       return makeJsModuleRecord(specifier, filename, packageData)
     }
+    if (extension === '.node') {
+      return makeNativeModuleRecord(specifier, filename, packageData)
```
EtDu, line 108: What's the purpose of shifting this below?

kumavis: its less common, so we'll hit the common case first. likely a tiny but unmeasured perf boost

PR #119, opened 2020-11-06
perf - updated results

kumavis:


kumavis:
freeze `compartment.globalThis` you coward!

PR #127, opened 2020-12-16
Create mergeConfig in Lavamoat-Core

EtDu:
- Creates new `mergeConfig` that uses `merge-deep` dep to recursively merge full config objects
- Add test for new `mergeConfig`


packages/tofu/src/util.js
```diff
@@ -122,6 +124,14 @@ function mergeConfig (configA, configB) {
   return newConfig
 }
 
+function mergeConfig (configA, configB) {
+  configA = JSON.parse(configA)
```
kumavis: why are we calling parse here? that wasn't part of the original fn. plz remove

packages/tofu/test/util.js
```diff
@@ -29,10 +42,48 @@ testMerge('non-overlapping', {
   'xyz.jkl': 'write'
 })
 
-function testMerge (label, configA, configB, expectedResultObj) {
+testMerge('merge with resources', {
```
kumavis: 👍

kumavis (approved):

kumavis (changes requested):
meant to mark as request changes

packages/tofu/src/util.js
EtDu: because we're passing in the string from readfile. Unless we want to use parse on the consumer side

kumavis: parse on consumer side

yarn.lock
```diff
@@ -9999,6 +9999,19 @@ yargs@^15.0.2, yargs@^15.3.1:
     y18n "^4.0.0"
     yargs-parser "^18.1.1"
 
+yargs@^16.0.0:
```
kumavis, line 10002: weird

kumavis (approved):
approve

PR #142, opened 2021-01-04
lavamoat-browserify compatibility for running under lavamoat-node

EtDu:
- Creates a test to simulate a lavamoat-browserify bundle running in lavamoat-node, with a new fixture for running the bundle with the JS example entry file
- Remakes `packageData.js` in lavamoat core to use createRequire instead of standard require, necessary so that packages required by lavamoat-browserify skip needing to have endowments in lavamoat-node's config.
- Creates new lavamoat config files for lavamoat-node


packages/core/src/packageData.js
```diff
@@ -24,12 +25,18 @@ function packageDataForModule (moduleData, rootPackageName) {
...
-  const packagePath = path.join(packageParentPath, packageName, 'package.json')
-  if (!packagePath) return
   // attempt to load package path
   let packageVersion
   try {
-    const packageJson = require(packagePath)
+    const NODE_MAJOR_VERSION = process.version.split('.')[0].split('v')[1]; 
+    console.log(typeof NODE_MAJOR_VERSION)
+    let requireFn
+    if (NODE_MAJOR_VERSION < 12) {
```
kumavis: something like
```js
if (createRequire) {
// use it
} else if (createRequireFromPath) {
// use it
} else {
throw unsupported error
}
```

packages/browserify/test/lavamoatNode.js
```diff
@@ -0,0 +1,8 @@
+const test = require('ava')
+const { execSync } = require('child_process');
+const path = require('path')
+
+test('basic - bundle works under lavamoat node', async (t) => {
+  const bundle = execSync("lavamoat ../../../test/fixtures/runBrowserifyWithEntry.js --debugMode --applyExportsDefense false", { cwd: path.resolve(__dirname, '../examples/01-simple-js/lavamoat-build'), maxBuffer: 8192 * 10000 })
```
kumavis: break out options into multiple lines

kumavis (changes requested):

kumavis (changes requested):
- [x] manual override makes me sad
- [x] move `01-simple-js/lavamoat-build` to like `02-secure-runtime` or somn

packages/core/src/packageData.js
```diff
@@ -2,6 +2,9 @@
 const path = require('path')
 const pathSeperator = require('path').sep
 const { isCore } = require('resolve')
+// Importing depracated API for use with old node versions
```
kumavis:
```suggestion
// importing deprecated API for use with old node versions
```

packages/core/test/builtin.js
```diff
@@ -28,8 +28,8 @@ test('builtin - basic access', async (t) => {
       }
     }
   })
-
   const result = await runScenario(scenario)
+  console.log(result)
```
kumavis: remove

kumavis (changes requested):

PR #143, opened 2021-01-11
Test refactor

EtDu:


kumavis:
includes a breaking change in config generation (skips `Object.prototype` props in globals)

packages/core/test/util.js
```diff
@@ -195,11 +196,13 @@ async function runScenario ({ scenario }) {
   return testResult
 }
 
-async function prepareScenarioOnDisk ({ scenario }) {
+async function prepareScenarioOnDisk ({ scenario, opts = {} }) {
   const { path: projectDir } = await tmp.dir()
   const filesToWrite = Object.values(scenario.files)
-  filesToWrite.push({ file: 'lavamoat-config.json', content: stringify(scenario.config) })
-  filesToWrite.push({ file: 'lavamoat-config-override.json', content: stringify(scenario.configOverride) })
+  if (!opts.writeAutoConfig) {
```
kumavis: good

kumavis:
> Error: LavaMoat - required package not in whitelist: package "one" requested "two" as "two"

tests are currently failing -- we should catch the error in runScenario and re-throw with the scenario name, so we can easily see whats breaking

packages/core/test/util.js
kumavis: actually remove this and always write the config

PR #146, opened 2021-01-25
browserify + node README update

EtDu:
Pending:
- Change lavamoat-browserify options naming, add CLI shorthands
- Deploy viz demo and add the link inside README
- Fix browserify to create bundle with `autoconfig` enabled, instead of having to run the command twice


kumavis:
i fixed main (oops) so pull_request tests should pass when rerun

PR #150, opened 2021-01-26
core - kernel rework

kumavis:
- [x] refactor for readability
- [x] remove membrane defense (too slow)
- [x] one compartment per package
- [x] root package compartment `globalThis` used as source for endowments
- [ ] remove global store (?)


kumavis:
this includes a breaking change for core and all platforms

PR #173, opened 2021-02-09
CI artifacts fix

EtDu:
- Remove `injectConfigDebugData.js`
- use `script` tags instead of `import` for policies


kumavis (changes requested):
https://github.com/LavaMoat/LavaMoat/pull/173/files#diff-327cc2641b45ae2495187ff4d7781223873dd4d6d3d7445af647037d00d28b1eR109-R119

comment out this section since we're not using it right now

kumavis (approved):

PR #182, opened 2021-02-17
allow-scripts/deps - use yarn-logical-tree

kumavis:
need tests to ensure this actually works!


kumavis:
closing in favor of using arborist https://github.com/LavaMoat/LavaMoat/pull/184

PR #184, opened 2021-02-20
Lava tree

kumavis:
secure comprehension of dependency graph + parsing dependency paths + qualified name based on resolved location

is it weird to call it `@lavamoat/lava-tree`?
should it be `@lavamoat/lavatree`
naming things!


kumavis:
still blocked by mysterious lack of `resolved` field in certain cases https://github.com/npm/arborist/issues/239

kumavis:
abandoning this effort - too package manager dependent

tmpfs:
I already have some work done in `basalt` for rendering the dependency graph as a tree, currently it is `basalt ls entry.js` but I think `tree` is a better command name. If there are features here I haven't done yet that you want let me know and I will fold them into `basalt tree`. Currently the output is like `tree(1)`, for example:

```
tests/fixtures/basic-tree/main.js
├── ./foo.js
│   └── ./bar.js
│       └── ./qux.js
└── ./bar.js
    └── ./qux.js
```

PR #206, opened 2021-03-24
Fix Function constructor this value when using as constructor

EtDu:
`this` value of Function constructors should remain unchanged when using as a constructor, we were previously unwrapping them to the package compartment's global object


kumavis (approved):
very constructive functionality. now we can use the function constructor to construct constructor functions. hope this was a funstructive bug for you

PR #209, opened 2021-03-30
firefox code test

EtDu:


packages/core/src/makeGetEndowmentsForConfig.js
```diff
@@ -171,7 +171,13 @@ function makeGetEndowmentsForConfig ({ createFunctionWrapper }) {
...
         // replace the "receiver" value if it points to fake parent
         const receiverRef = receiver === sourceGlobal ? targetGlobal : receiver
-        const result = Reflect.apply(sourcePropDesc.get, receiverRef, [])
+        const desc = Reflect.getOwnPropertyDescriptor(targetGlobal, 'chrome')
+        let result
+        if ('get' in desc) {
+          result = Reflect.apply(sourcePropDesc.get, receiverRef, [])
+        } else {
+          result = Reflect.getOwnPropertyDescriptor(targetGlobal, 'chrome').value
+        }
```
kumavis: `targetGlobal` should prolly be `sourceGlobal`.
should not be a hard-coded `'chrome'`
if we're getting the property descriptor again, we should use that new property descriptor

I suppose we could actually just do 
```js
Reflect.get(sourceGlobal, key, receiverRef)
```
and skip doing the prop desc lookup + if statement. one liner

kumavis: should add a comment as to why we're doing this lookup again
"sometimes getters replace themselves with static properties"

packages/core/test/globals.js
```diff
@@ -206,3 +206,65 @@ test('globals - endowing properties on the globalThis prototype chain', async (t
   const testResult = await runScenario({ scenario })
   t.is(testResult, 123, 'expected result, did not error')
 })
+
+test.only('globals - firefox code works in the wild', async (t) => {
```
kumavis: "firefox addon chrome api lazy getter works"

kumavis (changes requested):

packages/node/test/index.js
```diff
@@ -96,7 +96,6 @@ test('execute - resolutions', async (t) => {
     cwd: projectRoot,
     args: [entryId]
   })
-  t.is(output.stderr, '', 'should not have any error output')
```
EtDu, line 99: This is because Node 12 spits this in standard err:
`(node:47504) [DEP0016] DeprecationWarning: 'GLOBAL' is deprecated, use 'global'`

I could not find any instances where we are using `GLOBAL`

packages/core/test/globals.js
```diff
@@ -91,11 +94,6 @@ test('globals - ensure circular refs on package compartment global', async (t) =
       const testResult = xyz === globalThis
       console.log(JSON.stringify(testResult, null, 2))
     },
-    context: {
```
EtDu, line 94: This is because we invoke every getter, so naturally it triggers the error. I think we can skip this check anyway. `xyz` would not be a `globalThis` reference if the getter could be invoked

packages/core/src/kernelCoreTemplate.js
```diff
@@ -248,6 +248,18 @@
       if (commonPrototypeIndex === -1) throw new Error('Lavamoat - unable to find common prototype between Compartment and globalRef')
       // we will copy endowments from all entries in the prototype chain, excluding Object.prototype
       const endowmentSources = globalProtoChain.slice(0, commonPrototypeIndex)
+
+      endowmentSources.forEach(source => {
+        const descriptors = Object.getOwnPropertyDescriptors(source)
+        Object.values(descriptors).forEach(desc => {
+          if ('get' in desc) {
+            // call all getters, in case of behavior change (such as with FireFox lazy getters)
```
kumavis: move comment to above the `endowmentSources.forEach(source => {`

kumavis (changes requested):

PR #216, opened 2021-05-04
Add project root to lavamoat browserify options

EtDu:
- Adds lavamoat option projectRoot for lavamoat-browserify
- Fixes terminology (config --> policy)


kumavis (approved):
breaking change for core, browserify

PR #243, opened 2021-08-10
Removing unused vars in core/ & lavapack/

ryanml:


kumavis:
`lavapack/src/runtime.js` is a built file

kumavis:
these should be captured in an eslint ignore

kumavis:
thanks for taking a look/

kumavis:
these are "vendorized" external code that should not be modified `packages/core/lib/*`

PR #255, opened 2021-09-15
Kernel - add runWithPrecompiledModules option for environments without eval

kumavis:


packages/lavapack/src/runtime-template.js
```diff
@@ -68,7 +69,14 @@
         throw new Error(`LavaMoat - loadBundle encountered redundant module definition for id "${moduleId}"`)
       }
       // add the module
-      moduleRegistry.set(moduleId, { type: 'js', id: moduleId, deps: moduleDeps, source: `(${initFn})`, package: packageName })
+      moduleRegistry.set(moduleId, {
+        type: 'js',
+        id: moduleId,
+        deps: moduleDeps,
+        // source: `(${initFn})`,
```
kumavis, line 76: rm comment

PR #256, opened 2021-09-20
Support type in module record options.

tmpfs:
Prepare to support ESM (via static module record) as well as CJS.

The value for `type` should be either `cjs` or `esm`.


kumavis (approved):

kumavis:
currently using the `js` type in other parts of lavamoat, like handling the `builtins` in nodejs.

tmpfs:
> currently using the `js` type in other parts of lavamoat, like handling the `builtins` in nodejs.

Thanks, I wasn't sure whether it was being used or just a placeholder. I have updated to use `js` here by default and in the bundler code.

packages/lavapack/src/runtime.js
```diff
@@ -6423,13 +6423,13 @@ function makePrepareRealmGlobalFromConfig ({ createFunctionWrapper }) {
...
     // verify + load in each module
-    for (const [moduleId, moduleDeps, initFn, { package: packageName }] of newModules) {
+    for (const [moduleId, moduleDeps, initFn, { package: packageName, type }] of newModules) {
       // verify that module is new
       if (moduleRegistry.has(moduleId)) {
         throw new Error(`LavaMoat - loadBundle encountered redundant module definition for id "${moduleId}"`)
       }
       // add the module
-      moduleRegistry.set(moduleId, { type: 'js', id: moduleId, deps: moduleDeps, source: `(${initFn})`, package: packageName })
+      moduleRegistry.set(moduleId, { type: type || 'cjs', id: moduleId, deps: moduleDeps, source: `(${initFn})`, package: packageName })
```
kumavis: still cjs here

tmpfs: Oops, forgot to rebuild the runtime, done now :+1:

kumavis (approved):

PR #278, opened 2021-12-07
Secure Package Naming with 'A'a

kumavis:
- [x] create `@lavamoat/aa`
- [x] implement in browserify
- [x] implement in node
- [x] use `browser-resolve` in browserify
- [x] apply policy-override before policy generation only
- [ ] rename `<root>` to `$root$`
- [ ] fix skipped bundle factoring test
- [ ] fix skipped browserify process test
- [x] remove canonicalNameMap debug logging
- [ ] test against mm
- [ ] no builtins in browserify

PR #282, opened 2022-01-27
Policy - apply policy override only when generating policy

kumavis:


kumavis:
merged into https://github.com/LavaMoat/LavaMoat/pull/278

PR #300, opened 2022-02-09
Next Major Release Blockers unblocked!

kumavis:
- [x] rename <root> to $root$
- [x] fix test flakiness (some race condition around the browserify cache?)
- [x] fix skipped browserify process test
- [x] no builtins in browserify (policy?)
- [x] fix yarn concurrency issues in test
- [ ] test against mm
- [x] fix skipped bundle factoring test


kumavis:
oof - that was a pain. mischief managed.

PR #309, opened 2022-03-07
aa - memoize expensive operations in eachPackageInLogicalTree

naugtur:
Some tests were failing for me locally, but they failed on main as well.

I think before we merge this we may need to:
- [ ] introduce cache invalidation or instantiate cache on each inbound call to aa package
- [ ] figure out why there's a weird failure in node12

### DIFF
--- a/packages/aa/src/index.js
+++ b/packages/aa/src/index.js
@@ -1,4 +1,4 @@
-const { promises: fs } = require('fs')
+const { readFileSync } = require('fs')
 const path = require('path')
 const nodeResolve = require('resolve')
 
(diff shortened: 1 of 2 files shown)


kumavis:
interesting that promise overhead outweighs disk concurrency gains

packages/aa/src/index.js
```diff
@@ -52,25 +52,53 @@ async function loadCanonicalNameMap ({ rootDir, includeDevDeps, resolve } = {})
...
+    return depsToWalk
+  }
+}
+
 /**
  * @param {object} options
  * @returns {AsyncIterableIterator<{packageDir: string, logicalPathParts: string[]}>}
  */
 // TODO: optimize this to not walk the entire tree, can skip if the best known logical path is already shorter
 async function * eachPackageInLogicalTree ({ packageDir, logicalPath = [], includeDevDeps = false, visited = new Set(), resolve = nodeResolve }) {
```
naugtur, line 97: I ended up dropping the changes to prevent DEOPT because they were ugly and didn't visibly improve the time.
This function gets deoptimized to polymorphic because logicalPath changes from array of unknown to array of strings. 
`logicalPath = ['']` and the logic to drop the empty string on the first pass would make the function monomorphic. 

No visible speed improvement from that though.

kumavis:
can make the whole thing sync now, doesn't need to be an async iterator

packages/aa/src/index.js
```diff
@@ -52,25 +52,53 @@ async function loadCanonicalNameMap ({ rootDir, includeDevDeps, resolve } = {})
...
+    return depsToWalkCache.get(key)
+  } else {
+    const packageJsonPath = path.join(packageDir, 'package.json')
+    // If this function used async, it'd have to be awaited, which would mean cache lookup 
+    // would need to happen outside the function to save on performance of spawning a promise 
+    // for each cache lookup.
+    const rawPackageJson = readFileSync(packageJsonPath, 'utf8')
+    const packageJson = JSON.parse(rawPackageJson)
+    const depsToWalk = [
+      ...Object.keys(packageJson.dependencies || {}),
```
naugtur, line 84: `.dependencies` makes the function that reads it megamorphic, so there's theoretically a performance benefit from it, but I didn't see much. 
The biggest performance issues were:
- resolve.sync making statSync checks (which throws and catches internally for each location where the file is not, so it's awfully slow)
- reading and parsing json many times around.

naugtur:
> interesting that promise overhead outweighs disk concurrency gains

It may not, but the vast majority of lookups are repetitions and we don't want a new promise to check if an item is in a map. 

I could externalize the cache lookup and run fs async, but I had a version like that and the difference was not noticeable.

kumavis:
also surprised the packagejson parsing memorization improved performance more than the resolution memoization improved performance (per comment elsewhere)

kumavis (approved):
i love it, tysm

PR #332, opened 2022-04-21
Naugtur perf experiments

naugtur:
in my first commit, I'm getting rid of the recursion (and bringing back some asynchronicity so that measurements can be collected at all) - slight overall improvement in performance ~20%

in my second commit, I make it synchronous again and get rid of some repetition by not revisiting packages that have already been visited. - 90% improvement in performance, 50% in RAM. Not sure about correctness. there's a tiny decrease in the number of package.jsons visited along the way, but the result looks similar enough. We're not getting the shortest paths that way though. Maybe if the order of visiting nodes was better...
```
<   '/home/naugtur/work/metamask/metamask-extension/node_modules/wrappy' => 'pump>once>wrappy',
---
>   '/home/naugtur/work/metamask/metamask-extension/node_modules/wrappy' => 'read-installed>readdir-scoped-modules>dezalgo>wrappy',
```

The second change could be applied to the original code too (returning empty instead of cached) to see what happens.

in my third commit I'm adding clearing caches, but it doesn't seem to help much - stuff gets garbage-collected only when the function ends running. Which might be useful anyway.

A combination of some of these changes may be beneficial.


kumavis:
>  I [...] get rid of some repetition by not revisiting packages that have already been visited.  [...]  Not sure about correctness. 

yeah this is wrong as is but i think we can, 
1. improve visiting order (as you suggest)
2. compare best dep path length and only continue if the current one is shorter

naugtur:
> * improve visiting order (as you suggest)
> * compare best dep path length and only continue if the current one is shorter

Breadth-first should suffice - a path will never be revisited if it's not shortest. My only concern with that is possible dead ends.

Also, I believe we could save a lot of time and memory by reading from lockfile instead. Worth prototyping?

PR #338, opened 2022-05-25
tofu - fix optional chaining

kumavis:


packages/tofu/src/util.js
```diff
@@ -45,8 +47,12 @@ function getNameFromNode (node) {
   }
 }
 
-function isDirectMemberExpression (node) {
-  return node.type === 'MemberExpression' && !node.computed
+function isMemberLikeExpression (node) {
+  return node.type === 'MemberExpression' || node.type === 'OptionalMemberExpression'
```
naugtur, line 51: I was hoping @babel/types' own isMemberExpression or a similar function would answer true for both, but it's not there :(

PR #347, opened 2022-06-08
Fix alphabetical ordering logic

weizman:
`a>b` is checked twice and `a<b` is never checked, seems like a mistake


kumavis:
nice find!

kumavis (approved):

PR #355, opened 2022-07-11
Handle owner document access

weizman:


packages/core/src/makeGeneralUtils.js
```diff
@@ -1,5 +1,20 @@
...
+  if (!isInstanceOfNode) {
+    return
+  }
+  // apply patch only once
+  const isAlreadyPatched = node.hasOwnProperty('ownerDocument')
+  if (isAlreadyPatched) {
+    return
+  }
+  // when ownerDocument prop is being accessed return proxy document instead of real document
+  Object.defineProperty(node, 'ownerDocument', { get: () => doc })
```
naugtur, line 15: without `writable: false` can I simply 
```js
delete node.ownerDocument
```
and go bask to accessing it via the prototype?

packages/core/src/makeGeneralUtils.js
```diff
@@ -14,10 +29,12 @@ function makeGeneralUtils () {
         // handle function calls
         // unwrap to target value if this value is the source package compartment's globalThis
         const thisRef = unwrapTest(this) ? unwrapTo : this
-        return Reflect.apply(sourceValue, thisRef, args)
+        const val = Reflect.apply(sourceValue, thisRef, args)
+        patchDocumentAccess(this, val);
```
naugtur, line 33: What about the cases where a DOM node is not returned from a function? 
As in:
```js
const thisIsSafe = getMeADomNode()
const thisIsNot = thisIsSafe.firstChild.ownerDocument
```

PR #360, opened 2022-08-08
Start compartment globalThis scuttling (browserify only)

weizman:
This PR introduces @kumavis's idea for scuttling the `globalThis` object of the start compartment in LavaMoat.

## Motivation

Since each package LavaMoat wraps has its own compartment and own set of intrinsics (including the app that is protected which is referred to as `root` package), this effectively means the real `globalThis` object is out of use and serves almost exclusively attackers only.

In other words, the main app has access to everything the globalThis object has to offer, but it goes through LavaMoat to do so which provides everything from the compartment - not the actual globalThis object.

However, if a package manages to access the real globalThis object which no code in the app really uses, it has access to sensitive APIs it shouldn't have.

Based on how this serves no purpose other then a malicious one, it would make sense to offer as a configurable feature in LavaMoat the option to disarm the globalThis object completely, by redefining all of its properties to not return their real values (aka scuttling).

## Implementation 

So when scuttling is turned on, any code that runs within the app and manages to access the globalThis object, will encounter a thrown error when trying to access anything it offers:

Screen Shot 2022-09-18 at 15 01 20

## Issues

If you encounter an unexpected issue when running an app under scuttling mode, please let us know, that should not be the case. 

**If the app is trying to access a property that it should have access to, you can easily fix that** by adding it to the `lavamoatOpts.scuttleGlobalThisExceptions` config property (see `usage` section below)

> Many have reported issues with WebDriver manipulation of MM when this feature is on, please refer to my [comment below](https://github.com/LavaMoat/LavaMoat/pull/360#issuecomment-1547726986) for further info on this

## Support

~~Currently this feature is only usable for browser based apps built with `browserify`~~
With #391 `node`/`core`/`browserify` support scuttling

## Usage

By adding the `scuttleGlobalThis=true` config property, scuttling can be turned on (see [here](https://github.com/LavaMoat/LavaMoat/pull/360/files#diff-ef5b28753cd110237bb71c073c19be3c94803ad7f42501cd4ae063a96dc0454dR7))

Optionally, you can also set `scuttleGlobalThisException=[]` property to skip certain props from being scuttled.

```javascript
const lavamoatOpts = {

  // turn scuttling off
  scuttleGlobalThis: false, // true,

  // set scuttle exceptions
  scuttleGlobalThisException: [], // ['fetch', 'alert'],
}
```

## WIP

Some later related PRs:

* #382 
  * some fixes requested by @kumavis after #360 was merged
* #391
  * some important fixes and updates, partially related to bugs introduces by #382
* #405
  * support reg exps as excepted scuttling props rather than just hardcoded strings
* #462 
  * Support configurable hooking into LavaMoat scuttling


packages/core/src/kernelCoreTemplate.js
```diff
@@ -72,6 +72,9 @@
     const rootPackageName = '$root$'
     const rootPackageCompartment = createRootPackageCompartment(globalRef)
 
+    // scuttle globalThis right after we used it to create the root package compartment
+    scuttleGlobalThis()
```
naugtur: Feels like we'd want this to be configurable in the final version

packages/core/src/kernelCoreTemplate.js
```diff
@@ -82,6 +85,31 @@
...
+        'location', 'top', 'global', 'self', 'window', 'Infinity', 'NaN', 'document',
+        // support LM,SES exported APIs
+        'LavaPack', 'Compartment',
+        // support polyfill
+        'globalThis',
+      ]
+
+      for (const prop of props) {
+        if (!avoids.includes(prop)) {
+          Object.defineProperty(window, prop, {value: undefined})
```
naugtur: How about a getter that throws with helpful information?

packages/core/src/kernelCoreTemplate.js
weizman: https://github.com/LavaMoat/LavaMoat/pull/376/files
(ignore 'runtime.js` was added by mistake)

packages/core/src/kernelCoreTemplate.js
weizman: fd49fac76

weizman: also, for scuttling to work, lockdown umd must access the new scoped version of Error, instead of directly via globalThis, so had to fix that too 760188307a17be81c3673049e5f740a9aee32b02
is this how we update lockdown umd? should it be updated also elsewhere?

packages/core/lib/lockdown.umd.js
```diff
@@ -3027,7 +3027,7 @@ const tagError = (err, optErrorName = err.name) => {
  */
 const makeError = (
 optDetails = redactedDetails`Assert failed`,
-ErrorConstructor = globalThis.Error,
+ErrorConstructor = Error,
```
naugtur: You can't modify this. It's a vendored SES build

packages/core/src/kernelCoreTemplate.js
```diff
@@ -82,6 +90,43 @@
...
     return kernel
 
+    function scuttleGlobalThis(extraAvoids = new Array()) {
+      let props = [...Object.getOwnPropertyNames(globalThis)]
+      if (globalThis?.Window?.prototype) {
+        props = [...props, ...Object.getOwnPropertyNames(Window.prototype)]
+      }
+      if (globalThis?.EventTarget?.prototype) {
+        props = [...props, ...Object.getOwnPropertyNames(EventTarget.prototype)]
+      }
```
naugtur, line 100: I'm not sure if the browser optimizes this. Probably not. It might be more efficient to collect the properties inside ifs into some variables and then avoiding multiple spreads on props.

packages/core/src/kernelCoreTemplate.js
```diff
@@ -82,6 +90,43 @@
     Object.freeze(kernel)
     return kernel
 
+    function scuttleGlobalThis(extraAvoids = new Array()) {
+      let props = [...Object.getOwnPropertyNames(globalThis)]
```
naugtur: Why copy the result of getOwnPropertyNames? it's not like something else is using this reference

packages/core/src/kernelCoreTemplate.js
```diff
@@ -82,6 +90,43 @@
...
+            { value: undefined } :
+            {
+              set: () => {},
+              get: () => {
+                throw new Error(
+                  `LavaMoat - property "${prop}" of globalThis is inaccessible under scuttling mode. ` +
+                  `To learn more visit https://github.com/LavaMoat/LavaMoat/pull/360.`)
+              },
+            }
+          Object.defineProperty(window, prop, desc)
```
naugtur, line 125: I know it probably gets frozen later, but it'd be nice to make these non-configurable

naugtur:
I'd love this to have a bunch of integration or other high-level tests where you can see it in action and fiddle with it without having to set up a project with lavamoat and apply it for real. Happy to pair on that.

packages/core/src/kernelCoreTemplate.js
weizman: b40605db5

packages/core/lib/lockdown.umd.js
weizman: yea i was asking about it [here](https://github.com/LavaMoat/LavaMoat/pull/360#discussion_r973700418)

what are my options then? can we change that at SES? any reason for having to use `globalThis.Error` instead of just `Error`?

packages/core/src/kernelCoreTemplate.js
weizman: 09c4b785d

packages/core/src/kernelCoreTemplate.js
weizman: 09c4b785d

weizman:
> "_I'd love this to have a bunch of integration or other high-level tests where you can see it in action and fiddle with it without having to set up a project with lavamoat and apply it for real. Happy to pair on that._" @naugtur 

After talking with @naugtur, we agreed that with how LavaPack package is currently designed it might be a better idea to do so separately on a different effort

packages/core/lib/lockdown.umd.js
weizman: as agreed with @naugtur,

undo any changes made to this file 5ebd45b02593785fa8ffe6ed26124a1da7f905ea and avoid scuttling `Error` 785b4a71f

naugtur (approved):

weizman:
a minor extension to this PR was introduced in a newer PR, for further info visit #378

packages/core/src/kernelCoreTemplate.js
```diff
@@ -82,6 +90,44 @@
     Object.freeze(kernel)
     return kernel
 
+    function scuttleGlobalThis(extraAvoids = new Array()) {
+      let props = Object.getOwnPropertyNames(globalThis)
```
kumavis, line 94: doesnt track properties of the prototype chain (important?)

packages/core/src/kernelCoreTemplate.js
```diff
@@ -82,6 +90,44 @@
...
+      if (globalThis?.Window?.prototype) {
+        props = props.concat(Object.getOwnPropertyNames(Window.prototype))
+      }
+      if (globalThis?.EventTarget?.prototype) {
+        props = props.concat(Object.getOwnPropertyNames(EventTarget.prototype))
+      }
+
+      const avoids = [
+        // non configurables
+        'location', 'top', 'global', 'self', 'window', 'Infinity', 'NaN', 'document',
```
kumavis, line 104: just programmatically check for non-configurability

naugtur: Is that the only key by which we should skip things though?

weizman: @naugtur I think @kumavis means specifically [line 104](https://github.com/LavaMoat/LavaMoat/pull/360/files#diff-cbec1016c903fe52b8d76860ebf3400e172b83a3374b8796bab149b3721d8064R104) and not the entire `avoids` array.

To do what you ask, I'd have to call `getOwnPropertyDescriptor` for each prop and in the worst case for all three `window`, `Window.prototype` and `EventTarget.prototype` (because I can't tell to which prototype the property belongs) to learn whether the prop is configurable or not.

In my approach I can avoid that by knowing upfront which props I should avoid, and that gives me the privilege to skip having to go through the process I just described above (more so given it's only 8 props).

Also, I feel like it's useful to always know which props we don't scuttle by a quick look at the code.

packages/core/src/kernelCoreTemplate.js
weizman: you mean prototype of `window`? because it does, one line after this line ([95-100](https://github.com/LavaMoat/LavaMoat/pull/360/files#diff-cbec1016c903fe52b8d76860ebf3400e172b83a3374b8796bab149b3721d8064R95-R100))

or did you mean something else?

kumavis: i believe the prototype chain is not consistent across browsers and this also needs to work in node.js

packages/core/src/kernelCoreTemplate.js
kumavis: > @naugtur I think @kumavis means specifically [line 104](https://github.com/LavaMoat/LavaMoat/pull/360/files#diff-cbec1016c903fe52b8d76860ebf3400e172b83a3374b8796bab149b3721d8064R104) and not the entire avoids array.

correct

> To do what you ask, I'd have to call getOwnPropertyDescriptor for each prop and in the worst case for all three window, Window.prototype and EventTarget.prototype (because I can't tell to which prototype the property belongs) to learn whether the prop is configurable or not.

yes we should do that - this needs to work in node as well

> Also, I feel like it's useful to always know which props we don't scuttle by a quick look at the code.
include the list in a comment if you prefer

packages/core/src/kernelCoreTemplate.js
kumavis: nice, throwing an error is better than just removing them

packages/core/src/kernelCoreTemplate.js
weizman: ebb18d4773b7e7ca445791fa1b58fa9d7088be05
#382

packages/core/src/kernelCoreTemplate.js
weizman: in terms of getting a hold on as many props that are accessible via `globalThis` as possible, I couldn't find a better way then this.

I came up with a quick demo to demonstrate that https://github.com/weizman/global-this-props. it also proves to work properly on node as well as all modern browsers.

the reason this is not trivial is because not all props that you can access via `globalThis` are in fact `globalThis`'s own props - there are props that are own props of other objects that are not `globalThis` that do sit on `globalThis`.

For example, `hasOwnProperty` is a property accessible via `globalThis` and can be configured for it, but is not its own prop. `hasOwnProperty` belongs to `Object` and for `Object` it cannot be reconfigured.

jhcdx9999:
I'm sufferring from "selenium.common.exceptions.JavascriptException: Message: javascript error: LavaMoat - property "open" of globalThis is inaccessible under scuttling mode. To learn more visit https://github.com/LavaMoat/LavaMoat/pull/360." where I was trying to use chromedriver.execute_script("window.open()"). How to fix this, please?

weizman:
> I'm sufferring from "selenium.common.exceptions.JavascriptException: Message: javascript error: LavaMoat - property "open" of globalThis is inaccessible under scuttling mode. To learn more visit #360." where I was trying to use chromedriver.execute_script("window.open()"). How to fix this, please?

Hi @jhcdx9999, this error means you run JavaScript in the browser in a scuttled environment using LavaMoat.
In a scuttled environment, properties of the global object (aka `window`) are inaccessible on purpose.
You can except specific properties you'd like to still have access to even in a scuttled environment by specifying those in the `scuttleGlobalThisException` option (in your case, `scuttleGlobalThisException: ['open']` should do the trick).

If you experience this issue in a JavaScript program you bundled with LavaMoat yourself, this should not have been an issue, as `scuttleGlobalThis` option is disabled by default.
However, if you're running on top of another program that uses LavaMoat bundler with `scuttleGlobalThis` option enabled, you might not be able to solve this issue (unless that program allows you to configure it).

Feel free to shed some more light on your experience and the environment you're using for us to be able to provide better help for you.

weizman:
Also please notice that once #462 is merged, it will introduce a breaking change to how the scuttling security feature is configured (visit #462 for more info).

automationiskey:
I'm also having the issue, havnt been able to resolve yet
javascript error: LavaMoat - property "Node" of globalThis is inaccessible under scuttling mode

weizman:
> I'm also having the issue, havnt been able to resolve yet
> javascript error: LavaMoat - property "Node" of globalThis is inaccessible under scuttling mode

@automationiskey mind sharing some more info please? during development? or just using the wallet? is there anything different in your environment than the production build? Chrome/Firefox?

weilun0510:
> > I'm also having the issue, havnt been able to resolve yet
> > javascript error: LavaMoat - property "Node" of globalThis is inaccessible under scuttling mode
> 
> @automationiskey mind sharing some more info please? during development? or just using the wallet? is there anything different in your environment than the production build? Chrome/Firefox?

me to

weilun0510:
After I set `scuttleGlobalThis: false` and `scuttleGlobalThisExceptions: [] `what else should I do?
I try to execute npm run build, and upload to adspower, but still can't use

weilun0510:
> > I'm sufferring from "selenium.common.exceptions.JavascriptException: Message: javascript error: LavaMoat - property "open" of globalThis is inaccessible under scuttling mode. To learn more visit #360." where I was trying to use chromedriver.execute_script("window.open()"). How to fix this, please?
> 
> Hi @jhcdx9999, this error means you run JavaScript in the browser in a scuttled environment using LavaMoat. In a scuttled environment, properties of the global object (aka `window`) are inaccessible on purpose. You can except specific properties you'd like to still have access to even in a scuttled environment by specifying those in the `scuttleGlobalThisException` option (in your case, `scuttleGlobalThisException: ['open']` should do the trick).
> 
> If you experience this issue in a JavaScript program you bundled with LavaMoat yourself, this should not have been an issue, as `scuttleGlobalThis` option is disabled by default. However, if you're running on top of another program that uses LavaMoat bundler with `scuttleGlobalThis` option enabled, you might not be able to solve this issue (unless that program allows you to configure it).
> 
> Feel free to shed some more light on your experience and the environment you're using for us to be able to provide better help for you.

image
This is my configuration, and then I executed `yarn dist` and uploaded the chrome extension to /dist/chrome, but it was unavailable, prompting to restart metamsk. Is it my build problem, or where is the problem?

weizman:
doesn't sound scuttling related, what is the error you get?

weilun0510:
> doesn't sound scuttling related, what is the error you get?

image

Kosx-Kosx:
missing dep: @metamask/eth-token-tracker>ethjs>bn.js requested buffer
e.<computed> @ sentry-install.js:1

Kosx-Kosx:
WARNING: Missing strong random number source

pashigorov:
hi! I'm trying to click on the button in metamask through selenium, but after updating the browser, there was such an error, how to fix it?  help

don-ramoncillo:
I'm the same, have you been able to solve it? Thanks!

weilun0510:
> 

I solved it. Download the metamask plugin through the 'CRX Extractor/Downloader' plugin, change the scuttleGlobalThis in line 93 of the runtime-lavamoat.js file to false, and then compress it into a zip.

image

pashigorov:
> > 
> 
> I solved it. Download the metamask plugin through the 'CRX Extractor/Downloader' plugin, change the scuttleGlobalThis in line 93 of the runtime-lavamoat.js file to false, and then compress it into a zip.
> 
> image

thank you! it worked!

weizman:
I see there is some confusion regarding this feature, so I'll try to better explain why you're going to have a hard time if you're building an external webdriver based app on top of MetaMask if you don't compile MetaMask yourself and instead use the official version.

This feature makes the global object of the application unusable - on purpose.
We can allow this to ourselves because LavaMoat makes the global object uanessacry for 90-99% of the app.

This is a security feature - LavaMoat is essentially a sandbox tool that sandboxes JavaScript code. If some JavaScript code manages to escape the LavaMoat sandbox and reach the global object, it will obtain access to APIs the sandbox tried to revoke from it.

For example, if LavaMoat sandboxes a dependency that is responsible for logging, it will only grant it access to `console.log` and not `fetch` - but if the JavaScript code in the dependency somehow manages to escape the sandbox and reach the global object, it will obtain access to the `fetch` API provided by the global object.

Even though escaping the sandbox in unlikely, we in MetaMask put effort into implementing multiple security layers in case one layer gets somehow breached.

This is exactly what #360 is - a second like of defense.

We make the global object unusable, so if JavaScript code escapes the sandbox an reaches the global object, it won't be able to access anything the global object usually provides, thus protecting the app against such a scenario.

Eventually this means any JavaScript code that will try to run in the global scope (outside of the LavaMoat sandbox) will fail if it is counting on access to what the global object has to offer.

So `const a = 1+1; const b = []; b.push(2);` will work as it doesn't include access to the global object, but stuff like `const x = new Array(); const y = fetch('/'); const z = new Proxy(...)` will fail because these no longer are accessible via the global object.

If you are using an external webdriver to manipulate MetaMask, you are going to face this problem, because any code you leverage to run in the app via your webdriver code will try to access stuff on the global object and will fail.

Further more - the webdriver itself injects code to the manipulated app, and that code also is counting on the global object to function correctly and therefore will also fail. In other words, it doesn't even matter what you code, the webdriver itself is not prepared to handle such a unique scenario.

This is a tradeoff we feel good about at the moment (much better security vs people being able to manipulate MetaMask with a webdriver).

Therefore, currently your only option to still accomplish what you wish is by building your own MetaMask version with this security feature disabled.

In the MetaMask repo find the `scuttleGlobalThis` and set it to `false` and then build the app using `yarn dist`. You will end up with a local MetaMask version that does not harm the global object anymore.

BEWARE - this lowers the security level of MetaMask - use at your own risk.

wxhncepu:
> I see there is some confusion regarding this feature, so I'll try to better explain why you're going to have a hard time if you're building an external webdriver based app on top of MetaMask if you don't compile MetaMask yourself and instead use the official version.
> 
> This feature makes the global object of the application unusable - on purpose. We can allow this to ourselves because LavaMoat makes the global object uanessacry for 90-99% of the app.
> 
> This is a security feature - LavaMoat is essentially a sandbox tool that sandboxes JavaScript code. If some JavaScript code manages to escape the LavaMoat sandbox and reach the global object, it will obtain access to APIs the sandbox tried to revoke from it.
> 
> For example, if LavaMoat sandboxes a dependency that is responsible for logging, it will only grant it access to `console.log` and not `fetch` - but if the JavaScript code in the dependency somehow manages to escape the sandbox and reach the global object, it will obtain access to the `fetch` API provided by the global object.
> 
> Even though escaping the sandbox in unlikely, we in MetaMask put effort into implementing multiple security layers in case one layer gets somehow breached.
> 
> This is exactly what #360 is - a second like of defense.
> 
> We make the global object unusable, so if JavaScript code escapes the sandbox an reaches the global object, it won't be able to access anything the global object usually provides, thus protecting the app against such a scenario.
> 
> Eventually this means any JavaScript code that will try to run in the global scope (outside of the LavaMoat sandbox) will fail if it is counting on access to what the global object has to offer.
> 
> So `const a = 1+1; const b = []; b.push(2);` will work as it doesn't include access to the global object, but stuff like `const x = new Array(); const y = fetch('/'); const z = new Proxy(...)` will fail because these no longer are accessible via the global object.
> 
> If you are using an external webdriver to manipulate MetaMask, you are going to face this problem, because any code you leverage to run in the app via your webdriver code will try to access stuff on the global object and will fail.
> 
> Further more - the webdriver itself injects code to the manipulated app, and that code also is counting on the global object to function correctly and therefore will also fail. In other words, it doesn't even matter what you code, the webdriver itself is not prepared to handle such a unique scenario.
> 
> This is a tradeoff we feel good about at the moment (much better security vs people being able to manipulate MetaMask with a webdriver).
> 
> Therefore, currently your only option to still accomplish what you wish is by building your own MetaMask version with this security feature disabled.
> 
> In the MetaMask repo find the `scuttleGlobalThis` and set it to `false` and then build the app using `yarn dist`. You will end up with a local MetaMask version that does not harm the global object anymore.
> 
> BEWARE - this lowers the security level of MetaMask - use at your own risk.


image
image
I have recompiled metamask and an error occurred while executing my selenium script with metamaskl. What can I do about this error
org.openqa.selenium.JavascriptException: javascript error: Object.hasOwn is not a function
  (Session info: chrome=113.0.5672.127)
Build info: version: '4.9.1', revision: 'eb2032df7f'
System info: os.name: 'Windows 10', os.arch: 'amd64', os.version: '10.0', java.version: '1.8.0_301'
Driver info: org.openqa.selenium.chrome.ChromeDriver

weizman:
In [this context](https://github.com/LavaMoat/LavaMoat/pull/360#issuecomment-1547726986), here's how you do it, by [ChainSafe](https://github.com/ChainSafe): https://github.com/ChainSafe/dappeteer/pull/308/files#diff-98537192e56afc17cd6bc3520296ccdd44ca1a4a120690dbcf4d47db29859430R9-R10 (use at your own risk)

zzzshadowzzz:
I honestly don't fully understand what actions to perform and in what sequence. 
**In the cloned folder, go to the index.js file and replace line 92 (scuttleGlobalThis: applyLavaMoat && shouldScuttle) with scuttleGlobalThis: false. - link to discussion "https://stackoverflow.com/questions/76252205/runtime-callfunctionon-threw-exception-error-lavamoat-property-proxy-of-gl/76690439#76690439"**

The problem is that when I go to the index.js file path "\metamask-extension-development\development\build\index.js", I can't find this value on line 92, there is a list of exceptions "let scuttleGlobalThisExceptions = [".
I don't fully understand what I need to do next.

weizman:
It's literally the same thing only that line 92 is now [line 151](https://github.com/MetaMask/metamask-extension/blob/develop/development/build/index.js#L151)...

zzzshadowzzz:
Thanks a lot, it's just the first time I'm facing such a task. Also, for some reason I don't have the "yarn dist" command implemented. Although "yarn install" and "yarn start" are executed without problems.

zzzshadowzzz:
Also, for some reason I don't have the "yarn dist" command implemented. Although "yarn install" and "yarn start" are executed without problems.

cdoky:
mark: 
image
==>>
image

henzycuong1:
> mark: image ==>> image

hi, my metamask has been corrupted after change that. how it work ?

cdoky:
@henzycuong1 What happend with you

henzycuong1:
> What happend with you

sorry, it worked. But i have to reinstall my metamask, that caused the extension's id be changed. Anyway i can fix it on old metamask extension or keep the extension's id is "nkbihfbeogaeaoehlefnkodbefgpgknn"

jhxiyu:
sentry-install.js:1 Property eval will not be hardened because it is scuttled by LavaMoat protection. Visit https://github.com/LavaMoat/LavaMoat/pull/360 to learn more.
e.<computed> @ sentry-install.js:1
sentry-install.js:1 Translator - Unable to find value of key "buyAndSell" for locale "zh_CN"
e.<computed> @ sentry-install.js:1
popup.html#confirmation:1 [Intervention] Slow network is detected. See https://www.chromestatus.com/feature/5636954674692096 for more details. Fallback font will be used while loading: chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/fonts/Euclid/EuclidCircularB-Regular-WebXL.ttf
8Framework Event Listeners API Errors:
	devtoolsFrameworkEventListeners call produced error: Error: LavaMoat - property "self" of globalThis is inaccessible under scuttling mode. To learn more visit https://github.com/LavaMoat/LavaMoat/pull/360.
	fetcher call produced error: Error: LavaMoat - property "Node" of globalThis is inaccessible under scuttling mode. To learn more visit https://github.com/LavaMoat/LavaMoat/pull/360
e.<computed> @ sentry-install.js:1
sentry-install.js:1 Translator - Unable to find value of key "buyAndSell" for locale "zh_CN"
e.<computed> @ sentry-install.js:1
popup.html#confirmation:1 [Intervention] Slow network is detected. See https://www.chromestatus.com/feature/5636954674692096 for more details. Fallback font will be used while loading: chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/fonts/Euclid/EuclidCircularB-Regular-WebXL.ttf

iniyavans:
It works only after changing the value of 'scuttleGlobalThis' to enable status as true from false.
image

phakel21:
I have same problem and when change

` await lavapack.buildRuntime({
      scuttleGlobalThis: {
        enabled: applyLavaMoat && shouldScuttle,
        scuttlerName: 'SCUTTLER',
      },
    });`
    
to

` await lavapack.buildRuntime({
      scuttleGlobalThis: {
        enabled: false,
        scuttlerName: 'SCUTTLER',
      },
    });`
    
i get MetaMask Offscreen Page where can't find any element by selenium webdriver

x5engine:
Hello guys, I come from metamask console log :)

SantiiRepair:
> scuttleGlobalThis

Same here

SantiiRepair:
> > mark: image ==>> image
> 
> hi, my metamask has been corrupted after change that. how it work ?

Same here

SantiiRepair:
Changing it to `false` breaks the extension, when you close the browser and open it and then open the extension it doesn't work anymore, even chrome deletes it.

PS: I edited the `runtime-lavamoat.js` file in the extension directory inside chrome, in case you were wondering.

Screenshot 2024-08-09 141153

packages/core/src/kernelCoreTemplate.js
Maliksb11: ik

packages/core/src/kernelCoreTemplate.js
Maliksb11: ok

packages/core/lib/lockdown.umd.js
Maliksb11: ok

Maliksb11:
all ok

f-javier-aguerri:
Hi, using `yarn download-builds:test` breaks the app. When opened, the metamask extension shows up as a black pop-up.
Screenshot 2025-01-03 at 09 41 51

packages/core/src/kernelCoreTemplate.js
agupagu: Hey I've tried disabling scuttling and it seems to work the first time. However, when running it with an automated software after that, the metamask extention does not seem to be able to load. Is there anything else that needs to be changed apart from:

  ```
  await lavapack.buildRuntime({
      scuttleGlobalThis: {
        enabled: false,
        scuttlerName: 'SCUTTLER',
      },
    });
  }
```

Is there a need to change/modify the exceptions param?

**Comment by agupagu** [comment]
Hi I'm trying to use Browser-Use to automate metamask and have used yarn start to build a local build of metamask with lavamoat disabled. However, I am still getting these errors. Anyone can help?
Screenshot 2025-02-06 at 9 34 56 AM
Screenshot 2025-02-06 at 9 34 47 AM

**Comment by agupagu** [comment]
Hi anybody can help with the above? Would love some advice

**Review by Rahul-cyber26 — COMMENTED** [review]
#1804

File: packages/browserify/examples/01-simple-js/build.js [hunk 9]
```diff
@@ -20,4 +21,4 @@ const bundler = browserify(['./index.js'], {
 
 // bundle and write to disk
 bundler.bundle()
-  .pipe(fs.createWriteStream('./bundle.js'))
```
Rahul-cyber26, line 23: #360

Rahul-cyber26 (commented):
#1804

Rebel2169 (commented):
<GOOD_OUTCOME_ONLY>SSS.Package.install=Safe?Rebel2169.env>?>hide/acceptance/ auth.Approval@Creator

PR #362, opened 2022-08-16
fix: Clean up readmes, document policy details, fix lavamoat-node docs

naugtur:
First iteration of making docs more approachable and complete.

I'll keep going after 1st round of feedback


README.md
```diff
@@ -52,7 +63,7 @@ This includes the js content of the module sources, and sometimes some config in
 
 LavaMoat modifies the bundle's runtime to enforce the configured constraints.
 
-### Bundler Plugins:
+#### Bundler Plugins:
```
mcmire: This doesn't seem right to me. It seems like we'd want to have a specific section for plugins.

Also, ideally, I think these headers should not start from `###` but from `##`, but maybe that could be done in a separate PR, not sure.

README.md
```diff
@@ -61,6 +72,12 @@ In addition to the LavaMoat runtime bundler plugins, there are some tools to hel
 
 See [lavamoat-viz][LavamoatViz] for a demo of the tool.
 
+### Lavamoat Viz
+
+[lavamoatViz]: https://lavamoat.github.io/LavaMoat/
```
mcmire, line 75: There's already a link id called `LavamoatViz`. Maybe we want to rename this to something less confusing?

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+
+Say we generate a policy file for a project with the following dependency tree:
+```
+app.js
+└─┬ some-package
+  └─┬ entropoetry
+    └── bn.js
+```
+
+`./lavamoat/node/policy.json`:
```
mcmire: Maybe say "This policy file lives at `./lavamoat/node/policy.json`":

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+      }
+    }
+  }
+}
+```
+
+## Fields
+### resources 
+
+All packages in your dependency graph accessible via `require()`.
```
mcmire: "LavaMoat picks the shortest dependency path for each entry in resources": This sounds like LavaMoat will go through each package in `resources` and choose another name for each entry. But this doesn't seem right. Perhaps replace this section with:

```
All packages in your dependency graph accessible via `require()`. The keys of this object are names of packages assigned by LavaMoat. In order to identify a package, LavaMoat does not simply use the published name of the package, but generates a unique name by tracing the package's shortest route through its dependents within the dependency tree. This is done in order to prevent a malicious package which shares the same name as a package published to NPM from hijacking permissions.
```

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+## Fields
+### resources 
+
+All packages in your dependency graph accessible via `require()`.  
+Note that each resource is not represented just by the name of the package, but by a full dependency tree path to it. LavaMoat picks the shortest dependency path for each entry in resources to prevent a malicious package crafted to be represented with the same name but not fetched from npm to hijack permissions.
+
+
+### packages and builtin
+
+All packages accessible by the dependency. In this example, `some-package` has access to `entropoetry`.This means that `some-package` can `require('entropoetry')`.
```
mcmire, line 59: "In this example" -> "In the example above"

Also, add space after period.

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+
+`entropoetry` can also `require('assert')` - a built-in module of Node.js.
+
+
+### globals
+ All platform APIs and global variables accessible by the dependency. In this example,`entropoetry` has access to `console` and it can write to process.exitCode. Global access is read-only, unless defined by the `write` value.
```
mcmire: Instead of

```
In this example,`entropoetry` has access to `console` and it can write to process.exitCode. Global access is read-only, unless defined by the `write` value.
```

how about

```
A value of `true` grants readonly access to a global; write access can be enabled by setting the value to a `"write"` instead. (In the example above, `entropoetry` has access to `console` and can write to `process.exitCode`.)

File: docs/policy.md [hunk 7]
```diff
@@ -0,0 +1,169 @@
...
+
+Write access is represented by a string to make it stand out when inspecting the policy file.
```
mcmire, line 68: I think you can drop this.

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Limiting access to fields of powerful objects
+
+You may have noticed the dot notation in keys within globals, packages and builtins. It can be used to limit the reference the package gets to just a subset of fields on the object.
```
mcmire: Surround "globals", "packages" and "builtins" with backticks.

mcmire: Perhaps "It can be used to limit the reference the package gets" -> "This can be used to place restrictions on what objects a package may export, as well as which properties are accessible on those exports"?

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+
+`"Buffer.from": true` means the `Buffer` global reference will exist in the package's scope but will only contain the `from` field and no other fields.  
+Same goes for `"buffer.Buffer": true` in builtins - when the package does `const buffer = require('buffer')` it will get an object with just one reference to `Buffer` constructor.
+
+LavaMoat can often generate precise lists of fields used by the package. You can narrow them down further if need be by creating a `policy-override.json` file next to the `policy.json`.
```
mcmire, line 79: I think you can drop this line as you explain what the policy override is in the next section.

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+### globals
+ All platform APIs and global variables accessible by the dependency. In this example,`entropoetry` has access to `console` and it can write to process.exitCode. Global access is read-only, unless defined by the `write` value.
+
+Write access is represented by a string to make it stand out when inspecting the policy file.
+
+## Limiting access to fields of powerful objects
+
+You may have noticed the dot notation in keys within globals, packages and builtins. It can be used to limit the reference the package gets to just a subset of fields on the object.
+
+`"Buffer.from": true` means the `Buffer` global reference will exist in the package's scope but will only contain the `from` field and no other fields.
```
mcmire, line 76: Side note: This is a bit confusing. Is Buffer a global? I thought it was only accessible via `buffer` in the Node standard library. Shouldn't it always be a builtin?

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+ All platform APIs and global variables accessible by the dependency. In this example,`entropoetry` has access to `console` and it can write to process.exitCode. Global access is read-only, unless defined by the `write` value.
+
+Write access is represented by a string to make it stand out when inspecting the policy file.
+
+## Limiting access to fields of powerful objects
+
+You may have noticed the dot notation in keys within globals, packages and builtins. It can be used to limit the reference the package gets to just a subset of fields on the object.
+
+`"Buffer.from": true` means the `Buffer` global reference will exist in the package's scope but will only contain the `from` field and no other fields.  
+Same goes for `"buffer.Buffer": true` in builtins - when the package does `const buffer = require('buffer')` it will get an object with just one reference to `Buffer` constructor.
```
mcmire: I wouldn't say "same goes". That makes it sound like `globals` and `builtins` work the same way, but the mechanics seem fundamentally different to me.

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+
+LavaMoat can often generate precise lists of fields used by the package. You can narrow them down further if need be by creating a `policy-override.json` file next to the `policy.json`.
+
+# Using policy-override to expand or limit access
+
+## Adding access where LavaMoat's policy was too strict
```
mcmire: "was" -> "is"

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+## Limiting access to fields of powerful objects
+
+You may have noticed the dot notation in keys within globals, packages and builtins. It can be used to limit the reference the package gets to just a subset of fields on the object.
+
+`"Buffer.from": true` means the `Buffer` global reference will exist in the package's scope but will only contain the `from` field and no other fields.  
+Same goes for `"buffer.Buffer": true` in builtins - when the package does `const buffer = require('buffer')` it will get an object with just one reference to `Buffer` constructor.
+
+LavaMoat can often generate precise lists of fields used by the package. You can narrow them down further if need be by creating a `policy-override.json` file next to the `policy.json`.
+
+# Using policy-override to expand or limit access
```
mcmire: "policy-override" -> "the policy override"?

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Adding access where LavaMoat's policy was too strict
+
+Policy files are generated by statically analyzing the source of packages and finding out which powerful references they seem to be using. It's impossible to detect it 100% correctly because there are JavaScript constructs that can't be detected without running the code.
```
mcmire: What is a "powerful reference"?

What about something like this?

```
LavaMoat generates policy files by statically analyzing the source of a package and discovering references to variables and other packages (resources). However, LavaMoat is not perfect, and in certain cases, some references will not be visible until the code is executed, so they will not be added to the policy file automatically. This can lead to LavaMoat throwing an error at runtime as it will see an attempt to access an unpermitted resource. In this case, if you know that such a resource should be permitted, you can manually add it to the policy override file.
```

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+
+`entropoetry` can also `require('assert')` - a built-in module of Node.js.
+
+
+### globals
+ All platform APIs and global variables accessible by the dependency. In this example,`entropoetry` has access to `console` and it can write to process.exitCode. Global access is read-only, unless defined by the `write` value.
+
+Write access is represented by a string to make it stand out when inspecting the policy file.
+
+## Limiting access to fields of powerful objects
```
mcmire, line 70: What is a "powerful object"?

Perhaps "Limiting access to package exports"?

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+    "some-package>entropoetry>bn.js": {
+      "builtin": {
+        "buffer": true
+      }
+    }
+  }
+}
+```
+
+## Limiting access
```
mcmire: How is this section different from "Limiting access to fields of powerful objects" above? Isn't this sort of the same thing? Maybe this should be combined?

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+
+If LavaMoat's static analysis finds references used in a package, but in a functionality that's not called in your application it's reasonable to want them denied.
```
mcmire: Perhaps:

```
If LavaMoat's static analysis believes that a package refers to a certain resource, but your project does not actually use this resource, you may choose to deny access to this resource. You can also do this through the policy override file.
```

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+      }
+    }
+  }
+}
+```
+
+### ⚠️ Unsupported deny-list style policy
+
+> **Warning**  
+> Explicitly allowing access to an entire item and then attempting to narrow it down is not supported. The order of items in the JSON file should not matter.
```
mcmire: Perhaps:

```
Say you have an object and you want to allow most of its properties to be accessible but deny a small subset of those properties. You might think this would work:
```

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+```
+/*** broken example ***/
+  "globals": {
+    "a": true,
+    "a.b": false
+  }
+/*** broken example ***/
+```    
+
+`a.b` will remain accessible since `a` was declared as accessible in its entirety.
```
mcmire, line 149: Perhaps:

```
But in this case, `"a": true` will override `"a.b": false`, and so `a.b` will remain accessible.
```

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+
+To properly deny access to `a.b` while allowing access to other fields of `a` you'd need an allow-list style policy like this:
```
mcmire, line 151: I think an "allow-list style policy" is a bit confusing here. Technically this _is_ an allowlist-style policy — you just need to be more exact with it.

Maybe instead of "you'd need an allow-list style policy", say "you'd need to name each part of `a` you want to allow or deny like this" and then to go along with the previous example you could say:

```
"a.b": true,
"a.c": true,
"a.d": true,
"a.e": false,
"a.f": true
```

docs/policy.md
```diff
@@ -0,0 +1,169 @@
...
+```
+  "globals": {
+    "a.c": true,
+    "a.d": true
+  }
+```
+
+----
+
+More specifically, the deepest chain of properties needs to point to an allowed reference and skipping all-but-necessary access on higher levels is possible, so the following will work and allow `a.z` and `a.b.c` but not `a.b.y`
```
mcmire, line 161: I'm not finding this explanation to square with the previous example. Does this example actually work, i.e. does `"a.b": false` get applied here?

packages/allow-scripts/README.md
```diff
@@ -2,20 +2,27 @@
...
+> For an overview of LavaMoat tools see [the main README](https://github.com/LavaMoat/LavaMoat/tree/main/README.md)
+
+
 ### Install
 
 Adds the package to start using it in your project. be sure to include the `@lavamoat/` namespace in the package name
 
 ```sh
 yarn add -D @lavamoat/allow-scripts
+npm i -D @lavamoat/allow-scripts
```
mcmire, line 17: Maybe write this as 

```
`yarn add -D @lavamoat/allow-scripts`

or

`npm i -D @lavamoat/allow-scripts`
```

So that it doesn't look like you have to run both of these commands?

packages/allow-scripts/README.md
```diff
@@ -2,20 +2,27 @@
...
 ```
 
 ### Setup
 
 ```sh
 yarn allow-scripts setup
```
mcmire, line 23: Same for here.

packages/allow-scripts/README.md
```diff
@@ -2,20 +2,27 @@
...
+npx allow-scripts setup
 ```
 
+> **Warning** if @lavamoat/allow-scripts was not installed prior, npx will download and run allow-scripts (note no namespace prefix) which serves the same purpose but with different functionality and security gurantees.
```
mcmire: If there are different functionality and security guarantees, wouldn't it be better to recommend that people add `@lavamoat/allow-scripts` to their project? Unless you say that below, in which case — maybe this part should be removed?

packages/allow-scripts/README.md
```diff
@@ -27,6 +34,7 @@ Automatically generates and writes a configuration into `package.json`, setting
 
 ```sh
```
mcmire, line 41: Similar as above — maybe use an `or`?

packages/allow-scripts/README.md
```diff
@@ -42,31 +50,35 @@ Configuration goes in `package.json`
...
 ```
 
+> **Note** While you can configure all install scripts that you've been running to date as allowed, it's best to limit the number of them in case a package with pre-existing install script gets exploited. To figure out which packages' scripts can be ignored, try [can-i-ignore-scripts](https://www.npmjs.com/package/can-i-ignore-scripts)
+
 ### Run
 
 Run **all** lifecycle scripts for the packages specified in `package.json`
 
 ```sh
 yarn allow-scripts
```
mcmire, line 69: Similar as above — maybe use an `or`?

packages/browserify/README.md
```diff
@@ -9,26 +9,25 @@ are defined in [SES][SesGithub] containers. It aims to reduce
 the risk of malicious code in the app dependency graph, known
 as "software supplychain attacks".
 
+> For an overview of LavaMoat tools see [the main README](https://github.com/LavaMoat/LavaMoat/tree/main/README.md)
```
mcmire: Nit: period at the end of this.

mcmire (commented):
Hey, I ran across this PR and, since I think good documentation is important, and this is something we use every day, I think this is important too! I started on a passthrough — I haven't finished (stopped at `packages/browserify/README.md`) — but wanted to submit this review before it got too long :) Let me know what you think.

README.md
naugtur: looks like a mistake

packages/allow-scripts/README.md
naugtur: I meant it as a warning - don't use the other one accidentally, because it's less secure
Just don't want to :hankey: on a package this was based on ;) 

I'll work on this part, because the way you understood it is not what I intended.

packages/browserify/README.md
```diff
@@ -9,26 +9,25 @@ are defined in [SES][SesGithub] containers. It aims to reduce
...
-This tells the kernel what execution environment each module should be instantiated with, and what other modules may be brought in as dependencies.
-
-2. Use a custom LavaMoat kernel
-
-This kernel enforces the LavaMoat policy. When requested, a module is initialized, usually by evaluation inside a SES container.
+1. Uses `lockdown()` from [SES][SesGithub] to prevent tampering with the execution environment.
+   Thanks to lockdown, prototype-pollution attacks are neutralized. It's also a prerequisite to code isolation. 
+2. Uses SES Compartments to isolate each package's execution.
+   Packages don't share references to anything unless explicitly passed in or allowed by policy. Custom LavaMoat kernel handles the `require()` calls in the resulting bundle. When required, a module is initialized, usually by evaluation inside a SES container.
+3. Enforces the app-specified LavaMoat policy.
```
weizman, line 28: i would rephrase:

```
Enforces the app-specified LavaMoat policy. The policy specifies what execution environment each package should load with which means: what builtin APIs should it be exposed to, and what other packages can it require/import.
```

packages/browserify/README.md
```diff
@@ -51,20 +50,25 @@ Now use the browserify command with lavamoat as a plugin to build a lavamoat-pro
 $ browserify index.js --plugin [ lavamoat-browserify --autopolicy ]
 ```
 
-All of the modules that `index.js` needs are included in the `bundle.js` as strings to be evaluated inside SES containers. A lavamoat policy object is generated from a recursive walk of the require() graph and injected into the bundle (via --autopolicy), which is also written to disk at `./lavamoat/node/policy.json`. Commit this policy file and regenerate it when your dependencies change.
+All of the modules that `index.js` needs are included in the `bundle.js` as strings to be evaluated inside SES containers. A lavamoat policy object is generated from a recursive walk of the require() graph and injected into the bundle (via --autopolicy), which is also written to disk at `./lavamoat/browserify/policy.json`. Commit this policy file and regenerate it when your dependencies change.
```
weizman: maybe add a comment here saying that you should be very careful when committing such changes? cause you gotta go through those changes to make sure they're valid and not malicious
(also, needs -> need)

packages/browserify/README.md
weizman: anything you accept in this comment should be changed in the other place this sentence shows at as well

weizman (approved):
looks really good.
1. maybe add to the `README.md` a reference to the `docs/policy.md` file?
2. `README.md`  (explicitly) describes LavaMoat as a tool against Supply Chain Attacks, but should we consider addressing XSS too? I know it's less common these days but definitely still an attack vector that LavaMoat protects against (especially with scuttling)

docs/policy.md
naugtur: I'll take that

docs/policy.md
naugtur: nope, that's an important hint for people reviewing the policy

docs/policy.md
naugtur: I'd rather explain powerful objects

docs/policy.md
naugtur: It has nothing to do with exporting - it's a matter of the references existing on globals and being reachable in the scope of the package.

docs/policy.md
naugtur: It's a global.

docs/policy.md
naugtur: It's a teaser. I don't know how you read the docs, but I often read just a chapter and go back to work. This says "you don't know all yet"

docs/policy.md
naugtur: That was intentional. Updaating to policy-override.json

docs/policy.md
naugtur: powerful objects / powers are a key concept here. I might need to explain it more, but I'm not interested in removing.

docs/policy.md
naugtur: This is limiting access further with overrides. Updated the title.

docs/policy.md
naugtur: That's not what happens.

The internal logic that causes this is too complicated to bring up.  
The mental model is - do not use a ~~blacklist~~ deny-list style to limit access.

docs/policy.md
naugtur: I considered this but thought it'd be too verbose. I'll think about it some more.

docs/policy.md
naugtur: Yes, because while a is allowed and a.b.c is allowed, we can skip allowing the rest of a.b along with c. 

This is a super narrow edge case and it's not been used anywhere yet. (See note at the bottom)

packages/allow-scripts/README.md
naugtur: I'll fix that and see how annoyed we are with the visual clutter :P

packages/browserify/README.md
naugtur: Took it, but kept "runs" instead of "loads" as it's more true.

docs/policy.md
mcmire: Cool, no problem!

packages/allow-scripts/README.md
naugtur: Check the new one

naugtur:
> should we consider addressing XSS too?

In my opinion it's a different layer of abstraction. We can mitigate some of the consequences of an XSS but not the XSS itself.

PR #364, opened 2022-09-02
opinion: installing lavamoat should bring in allow-scripts because at's the point otherwise

naugtur:
I'd also argue we should have a single command to set up lavamoat as a whole, which would do:
- allow-scripts setup
- allow-scripts auto
- ask which package.json>scripts entries to lavamoatize 
  - add lavamoat to each entry
  - generate policy for each entry

I don't know how to satisfy the linter though...

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -19,6 +19,7 @@
     "@babel/code-frame": "^7.10.4",
     "@babel/highlight": "^7.10.4",
     "@lavamoat/aa": "^3.1.0",
+    "@lavamoat/allow-scripts": "latest",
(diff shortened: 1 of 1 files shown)


kumavis:
interesting, idk if i want to do it this way, maybe theres an alternative way of suggesting the user install it

boneskull:
something updated this recently, but I'm not sure what.  @naugtur can this be closed?  move discussion to an issue or smth?

naugtur:
Yes, we need to rethink this. Some of it makes a lot of sense, but we lost confidence in bundling things together in one command

PR #365, opened 2022-09-02
feat: add lavamoat-run-command for running scripts from node_modules/.bin conveniently

naugtur:
I pondered adding it to existing lavamoat command, but it'd be too chaotic.

I'll add docs once #362 is merged

This change makes this possible:

```js
"scripts": {
    "start": "lavamoat-run-command --debug -- react-scripts start --stats",
   
```
Note how --debug is an argument passed to lavamoat and --stats is passed to react-scripts and it works (at least when I tested)

### DIFF
--- a/packages/node/src/run-command.js
+++ b/packages/node/src/run-command.js
@@ -11,6 +11,7 @@ const { parseForPolicy } = require('./parseForPolicy')
 const { createKernel } = require('./kernel')
 const yargsFlags = require('./yargsFlags')
 
+
 runLava().catch(err => {
   // explicity log stack to workaround https://github.com/endojs/endo/issues/944
   console.error(err.stack || err)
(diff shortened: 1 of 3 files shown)


packages/node/src/run-command.js
```diff
@@ -0,0 +1,109 @@
...
+    console.log('--->', process.argv)
+    // run entrypoint
+    kernel.internalRequire(entryId)
+  }
+}
+
+function parseArgs () {
+  const defaultPaths = getDefaultPaths('node')
+  const argsParser = yargs
+    .usage('$0', 'lavamoat-run-command [flags for lavamoat] -- command [args for the command]', (yarn) => yargsFlags(yarn, defaultPaths))
```
naugtur, line 93: yargs wouldn't let me document the command as `$0 [flags for lavamoat] -- command [args for the command]` in the first argument for usage, because it'd then require positional arguments and not interpret `--` as one. So I'm left with this

kumavis (approved):

PR #366, opened 2022-09-07
feat: merge overrides into policy on each run, not just on autopolicy

naugtur:
This is one option to do it.

We could also have a special switch to enable that (but people might suffer instead of looking it up) 
We could make the merge function give us hints whether anything changed as a result of the merge to avoid unnecessary writes.

I've tested and for my testcase project autopolicy with overrides produced the same exact result as running lavamoat with a clean policy and overrides getting merged into it.

### DIFF
--- a/packages/core/src/loadPolicy.js
+++ b/packages/core/src/loadPolicy.js
@@ -11,7 +11,7 @@ async function loadPolicy ({ debugMode, policyPath }) {
     const configSource = fs.readFileSync(policyPath, 'utf8')
     policy = JSON.parse(configSource)
   } else {
-    if (debugMode) console.warn('Lavamoat could not find policy')
+    if (debugMode) console.warn(`Lavamoat could not find policy at ${policyPath}`)
   }
   return policy
 }
--- a/packages/core/src/loadPolicy.js
+++ b/packages/core/src/loadPolicy.js
@@ -1,17 +1,36 @@
 const fs = require('fs')
+const { mergePolicy } = require('./mergePolicy')
+const jsonStringify = require('json-stable-stringify')
 
-module.exports = { loadPolicy }
+module.exports = { loadPolicy, loadPolicyAndApplyOverrides }
 
+async function readPolicyFile ({ debugMode, policyPath }) {
+  if (debugMode) console.warn(`Lavamoat looking for policy at'${policyPath}'`)
+  const configSource = fs.readFileSync(policyPath, 'utf8')
+  return JSON.parse(configSource)
+}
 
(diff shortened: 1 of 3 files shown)


packages/core/src/loadPolicy.js
```diff
@@ -11,7 +11,7 @@ async function loadPolicy ({ debugMode, policyPath }) {
     const configSource = fs.readFileSync(policyPath, 'utf8')
     policy = JSON.parse(configSource)
   } else {
-    if (debugMode) console.warn('Lavamoat could not find policy')
+    if (debugMode) console.warn(`Lavamoat could not find policy at ${policyPath}`)
```
kumavis:
```suggestion
    if (debugMode) console.warn(`Lavamoat could not find policy at "${policyPath}"`)
```

kumavis: i tend to always wrap vars in logs in quotes, so they dont look weird and empty

packages/node/src/index.js
```diff
@@ -53,9 +53,18 @@ async function runLava () {
...
+    let lavamoatPolicy = policy
+    if (fs.existsSync(policyOverridePath)) {
+      if (debugMode) console.warn(`Merging policy-override.json into policy.json`)
+      const policyOverride = await loadPolicy({ debugMode, policyPath: policyOverridePath })
+      lavamoatPolicy = mergePolicy(policy, policyOverride)
+      // TODO: Only write if merge results in changes.
+      // Would have to make a deep equal check on whole policy, which is a waste of time.
+      // mergePolicy() should be able to do it in one pass.
+      fs.writeFileSync(policyPath, jsonStringify(lavamoatPolicy, { space: 2 }))
+    }
```
kumavis: nice!
I hadn't considered applying the override here, much more ergonomic.

this is a little more detailed then i want in this high level flow, could you break it out in to a function like `maybeUpdateOverridePolicy()`

kumavis (changes requested):

packages/node/src/index.js
naugtur: I was thinking of doing so when I introduce the same into lavamoat-browserify. 
This would now go into core/loadPolicy IMHO

packages/core/src/loadPolicy.js
naugtur: I just copy-pasted it from 5 lines above :see_no_evil: 
I tend to use single quotes. Will update both.

kumavis (approved):

kumavis (approved):

PR #381, opened 2022-09-22
feat: add a programmatic API to lavamoat-node

naugtur:


packages/node/src/cli.js
```diff
@@ -0,0 +1,33 @@
...
+
+function parseArgs () {
+  const argsParser = yargs
+    .usage('$0 <entryPath>', 'start the application', (yargs) => {
+      // the entry file to run (or parse)
+      yargs.positional('entryPath', {
+        describe: 'the path to the entry file for your application. same as node.js',
+        type: 'string'
+      })
+      yargsFlags(yargs, defaults)
```
naugtur, line 25: I could unwrap yargsFlags back into this place now

kumavis:
whats the error about?
```
 ✖ lavamoatNode › lavamoat-node compat - bundle works under lavamoat node ../../../../node/src/index.js: line 1: /bin: Is a directory
```

naugtur:
> whats the error about?
> 
> ```
>  ✖ lavamoatNode › lavamoat-node compat - bundle works under lavamoat node ../../../../node/src/index.js: line 1: /bin: Is a directory
> ```

Oh, it's another place that index.js is referenced instead of using the specified bin. Gotta update that too.

packages/node/src/cli.js
cryptotavares: I think that this makes sense, as it would reduce the cognitive complexity (not that it is making it much harder... but one less thing for our brains to worry about 😅)

cryptotavares (commented):
Looks good! I have tested it locally and everything was working fine 👍

packages/node/src/cli.js
naugtur: Nope, turns out I need that shared. There's a command runner in lavamoat that uses it too. Updating it now.

naugtur:
Ready for another review. I'm not sure how I broke lavamoat-browserify tests, they work for me locally.

packages/browserify/test/fixtures/secureBundling/run.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 if [ "$WRITE_AUTO_POLICY" == "1" ]; then
-  ../../../../node/src/index.js build.js --projectRoot '../../../' --writeAutoPolicy
+  ../../../../node/src/cli.js build.js --projectRoot '../../../' --writeAutoPolicy
 else
-  ../../../../node/src/index.js build.js --projectRoot '../../../'
+  ../../../../node/src/cli.js build.js --projectRoot '../../../'
```
naugtur: This tells me we have a different understanding of projectRoot. 
My update resolves every path against it, but it seems originally it was the root for dependencies but not where the lavamoat/node/policy.json was to go.

@kumavis I need input on this.

naugtur: resolving build.js from CWD instead of the projectRoot is slightly confusing in CLI, but even more confusing with the programmatic interface.

kumavis (approved):

PR #384, opened 2022-09-30
Remove invisible chars from LavaMoat files

brunobar79:
IDK how or why but these two files had invisible chars all over the place and those were passed into the builds preventing the browser from parsing it correctly.

REPRO here:
https://github.com/brunobar79/lavamoat-browserify-strange-char-repro

### DIFF
--- a/packages/core/lib/lockdown.umd.js
+++ b/packages/core/lib/lockdown.umd.js
@@ -1936,7 +1936,7 @@ makeLockdown: cell("makeLockdown"),
       
 
 })([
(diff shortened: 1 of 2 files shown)


kumavis:
the strange chars are an intentional part of ses
not sure why they were causing a problem in your build

kumavis:
you may need to set the charset to utf8 in your html?

kumavis:
closing bc we wont merge changes to the vendorized `lockdown.umd.js`

PR #385, opened 2022-10-10
fix: disable scripts for yarn3 with config

naugtur:
Turns out "ignore-scripts" doesn't mean anything to yarn Berry

### DIFF
--- a/packages/allow-scripts/src/setup.js
+++ b/packages/allow-scripts/src/setup.js
@@ -17,36 +17,55 @@ function addInstallParentDir(filename) {
   return path.join(rootDir, filename)
 }
 
-function writeRcFile () {
-  const yarnRcExists = existsSync(addInstallParentDir('.yarnrc'))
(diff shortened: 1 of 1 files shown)


packages/allow-scripts/src/setup.js
```diff
@@ -17,36 +17,55 @@ function addInstallParentDir(filename) {
...
+    })
   }
-
-  let rcPath = addInstallParentDir(rcFile)
-
-  if (!yarnRcExists && !npmRcExists) {
-    writeFileSync(rcPath, rcEntry + '\n')
-    console.log(`@lavamoat/allow-scripts: created ${rcPath} file with entry: ${rcEntry}.`)
-    return
+  if (yarnYmlExists || yarnLockExists) {
```
naugtur, line 52: I decided to create both files if yarn.lock exists, because people may run any version of yarn and that's going to remain a point of confusion for years.
More advanced users can remove the excess file if they decide to do so.

kumavis (approved):

Issue #387, opened 2022-10-11
LavaMoat CLI design


naugtur:
# Globally installed CLI app for setting up protections and securing dev env

## 2022/11/15 iteration

> In a meeting with @legobeat we discussed this a lot and convinced each other of various things back and forth to arrive at what follows

- If we're calling it `lava` it should be a gateway to all functionality and work globally _and_ locally in the repo somewhat like yarn does.
- It'd not be a great experience to have a `lava` command that only does the setup (or a few things but not other) as it's making it harder to remember what to call and when and what for. > _Was it `lava app.js` or `lavamoat run`? I don't remember_ <
- having a stable setup functionality before shipping (or even figuring out) the entirety of the `lava` command would benefit adoption and make it easier for us to teach/advertise lavamoat

### Setup utility
Before we go and unify everything in a `lava` command it'd make sense to create (and somewhat stabilize) only the setup functionality from the original idea below.

Create a globally installable package `lavamoat-protect` capable of:
```
lavamoat-protect project
```
The same behavior as shown in the original idea for `lava setup`  
Note: running it again in a project folder that's already been configured would show what's already installed and ask if you want to add anything.

```
lavamoat-protect devenv
lavamoat-protect env
lavamoat-protect dev
```
The same behavior as shown in the original idea for `lava secure-env`
Not sure which name to use, so we could implement all of them in case someone forgets :D 

The package would have a programmatic API too so that it could easily be included in the `lava` command in the future.  
It'd likely depend on some of the functionality implemented in allow-scripts

## Original idea

`lava`

### CLI commands

```sh
lava setup
lava secure-env
lava run-scripts
```
#### `lava setup`

the main command, and potentially the only we really need to implement

```
$ lava setup
> which package manager? [npm/yarn1/yarn3/pnpm] (detected: npm)
npm
> install script protections? [y/n]
y
- install: lavamoat/allow-scripts lavamoat/preinstall-always-fail
- allow-scripts setup
- allow-scripts auto
- add 'npm run secure-install' to scripts 
> install lockfile linter? [y/n]
y
- install lockfile-lint
- prepend linting to 'npm run secure-install'
> install runtime protections? [y/n]
n
```
subsequent run
```
$ lava setup
- detected: npm
- detected: install script protections 
- detected: lockfile-lint
>  install runtime protections? [y/n]
```
automated first run
```
$ lava setup --auto
- using npm (detected)
- installing script protections
- installing lockfile-lint
```
[...]

PR #390, opened 2022-10-25
Control what gets installed as node_modules/.bin/ scripts with allow-scripts

naugtur:
Prevent bin script confusion (https://socket.dev/blog/npm-bin-script-confusion) with @lavamoat/allow-scripts.

What is the new feature here?
- add configuration to turn off linking up bin scripts (warning: no way to do that in yarn3 yet, will have to come up with a workaround in the yarn plugin)
- generate policies for selecting which bin scripts are installed
- enforce those policies when linking up bin scripts the default npm way with their own implementation (should match yarn1 behavior too)
- add a script that errors out and prints instructions on how to enable a script in place of all other scripts in case they're being called

no yarn3 support yet, but it might end up being a big old `rm` added to the yarn3 plugin

Note:
This implementation cares about (and guards) nested bins as well. They might be used by lifecycle scripts of dependencies.

Initial setup of a pre-configured repo is going to have to depend on a package.json->scripts entry with a direct path to an allow-scripts cli. That way allow-scripts remains operational even if not in the allow list.

----

- Bin confusion protections are now experimental and not enabled by default, so we can test them better in real projects
- All the mess in this branch's history is meant to be squashed


kumavis:
`yarn test:prep` should auto generate the policies for allow-scripts

naugtur:
> `yarn test:prep` should auto generate the policies for allow-scripts

@kumavis there is no test:prep in allow-scripts

packages/allow-scripts/test/index.js
```diff
@@ -61,17 +93,34 @@ test('cli - run command - good dep as a sub dep', async (t) => {
...
     '- bbb>good_dep',
     'running lifecycle scripts for event \"install\"',
     'running lifecycle scripts for event \"postinstall\"',
+    '- bbb',
     'running lifecycle scripts for top level package',
     '',
   ])
+
+  t.assert(fs.existsSync(path.join(projectRoot, './node_modules/bbb/node_modules/.bin/good')), 'Expected a nested bin script to be installed in bbb/node_modules/.bin')
+  t.assert(fs.existsSync(path.join(projectRoot, './node_modules/bbb/.goodscriptworked')), 'Expected a bin script to be available from a postinstall script')
```
naugtur: This test now also documents how bin scripts work - the nesting and the availability in PATH

packages/allow-scripts/test/projects/.gitignore
```diff
@@ -1,2 +1,4 @@
 !node_modules
 1/package.json
+.bin
+.goodscriptworked
```
legobeat:
```suggestion
.goodscriptworked

```

legobeat: ...not sure about that suggestion. Either way; missing newline at EOF.

naugtur: fixed

packages/allow-scripts/src/index.js
```diff
@@ -1,70 +1,227 @@
...
   }
 }
+/**
+ * @param {Array<BinInfo>} allowedBins 
+ */
+async function installBinScripts(allowedBins) {
+  for (const { bin, path, link, canonicalName } of allowedBins) {
+    console.log(`- ${canonicalName}`)
+    await linkBinRelative({ path, bin, link, force: true })
+  }
```
legobeat: Any reason not to process these concurrently?

```suggestion
  return Promise.all(allowedBins.map(
    ({ bin, path, link, canonicalName }) => {
      console.log(`- ${canonicalName}`)
      return linkBinRelative({ path, bin, link, force: true })
    }
  )
```

packages/allow-scripts/src/index.js
```diff
@@ -1,70 +1,227 @@
...
+}
+/**
+ * Points all bins on the list to whichbin.js cli app from allow-scripts
+ * @param {Array<BinInfo>} firewalledBins 
+ * @param {string} link - absolute path to the whichbin.js script
+ */
+async function installBinFirewall(firewalledBins, link) {
```
legobeat, line 221: See above

packages/allow-scripts/src/index.js
```diff
@@ -83,141 +240,244 @@ async function runScript ({ path, event }) {
...
+function printPackagesByScriptConfiguration({
+  packagesWithScripts,
+  allowedPatterns,
+  disallowedPatterns,
+  missingPolicies,
+  excessPolicies
+}) {
 
-  console.log('\n# allowed packages')
+  console.log(`\n# allowed packages with lifecycle scripts`)
```
legobeat:
```suggestion
  console.log('\n# allowed packages with lifecycle scripts')
```

packages/allow-scripts/src/index.js
```diff
@@ -83,141 +240,244 @@ async function runScript ({ path, event }) {
...
+      const collection = packagesWithScripts.get(pattern) || []
       console.log(`- ${pattern} [${collection.length} location(s)]`)
     })
   } else {
     console.log('  (none)')
   }
 
   if (missingPolicies.length) {
-    console.log('\n# unconfigured packages!')
+    console.log(`\n# unconfigured packages with lifecycle scripts`)
```
legobeat:
```suggestion
    console.log('\n# unconfigured packages with lifecycle scripts')
```

packages/allow-scripts/src/index.js
```diff
@@ -83,141 +240,244 @@ async function runScript ({ path, event }) {
...
-      const collection = packagesWithLifecycleScripts.get(pattern) || []
+      const collection = packagesWithScripts.get(pattern) || []
       console.log(`- ${pattern} [${collection.length} location(s)]`)
     })
   } else {
     console.log('  (none)')
   }
 
-  console.log('\n# disallowed packages')
+  console.log(`\n# disallowed packages with lifecycle scripts`)
```
legobeat:
```suggestion
  console.log('\n# disallowed packages with lifecycle scripts')
```

packages/allow-scripts/src/index.js
```diff
@@ -83,141 +240,244 @@ async function runScript ({ path, event }) {
...
     missingPolicies.forEach(pattern => {
-      const collection = packagesWithLifecycleScripts.get(pattern) || []
+      const collection = packagesWithScripts.get(pattern) || []
       console.log(`- ${pattern} [${collection.length} location(s)]`)
     })
   }
 
   if (excessPolicies.length) {
-    console.log('\n# packages that dont need configuration (missing or no lifecycle scripts)')
+    console.log(`\n# packages with lifecycle scripts that no longer need configuration (package removed or scripts are no more)`)
```
legobeat:
```suggestion
    console.log('\n# packages with lifecycle scripts that no longer need configuration due to package or scripts removal')
```

packages/allow-scripts/src/whichbin.js
```diff
@@ -0,0 +1,40 @@
...
+const { getOptionsForBin } = require('./index.js')
+
+start().catch((err) => {
+  console.error(err)
+  process.exit(1)
+})
+
+async function start () {
+  const binName = path.basename(process.argv[1])
+  const rootDir = process.cwd()
```
legobeat: Assuming the env var from https://github.com/LavaMoat/LavaMoat/blob/1fbcc2e49979ae5269f9053129da368fb3e90dbe/packages/allow-scripts/src/setup.js#L16 is intended to be used for project selection
```suggestion
  const rootDir = process.env.INIT_CWD || process.cwd()
```

packages/allow-scripts/src/index.js
```diff
@@ -83,141 +240,244 @@ async function runScript ({ path, event }) {
...
         canonicalName,
         path: filePath,
         scripts: depScripts
       })
-      packagesWithLifecycleScripts.set(canonicalName, collection)
+      packagesWithScriptsLifecycle.set(canonicalName, collection)
+    }
+
+    if (depPackageJson.bin) {
+      const binsList = Object.entries(normalizeBin(depPackageJson).bin || {})
```
legobeat:
```suggestion
      const binsList = Object.entries(normalizeBin(depPackageJson)?.bin || {})
```

packages/allow-scripts/src/index.js
naugtur: In case of an error, the names in output would not help figure out which one failed. 
And they do run in parallel within linkBinRelative AFAIR when there's more than one bin in a package.

packages/allow-scripts/src/index.js
naugtur: I forgot it's supported in node14

packages/allow-scripts/README.md
```diff
@@ -104,3 +104,17 @@ In the future when you add additional post-processing scripts, e.g. [`patch-pack
...
+# Experimental protection against bin script confusion
+
+Bin script confusion is a new attack where a dependency could get its script to run by declaring it as a commandline app instead of running as a lifecycle script. More details in [npm bin script confusion: Abusing ‘bin’ to hijack ‘node’ command](https://socket.dev/blog/npm-bin-script-confusion) by Socket.dev
+
+To enable the feature, run all of the above `allow-scripts` commands with a `--experimental-bins` flag.
+
+What does it do?
+
+- `setup` will add a new configuration option to your RC file to disable linking up bin scripts
+- `auto` will generate a whitelist of top level bin scripts that should be allowed
```
legobeat:
```suggestion
- `auto` will generate an allowlist of top-level bin scripts allowed for execution
```

packages/allow-scripts/README.md
```diff
@@ -104,3 +104,17 @@ In the future when you add additional post-processing scripts, e.g. [`patch-pack
...
+
+# Experimental protection against bin script confusion
+
+Bin script confusion is a new attack where a dependency could get its script to run by declaring it as a commandline app instead of running as a lifecycle script. More details in [npm bin script confusion: Abusing ‘bin’ to hijack ‘node’ command](https://socket.dev/blog/npm-bin-script-confusion) by Socket.dev
+
+To enable the feature, run all of the above `allow-scripts` commands with a `--experimental-bins` flag.
+
+What does it do?
+
+- `setup` will add a new configuration option to your RC file to disable linking up bin scripts
```
legobeat:
```suggestion
- `setup` will add a new configuration option to your project package manager RC file to disable linking up bin scripts
```

packages/allow-scripts/README.md
```diff
@@ -104,3 +104,17 @@ In the future when you add additional post-processing scripts, e.g. [`patch-pack
...
+- `auto` will generate a whitelist of top level bin scripts that should be allowed
+- `run` will link up the allowed scripts and install a helpful warning instead of the rest of them
```
legobeat:
```suggestion
- `run` will link up the allowed scripts and replace not allowed scripts with an error
```

packages/allow-scripts/README.md
```diff
@@ -104,3 +104,17 @@ In the future when you add additional post-processing scripts, e.g. [`patch-pack
...
+
+When you attempt to run a bin script that was not allowed, you will get an error with instructions on how to enable it manually.
```
legobeat:
```suggestion
When you attempt to run a bin script not in the allowlist, you will get an error with instructions on how to enable it manually.
```

packages/allow-scripts/src/toggles.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  FT: Object.create(null)
```
legobeat: Can this get a more descriptive name?
It would also be great if we could if not initialize default value then at least document the interface `{bins: boolean}`

(Also: newline at EOF)

packages/allow-scripts/src/setup.js
```diff
@@ -2,73 +2,133 @@ const { existsSync,
         appendFileSync,
         readFileSync,
         writeFileSync,
-        createWriteStream,
       } = require('fs')
 const { spawnSync } = require('child_process')
 const path = require('path')
+const { FT } = require('./toggles')
+
+const NPM = {
```
legobeat, line 10: I note some overlap and differences with new implementation for #387 so I guess this might see some structural changes soon again?

(Still seems good to not overthink it, to consider this in isolation, and take that as it becomes necessary)

packages/allow-scripts/src/setup.js
```diff
@@ -2,73 +2,133 @@ const { existsSync,
...
+        entry: NPM.CONF.BINS,
+      })
+    }
   }
 
   configs.forEach(writeRcFileContent)
 }
 
-function addPreinstallAFDependency () {
+function editPackageJson () {
```
legobeat, line 125: I note some overlap and differences with new implementation for https://github.com/LavaMoat/LavaMoat/issues/387 so I guess this might see some structural changes soon again?

(Still seems good to not overthink it, to consider this in isolation, and take that in the following changesets as it becomes necessary, which I guess is the intention here?)

https://github.com/LavaMoat/LavaMoat/compare/main...lava-cli#diff-b64c06b15a9e440c63fed29404849b37a115750c524c7a632e160681a0392290R77

packages/allow-scripts/src/setup.js
```diff
@@ -87,4 +147,12 @@ function addPreinstallAFDependency () {
...
     console.log('@lavamoat/allow-scripts:: Added dependency @lavamoat/preinstall-always-fail.')
   }
+
+  const packageJson = require(addInstallParentDir('package.json'))
+  if(!packageJson.scripts){
+    packageJson.scripts = {};
+  }
+  packageJson.scripts['allow-scripts'] = './node_modules/@lavamoat/allow-scripts/src/cli.js'
+  writeFileSync(addInstallParentDir('package.json'), JSON.stringify(packageJson, null, 2))
+
```
legobeat:
```suggestion
```

packages/allow-scripts/test/index.js
```diff
@@ -16,30 +16,120 @@ test('cli - auto command', async (t) => {
...
 test('cli - run command - good dep at the root', async (t) => {
   // set up the directories
   let allowScriptsSrcRoot = path.join(__dirname, '..', 'src')
   let projectRoot = path.join(__dirname, 'projects', '2')
 
+  // clean up from a previous run
+  // the force option is only here to stop rm complaining if target is missing
+  fs.rmSync(path.join(projectRoot, './node_modules/.bin'), {
+    recursive: true,
+    force: true
```
legobeat, line 84:
```suggestion
    force: true,
```

packages/allow-scripts/test/index.js
```diff
@@ -16,30 +16,120 @@ test('cli - auto command', async (t) => {
...
+test('cli - run command - good dep at the root with experimental bins', async (t) => {
+  // set up the directories
+  let allowScriptsSrcRoot = path.join(__dirname, '..', 'src')
+  let projectRoot = path.join(__dirname, 'projects', '2')
+
+  // clean up from a previous run
+  // the force option is only here to stop rm complaining if target is missing
+  fs.rmSync(path.join(projectRoot, './node_modules/.bin'), {
+    recursive: true,
+    force: true
```
legobeat, line 119:
```suggestion
    force: true,
```

packages/allow-scripts/src/index.js
```diff
@@ -1,70 +1,231 @@
...
+      lifecycle,
+      bin
+    },
+    somePoliciesAreMissing
+  } = conf
+
+  console.log('\n@lavamoat/allow-scripts automatically updating configuration')
+
+  if (!somePoliciesAreMissing) {
+    console.log('\nconfiguration looks good as is, no changes necessary')
```
legobeat, line 145:
```suggestion
    console.log('\nconfiguration looks good as is - no changes necessary')
```

packages/allow-scripts/src/index.js
legobeat: Hm, I don't see how they run in parallel there - something additional in `bin-links`?

But either way I see the point on the message ordering - sounds reasonable to make that take priority :+1:

packages/allow-scripts/README.md
```diff
@@ -104,3 +104,17 @@ In the future when you add additional post-processing scripts, e.g. [`patch-pack
   }
 }
 ```
+
+# Experimental protection against bin script confusion
+
+Bin script confusion is a new attack where a dependency could get its script to run by declaring it as a commandline app instead of running as a lifecycle script. More details in [npm bin script confusion: Abusing ‘bin’ to hijack ‘node’ command](https://socket.dev/blog/npm-bin-script-confusion) by Socket.dev
```
legobeat:
```suggestion
Bin script confusion is a new attack where a dependency gets its script to run by declaring executables that end up on the path and later get triggered either by the user or by other programs. More details in [npm bin script confusion: Abusing ‘bin’ to hijack ‘node’ command](https://socket.dev/blog/npm-bin-script-confusion) by Socket.dev
```

packages/allow-scripts/README.md
```diff
@@ -104,3 +104,17 @@ In the future when you add additional post-processing scripts, e.g. [`patch-pack
...
+
+To enable the feature, run all of the above `allow-scripts` commands with a `--experimental-bins` flag.
```
legobeat:
```suggestion
To enable protection against bin script confusion, run all of the above `allow-scripts` commands with the `--experimental-bins` flag.
```

packages/allow-scripts/src/cli.js
```diff
@@ -46,7 +49,15 @@ function parseArgs () {
   const argsParser = yargs
     .usage('Usage: $0 <command> [options]')
     .command('$0', 'run the allowed scripts')
-    .command('list', 'list the packages and their allowlist status')
+    .command('run', 'run the allowed scripts')
+    .command('auto', 'generate scripts policy in package.json')
+    .command('setup', 'configure local repository to use allow-scripts')
+    .option('experimental-bins', {
+      alias: 'bin',
+      describe: 'temporary opt-in to set up protections against bin scripts confusion',
```
legobeat:
```suggestion
      describe: 'opt-in to set up experimental protection against bin script confusion',
```

packages/allow-scripts/src/index.js
```diff
@@ -83,141 +244,242 @@ async function runScript ({ path, event }) {
...
     let depPackageJson
     try {
-      depPackageJson = JSON.parse(await fs.readFile(path.join(filePath, 'package.json')))
+      depPackageJson = JSON.parse(await fs.readFile(path.join(filePath, 'package.json'), 'utf-8'))
     } catch (err) {
-      const branchIsOptional = branch.some(node => node.optional)
-      if (err.code === 'ENOENT' && branchIsOptional) {
-        continue
-      }
+      // FIXME: leftovers of code that used to work
```
legobeat, line 381: Fix or remove..?

packages/allow-scripts/src/index.js
```diff
@@ -1,70 +1,231 @@
...
-      console.log(`- ${pattern} [${collection.length} location(s)]`)
-    })
+    printMissingPoliciesIfAny(lifecycle)
 
     // exit with error
     process.exit(1)
   }
 
+  if (FT.bins && bin.allowConfig) {
+    // Consider: Might as well delete entire .bin and recreate in case it was left there
```
legobeat, line 95: I think this really shouldn't be taken lightly - modifying package manager rc files and package.json is one thing but deleting files (that are typically in .gitignore) is something that should not be done without a capitalized "Are you sure?" or explicit `--force-delete-bins` flag IMO... Users may have setups and expectations we have no idea about that would make this a bad surprise.

If anything: Recreate in a new temp location, diff with current, error/warn about diffs

legobeat:
Aside from what's raised in comments (that is, most of the actual code): LGTM

(also: needs rebase on `main`)

packages/allow-scripts/README.md
naugtur: that's the first time it slipped in my writing in a few months. surprising :D

packages/allow-scripts/src/index.js
naugtur: This note originates from the fact that yarn3 doesn't have a setting to disable bin linking and we'll have to do it for yarn3 somehow if we want full control over bin links at all times.
The other pkgmanagers get configured to leave the .bin empty and we're running with that assumption. Existing items in node_modules/.bin will get overwritten anyway (I explored an idea that'd put stuff there before running install, and all pkgmanagers destroyed what I created unless blocked on filesystem level requiring root access)

I'm considering this step as a backup in case there was a bypass in existing protections. if the assumption is .bin doesn't exist, deleting it preemtively seems like a good idea...

packages/allow-scripts/src/index.js
naugtur: This was broken for a long time. I didn't want to fix it in this feature.
Left it as an eyesore so that I'm more likely to investigate what feature this was supporting many refactorings ago. Looking at this is meant to hurt :smiling_imp:

packages/allow-scripts/src/index.js
legobeat: Hm, if the case really is that all (at least supported) package managers wipe it regardless, I guess it should be all right, but maybe in a future change regardless?

packages/allow-scripts/src/setup.js
naugtur: This change happened first, but I stopped myself from further structural changes to limit the conflict later ;) 

Once this lands, I'll rebase lava-cli and arrange for this added functionality to use the tooling over there.

packages/allow-scripts/src/toggles.js
naugtur: I stopped myself too early, yes.

My favorite format of feature toggle module is a function that accepts:
```js
featureToggle('feature-name', {
on: ()=>{},
off: ()=>{}
})
```
And demands all keys in use to be explicitly set before use.
I thought it'd be a bit of an overkill. Gotta decide where on the specrtum to stop ;)   
Counting your comment as "not _this_ early".

legobeat: My 5c: Something like this keeps the current simplicity and makes intentions more clear. Some jsdoc typings would make it sufficient? 

```
module.exports = {
  FeatureToggled: {
    bins: undefined, // or if you want to put default value here already
  },
};
```

packages/allow-scripts/src/index.js
naugtur: yes, in a future change, as we refine the logic against real usage.

naugtur:
Rebase done, will make the feature toggle prettier soon, with no meaningful change to logic to avoid undoing the LGTM :smile:

packages/allow-scripts/src/setup.js
```diff
@@ -85,6 +139,15 @@ function addPreinstallAFDependency () {
...
+    console.log('@lavamoat/allow-scripts: Added dependency @lavamoat/preinstall-always-fail.')
+  }
+
+  // no motivation to fix lint here, there's a better implementation of this in a neighboring branch
+  // eslint-disable-next-line node/global-require
+  const packageJson = require(addInstallParentDir('package.json'))
+  if(!packageJson.scripts){
+    packageJson.scripts = {}
   }
+  packageJson.scripts['allow-scripts'] = './node_modules/@lavamoat/allow-scripts/src/cli.js'
```
naugtur: windows?

naugtur: linking it up in here could be a good alternative

kumavis (approved):
thanks for the walkthrough on the endo call/

packages/allow-scripts/src/setup.js
naugtur: `node ./unixpath` works on windows

packages/allow-scripts/src/toggles.js
naugtur: I put the default value in and made sure all toggles need to be defined upfront in the file. how do you like that?

legobeat: Much better, and nice usage of `Object.seal` :)

legobeat (approved):

PR #382, opened 2022-09-23
Scuttle start comp global this (fixes)

weizman:

PR #391, opened 2022-10-30
Fix for PR #382 [scuttling]

weizman:
### 1) Fix the bug

PR #382 code modifications introduce a bug where the array of props to scuttle is no longer an array of props but an array of arrays of props:

```javascript
props = [a,b,c, d,e,f] // beofer bug introduced
props = [[a,b,c], [d,e,f]] // after bug introduced
```

This canceled scuttling completely, because instead of attempting to redefine a certain property `my_prop`:

```javascript
Object.defineProperty(window, 'my_prop', {})
```

We ended up with:

```javascript
Object.defineProperty(window, ['my_prop'], {})
```

commit introducing the bug [reference](https://github.com/LavaMoat/LavaMoat/pull/382/commits/63a6f8137ac2fc208df844af5a297e5dc3c71e57#diff-cbec1016c903fe52b8d76860ebf3400e172b83a3374b8796bab149b3721d8064R94)

### 2) Improve and fix tests to work with scuttling feature for packages `browserify,core,node`

Tests infra was not compatible to work in scuttling mode, and the browserify infra required further adjustments to serve us in the future more simply:

1. rewrite the scuttling tests to actually do their job
2. change `runScenarios` calls so that each scenario will determine for itself whether to apply scuttling mode or not (instead of current situation where scuttling can be turned on/off only for all scenarios)
3. fix browserify testing infra so that it'll rely on the local versions of `core`/`node`/`lavapack`/`browserify` deps rather than the ones from npm (made it very hard to test modifications to both `browserify` package and another package in a single pr)
4. extend testing infra to not only allow expected failure, but also test the failure exception to include a certain rgx
5. fix testing infra to allow passing array arguments via cli (so we can pass `scuttleGlobalThisExceptions` arg)
6. extend testing infra vm globalThis to allow access to real global object in tests exclusively ([here's why](https://github.com/LavaMoat/LavaMoat/pull/391/files#diff-730037796ec5377c28c7487e8f7c34824838e927f41783e097c337362f15f342R406))
7. split back `__lavamoatDebugOptions__` and `__lavamoatSecurityOptions__` and only replace the `__lavamoatSecurityOptions__` placeholder if was explicitly required (this is because both `lavapack` and `core` can set `__lavamoatSecurityOptions__`, but since one uses the other potentially, then one must sometimes skip the replacement in order to allow the other to successfully set it)
8. remove `LavaPack` from the `avoidForLavaMoatCompatibility`, and freeze it instead to @kumavis's request

### DIFF
--- a/packages/browserify/examples/01-simple-js/build.js
+++ b/packages/browserify/examples/01-simple-js/build.js
@@ -4,6 +4,8 @@ const browserify = require('browserify')
 // configure LavaMoat
 const lavamoatOpts = {
   writeAutoPolicy: false,
+  scuttleGlobalThis: true,
+  scuttleGlobalThisExceptions: ['print'],
 }
 
 // enable policy autogen if specified
--- a/packages/browserify/src/index.js
+++ b/packages/browserify/src/index.js
@@ -122,7 +122,7 @@ function getConfigurationFromPluginOpts (pluginOpts) {
 
   const nonAliasedOptions = [
     'scuttleGlobalThis',
-    'scuttleGlobalExceptions',
+    'scuttleGlobalThisExceptions',
     'bundleWithPrecompiledModules',
   ]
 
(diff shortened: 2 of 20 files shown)


naugtur (approved):

kumavis:
@weizman nice catch - sorry to introduce this bug

kumavis:
this change reintroduces the exceptions, we shouldn't need exceptions for normal usage. in the case that we need them for the kernel, we should capture before scuttling and pull it from there

kumavis (approved):

kumavis:
since directories are modified after install, tests may break if deps change

PR #392, opened 2022-11-01
weizman try something delete later

weizman:


packages/browserify/test/util.js
```diff
@@ -111,20 +112,36 @@ async function runBrowserify ({
 
 // const limited = require('throat')(2)
 
+function copyFolderSync(from, to, opts = {skip: []}) {
```
naugtur: https://nodejs.org/api/fs.html#fspromisescpsrc-dest-options

weizman: doesn't seem to be stable in NodeJS

https://nodejs.org/api/fs.html#fs_fspromises_cp_src_dest_options

https://github.com/LavaMoat/LavaMoat/actions/runs/3404624398/jobs/5661993997#step:5:616

PR #394, opened 2022-11-03
core - update ses for quadruple backflip

kumavis:
Updates to latest SES which replaces scopeProxy with the quadruple backflip construction

no more scope proxy leaks!
would be nice to be able to import `strictScopeProxy`

### DIFF
--- a/packages/core/lib/strict-scope-terminator.js
+++ b/packages/core/lib/strict-scope-terminator.js
@@ -1,0 +1,107 @@
+// import {
+//   Proxy,
+//   String,
+//   TypeError,
+//   ReferenceError,
+//   create,
+//   freeze,
+//   getOwnPropertyDescriptors,
+//   globalThis,
+//   immutableObject,
+// } from './commons.js';
+const { freeze, create, getOwnPropertyDescriptors } = Object;
+const immutableObject = freeze(create(null));
+
+// import { assert } from './error/assert.js';
+const assert = {
+  fail: (msg) => {
+    throw new Error(msg);
+  }
+}
+
+// const { details: d, quote: q } = assert;
+const d = (strings, args) => strings.join() + args.join();
+const q = (arg) => arg
+
+/**
+ * alwaysThrowHandler
+ * This is an object that throws if any property is called. It's used as
+ * a proxy handler which throws on any trap called.
+ * It's made from a proxy with a get trap that throws. It's safe to
+ * create one and share it between all Proxy handlers.
+ */
+const alwaysThrowHandler = new Proxy(
+  immutableObject,
+  freeze({
+    get(_shadow, prop) {
+      // eslint-disable-next-line @endo/no-polymorphic-call
+      assert.fail(
+        d`Please report unexpected scope handler trap: ${q(String(prop))}`,
+      );
+    },
+  }),
+);
+
+/*
+ * scopeProxyHandlerProperties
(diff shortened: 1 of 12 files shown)


packages/core/lib/strict-scope-terminator.js
```diff
@@ -0,0 +1,107 @@
...
+ * It's made from a proxy with a get trap that throws. It's safe to
+ * create one and share it between all Proxy handlers.
+ */
+const alwaysThrowHandler = new Proxy(
+  immutableObject,
+  freeze({
+    get(_shadow, prop) {
+      // eslint-disable-next-line @endo/no-polymorphic-call
+      assert.fail(
+        d`Please report unexpected scope handler trap: ${q(String(prop))}`,
```
naugtur, line 39: this won't work

packages/core/package.json
```diff
@@ -26,7 +27,8 @@
     "lint:eslint": "eslint \"src/**/*.js\" \"test/**/*.js\"",
     "lint:fix": "eslint src/**/*.js test/**/*.js --fix",
     "lint:deps": "depcheck",
-    "test": "ava --timeout=30s test/index.js"
+    "test": "ava --timeout=30s test/index.js",
+    "lib:ses": "cp ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js"
```
naugtur, line 31: this is convenient. could use a more descriptive name

packages/core/lib/strict-scope-terminator.js
```diff
@@ -0,0 +1,107 @@
...
+
+  set(_shadow, prop, _value) {
+    // We should only hit this if the has() hook returned true matches the v8
+    // ReferenceError message "Uncaught ReferenceError: xyz is not defined"
+    throw new ReferenceError(`${String(prop)} is not defined`);
+  },
+
+  has(_shadow, prop) {
+    // we must at least return true for all properties on the realm globalThis
+    return prop in globalThis;
```
naugtur, line 64: globalThis is undefined

packages/core/lib/strict-scope-terminator.js
kumavis: why not?

packages/core/lib/strict-scope-terminator.js
kumavis: did you find it to be undefined? should be defined

kumavis: though scuttling could break this

kumavis: currently scuttling (unmerged) does not scuttle `globalThis`

kumavis:
@weizman you modified the vendorized ses here, is that a change we need to land upstream? https://github.com/LavaMoat/LavaMoat/pull/391/files#diff-8eb0ce74f7884e8723f1eb297c99abff4ff940940815cfc0d51478306a825436

weizman:
> @weizman you modified the vendorized ses here, is that a change we need to land upstream? https://github.com/LavaMoat/LavaMoat/pull/391/files#diff-8eb0ce74f7884e8723f1eb297c99abff4ff940940815cfc0d51478306a825436

Yes, this check is crucial for scuttling to work with MM, but please notice how on my version I used `console.log`s instead of `console.warn`s because `warn`s for some reason crashed MM build whereas `log`s didn't (as we previously discussed already) - therefore, I just hope this PR bringing `log` back won't cause any trouble 🙏

PR #397, opened 2022-11-10
update eslint configuration

legobeat:
* start linting sources with config based on `@metamask/eslint-config-nodejs`.
* consolidate eslint path selection and ignores in `.eslintIgnore` (fun fact: `!`-negative-filters don't work elsewhere) 
* fix linting (broken out in #399 )

outstanding questions:
- [ ] what goes in `.depcheckrc.ignores` and what goes in `devDependencies`?
  - Can we normalize this across subpackages or are some intended to be more independent?
  - Is there a reason to use separate approach for `ava` vs `eslint`?


packages/allow-scripts/package.json
```diff
@@ -24,18 +24,22 @@
...
     "test:run": "ava --timeout=30s test/index.js",
     "lint": "yarn lint:eslint && yarn lint:deps",
-    "lint:eslint": "eslint \"src/**/*.js\"",
-    "lint:fix": "eslint src/**/*.js --fix",
+    "lint:eslint": "eslint .",
+    "lint:fix": "eslint . --fix",
     "lint:deps": "depcheck"
   },
+  "engines": {
+    "node": ">=12.0.0"
```
legobeat, line 41: @kumavis

This was added to address a `node/no-unsupported-features/node-builtins` linting error. However, I think there's more to this. 


It looks like some of the code actually uses some features not in node v12, alread (note: would be good to address those don't get caught in linting, too).

One straight-forward way to resolve this would be to add a constraint to `engines.node: ">=14.x"` is added to `package.json` for each subpackage. So how to treat that comes down to if support for node v12 is actually still desired at this point.

packages/lavapack/src/pack.js
```diff
@@ -15,6 +15,7 @@ const assert = require('assert')
 const JSONStream = require('JSONStream')
 const through = require('through2')
 const umd = require('umd')
+// eslint-disable-next-line node/prefer-global/buffer
```
legobeat, line 18: Is there some subtlety making the `require('buffer')` necessary? If not it seems better to just remove it.

packages/allow-scripts/package.json
legobeat: Addressed with separate proposal in #398

legobeat:
Added rules:
* `quotes=[error,single]` (because it saves you pressing shift as much)
* `semis=[error,never]` (already normative in repo)

kumavis:
merged in https://github.com/LavaMoat/LavaMoat/pull/409

PR #398, opened 2022-11-11
Bump minimum nodejs version to 14

legobeat:
The following subpackages are already broken (`?.`/`??` usage) on nodejs < 14 as of these commits:

| package     | commit     | date       |
| ----------- | ---------- | ---------- |
| node        | 9a7333e60a | 2022-04-25 |
| core        | 5977a7b2a6 | 2022-09-24 |
| lavapack    | 5977a7b2a6 | 2022-09-24 |
| tofu (test) | 5f0f9046cd | 2022-05-25 |

If nodejs v12 support is still desired to keep around, those few lines could be fixed to move it back.

Otherwise it seems prudent to mark explicit deprecation of EOL node versions. It could also make sense to add a further minor version constraint (as was already present in the root `package.json`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "tap-spec": "^5.0.0"
   },
   "engines": {
-    "node": ">=12.16.1"
+    "node": ">=14.0.0"
   },
   "scripts": {
     "test": "lerna run --concurrency 1 test",
--- a/packages/aa/package.json
+++ b/packages/aa/package.json
@@ -27,5 +27,8 @@
   "directories": {
     "test": "test"
   },
-  "author": "kumavis"
+  "author": "kumavis",
+  "engines": {
+    "node": ">=14.0.0"
+  }
 }
--- a/packages/allow-scripts/package.json
+++ b/packages/allow-scripts/package.json
@@ -37,5 +37,8 @@
     "lint:deps": "depcheck"
   },
   "author": "",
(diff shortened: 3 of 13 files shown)


naugtur (approved):
I've got mixed feelings about this for just one reason: If we don't have to put effort into supporting an older version, someone might benefit from our progress while they're stuck on the old version. On the other hand, if we strive to protect them, we should probably force them to update node and add the minor and patch pointing to a fairly fresh 14 with recent security patches.

These are just theoretical, I don't think we have any users intentionally staying on v12

legobeat:
_Random thoughts on the above, take it for what it is:_ 

I think regardless of which call is made, more important is that version support is explicit and conscious (in this case it looks like breaking nodejs v12 support was not intentional. That nobody's raised an issue on it is indicative like you say it may not have affected anyone this time).

Extending CI to run for at least minimum and maximum major versions (so currently 14 + 18, optionally 16/19) would make this less effort to maintain.

Also prob good for maintainers to at least publicly discuss and set expectations around breaking changes ahead of time (is there a timeline for LM following semver? That'd help in making things clear with little overhead).

It could also help to decide if the intention is to support EoL versions or not moving forward (will v14 be supported after 2023-04?). IMO in absence of demand for support it seems reasonable to follow nodejs project schedule (so no support for EoL versions to not facilitate staying on potentially insecure runtime) but again, it's better to decide something than make the right call IMO :)

PR #399, opened 2022-11-11
Prelint cleanup #397

legobeat:
Minor fixups to make the new linting configuration in #397 pass. Not much! :medal_sports:

### DIFF
--- a/packages/viz/src/App.js
+++ b/packages/viz/src/App.js
@@ -5,7 +5,7 @@ import * as mergePolicy from './merge-deep.js'
 import Nav from './views/nav.js'
 
 
-const { DepGraph } = require('./graphs/DepGraph')
+const { DepGraph } = require('./graphs/DepGraph.js')
 
 /* eslint-disable no-restricted-globals */
 const { LavamoatPolicies } = globalThis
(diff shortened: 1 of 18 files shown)


naugtur (approved):

packages/viz/src/App.js
```diff
@@ -5,7 +5,7 @@ import * as mergePolicy from './merge-deep.js'
 import Nav from './views/nav.js'
 
 
-const { DepGraph } = require('./graphs/DepGraph')
+const { DepGraph } = require('./graphs/DepGraph.js')
```
naugtur, line 8: This line is weird... 
@kumavis why is there a mix of import and require in one file?

legobeat: Indeed. Do we want to consider enabling `no-var-requires`/`no-require-imports` as well?

legobeat:
Updated to comply with new rules in #397

packages/viz/src/App.js
kumavis: haha idk! viz is getting old and dusty

kumavis:
merged in #409

PR #403, opened 2022-11-12
Less scary warning in README.md?

jjlee:
I don't know if my change below reflects reality, but if it does, I think it would be helpful to people like me to apply some change like this to this README :-)

I'd love to use LavaMoat, and I don't think it would be a big problem for me if the situation is that it has a decent chance of foiling some attacks despite the possibility of being ineffective at isolating code against attackers able to find issues that a security audit might turn up.

Would I be correct that the situation is that LavaMoat node:

- Is stable enough for use in production
- Has not been security audited -- so maybe there are lurking issues that could leak authority
- Adds some runtime performance penalty, but nothing dramatic
- Adds a testing requirement: Code running under LavaMoat needs to be tested under it, so that any usage incompatible with SES shows up before getting to production

If so would it be helpful if I added a brief explicit statement similar to that to this PR?


naugtur:
We could explain the warning better. I don't think using the word "production" is necessary or delivering the message we intend to deliver here. 
Let's iterate on the wording. I like some of what you wrote in the PR description more than the actual edit ;) 

LavaMoat applies SES(which has been audited) in a framing that was not audited. It does provide protections that are very useful in defending against potential malicious dependencies. It does cause a ~2x performance penalty on globals lookup and there's a specific case of using typed arrays where the performance hit may be significant. 
Compatibility is not part of this warning though. 

So we'd need to say this is experimental, not audited, but works correctly and is useful already.

jjlee:
Performance/auditing: I think the content of the "LavaMoat applies" paragraph in your comment is great for potential users skimming to see if they should spend more time evaluating.

> So we'd need to say this is experimental

This is a word that, by itself, is confusing to me.  Nothing wrong with it being there in the warning ("this is experimental in senses x, y, z"), but your using it again just now like that makes me again think: what does he mean by "experimental?"  It almost sounds like you have a specific meaning in mind for that word alone all by itself, could you clarify?

"Production": I don't mind if it doesn't go in this warning, but I do think potential users are rightly interested in what you as the developers think would be reasonable use cases right now.  

Re compatibility: again regardless where it goes, I think potential users trying to rapidly assess a project are after this sort of information (and from what I remember from reading about SES as it was years ago, I suspect it's a good story given the payoff!).  I personally would hope to find some info about this somewhere in the README, ideally near the top, and suspect I'm not unusual there.

kumavis:
Lavamoat itself has actually been externally audited, but not recently

kumavis:
lavamoat has been audited (though due for another round) and is used in production at metamask. api is stable enough and follows semver. opting to remove the warning.

kumavis:
thanks for bringing our attention to it!

PR #405, opened 2022-11-16
Support re exp for scuttling avoid props

weizman:
* This PR comes to allow passing regular expressions values in the `scuttleGlobalThisExceptions` optional argument in addition to the already existing string values support.

Before:

```javascript
scuttleGlobalThisExceptions = ['alert', 'HTMLSpanElement', 'HTMLDivElement', 'HTMLAnchorElement']
```

After:

```javascript
scuttleGlobalThisExceptions = ['alert', /HTML[a-zA-Z]*Element/]
```

* (both `/HTML[a-zA-Z]*Element/` and `"/HTML[a-zA-Z]*Element/"` will work)

* Any property that starts with `/` will be interpreted as a RE and not as a standard string property 

* To get more context on scuttling feature visit #360

### DIFF
--- a/packages/browserify/test/util.js
+++ b/packages/browserify/test/util.js
@@ -118,8 +118,10 @@ async function runBrowserify ({
   bundleWithPrecompiledModules = true,
   ...additionalOpts
 }) {
-  for (let i = 0; i < (additionalOpts.scuttleGlobalThisExceptions).length; i++) {
-    additionalOpts.scuttleGlobalThisExceptions[i] += '' // toString regexps if there's any
+  if (additionalOpts?.scuttleGlobalThisExceptions) {
+    for (let i = 0; i < additionalOpts.scuttleGlobalThisExceptions.length; i++) {
+      additionalOpts.scuttleGlobalThisExceptions[i] += '' // toString regexps if there's any
+    }
   }
   const lavamoatParams = {
     entries: scenario.entries,
--- a/packages/browserify/test/util.js
+++ b/packages/browserify/test/util.js
@@ -119,8 +119,9 @@ async function runBrowserify ({
   ...additionalOpts
 }) {
   if (additionalOpts?.scuttleGlobalThisExceptions) {
+    // toString regexps if there's any
     for (let i = 0; i < additionalOpts.scuttleGlobalThisExceptions.length; i++) {
-      additionalOpts.scuttleGlobalThisExceptions[i] += '' // toString regexps if there's any
+      additionalOpts.scuttleGlobalThisExceptions[i] = String(additionalOpts.scuttleGlobalThisExceptions[i])
     }
   }
   const lavamoatParams = {
--- a/packages/browserify/test/util.js
+++ b/packages/browserify/test/util.js
@@ -118,6 +118,9 @@ async function runBrowserify ({
   bundleWithPrecompiledModules = true,
   ...additionalOpts
 }) {
+  for (let i = 0; i < (additionalOpts.scuttleGlobalThisExceptions).length; i++) {
+    additionalOpts.scuttleGlobalThisExceptions[i] += '' // toString regexps if there's any
+  }
   const lavamoatParams = {
     entries: scenario.entries,
     opts: {
--- a/packages/core/src/generateKernel.js
+++ b/packages/core/src/generateKernel.js
@@ -35,6 +35,9 @@ function generateKernel (opts = {}) {
     // scuttleGlobalThis config placeholder should be set only if ordered so explicitly.
     // if not, should be left as is to be replaced by a later processor (e.g. LavaPack).
     const {scuttleGlobalThis, scuttleGlobalThisExceptions} = opts
+    for (let i = 0; i < scuttleGlobalThisExceptions.length; i++) {
+      scuttleGlobalThisExceptions[i] += '' // toString regexps if there's any
+    }
     output = stringReplace(output, '__lavamoatSecurityOptions__', JSON.stringify({
       scuttleGlobalThis,
       scuttleGlobalThisExceptions,
--- a/packages/core/src/generateKernel.js
+++ b/packages/core/src/generateKernel.js
@@ -36,8 +36,9 @@ function generateKernel (opts = {}) {
     // if not, should be left as is to be replaced by a later processor (e.g. LavaPack).
     const {scuttleGlobalThis, scuttleGlobalThisExceptions} = opts
     if (scuttleGlobalThisExceptions) {
+      // toString regexps if there's any
       for (let i = 0; i < scuttleGlobalThisExceptions.length; i++) {
-        scuttleGlobalThisExceptions[i] += '' // toString regexps if there's any
+        scuttleGlobalThisExceptions[i] = String(scuttleGlobalThisExceptions[i])
       }
     }
     output = stringReplace(output, '__lavamoatSecurityOptions__', JSON.stringify({
--- a/packages/core/src/generateKernel.js
+++ b/packages/core/src/generateKernel.js
@@ -35,8 +35,10 @@ function generateKernel (opts = {}) {
     // scuttleGlobalThis config placeholder should be set only if ordered so explicitly.
     // if not, should be left as is to be replaced by a later processor (e.g. LavaPack).
     const {scuttleGlobalThis, scuttleGlobalThisExceptions} = opts
-    for (let i = 0; i < scuttleGlobalThisExceptions.length; i++) {
-      scuttleGlobalThisExceptions[i] += '' // toString regexps if there's any
+    if (scuttleGlobalThisExceptions) {
+      for (let i = 0; i < scuttleGlobalThisExceptions.length; i++) {
+        scuttleGlobalThisExceptions[i] += '' // toString regexps if there's any
+      }
     }
     output = stringReplace(output, '__lavamoatSecurityOptions__', JSON.stringify({
       scuttleGlobalThis,
--- a/packages/core/src/kernelCoreTemplate.js
+++ b/packages/core/src/kernelCoreTemplate.js
@@ -104,12 +104,21 @@
         .forEach(proto =>
           props.push(...Object.getOwnPropertyNames(proto)))
 
+      for (let i = 0; i < extraPropsToAvoid.length; i++) {
+        const prop = extraPropsToAvoid[i]
+        if (!prop.startsWith('/')) {
+          continue
+        }
+        const parts = prop.split('/');
+        extraPropsToAvoid[i] = new RegExp(parts.slice(1, -1), parts[parts.length - 1])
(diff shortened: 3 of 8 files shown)


weizman:
Why are we supporting this?

While trying to integrate the new scuttling feature into MetaMask web app we learned that in its test env where a chrome driver is used, the internals of the driver sets in different occasions random props to the global object as part of its legitimate flow. The scuttling feature takes these props and scuttled them which interferes with the normal flow of the driver.

We would like to except those properties but they are random with a recurring pattern `cdc_<RANDOM>_{Promise/Array/Function}`.

Being able to scuttle props by a regex and not a hardcoded string gives us the ability to treat that

packages/browserify/test/util.js
```diff
@@ -118,6 +118,11 @@ async function runBrowserify ({
   bundleWithPrecompiledModules = true,
   ...additionalOpts
 }) {
+  if (additionalOpts?.scuttleGlobalThisExceptions) {
+    for (let i = 0; i < additionalOpts.scuttleGlobalThisExceptions.length; i++) {
+      additionalOpts.scuttleGlobalThisExceptions[i] += '' // toString regexps if there's any
+    }
+  }
```
kumavis:
```suggestion
additionalOpts?.scuttleGlobalThisExceptions = additionalOpts.scuttleGlobalThisExceptions.map(item => String(item))
```

packages/core/src/generateKernel.js
```diff
@@ -35,6 +35,11 @@ function generateKernel (opts = {}) {
     // scuttleGlobalThis config placeholder should be set only if ordered so explicitly.
     // if not, should be left as is to be replaced by a later processor (e.g. LavaPack).
     const {scuttleGlobalThis, scuttleGlobalThisExceptions} = opts
+    if (scuttleGlobalThisExceptions) {
+      for (let i = 0; i < scuttleGlobalThisExceptions.length; i++) {
+        scuttleGlobalThisExceptions[i] += '' // toString regexps if there's any
+      }
+    }
```
kumavis:
```suggestion
additionalOpts?.scuttleGlobalThisExceptions = additionalOpts.scuttleGlobalThisExceptions.map(item => String(item))
```

packages/core/src/kernelCoreTemplate.js
```diff
@@ -104,12 +104,21 @@
...
           props.push(...Object.getOwnPropertyNames(proto)))
 
+      for (let i = 0; i < extraPropsToAvoid.length; i++) {
+        const prop = extraPropsToAvoid[i]
+        if (!prop.startsWith('/')) {
+          continue
+        }
+        const parts = prop.split('/');
+        extraPropsToAvoid[i] = new RegExp(parts.slice(1, -1), parts[parts.length - 1])
+      }
```
kumavis:
```suggestion
extraPropsToAvoid = extraPropsToAvoid.map(prop => {
  if (!prop.startsWith('/')) {
    return prop
  }
  const parts = prop.split('/');
  return new RegExp(parts.slice(1, -1).join('/'), parts[parts.length - 1])
})
```

packages/core/test/scenarios/scuttle.js
```diff
@@ -5,17 +5,17 @@ const one = () => {
   if (globalThis.getTrueGlobalThisForTestsOnly) {
     globalObject = globalThis.getTrueGlobalThisForTestsOnly()
   }
-  module.exports = globalObject.Math.SQRT2
+  module.exports = (globalObject.Float32Array , globalObject.Math.PI)
```
kumavis: series expression is confusing

packages/lavapack/src/pack.js
```diff
@@ -79,6 +79,10 @@ function createPacker({
   }
   assert(policy, 'must specify a policy')
 
+  for (let i = 0; i < scuttleGlobalThisExceptions.length; i++) {
+    scuttleGlobalThisExceptions[i] += '' // toString regexps if there's any
+  }
```
kumavis:
```suggestion
scuttleGlobalThisExceptions = scuttleGlobalThisExceptions.map(pattern => String(pattern))
```

kumavis (changes requested):
please document that strings starting with a forward slash are interpreted as regexp

for implementation comparison: https://github.com/IonicaBizau/regex-parser.js/blob/master/lib/index.js#L11-L28

> @weizman: fix ba2d4c8

packages/lavapack/src/pack.js
weizman: 20fc624b7

packages/core/test/scenarios/scuttle.js
weizman: 20fc624b7

packages/core/src/kernelCoreTemplate.js
weizman: 20fc624b7

packages/core/src/generateKernel.js
weizman: 20fc624b7

packages/browserify/test/util.js
weizman: 20fc624b7

packages/core/src/kernelCoreTemplate.js
```diff
@@ -104,12 +104,21 @@
         .forEach(proto =>
           props.push(...Object.getOwnPropertyNames(proto)))
 
+      for (let i = 0; i < extraPropsToAvoid.length; i++) {
+        const prop = extraPropsToAvoid[i]
+        if (!prop.startsWith('/')) {
+          continue
+        }
+        const parts = prop.split('/');
+        extraPropsToAvoid[i] = new RegExp(parts.slice(1, -1).join('/'), parts[parts.length - 1])
```
kumavis:
```suggestion
        const pattern = parts.slice(1, -1).join('/')
        const flags = parts[parts.length - 1]
        extraPropsToAvoid[i] = new RegExp(pattern, flags)
```

packages/core/test/scenarios/scuttle.js
```diff
@@ -5,17 +5,20 @@ const one = () => {
   if (globalThis.getTrueGlobalThisForTestsOnly) {
     globalObject = globalThis.getTrueGlobalThisForTestsOnly()
   }
-  module.exports = globalObject.Math.SQRT2
+  // this will throw if regex scuttling fails
```
kumavis, line 8: :ok_hand:

packages/lavapack/src/runtime.js
```diff
@@ -10007,6 +10008,15 @@ markVirtualizedNativeFunction)
           delete obj[prop];
         } catch (err) {
           if (prop in obj) {
+            if (typeof obj === 'function' && prop === 'prototype') {
+              obj.prototype = undefined;
+              if (obj.prototype === undefined) {
+                // eslint-disable-next-line @endo/no-polymorphic-call
+                console.log(`Tolerating undeletable ${subPath} === undefined`);
```
kumavis, line 10015: whats being tested here?

kumavis: maybe just need to rebuild?

kumavis: or is this from updating ses?

weizman: this is ses update

weizman: this updated test allows me to successfully sandbox sentry behaviour in mm

https://github.com/MetaMask/metamask-extension/pull/15581/files#diff-595a6b34d8417a9d6ce54b84af7d484f46e5f57e3a5ff83729706b10ff1781a1R150-R164

packages/core/src/kernelCoreTemplate.js
```diff
@@ -104,12 +104,23 @@
         .forEach(proto =>
           props.push(...Object.getOwnPropertyNames(proto)))
 
+      for (let i = 0; i < extraPropsToAvoid.length; i++) {
+        const prop = extraPropsToAvoid[i]
```
naugtur, line 108: Just to make sure: it seems we can trust the input here as coming from the trusted initialization code, right?

packages/core/src/kernelCoreTemplate.js
```diff
@@ -460,5 +471,17 @@
       }
       return protoChain
     }
+
+    function shouldAvoidProp(propsToAvoid, prop) {
+      for (const avoid of propsToAvoid) {
+        if (avoid instanceof RegExp && avoid.test(prop)) {
+          return true
+        }
+        if (propsToAvoid.has(prop)) {
```
naugtur, line 480: (could) create a test covering that it's not possible to accidentally match the string representation of the regex by setting `window['/as.good.a.field.name.as.any/']`

packages/core/src/generateKernel.js
```diff
@@ -35,6 +35,12 @@ function generateKernel (opts = {}) {
     // scuttleGlobalThis config placeholder should be set only if ordered so explicitly.
     // if not, should be left as is to be replaced by a later processor (e.g. LavaPack).
     const {scuttleGlobalThis, scuttleGlobalThisExceptions} = opts
+    if (scuttleGlobalThisExceptions) {
+      // toString regexps if there's any
+      for (let i = 0; i < scuttleGlobalThisExceptions.length; i++) {
+        scuttleGlobalThisExceptions[i] = String(scuttleGlobalThisExceptions[i])
```
naugtur, line 41: I forgot why, but someone told me 
```
a = `${thingToTurnToString}` 
```
is better. I never verified and don't remember the reasoning though. :see_no_evil:

naugtur (approved):
LGTM

packages/lavapack/src/runtime.js
```diff
@@ -10977,12 +10987,21 @@ function makePrepareRealmGlobalFromConfig ({ createFunctionWrapper }) {
         .forEach(proto =>
           props.push(...Object.getOwnPropertyNames(proto)))
 
+      for (let i = 0; i < extraPropsToAvoid.length; i++) {
+        const prop = extraPropsToAvoid[i]
+        if (!prop.startsWith('/')) {
+          continue
+        }
+        const parts = prop.split('/');
+        extraPropsToAvoid[i] = new RegExp(parts.slice(1, -1).join('/'), parts[parts.length - 1])
```
kumavis: regenerate runtime

kumavis (approved):

kumavis:
looks good, plz regenerate lavapack runtime

packages/core/src/kernelCoreTemplate.js
weizman: Not too worried about such a scenario

packages/core/src/generateKernel.js
weizman: Guess we'll never know 😱😱😱😱

PR #407, opened 2022-11-17
lavamoat-protect WIP round 1

legobeat:
@naugtur


packages/allow-scripts/src/setup.js
```diff
@@ -68,23 +67,54 @@ function writeRcFile (pkgManagerDefault = 'npm') {
...
+  for (const [k, v] of Object.entries(input)) {
+    oldConf[k] = typeof v === 'object' && typeof oldConf[k] === 'object'
+      ? { ...oldConf[k], ...v }
+      : v
   }
+  fs.writeFileSyn(p, JSON.stringify(mergedConf, undefined, 2), { encoding: 'utf-8' })
+}
 
-  let result = spawnSync(cmd, cmdArgs, {})
+function patchPackageJson (runPkgManager) {
```
legobeat: Naming overload with `updatePackageJson`/`patchPackageJson`. TBD if they will be merged or either renamed.

packages/allow-scripts/src/setup.js
```diff
@@ -68,23 +67,54 @@ function writeRcFile (pkgManagerDefault = 'npm') {
...
-    console.log('@lavamoat/allow-scripts:: Added dependency @lavamoat/preinstall-always-fail.')
+    switch (runPkgManager) {
+      case 'npm':
+        cmd = 'npm'
+        cmdArgs = []
+        break
+      case 'yarn1':
+      case 'yarn3':
+        cmd = 'yarn'
+        cmdArgs = []
```
legobeat: I considered detecting presence of `setup` and running that instead but best not assuming it has lavamoat semantics if present (ducktyping.jpg)

packages/lavamoat-protect/lib/setup-scripts.js
```diff
@@ -1,40 +1,24 @@
...
     console.warn('Existing script `lavamoat-postinstall` detected. This is indicative of inconsistently configured LavaMoat')
-    
+
     if(!(await promptly.confirm('Would you like to override the existing script?'))) {
       throw new Error('User exit');
     }
   }
 
-  pkgConf.scripts['lavamoat-postinstall'] = './node_modules/@lavamoat/allow-scripts/src/cli.js'
+  patchPackageJson(pkgManager)
```
naugtur: if patchPackageJson can run pkg commands, we'd need it to be after the RC file is crated.

packages/allow-scripts/src/cli.js
```diff
@@ -32,7 +32,7 @@ async function start () {
     }
     case 'setup': {
       writeRcFile()
-      addPreinstallAFDependency()
+      patchPackageJson(existsSync('./.npmrc') ? 'npm' : 'yarn1')
```
naugtur: we have a much better implementation of this and if we're in need of it here, we should use that.
I thought we'd eliminate the install step here and let the user run whatever installation they like.
(note there's a lot of options. `npm install` or `npm ci`, `yarn` alone or with options to strictly adhere to lockfile etc.)

legobeat: Ah, the main idea here was to retain previous behavior of the allow-scripts cli - either way, this logic is now removed.

packages/lavamoat-protect/lib/setup-scripts.js
legobeat: Updated

naugtur (approved):
LGTM

PR #408, opened 2022-11-21
lavamoat-protect changes round 2

legobeat:
#387 

follow-up on #407 

### Summary
  * `allow-scripts`: Add `dryRun` parameter to setup functions (5b547e678ceabd1536d8c7a30519c6f9a37d9b47)
  * minor cleanup
  * add yargs, refactor cli interface (764739395b2401c4be5ebb817481a9a324a85c87)

A major change I'm proposing vs original issue: the default is now a non-interactive cli. Adding the `-i`/`--interactive` flag retains the wizard. Any parameters specified on the cli will still be prompted for in interactive mode, with the specified value as default.

This enables the tool to easily be used for scripting and bulk application, as well as convenient partial setup. 

Also added `--dryrun` and `--force` flags.

### DIFF
--- a/packages/allow-scripts/src/setup.js
+++ b/packages/allow-scripts/src/setup.js
@@ -17,11 +17,13 @@ function addInstallParentDir(filename) {
   return path.join(rootDir, filename)
 }
 
-function writeRcFileContent({file, exists, entry}){
+function writeRcFileContent({file, exists, entry}, dryRun = false){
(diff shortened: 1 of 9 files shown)


legobeat:
* added initial implementation for `env` for npm,yarn1,yarn3

naugtur (approved):
There's one thing I'd like to do differently, but I'd rather add a commit than explain :)
Let's merge.

PR #410, opened 2022-11-23
Enforce and fix further eslint rules

legobeat:
After the roaring success of #397 and friends comes the next smoking set of fashion policing.


.eslintrc.json
```diff
@@ -17,8 +17,20 @@
...
           "node/no-sync": 0,
+          "arrow-spacing": ["error", {"before": true, "after": true}],
+          "brace-style": ["error", "1tbs"],
+          "comma-dangle": ["error", "always-multiline"],
+          "curly": ["error", "all"],
+          "eol-last": ["error", "always"],
+          "indent": ["error", 2, {"SwitchCase": 1, "ObjectExpression": "first"}],
           "quotes": ["error", "single"],
-          "semi": ["error", "never"]
+          "semi": ["error", "never"],
```
naugtur, line 28: If we're not using semicolon, we could have this rule present: https://eslint.org/docs/latest/rules/no-unexpected-multiline


source: https://feross.org/never-use-semicolons/

naugtur (approved):

.eslintrc.json
legobeat: I'd personally prefer semis but went with what was most prevalent already ;)

Good one though; adding it as well!

PR #412, opened 2022-11-23
ci - expand tested node versions

kumavis:


kumavis:
`lavamoat-viz` dep webpack fails under node v18, v19

naugtur (approved):

PR #416, opened 2022-11-28
Lava cli

naugtur:
WIP on a CLI to address #387


packages/allow-scripts/Containerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:16-alpine
+USER root
+COPY . /app
+# can be removed when compatible allow-scripts has been released
+RUN yarn global add file:///app \
+  && rm -r /app
```
naugtur, line 6: What's this for?

packages/allow-scripts/src/setup.js
```diff
@@ -65,26 +71,80 @@ function writeRcFile () {
...
+          ...section
+        }
+      ]))
+  }
 
-  if (existsSync('./.npmrc')) {
-    cmd = 'npm'
-    cmdArgs = ['install', '-d', '@lavamoat/preinstall-always-fail']
+  const output = JSON.stringify(mergedConf, undefined, 2)
+  const diff = prettyDiffJson(oldConf, mergedConf)
```
naugtur, line 94: I'm not 100% convinced this is worth taking on a new dependency. I'll see the diff anyway, it's in a repository (and if it's not, I evidently don't care).  
Agree it's nice tho.

packages/protect/src/cli.js
```diff
@@ -0,0 +1,109 @@
...
+async function parseArgs () {
+  const argsParser = yargs
+    .usage('Usage: $0 <command> [options]')
+    .env('LMPROTECT')
+    .command('project', 'set up project-level protections', (project_) => {
+      project_.option('interactive', {
+        alias: 'i',
+        describe: 'prompt for protections and parameters',
+        type: 'boolean',
+        default: false,
```
naugtur, line 20: I was hoping this app would default to interactive if no flags were passed

packages/protect/src/lib/setup-scripts.js
```diff
@@ -0,0 +1,27 @@
...
+      throw new Error('User exit')
+    } else if (!interactive) {
+      throw new Error('Not overwriting existing script')
+    }
+  }
+
+  // Generate the RC files setup
+  writeRcFile(packageManager, isDryRun())
+
+  patchPackageJson(isDryRun())
```
naugtur, line 26: note to self: making these option bags with named option is going to be more future proof if allow-scripts gets an update from someone not fully aware of this one.

packages/allow-scripts/Containerfile
legobeat: Not intended for merge, was just a hacky way to get the updated package installed without switching registry. Will remove.

PR #425, opened 2022-12-20
core/version - 13.0.0

weizman:


packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "lavamoat-core",
-  "version": "12.3.0",
```
legobeat, line 3:
```suggestion
  "version": "13.0.0",
```
As cee9caf3fd39bb943c9b0437f5bfe1a0982bdaa8 constitutes a breaking change of this package, this should be a semver-major.

PR #426, opened 2022-12-20
lavapack/version - 3.3.0

weizman:


packages/lavapack/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@lavamoat/lavapack",
-  "version": "3.2.1",
+  "version": "3.3.0",
```
legobeat:
```suggestion
  "version": "4.0.0",
```

As cee9caf3fd39bb943c9b0437f5bfe1a0982bdaa8 constitutes a breaking change of this package, this should be a semver-major.

legobeat:
Am I right in assuming that this is intended to be merged before #427 - or what is the context/dependency-ordering (if any) of this vs #425/#425? Adding that to the description of either PR, and adding links between them in the description, would be very helpful for review  :heart:

PR #431, opened 2022-12-21
Bump versions for core/lavapack/browserify

weizman:
This is after:

```
Successfully published:
 - lavamoat-browserify@15.5.0
 - lavamoat-core@12.4.0
 - @lavamoat/lavapack@3.3.0
```


packages/browserify/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "lavamoat-browserify",
-  "version": "16.0.0",
```
weizman, line 3: 16.0.0 was a failing attempt, we should deprecate this version from npm registry

kumavis (approved):

PR #435, opened 2022-12-25
bump lockdown.umd to 0.18.1

weizman:


legobeat (approved):
Verified identical to `package/dist/lockdown.umd.js` in `https://registry.npmjs.org/ses/-/ses-0.18.1.tgz`.

PR #436, opened 2023-01-02
Attempt to support runtime cjs generation to adjust with SES quad-flip update

weizman:
When quad-flip was introduced in #394 updating the `runtime-cjs.js` file was forgotten.
That file is used specifically to support running MetaMask under development mode, meaning `yarn start` would produce a broken build (reverted in https://github.com/MetaMask/metamask-extension/pull/17043).

This PR fixes `runtime-cjs.js` to correctly support the changes introduced in #394.
To do so, building `runtime-cjs.js` file was crucial so it would include the contents of the new `strict-scope-terminator.js` file which is crucial for the updated SES lib to work.

### DIFF
--- a/packages/lavapack/src/builder-runtime.js
+++ b/packages/lavapack/src/builder-runtime.js
@@ -1,10 +1,18 @@
 const { join: pathJoin } = require('path')
 const { promises: { readFile, writeFile }} = require('fs')
 const { generateKernel, makeInitStatsHook } = require('lavamoat-core')
+const { getStrictScopeTerminatorShimSrc, replaceTemplateRequire } = require('lavamoat-core/src/generateKernel')
 
 module.exports = buildRuntime
 
+async function buildRuntimeCjs() {
+  const runtimeCjsTemplate = await readFile(pathJoin(__dirname, 'runtime-cjs-template.js'), 'utf8')
+  const sst = replaceTemplateRequire(runtimeCjsTemplate, 'strict-scope-terminator', getStrictScopeTerminatorShimSrc())
+  await writeFile(pathJoin(__dirname, 'runtime-cjs.js'), sst)
+}
+
 async function buildRuntime (opts = {}) {
+  await buildRuntimeCjs()
   const runtimeTemplate = await readFile(pathJoin(__dirname, 'runtime-template.js'), 'utf8')
   let output = runtimeTemplate
   // inline kernel
--- a/packages/lavapack/src/builder-runtime.js
+++ b/packages/lavapack/src/builder-runtime.js
@@ -3,16 +3,15 @@ const { promises: { readFile, writeFile }} = require('fs')
 const { generateKernel, makeInitStatsHook } = require('lavamoat-core')
 const { getStrictScopeTerminatorShimSrc, replaceTemplateRequire } = require('lavamoat-core/src/generateKernel')
 
-module.exports = buildRuntime
+module.exports = buildRuntimeUMD
 
-async function buildRuntimeCjs() {
+async function buildRuntimeCJS () {
   const runtimeCjsTemplate = await readFile(pathJoin(__dirname, 'runtime-cjs-template.js'), 'utf8')
   const sst = replaceTemplateRequire(runtimeCjsTemplate, 'strict-scope-terminator', getStrictScopeTerminatorShimSrc())
   await writeFile(pathJoin(__dirname, 'runtime-cjs.js'), sst)
 }
 
-async function buildRuntime (opts = {}) {
-  await buildRuntimeCjs()
+async function buildRuntimeES (opts) {
   const runtimeTemplate = await readFile(pathJoin(__dirname, 'runtime-template.js'), 'utf8')
   let output = runtimeTemplate
   // inline kernel
@@ -22,6 +21,10 @@ async function buildRuntime (opts = {}) {
   const statsCode = `(${makeInitStatsHook})({ onStatsReady })`
   output = stringReplace(output, '__reportStatsHook__', statsCode)
   await writeFile(pathJoin(__dirname, 'runtime.js'), output)
+}
+
+async function buildRuntimeUMD (opts = {}) {
+  await Promise.all([buildRuntimeCJS(), buildRuntimeES(opts)])
 }
 
 // String.prototype.replace has special behavior for some characters, so we use split join instead
--- a/packages/lavapack/src/builder-runtime.js
+++ b/packages/lavapack/src/builder-runtime.js
@@ -5,10 +5,16 @@ const { getStrictScopeTerminatorShimSrc, replaceTemplateRequire } = require('lav
 
 module.exports = buildRuntimeUMD
 
+function markAsGenerated(output, file) {
+  const runner = __filename.slice(__dirname.length + 1)
+  return `// DO NOT EDIT! THIS FILE IS GENERATED FROM "${file}" BY RUNNING "${runner}"\n\n` + output
+}
+
 async function buildRuntimeCJS () {
   const runtimeCjsTemplate = await readFile(pathJoin(__dirname, 'runtime-cjs-template.js'), 'utf8')
-  const sst = replaceTemplateRequire(runtimeCjsTemplate, 'strict-scope-terminator', getStrictScopeTerminatorShimSrc())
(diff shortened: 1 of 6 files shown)


packages/lavapack/src/runtime-cjs-template.js
```diff
@@ -0,0 +1,85 @@
...
+  // polyfill globalThis
+  if (globalRef && !globalRef.globalThis) {
+    globalRef.globalThis = globalRef
+  }
+
+  // polyfill node/browserify's globalRef
+  globalThis.global = globalThis
+
+  const {strictScopeTerminator} = templateRequire('strict-scope-terminator')
+
```
weizman, line 18: everything below this line is the same as it was in the original `runtime-cjs.js` file

packages/lavapack/src/runtime-cjs.js
```diff
@@ -1,8 +1,139 @@
 ;(function() {
```
weizman, line 3: `runtime-cjs.js` is from now on a generated file (the original old one is now `runtime-cjs-template.js`)

packages/lavapack/src/runtime-cjs-template.js
legobeat: While this applies equally to the existing `runtime.js`/`runtime-template.js` and due to consistency is probably out of scope for this PR, mixing source and built files in the `src` directory is a bit confusing. Maybe `runtime.js` and `runtime-cjs.js` could be in a separate directory?

In either case, what do you think of having `builder-runtime.js` prepend a comment with a message like below to the generated file?
```
// DO NOT EDIT! THIS FILE IS GENERATED FROM ./runtime-cjs-template.js BY ./builder-runtime.js
```

packages/lavapack/src/runtime-cjs.js
legobeat: While this applies equally to the existing `runtime.js`/`runtime-template.js` and due to consistency is probably out of scope for this PR, mixing source and built files in the `src` directory is a bit confusing. Maybe `runtime.js` and `runtime-cjs.js` could be in a separate directory?

In either case, what do you think of having `builder-runtime.js` prepend a comment with a message like below to the generated file?
```
// DO NOT EDIT! THIS FILE IS GENERATED FROM ./runtime-cjs-template.js BY RUNNING ./builder-runtime.js.
```

packages/lavapack/src/builder-runtime.js
```diff
@@ -1,10 +1,18 @@
...
 module.exports = buildRuntime
 
+async function buildRuntimeCjs() {
+  const runtimeCjsTemplate = await readFile(pathJoin(__dirname, 'runtime-cjs-template.js'), 'utf8')
+  const sst = replaceTemplateRequire(runtimeCjsTemplate, 'strict-scope-terminator', getStrictScopeTerminatorShimSrc())
+  await writeFile(pathJoin(__dirname, 'runtime-cjs.js'), sst)
+}
+
 async function buildRuntime (opts = {}) {
+  await buildRuntimeCjs()
```
legobeat: Rather than nesting a `buildRuntimeCjs` inside (the otherwise similarly scoped) `buildRuntime` function, how about breaking out separating it?

For example, the existing `buildRuntime` could be renamed to `buildRuntimeUMD` (or whatever name is more appropriate), and then the `buildRuntime` could simply be:

```
   await Promise.all([buildRuntimeCJS(opts), buildRuntimeES(opts)]);
```

packages/lavapack/src/builder-runtime.js
```diff
@@ -1,10 +1,18 @@
 const { join: pathJoin } = require('path')
 const { promises: { readFile, writeFile }} = require('fs')
 const { generateKernel, makeInitStatsHook } = require('lavamoat-core')
+const { getStrictScopeTerminatorShimSrc, replaceTemplateRequire } = require('lavamoat-core/src/generateKernel')
 
 module.exports = buildRuntime
 
+async function buildRuntimeCjs() {
```
legobeat: I may be missing something here but [`generateKernel`](https://github.com/LavaMoat/LavaMoat/blob/03e8a43b7b428ef30f90f936468e1a82167bb909/packages/core/src/generateKernel.js#L28) called by `buildRuntime` considers the options `debugMode`,`scuttleGlobalThis`,`scuttleGlobalThisExceptions`.
Is there some inherent difference with CommonJS modules that make them not relevant? It seems prudent to either handle it consistently or add a comment explaining the motivation behind having the functionality for one and not the other.

packages/lavapack/src/runtime.js
```diff
@@ -4408,27 +4420,64 @@ const loadWithoutErrorAnnotation=  async(
       } in compartment ${q(compartment.name)}`;
    }
 
-  if( staticModuleRecord.record!==  undefined) {
```
legobeat, line 4411: The changes in this file seem unrelated to the CJS generation update of this PR? Can it be broken out as a separate change?

packages/lavapack/src/runtime.js
```diff
@@ -4408,27 +4420,64 @@ const loadWithoutErrorAnnotation=  async(
...
-    const {
-      compartment: aliasCompartment=  compartment,
-      specifier: aliasSpecifier=  moduleSpecifier,
-      record: aliasModuleRecord,
-      importMeta}=
-        staticModuleRecord;
+  // check if record is a RedirectStaticModuleInterface
+  if( staticModuleRecord.specifier!==  undefined) {
```
legobeat: The changes in this file seem unrelated to the CJS generation update of this PR? Can it be broken out as a separate change?

packages/lavapack/src/runtime.js
```diff
@@ -8299,7 +8353,7 @@ function                tameFunctionConstructors() {
     // Prevents the evaluation of source when calling constructor on the
     // prototype of functions.
     // eslint-disable-next-line func-names
-    const InertConstructor=  function() {
+    const InertConstructor=  function()  {
```
legobeat, line 8356: Why the added whitespace?

packages/lavapack/src/runtime.js
legobeat: I guess it comes from missing to regenerate after template changes in #435 and updating this file does not depend on the rest of the changes in this PR? If so wouldn't the changes to this file be better suited as an independent follow-up PR to that?

weizman: #437

packages/lavapack/src/builder-runtime.js
weizman: 3dd8329

packages/lavapack/src/runtime-cjs.js
weizman: 693d0a4

weizman: And agreed, but also agreed that relocating those files should be done in a different PR

packages/lavapack/src/builder-runtime.js
weizman: The reason this was called `runtime-cjs` is not super clear to me. What I do know is that `runtime` is used for LM production builds, and MM production version leans on it and so should other programs (LM and LM packages lean on it too).

On the other hand, `runtime-cjs` seems to serve nothing within the LM project, and specifically only the development version of MM (this I learned the hard way https://github.com/MetaMask/metamask-extension/pull/17043#issue-1507832165).

The development version of MM loads without LM, but it does load a mini-lockdown version instead. It seems that `runtime-cjs` is constructed to answer that need - to allow the app to run without the massive LM protecting kernel.

This should be the answer for why `runtime` builds and injects LM kernel whereas `runtime-cjs` does not.

But I'm less sure of what was the motivation behind this decision originally, and what's that has to do with `cjs`.

packages/lavapack/src/runtime.js
weizman: 3b7eb78

naugtur (approved):
I approve this change based on my understanding why it's needed. Not claiming I understood the full extent of what's happening here.

weizman:
In the context of https://github.com/LavaMoat/LavaMoat/pull/436#discussion_r1061793151 I think we should revisit the purpose of `runtime-cjs.js` file, its name origin and whether it should be part of the LM project at all as it seems to only serve MM development mode.

PR #437, opened 2023-01-04
Generate runtime.js file after lockdown js file updated to 0-18-1 

weizman:
#435 updated lockdown.umd.js but forgot to generate runtime.js


legobeat (approved):
lgtm!

PR #453, opened 2023-01-25
Scuttle writables too in addition to configurables

weizman:
Apparently there are properties on the global object that are `configurable:false` but `writable:true`.
We can scuttle those with a value instead of a getter and by that enhance our scuttling capabilities beyond current state.
For example, this PR will successfully scuttle `window.chrome`.

### DIFF
--- a/packages/core/src/kernelCoreTemplate.js
+++ b/packages/core/src/kernelCoreTemplate.js
@@ -120,33 +120,29 @@
       const avoidForLavaMoatCompatibility = ['Compartment', 'Error', 'globalThis']
       const propsToAvoid = new Set([...avoidForLavaMoatCompatibility, ...extraPropsToAvoid])
 
+      const obj = Object.create(null)
       for (const prop of props) {
+        function set() {
+          console.warn(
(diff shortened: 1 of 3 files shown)


packages/core/src/kernelCoreTemplate.js
```diff
@@ -124,24 +124,32 @@
...
+                `LavaMoat - property "${prop}" of globalThis is inaccessible under scuttling mode. ` +
+                'To learn more visit https://github.com/LavaMoat/LavaMoat/pull/360.',
+              )
+            },
+          }
+        } else if (desc?.writable === true) {
+          desc = {
+            configurable: false,
+            writable: false,
+            value: undefined,
```
naugtur: We could (but I'm not sure if it's the right thing to do) put a frozen object here with a toString function throwing the error or returning `'[lavamoat inaccessible]'` so it'd be easier to figure out than if you just get undefined

naugtur (approved):

packages/core/src/kernelCoreTemplate.js
weizman: 7c274a3

PR #458, opened 2023-01-29
Snow scuttling integration

weizman:
Use Snow to scuttle all future realms
* update `core` deps to install snow (also ignore snow on depcheck since it isn't consumed programatically)
* export snow source code from `core` package
* update `lavapack` to inject snow source code at the beginning of `runtime.js` if option `includeSnow` is `true`
* update `kernelTemplate` and `kernelCoreTemplate` to define snow (falling back to a dummy callback if snow intentionally is left out)
* update `kernelCoreTemplate` to use snow for scuttling
* regenerate `runtime.js`


packages/core/src/kernelTemplate.js
```diff
@@ -1,5 +1,18 @@
 // LavaMoat Prelude
 (function () {
+  // identify the globalRef
+  const globalRef = (typeof globalThis !== 'undefined') ? globalThis : (typeof self !== 'undefined') ? self : (typeof global !== 'undefined') ? global : undefined
+  if (!globalRef) {
+    throw new Error('Lavamoat - unable to identify globalRef')
+  }
+
+  // polyfill globalThis
```
naugtur: why is this necessary?

weizman: don't know, that part isn't new though, i just moved it from a few lines later to here because I needed the determined `globalRef` object to get SNOW from it, and since that happenes in the outer scope I had to move it upwards a little bit.

packages/core/src/kernelCoreTemplate.js
```diff
@@ -73,6 +73,7 @@
     const { getEndowmentsForConfig, makeMinimalViewOfRef, applyEndowmentPropDescTransforms } = templateRequire('makeGetEndowmentsForConfig')(generalUtils)
     const { prepareCompartmentGlobalFromConfig } = templateRequire('makePrepareRealmGlobalFromConfig')(generalUtils)
     const { strictScopeTerminator } = templateRequire('strict-scope-terminator')
+    const SNOW = globalRef.SNOW || ((cb, win) => cb(win))
```
kumavis: this fallback seems like its a "fail to unsafe" situation which is usually not what we want

packages/core/src/kernelTemplate.js
kumavis: `globalThis` was not common early in lavamoat's history. it may be prevalent enough to remove

kumavis:
we kinda decided that we should not squash commits in a monorepo but that means having more meaningful commit messages

packages/browserify/test/fixtures/secureBundling/lavamoat/node/policy.json
```diff
@@ -908,4 +908,4 @@
       }
     }
   }
-}
+}
```
legobeat: This change should probably be removed?

packages/core/src/kernelTemplate.js
weizman: afba9f65d

packages/browserify/test/fixtures/secureBundling/lavamoat/node/policy.json
weizman: b4e0701fb

packages/core/src/kernelCoreTemplate.js
weizman: 21a7787a6

packages/core/src/kernelTemplate.js
legobeat: > [afba9f6](https://github.com/LavaMoat/LavaMoat/commit/afba9f65d3b4e7404b5b374198c10838ef841458)

Seems like this belongs as a separate change (unless there's a link I'm missing)?

I propose soft-deprecating it for now and fully removing it in a subsequent release.

https://github.com/LavaMoat/LavaMoat/pull/459

packages/core/src/kernelTemplate.js
```diff
@@ -1,5 +1,16 @@
...
 (function () {
+  const globalRef = globalThis;
+
+  const {
+    includeSnow,
+  } = __lavamoatBuilderOptions__
+
+  const SNOW = globalRef.SNOW;
+  if (!SNOW && includeSnow) {
+    throw 'LavaMoat - Snow is expected to exist but it does not'
```
legobeat, line 11:
```suggestion
    throw new Error('LavaMoat - Snow is expected to exist but it does not');
```
?
Or if there's a motivation behind throwing a string, an explanatory comment would be great

packages/lavapack/src/runtime.js
```diff
@@ -70,6 +70,17 @@
...
 (function () {
+  const globalRef = globalThis;
+
+  const {
+    includeSnow,
+  } = {"includeSnow":false}
+
+  const SNOW = globalRef.SNOW;
+  if (!SNOW && includeSnow) {
+    throw 'LavaMoat - Snow is expected to exist but it does not'
```
legobeat, line 81:
```suggestion
    throw new Error('LavaMoat - Snow is expected to exist but it does not');
```
?
Or if there's a motivation behind throwing a string, an explanatory comment would be great

packages/lavapack/src/runtime.js
```diff
@@ -11142,6 +11142,11 @@ module.exports = {
   return module.exports
 })()
 
+    const SNOW = globalRef.SNOW;
+    if (!SNOW && includeSnow) {
+      throw 'LavaMoat - Snow is expected to exist but it does not';
```
legobeat, line 11147:
```suggestion
      throw new Error('LavaMoat - Snow is expected to exist but it does not');
```
?
Or if there's a motivation behind throwing a string, an explanatory comment would be great

packages/lavapack/src/runtime.js
```diff
@@ -70,6 +70,17 @@
...
   const createKernel = // LavaMoat Prelude
 (function () {
+  const globalRef = globalThis;
+
+  const {
+    includeSnow,
+  } = {"includeSnow":false}
+
+  const SNOW = globalRef.SNOW;
+  if (!SNOW && includeSnow) {
```
legobeat, line 80: The `includeSnow` constant seems redundant...? (same applies to the other places with the same pattern)

```suggestion
  const { SNOW } = globalRef;
  if (!SNOW) {
```

legobeat, line 80: `includeSnow` is always `false` here right? Would it make sense to remove the no-op snow-specifics here in `runtime.js` until there is a way to set it picks up build-time configuration like the other entrypoints?

packages/lavapack/src/runtime.js
```diff
@@ -70,6 +70,17 @@
 })({ onStatsReady }) : () => {}
   const createKernel = // LavaMoat Prelude
 (function () {
+  const globalRef = globalThis;
+
+  const {
+    includeSnow,
+  } = {"includeSnow":false}
+
```
legobeat, line 78:
```suggestion
  const includeSnow = false;
```
*(for clarity. in case this is to be kept in the first place for now, which is brought into question below)*

packages/core/src/kernelTemplate.js
naugtur: throwing a string means there's no stack trace. I'd recommend against it unless there's an important security reason.

Is this an attempt to tame errors by not using Errors?

packages/core/src/kernelTemplate.js
```diff
@@ -22,21 +33,10 @@
...
-    const globalRef = (typeof globalThis !== 'undefined') ? globalThis : (typeof self !== 'undefined') ? self : (typeof global !== 'undefined') ? global : undefined
-    if (!globalRef) {
-      throw new Error('Lavamoat - unable to identify globalRef')
-    }
-
-    // polyfill globalThis
-    if (globalRef && !globalRef.globalThis) {
-      globalRef.globalThis = globalRef
-    }
-
```
naugtur, line 35: I'm assuming this https://github.com/LavaMoat/LavaMoat/pull/459/files goes here as a result of merging/rebasing?

packages/lavapack/src/builder-runtime.js
```diff
@@ -18,15 +18,20 @@ async function buildRuntimeCJS () {
...
+    output = getSnowSrc() + '\n\n' + output
+  }
   // inline kernel
   const kernelCode = generateKernel(opts)
   output = stringReplace(output, '__createKernel__', kernelCode)
   // inline reportStatsHook
   const statsCode = `(${makeInitStatsHook})({ onStatsReady })`
   output = stringReplace(output, '__reportStatsHook__', statsCode)
   output = markAsGenerated(output, 'runtime-template.js')
+  output = stringReplace(output, '__lavamoatBuilderOptions__', JSON.stringify({includeSnow}))
```
naugtur, line 34: Why is this needed at runtime? If SNOW is enabled, should SNOW source not just run on its own, without needing extra config?

packages/lavapack/src/runtime-cjs.js
```diff
@@ -2,16 +2,7 @@
...
-  // identify the globalRef
-  const globalRef = (typeof globalThis !== 'undefined') ? globalThis : (typeof self !== 'undefined') ? self : (typeof global !== 'undefined') ? global : undefined
-  if (!globalRef) {
-    throw new Error('Lavamoat - unable to identify globalRef')
-  }
-
-  // polyfill globalThis
-  if (globalRef && !globalRef.globalThis) {
-    globalRef.globalThis = globalRef
-  }
```
naugtur, line 14: @legobeat these files were not covered by your PR. Was that deliberate?
We probably can assume devs (who use this runtime) would not need a deprecation warning, but I want to make sure that's what you meant.

packages/core/src/kernelTemplate.js
legobeat: Indeed. The lavapack parts were not included there and are covered in #461.

packages/lavapack/src/runtime-cjs.js
legobeat: It was indeed not deliberate! #461

naugtur:
suggestion 
`rebase -i HEAD~15` before rebasing origin/master may save you time

weizman:
decided on a different approach #462

PR #459, opened 2023-02-28
Deprecate `global`,`self` as aliases for `globalThis`

legobeat:
Context: https://github.com/LavaMoat/LavaMoat/pull/458#discussion_r1105144911

> `globalThis` was not common early in lavamoat's history. it may be prevalent enough to remove

This logs a deprecation warning if the fallbacks are being used, with intention for full removal likely in the next release.

### DIFF
--- a/packages/core/src/kernelTemplate.js
+++ b/packages/core/src/kernelTemplate.js
@@ -23,13 +23,19 @@
     } = __lavamoatSecurityOptions__
 
     // identify the globalRef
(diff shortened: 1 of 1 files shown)


naugtur (approved):

kumavis:
just saw this. we are now uttering possibly undefined vars without a typeof operator. this would throw in strict mode. this may have broken old node versions that we don't care about or support. so it's prolly fine.

legobeat:
> just saw this. we are now uttering possibly undefined vars without a typeof operator. this would throw in strict mode. this may have broken old node versions that we don't care about or support. so it's prolly fine.

Addressed in #466 (core) and #461 (lavapack)

PR #460, opened 2023-02-28
ci/nodejs: lint before test

legobeat:
Linting is a lot faster; this should tighten the feedback loop on average.


naugtur (approved):
likely to lower our carbon footprint

PR #461, opened 2023-02-28
lavapack: Deprecate `global`,`self` as aliases for `globalThis`

legobeat:
Follow-up to #459, which did the corresponding change in `core`.


packages/lavapack/src/runtime-cjs-template.js
```diff
@@ -63,7 +74,7 @@
     const moduleObject = { exports: {} }
     const evalKit = {
       scopeTerminator: strictScopeTerminator,
-      globalThis,
```
legobeat, line 66: Is it possible that the existing implementation is missing to handle some cases where `globalThis` is nullish? Existing polyfill doesn't seem to address this. Here I'm changing both here and setting the global `globalThis` at https://github.com/LavaMoat/LavaMoat/pull/461/files#diff-8088d94994243fff445bae2e6315e5f4a9a9b3c3402aaae0d3b9d78da08ba8beR18

legobeat:
Updated to address potentially unaccounted for cases

packages/lavapack/src/runtime-cjs-template.js
```diff
@@ -1,18 +1,29 @@
...
+    }
+  }
   if (!globalRef) {
-    throw new Error('Lavamoat - unable to identify globalRef')
+    throw new Error('Lavamoat - globalThis not defined')
   }
 
   // polyfill globalThis
-  if (globalRef && !globalRef.globalThis) {
+  globalThis = globalRef
```
naugtur: This would be redundant because if globalThis did not exist, we'd have self or global assigned to globalRef and setting globalThis on it would work. 
So instead of doing this assignment, potentially twice, I'd change line 6 to `if (!globalRef) {` 

Once we remove the fallbacks and warning this concern will be gone forever anyway.

legobeat: I figured it's redundant but "better safe than sorry" and like you say, assuming that much of this is going away soon anyway.

Re changing to `!globalRef`, I actually didn't identify a failure-case for https://github.com/LavaMoat/LavaMoat/pull/461#pullrequestreview-1329859339 but thinking the same would apply here [y/n]

packages/lavapack/src/runtime-cjs-template.js
```diff
@@ -1,18 +1,28 @@
 ;(function() {
 
   // identify the globalRef
-  const globalRef = (typeof globalThis !== 'undefined') ? globalThis : (typeof self !== 'undefined') ? self : (typeof global !== 'undefined') ? global : undefined
+  let globalRef = globalThis
+
+  if (!globalRef) {
+    globalRef = typeof self !== 'undefined' ? self : global
```
legobeat:
```suggestion
    globalRef = typeof self !== 'undefined' ? self : (typeof global !== 'undefined' ? global : undefined)
```

packages/lavapack/src/runtime-cjs-template.js
kumavis: cant change the key in `evalKit`, it needs to remain `globalThis`

packages/lavapack/src/runtime-cjs-template.js
```diff
@@ -63,7 +73,7 @@
     const moduleObject = { exports: {} }
     const evalKit = {
       scopeTerminator: strictScopeTerminator,
-      globalThis,
+      globalRef,
```
kumavis:
```suggestion
      globalThis: globalRef,
```

packages/lavapack/src/runtime-cjs-template.js
```diff
@@ -1,18 +1,28 @@
...
 
   // polyfill globalThis
-  if (globalRef && !globalRef.globalThis) {
+  if (!globalRef.globalThis) {
     globalRef.globalThis = globalRef
   }
 
   // polyfill node/browserify's globalRef
-  globalThis.global = globalThis
+  if (!globalRef.global) {
```
kumavis:
```suggestion
  if (globalRef.global !== globalRef) {
```

packages/lavapack/src/runtime-cjs-template.js
```diff
@@ -1,18 +1,28 @@
 ;(function() {
 
   // identify the globalRef
-  const globalRef = (typeof globalThis !== 'undefined') ? globalThis : (typeof self !== 'undefined') ? self : (typeof global !== 'undefined') ? global : undefined
+  let globalRef = globalThis
```
kumavis: this would fail under strict mode

packages/lavapack/src/runtime-cjs-template.js
legobeat:
```suggestion
      globalThis: globalRef,
```

packages/lavapack/src/runtime-cjs-template.js
kumavis: to answer your question, yes. this was a mistake so we should use the `globalRef` reference in the places where we intend to use the global obj

packages/lavapack/src/runtime-cjs-template.js
```diff
@@ -26,7 +36,7 @@
     runModule: Object.freeze(runModule),
   })
 
-  Object.defineProperty(globalThis, 'LavaPack', {value: LavaPack})
+  Object.defineProperty(globalRef, 'LavaPack', {value: LavaPack})
```
kumavis, line 39: :+1:

packages/lavapack/src/runtime-cjs-template.js
legobeat: Yappari ne

legobeat: Amended to address this and other review comments.

packages/lavapack/src/runtime-cjs-template.js
legobeat: @naugtur 
[y/n]: y
Changed back again, hopfully everything in place this round ;)
https://github.com/LavaMoat/LavaMoat/pull/461#discussion_r1144151584

weizman (approved):

PR #464, opened 2023-03-07
Avoid calling crashing descriptors in sandboxed frames

weizman:
Discovered by @FrederikBolding , LavaMoat runtime causes an exception when is executed on a realm of a frame with a "sandbox" attribute. That's because LavaMoat tries to call all descriptors getters it can find of the global object, and that crashes for shared storage APIs in sandboxed frames due to security.

For those specific descriptors, we don't need that - this PR ignores those props.

### DIFF
--- a/packages/browserify/examples/01-simple-js/build.js
+++ b/packages/browserify/examples/01-simple-js/build.js
@@ -4,6 +4,7 @@ const browserify = require('browserify')
 // configure LavaMoat
 const lavamoatOpts = {
   writeAutoPolicy: false,
+  sandboxedIframeMode: false,
   scuttleGlobalThis: true,
   scuttleGlobalThisExceptions: ['print', /HTML[a-zA-Z]*Element/, 'prompt'],
 }
--- a/packages/browserify/src/index.js
+++ b/packages/browserify/src/index.js
@@ -119,6 +119,7 @@ function getConfigurationFromPluginOpts (pluginOpts) {
   }
 
   const nonAliasedOptions = [
+    'sandboxedIframeMode',
     'scuttleGlobalThis',
     'scuttleGlobalThisExceptions',
     'bundleWithPrecompiledModules',
@@ -158,6 +159,7 @@ function getConfigurationFromPluginOpts (pluginOpts) {
     pruneConfig: Boolean(pluginOpts.prunePolicy),
     debugMode: Boolean(pluginOpts.debugMode),
     statsMode: Boolean(pluginOpts.statsMode),
+    sandboxedIframeMode: Boolean(pluginOpts.sandboxedIframeMode),
     scuttleGlobalThis: pluginOpts.scuttleGlobalThis,
     scuttleGlobalThisExceptions: pluginOpts.scuttleGlobalThisExceptions,
     writeAutoPolicy: Boolean(pluginOpts.writeAutoPolicy || pluginOpts.writeAutoPolicyDebug),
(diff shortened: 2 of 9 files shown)


weizman:
Instead of having a predefined list of descriptors to ignore, I can just wrap the attempts to call all getters with a try-catch if that sounds better - open for discussion (performance-wise I prefer avoiding try-catch)

legobeat:
As a middle-path, how would it look to make the list of ignored descriptors configurable rather than hardcoded?

weizman:
I feel that's too specific too be configurable, no one should ever need to disable a simple call to a getter, we only introduce this here due to a very specific sandboxed-frames reason which introduces a finite list of descriptors to ignore.

weizman:
I do however feel better about introducing a new boolean config `sandboxedIframeMode` that when is `true` will make sure the runtime is adjusted to work within a sandboxed iframe (`false` by default)

legobeat:
> I feel that's too specific too be configurable, no one should ever need to disable a simple call to a getter, we only introduce this here due to a very specific sandboxed-frames reason which introduces a finite list of descriptors to ignore.

It's specific for these specific runtimes and environment. Lavapack may be used in other contexts where the same mechanisms apply for other global values?

Just noting that currently the API for lavamoat/core is fairly general over runtimes where as-is this introduces browser-specific semantics (`sandboxedIframeMode`) where I think the same end-result can be achieved without either having to do that or making the implementation more complex.

naugtur:
Can we benchmark the difference with try-catch and see if the fear for performance is legitimate? It'd be best, in case any other throwing entities happen to be present, to ignore all errors that happen on property lookup and definition. That'd be my default solution.

I agree this is too specific to expose in configuration, but iframe sandbox is only one reason to have things throw. There may be more. At this time I can't say I like the logic of any other solution that's not a try-catch.

If performance proves to be an issue, we could have a short list of known offenders and run the entire loop in a single try-catch, and if anything throws - rerun with try-catch on each item. That way we'd only punish environments we didn't expect.

weizman:
try-catch is actually not relevant anymore since the sandboxedIframe problem introduced 2 issues, and the solution to the 2nd one is to cancel these predefined descriptors from being endowed to compartments ([here](https://github.com/LavaMoat/LavaMoat/pull/464/files#diff-cbec1016c903fe52b8d76860ebf3400e172b83a3374b8796bab149b3721d8064R465)).

So I think we'd have to consider the other 2 options or maybe a new 3rd one.

weizman:
I personally feel ok about having a sanboxed-iframe specific solution, because in the future we might need to do other magic that might be more complicated in this context. I wish we could have created super agnostic code as @legobeat wishes, but I don't think that's very easy based on LM's current architecture (`core` "should" be platform agnostic, but it takes care of the things that change between platforms so... 🤷 )

legobeat:
Food for thought: Bad performance can be improved; bad APIs tend can hold you hostage. (*"in the future we might need to do other magic"* makes me even more cautious here)

> try-catch is actually not relevant anymore since the sandboxedIframe problem introduced 2 issues

Is there a way to understand what these two issues and their implications are?

> (core "should" be platform agnostic, but it takes care of the things that change between platforms so... shrug )

Be as it may with there being some platform-specific internals in `core`, exposing those internals to the user is a different matter. What I was proposing would be along the lines of substituting `sandboxedIframeMode: bool` with something like `mutableGlobals: string[]` - seems like it would solve the same (and potentially future other) problem as easily without molding the API for this particular hack.

weizman:
Update: I worked with @FrederikBolding on this again and I can't seem to reproduce the problem I had that got me into creating this PR. This means that try-catch might be enough for this after all.

Therefore I'm moving the fix to https://github.com/LavaMoat/LavaMoat/pull/468 and closing this (in case the problem pops up again, we'll just reopen this)

PR #465, opened 2023-03-07
Build runtime

weizman:


packages/lavapack/src/runtime.js
```diff
@@ -93,13 +93,19 @@
     } = __lavamoatSecurityOptions__
 
     // identify the globalRef
-    const globalRef = (typeof globalThis !== 'undefined') ? globalThis : (typeof self !== 'undefined') ? self : (typeof global !== 'undefined') ? global : undefined
+    let globalRef = globalThis
+
     if (!globalRef) {
-      throw new Error('Lavamoat - unable to identify globalRef')
+      globalRef = self || global
```
kumavis, line 99: referencing undeclared vars can lead to failures in strict mode

weizman: we already had "unsafe" access to `self` and `global` before this change #459

PR #466, opened 2023-03-08
core: fix globalThis polyfill

legobeat:
Motivation: https://github.com/LavaMoat/LavaMoat/pull/459#issuecomment-1457347132

> we are now uttering possibly undefined vars without a typeof operator. this would throw in strict mode. 

This reverts to using `typeof`

### DIFF
--- a/packages/core/src/kernelTemplate.js
+++ b/packages/core/src/kernelTemplate.js
@@ -25,13 +25,15 @@
     // identify the globalRef
     let globalRef = globalThis
 
-    if (!globalRef) {
-      globalRef = self || global
-      if (!globalRef) {
-        throw new Error('Lavamoat - globalThis not defined')
-      }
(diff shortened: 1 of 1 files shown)


packages/core/src/kernelTemplate.js
```diff
@@ -25,13 +25,15 @@
     // identify the globalRef
     let globalRef = globalThis
 
-    if (!globalRef) {
-      globalRef = self || global
-      if (!globalRef) {
-        throw new Error('Lavamoat - globalThis not defined')
-      }
+    if (typeof globalRef === 'undefined') {
+      globalRef = typeof self !== 'undefined' ? self : global
```
legobeat:
```suggestion
      globalRef = typeof self !== 'undefined' ? self : (typeof global !== 'undefined' ? global : undefined) 
```

kumavis (approved):

PR #467, opened 2023-03-08
Runtime browser support fix and a test for runtime ecma version

naugtur:
updated to not go as far as 2019 as we will no longer need to support it.

### DIFF
--- a/packages/lavapack/package.json
+++ b/packages/lavapack/package.json
@@ -24,6 +24,7 @@
     "ava": "^3.15.0",
     "eslint-plugin-ava": "^11.0.0",
     "eslint-plugin-node": "^11.1.0",
+    "espree": "^7.3.0",
     "mississippi": "^4.0.0",
     "source-map": "^0.7.3"
   },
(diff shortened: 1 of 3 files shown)


packages/lavapack/package.json
```diff
@@ -32,7 +33,7 @@
     "lint:eslint": "eslint .",
     "lint:fix": "eslint . --fix",
     "lint:deps": "depcheck",
-    "test": "ava --timeout=30s test/index.js",
+    "test": "ava --timeout=30s test/index.js && node test/runtimeEcma.js",
```
weizman: I recommend exporting to a separate yarn command and then have another main one that combines the two

packages/lavapack/test/runtimeEcma.js
```diff
@@ -0,0 +1,18 @@
+const ECMA_VERSION = 2019;
```
weizman: consider making this a sys.env var?

weizman (approved):
cool stuff

packages/lavapack/test/runtimeEcma.js
legobeat: @naugtur my 5c on the question "do we still want to keep this but change it to 2020?":

Having a compliance test like introduced here is great. Though I'd also like to put this in a more centralized (but still static!) place. How about adding a non-standard field in `package.json` for it?

naugtur: I explicitly meant for this to be defined in the test - the way tests usually are redundant so that they specify the same assumptions, but independently. That's the value of having tests and that's why I don't want to read the number from eslint config or elsewhere. This test must fail when/if we go above es2020

packages/lavapack/package.json
naugtur: ok

legobeat (approved):

packages/lavapack/test/runtimeEcma.js
legobeat: The in the current version does make it simpler  :+1:

PR #468, opened 2023-03-09
Wrap getters calling with try catch to avoid unhandled errors

weizman:
Calling getters might throw and we need to catch there errors so the program will continue to execute ([ctx](https://github.com/LavaMoat/LavaMoat/pull/464#issue-1612466023))

### DIFF
--- a/packages/core/src/kernelCoreTemplate.js
+++ b/packages/core/src/kernelCoreTemplate.js
@@ -374,7 +374,10 @@
         const descriptors = Object.getOwnPropertyDescriptors(source)
         Object.values(descriptors).forEach(desc => {
           if ('get' in desc) {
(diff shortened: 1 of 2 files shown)


FrederikBolding:
I can confirm that this fixed the issues I was having when executing in a sandboxed iframe.

naugtur:
Previously when we discussed this, try-catch in a loop was rejected because of performance impact. Should we benchmark this to actually know?

weizman:
I don't think we'll see any actual difference. It's an iteration of around 1000~ descriptors, nothing too crazy.
I ran a quick benchmark on Chrome where I ran the entire `runtime.js` with and without try-catch and there was no difference whatsoever.

naugtur (approved):

PR #469, opened 2023-03-12
[WIP] Enable more scuttling than just globalThis on demand

weizman:


weizman:
Ended up being irrelevant but might be useful one day

PR #470, opened 2023-03-13
Support Node builtins in `lavamoat-browserify`

FrederikBolding:
Support Node builtins in `lavamoat-browserify` by providing them in `loadModuleData`.

Potential upstream from getting snaps execution environments running in LavaMoat. Pending feedback.

### DIFF
--- a/packages/lavapack/package.json
+++ b/packages/lavapack/package.json
@@ -16,6 +16,7 @@
     "json-stable-stringify": "^1.0.2",
     "lavamoat-core": "^14.1.0",
     "readable-stream": "^3.6.0",
+    "resolve": "^1.22.1",
     "through2": "^4.0.2",
     "umd": "^3.0.3"
   },
--- a/packages/lavapack/package.json
+++ b/packages/lavapack/package.json
@@ -16,7 +16,6 @@
     "json-stable-stringify": "^1.0.2",
     "lavamoat-core": "^14.1.0",
     "readable-stream": "^3.6.0",
-    "resolve": "^1.22.1",
     "through2": "^4.0.2",
     "umd": "^3.0.3"
   },
--- a/packages/lavapack/src/runtime-template.js
+++ b/packages/lavapack/src/runtime-template.js
@@ -52,7 +52,7 @@
 
   function loadModuleData (moduleId) {
     // eslint-disable-next-line node/global-require
-    if (typeof window === 'undefined' && require('node:module').isBuiltin(moduleId)) {
+    if (typeof window === 'undefined' && typeof require !== 'undefined' && require('node:module').isBuiltin(moduleId)) {
       return {
         type: 'builtin',
         package: moduleId,
--- a/packages/lavapack/src/runtime-template.js
+++ b/packages/lavapack/src/runtime-template.js
@@ -51,6 +51,7 @@
 
 
   function loadModuleData (moduleId) {
+    // eslint-disable-next-line node/global-require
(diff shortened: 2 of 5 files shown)


packages/lavapack/package.json
```diff
@@ -16,6 +16,7 @@
     "json-stable-stringify": "^1.0.1",
     "lavamoat-core": "^14.0.0",
     "readable-stream": "^3.6.0",
+    "resolve": "^1.22.1",
```
naugtur: resolve uses https://socket.dev/npm/package/is-core-module internally. If we don't need other functionality of resolve elsewhere, let's minimize dependency. It's also easier to update that way.

yarn.lock
```diff
@@ -12279,7 +12281,7 @@ resolve@^1.1.4, resolve@^1.10.0, resolve@^1.14.2, resolve@^1.15.1, resolve@^1.17
     path-parse "^1.0.7"
     supports-preserve-symlinks-flag "^1.0.0"
 
-resolve@^1.10.1:
+resolve@^1.10.1, resolve@^1.22.1:
```
naugtur: There was another version of resolve used in the project. We could unify versions if we keep using resolve instead of is-core-module.

packages/lavapack/src/runtime-template.js
```diff
@@ -51,6 +51,19 @@
 
 
   function loadModuleData (moduleId) {
+    // eslint-disable-next-line node/global-require
+    if (typeof window === 'undefined' && require('resolve').isCore(moduleId)) {
```
FrederikBolding: I'm realizing now that this won't work with a standalone bundle since `resolve` is not going to be available. Could we bundle `is-core-module` with the runtime? 🤔 @naugtur

packages/lavapack/package.json
FrederikBolding: See https://github.com/LavaMoat/LavaMoat/pull/470/files#r1135433710

packages/lavapack/src/runtime-template.js
FrederikBolding: An alternative would be to use `node:module`: https://nodejs.org/api/module.html#moduleisbuiltinmodulename

Perhaps a cleaner solution?

FrederikBolding: I've pushed a commit that uses the Node.js API for this, it does require at least Node 16 though

legobeat: It's possible browserify is doing something I'm overlooking here but...
https://github.com/LavaMoat/LavaMoat/actions/runs/4415669229/jobs/7738939915#step:6:367

`packages/lavapack/src/runtime-template.js` is used for ES modules, where `require` is [not available](https://nodejs.org/dist/latest-v18.x/docs/api/esm.html#require).

I think CommonJS require overriding would be done here: https://github.com/LavaMoat/LavaMoat/blob/main/packages/lavapack/src/runtime-cjs-template.js#L80h

Nodejs 14 is going EoL at the end of next month, so dropping support for it then doesn't seem problematic to me.

FrederikBolding: I'm not sure how we could support ESM imports with this since dynamic ESM imports are async 🤔

FrederikBolding: I have disabled the Node.js builtin resolution if `require` is unavailable, not sure if there is another way we can make this happen

PR #471, opened 2023-03-20
Fix test scenario creation

weizman:
A long time ago I added `additionalOpts` to move across arguments such as scuttling related stuff.
It was a mistake, these arguments can (and should) be integrated into the `scenario` object.
This PR fixes this (which will also help with #462)


legobeat (approved):
lgtm!

PR #472, opened 2023-03-21
Snow scuttling integration updates (#462)

legobeat:
Targeting #462 

- backwards compatibility (https://github.com/LavaMoat/LavaMoat/pull/462#discussion_r1143108260)
- make "preloader" (e.g. SNOW) pluggable
  - choice of `onLoad` for new field is tentative and up for alternatives

### DIFF
--- a/packages/core/src/kernelCoreTemplate.js
+++ b/packages/core/src/kernelCoreTemplate.js
@@ -79,19 +79,17 @@
     const {
       enabled: scuttleGlobalThisEnabled,
       exceptions: scuttleGlobalThisExceptions,
-      recursive: scuttleGlobalThisRecursive,
+      onLoad: onLoadHandle,
     } = scuttleGlobalThis === true ? scuttleGlobalThisDefaults : scuttleGlobalThis
 
-    let snow = (cb, win) => cb(win)
-    if (scuttleGlobalThisRecursive) {
-      if (!globalRef.SNOW) {
+    let onLoad = (cb, global) => cb(global)
+    if (onLoadHandle) {
+      if (typeof globalRef[onLoadHandle] !== 'function') {
         throw new Error(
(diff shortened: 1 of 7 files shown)


packages/core/src/kernelCoreTemplate.js
```diff
@@ -79,19 +79,17 @@
     const {
       enabled: scuttleGlobalThisEnabled,
       exceptions: scuttleGlobalThisExceptions,
-      recursive: scuttleGlobalThisRecursive,
+      onLoad: onLoadHandle,
```
kumavis, line 82: global export name?

packages/lavapack/src/runtime.js
```diff
@@ -11152,19 +11152,17 @@ module.exports = {
...
       enabled: scuttleGlobalThisEnabled,
       exceptions: scuttleGlobalThisExceptions,
-      recursive: scuttleGlobalThisRecursive,
+      onLoad: onLoadHandle,
     } = scuttleGlobalThis === true ? scuttleGlobalThisDefaults : scuttleGlobalThis
 
-    let snow = (cb, win) => cb(win)
-    if (scuttleGlobalThisRecursive) {
-      if (!globalRef.SNOW) {
+    let onLoad = (cb, global) => cb(global)
```
kumavis, line 11158: convention is to have the callback be the last argument

packages/lavapack/src/runtime.js
```diff
@@ -11152,19 +11152,17 @@ module.exports = {
...
+    if (onLoadHandle) {
```
kumavis:
```suggestion
    if (onLoadHandle !== undefined) {
```

kumavis (commented):
maybe out of scope for this PR - `performScuttleGlobalThis` should be broken out into a package and imported in here, not implemented here

we want to use this cross platform, eg in nodejs and so we dont want the implementation in this specific runtime

PR #474, opened 2023-03-24
Skip JSON files only for policy generation

FrederikBolding:
Inverts the JSON file extension check to only check for `.json`. This allows for parsing of `.cjs`, `.ts` etc.

### DIFF
--- a/packages/core/src/generatePolicy.js
+++ b/packages/core/src/generatePolicy.js
@@ -97,7 +97,7 @@ function createModuleInspector (opts = {}) {
     // skip json files
     const filename = moduleRecord.file || 'unknown'
     const fileExtension = path.extname(filename)
-    if (fileExtension === '.json') {
+    if (!fileExtension.match(/^\.([cm]?js|ts)$/)) {
       return
     }
     // get ast (parse or use cached)
--- a/packages/core/src/generatePolicy.js
+++ b/packages/core/src/generatePolicy.js
@@ -97,7 +97,7 @@ function createModuleInspector (opts = {}) {
     // skip json files
     const filename = moduleRecord.file || 'unknown'
     const fileExtension = path.extname(filename)
-    if (fileExtension !== '.js') {
+    if (fileExtension === '.json') {
       return
     }
     // get ast (parse or use cached)


packages/core/src/generatePolicy.js
```diff
@@ -97,7 +97,7 @@ function createModuleInspector (opts = {}) {
     // skip json files
     const filename = moduleRecord.file || 'unknown'
     const fileExtension = path.extname(filename)
-    if (fileExtension !== '.js') {
+    if (fileExtension === '.json') {
```
legobeat: @FrederikBolding What do you think of still doing an allowlist approach but expanding it?

```suggestion
    if (!fileExtension.match(/^\.(cjs|jsx?|mjs|tsx?)$/)) {
```

FrederikBolding: I don't have a strong opinion on that. But @naugtur recommended that we just invert to an equality check. Either way is fine by me.

naugtur: I think we should avoid JSON as described in the comment above and allow everything else by default if it's imported in your bundle.
Also, this change reduces what's skipped in MetaMask-extension too, because there are some cjs files there now that we had to allow manually in override.

Then we could make sure errors in parsing are only logged as warnings for extensions different than the list of known and expected javascript extensions. I wouldn't put ts there tho. So if someone misconfigures stuff, we'd throw, but only if ts files are actual typescript, not extension left on a processed file.

@legobeat does this sound right?

legobeat: It still makes more sense to me to only parse files with known endings (and if there is a mismatch like `.ts` files actually being transpiled js, this is up to the user to resolve).

If there exist files that we know should not be parsed and we have a heuristic for it based on file-endings (and here it does, in this case `*.json`), it seems dangerous/undefined to parse files with arbitrary endings.

I was also hesitant to put `.jsx` and `.tsx` in the regex above - those should probably go and the resulting `[".cjs",".js",".mjs"]` (any additions?) would be short enough to be a reasonable `.includes`.

But if there's some complication I'm missing, don't let that be a blocker.

FrederikBolding: From a snaps point of view, the problem is that Browserify/babelify doesn't change the file extension of transpiled files. So this check fails. If we go with an allowlist approach that doesn't include `.ts` we will also have the same problem long-term in the MetaMask extension with more and more code being migrated to TypeScript.

I'm agnostic whether we choose the allowlist approach or simply ignore JSON. But we need a solution for parsing the Browserify `.ts` output regardless.

This could perhaps be handled by the LavaMoat browserify plugin before generating the policy, but that seems even more complicated.

legobeat:
```suggestion
    if (!fileExtension.match(/^\.([cm]?js|ts)$/)) {
```

legobeat (approved):

packages/core/src/generatePolicy.js
naugtur: Since we are not failing the entire process on parse failure and the files listed for parsing are coming from what's actually required not from scanning the whole folder, we're better off looking at too many over too few files. 

Is .tsx going to show up the same way or are these files renamed?

naugtur (approved):

PR #475, opened 2023-03-26
Add scuttler prop instead of fixed Snow integration

weizman:
Instead of `recursive` pass `scuttler` which will be a string for a reference on the global object for a scuttler function.
Also, a demo of how to use this option to scuttle all realms using snow.

TODO
- [x] make sure its clear this is experimental
- [x] document this experimental option
- [x] make sure node env can't pass this option (irrelevant in node)


packages/core/src/kernelCoreTemplate.js
```diff
@@ -117,7 +116,9 @@
         const flags = parts[parts.length - 1]
         scuttleGlobalThisExceptions[i] = new RegExp(pattern, flags)
       }
-      snow(realm => performScuttleGlobalThis(realm, scuttleGlobalThisExceptions), globalRef)
+      scuttler(globalRef, {
+        scuttle: realm => performScuttleGlobalThis(realm, scuttleGlobalThisExceptions),
```
legobeat: previously we had `(cb, realm)`; here we have `(realm, {scuttle:cb})`.

I would prefer a more typical `(realm, cb)` and do away with wrapping the function in an object. If we ever need to introduce more extensibility, I would see options as a separate argument, like `(realm, cb, {})`

packages/core/src/kernelCoreTemplate.js
```diff
@@ -74,24 +74,23 @@
...
       enabled: true,
       exceptions: [],
-      recursive: false,
+      scuttler: '',
     }
     const {
       enabled: scuttleGlobalThisEnabled,
       exceptions: scuttleGlobalThisExceptions,
-      recursive: scuttleGlobalThisRecursive,
+      scuttler: scuttleGlobalThisScuttler,
```
legobeat: TBH I find this block (`:79-83`) a bit terse and tricky to parse and my eyes keep jumping back and forth especially now that `scuttleGlobalThisScuttler` is a field that is not "the final word" in this scope anymore. 

I'd personally find it more readable to have a local options hash instead of local variables for each option.

Something like:
```
const scuttleOpts = scuttleGlobalThis === true
  ? scuttleGlobalThisDefaults
  : {
    ...scuttleGlobalThis,
    _scuttler: scuttleGlobalThis?.scuttler && globalRef[scuttleGlobalThis.scuttler] || scuttleGlobalThisDefaults._scuttler
  }
```

And then in the following local scope we would refer to e.g. `scuttleOpts.enabled` instead of `scuttleGlobalThisEnabled`. We would also be calling `scuttleOpts._scuttler`, here prefixed with an underscore to distinguish the function-referencing key to the name-string referencing one in the options.


Might just be that I'm not fully accustomed to the new syntax here, just noting.

packages/core/src/kernelCoreTemplate.js
weizman: `(cb, realm)` was proposed originally because the only "option" was to use snow, which accepts `(cb, realm)` because `cb` is required while `realm` is optional.
Now that we go with a more generic option that just snow, it makes sense to go by standard which is first `realm` and `cb` after.

I do agree with `(realm, cb, {})` however, that's a fair suggestion.

weizman: (You can see what I mean btw by visiting `packages/browserify/examples/01-simple-js/index.html:5` on this PR)

weizman: aea3bbc57

packages/core/src/kernelCoreTemplate.js
weizman: I think it's really a matter of taste. I'm mostly trying to remain as align as possible with the rest of the codebase previous choices.
You can lookup `} =` in the codebase - this style is all over pretty much.

leotm:
i wasn't able to test our example at first ^ so updated our examples separately as follow-up in
- https://github.com/LavaMoat/LavaMoat/pull/476

(both console errors from before unrelated to scuttler so looks good)

leotm:
GitHub double-posted comment and giving 500 trying to delete

packages/core/src/kernelCoreTemplate.js
legobeat: Sure, I guess my main point is that with the changes in this PR, to answer the question "what is the effective scuttler", the reader will need walk up through lines 93,85,83,77,63 and make the connection that the falsiness of the empty string in defaults changes the outcome in the default scenario. (or walk down and keep this in mind at `:87`).

In the proposed alternative above, the logic for the specific field is kept at the same location.

weizman: I accepted parts of your suggestion (left the part where an exception is thrown if a scuttler was specified but was not found) 6fe045adc

weizman:
added support restriction for `scuttler` in node env e01281dc2

weizman:
added documentation PR https://github.com/LavaMoat/LavaMoat/pull/477

leotm (approved):
lgtm, but would wait on @legobeat

packages/browserify/examples/01-simple-js/index.html
```diff
@@ -1,7 +1,15 @@
 <!DOCTYPE HTML>
 <html lang="en">
 <head>
-    <script src="https://cdn.jsdelivr.net/npm/@lavamoat/snow/snow.js"></script>
+    <script>
+      Object.defineProperty(window, 'SCUTTLER', {
```
leotm: just an afterthought

```suggestion
      Object.defineProperty(globalThis, 'SCUTTLER', {
```

altho same result, this might pos confuse users getting started less

since we talk about `globalThis` alot (same in SES Compartment docs)
- original pr https://github.com/LavaMoat/LavaMoat/pull/360 where we intro `scuttleGlobalThis` (not `scuttleGlobalWindow`)
- recently https://github.com/LavaMoat/LavaMoat/pull/459
- in new docs https://github.com/LavaMoat/LavaMoat/pull/477 also mentioning original pr

then again i could be wrong, since users might expect `window` since in browser env

curious on your thoughts

packages/browserify/examples/01-simple-js/build.js
```diff
@@ -4,7 +4,11 @@ const browserify = require('browserify')
 // configure LavaMoat
 const lavamoatOpts = {
   writeAutoPolicy: false,
-  scuttleGlobalThis: true,
+  scuttleGlobalThis: {
+    enabled: true,
+    exceptions: ['prompt'],
+    scuttler: 'SCUTTLER',
+  },
```
leotm: a second afterthought, i know scuttling was introduced before my time ([nov22](https://github.com/LavaMoat/LavaMoat/commit/e90070f2aa8265e5b7e0e90a8d3f3d54f7b0d714#diff-ef5b28753cd110237bb71c073c19be3c94803ad7f42501cd4ae063a96dc0454d))
but since scuttling's an optional and experimental ft
might our simple-js example make more sense stable with scuttling off?
then we could have e.g. `02-simple-js-scuttling` for it on
unless here to promote the bleeding edge :feelsgood:

packages/core/src/kernelCoreTemplate.js
```diff
@@ -74,24 +74,19 @@
...
-      enabled: scuttleGlobalThisEnabled,
-      exceptions: scuttleGlobalThisExceptions,
-      recursive: scuttleGlobalThisRecursive,
-    } = scuttleGlobalThis === true ? scuttleGlobalThisDefaults : scuttleGlobalThis
-
-    let snow = (cb, win) => cb(win)
-    if (scuttleGlobalThisRecursive) {
-      if (!globalRef.SNOW) {
+
+    const scuttleOpts = scuttleGlobalThis === true ? scuttleGlobalThisDefaults : scuttleGlobalThis
```
legobeat: This will still potentially modify element in the `exceptions` property of the supplied `scuttleGlobalThis` down at https://github.com/LavaMoat/LavaMoat/pull/475/files#diff-cbec1016c903fe52b8d76860ebf3400e172b83a3374b8796bab149b3721d8064R113.

Hence preferable to consolidate the instantion and transformations of options with a pattern like
```
const scuttleOpts = {
  // [...]
  scuttlerFunc: scuttleGlobalThis?.scuttler ? ///....,
  exceptions: scuttleGlobalThis?.exceptions?.map(e => scuttleExceptionToRegexp(e)) || scuttleGlobalThisDefaults.exceptions,
  
}
```

(the same input validation can still be done, just moved to happen before (or if there is reason to do so, after) the creation of `scuttleOpts`)

packages/lavapack/src/runtime.js
```diff
@@ -11147,24 +11147,19 @@ module.exports = {
...
-      enabled: scuttleGlobalThisEnabled,
-      exceptions: scuttleGlobalThisExceptions,
-      recursive: scuttleGlobalThisRecursive,
-    } = scuttleGlobalThis === true ? scuttleGlobalThisDefaults : scuttleGlobalThis
-
-    let snow = (cb, win) => cb(win)
-    if (scuttleGlobalThisRecursive) {
-      if (!globalRef.SNOW) {
+
+    const scuttleOpts = scuttleGlobalThis === true ? scuttleGlobalThisDefaults : scuttleGlobalThis
```
legobeat: Same as https://github.com/LavaMoat/LavaMoat/pull/475/files#r1152703084

packages/browserify/examples/01-simple-js/build.js
legobeat: Agreed that `01-simple-js` should not make use of experimental APIs. Adding this variation as a new example sounds good.

packages/core/src/kernelCoreTemplate.js
```diff
@@ -102,22 +97,22 @@
...
-      if (!Array.isArray(scuttleGlobalThisExceptions)) {
-        throw new Error(`LavaMoat - scuttleGlobalThis.exceptions must be an array, got "${typeof scuttleGlobalThisExceptions}"`)
+    if (scuttleOpts.enabled) {
+      if (!Array.isArray(scuttleOpts.exceptions)) {
+        throw new Error(`LavaMoat - scuttleGlobalThis.exceptions must be an array, got "${typeof scuttleOpts.exceptions}"`)
       }
       // turn scuttleGlobalThis.exceptions regexes strings to actual regexes
-      for (let i = 0; i < scuttleGlobalThisExceptions.length; i++) {
-        const prop = scuttleGlobalThisExceptions[i]
+      for (let i = 0; i < scuttleOpts.exceptions.length; i++) {
```
kumavis: for (const prop of ...)

packages/core/src/kernelCoreTemplate.js
```diff
@@ -74,24 +74,19 @@
     const scuttleGlobalThisDefaults = {
       enabled: true,
       exceptions: [],
-      recursive: false,
+      scuttler: '',
```
kumavis: weird default

packages/node/src/kernel.js
```diff
@@ -15,6 +15,9 @@ const nativeRequire = require
 module.exports = { createKernel }
 
 function createKernel ({ projectRoot, lavamoatPolicy, canonicalNameMap, debugMode, statsMode, scuttleGlobalThis, scuttleGlobalThisExceptions }) {
+  if (scuttleGlobalThis?.scuttler) {
+    throw new Error('Lavamoat - "scuttleGlobalThis.scuttler" option is not supported for node environment.')
```
kumavis: why not?

packages/browserify/examples/01-simple-js/index.html
```diff
@@ -1,7 +1,15 @@
...
+        value: (globalRef, scuttle) => {
+          console.log('scuttler (start)');
```
kumavis: debug code?

packages/lavapack/src/runtime.js
```diff
@@ -11147,24 +11147,19 @@ module.exports = {
...
         throw new Error(
-          'LavaMoat - scuttleGlobalThis is configured to be enabled recursively. ' +
-          'For that, Snow-JS must be included before LavaMoat executes. ' +
+          `LavaMoat - 'scuttler' function "${scuttleOpts.scuttler}" expected on globalRef.` +
           'To learn more visit https://github.com/LavaMoat/LavaMoat/pull/462.',
         )
       }
-      snow = globalRef.SNOW
+      scuttleOpts.scuttlerFunc = globalRef[scuttleOpts.scuttler]
     }
```
kumavis: maybe should create a configuration obj that holds the scuttlerFunc instead of mutating the options obj

packages/lavapack/src/runtime.js
```diff
@@ -11147,24 +11147,19 @@ module.exports = {
     const scuttleGlobalThisDefaults = {
       enabled: true,
       exceptions: [],
-      recursive: false,
+      scuttler: '',
```
kumavis: should rename to scuttlerName or something

packages/browserify/examples/01-simple-js/index.html
weizman: nope, just wanted to demonstrate the control you get in the scuttler process when you configure it (seems legit to use in an example)

packages/browserify/examples/01-simple-js/build.js
weizman: 3cec4e1ed

packages/lavapack/src/runtime.js
weizman: 28ff29b5a

packages/core/src/kernelCoreTemplate.js
weizman: bb15afce1

packages/lavapack/src/runtime.js
weizman: bb15afce1

packages/lavapack/src/runtime.js
weizman: bb15afce1

packages/core/src/kernelCoreTemplate.js
weizman: this prop is expected to be a string, this is the default value of type:string (thought it would make sense to stick to one type if possible)

packages/core/src/kernelCoreTemplate.js
```diff
@@ -77,7 +77,7 @@
       scuttlerName: '',
     }
 
-    const scuttleOpts = scuttleGlobalThis === true ? scuttleGlobalThisDefaults : scuttleGlobalThis
+    const scuttleOpts = Object.assign({}, scuttleGlobalThis === true ? scuttleGlobalThisDefaults : scuttleGlobalThis)
```
legobeat: Why not use the pattern as [here](https://github.com/LavaMoat/LavaMoat/pull/475#discussion_r1152703084) instead?

This isn't a deep copy, so for example the input `scuttleOpts.exceptions` still get mutated [here](https://github.com/LavaMoat/LavaMoat/pull/475/files/28ff29b5ae781bf3209b524934df2fe3e32810ce..bb15afce16a2e997e04e62bf7256c257959329e6#diff-cbec1016c903fe52b8d76860ebf3400e172b83a3374b8796bab149b3721d8064L113).

packages/lavapack/src/runtime.js
```diff
@@ -11150,7 +11150,7 @@ module.exports = {
       scuttlerName: '',
     }
 
-    const scuttleOpts = scuttleGlobalThis === true ? scuttleGlobalThisDefaults : scuttleGlobalThis
+    const scuttleOpts = Object.assign({}, scuttleGlobalThis === true ? scuttleGlobalThisDefaults : scuttleGlobalThis)
```
legobeat: Same as above, by https://github.com/LavaMoat/LavaMoat/pull/475/files/28ff29b5ae781bf3209b524934df2fe3e32810ce..bb15afce16a2e997e04e62bf7256c257959329e6#diff-69e3f93883cb0ad47e2b217ba9e92eb890d8855e05d1f6ac66c2051bb636cbdaR11186

packages/node/src/kernel.js
weizman: * we decided to allow using "scuttlerName" config option in node env
* since it's experimental, we decided to allow using it similar to how it's used for browser envs (by defining a prop on globalRef)
* later on we'd like to allow passing an actual reference instead of a string prop
* fix 6bd2c5230

packages/core/src/kernelCoreTemplate.js
weizman: you're right, my bad, thank you

https://github.com/LavaMoat/LavaMoat/pull/482

packages/lavapack/src/runtime.js
weizman: you're right, my bad, thank you

https://github.com/LavaMoat/LavaMoat/pull/482

PR #478, opened 2023-04-11
Update allow scripts

leotm:
- Support Yarn 3+ (Berry) via plugin
- Add `list` to CLI help menu
- Update console log for `--experimental-bins`


.eslintrc.json
```diff
@@ -48,6 +48,6 @@
     }
   ],
   "env": {
-    "es6": true
+    "es2020": true
```
legobeat, line 51: What's the motivation here? If reasonable it seems preferrable to keep backwards-compatatibility for this package.

packages/allow-scripts/src/yarn-berry-plugin.cjs
```diff
@@ -0,0 +1,20 @@
+// @ts-check
+
+const { execute } = require( "@yarnpkg/shell");
```
legobeat:
```suggestion
const { execute } = require('@yarnpkg/shell');
```

legobeat:
@leotm I know this is still draft but can we split the yarn berry changes in a separate PR from the CLI changes?

.eslintrc.json
boneskull: template literals, from looking at the changes.  lavamoat doesn't support any version of node that doesn't have template strings anyway.  `es2020` is equivalent to node.js v14.17.0+ and node 14.x is EOL anyway

legobeat:
@leotm just fyi i rebased this over at #766

leotm:
> @leotm just fyi i rebased this over at #766

thanks 🙈 closing here then regathering thoughts to finish over in 766

PR #482, opened 2023-04-20
refactor scuttler opts generation

weizman:
have in a single function that properly makes sure that:
* scuttle conf is regenerated into a new object to avoid mutating the original opts obj
* handle "enabled-by-default" where scuttleOpts === true
* set scuttlerFunc
* throw if scuttlerName is set but does not appear on globalRef

(ctx https://github.com/LavaMoat/LavaMoat/pull/475#discussion_r1172773280)

### DIFF
--- a/packages/core/src/kernelCoreTemplate.js
+++ b/packages/core/src/kernelCoreTemplate.js
@@ -71,24 +71,7 @@
     const { prepareCompartmentGlobalFromConfig } = templateRequire('makePrepareRealmGlobalFromConfig')(generalUtils)
     const { strictScopeTerminator } = templateRequire('strict-scope-terminator')
 
-    const scuttleGlobalThisDefaults = {
-      enabled: true,
-      exceptions: [],
-      scuttlerName: '',
-    }
-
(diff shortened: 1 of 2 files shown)


packages/core/src/kernelCoreTemplate.js
```diff
@@ -125,6 +97,40 @@
     Object.freeze(kernel)
     return kernel
 
+    function generateScuttleOpts(originalOpts) {
```
weizman, line 100: add comment to explain this

packages/lavapack/src/runtime.js
leotm, line 1: _nit: i guess should run `yarn build` again after last commit, but only comment updated, so safe to ignore_

Screenshot 2023-05-23 at 13 39 20

leotm: noob question: do we always update the (generated) `lavapack` build-runtime after changes in `core`? or is it also after changes in other packages

weizman: currently we do, we should however refactor LavaMoat to not include runtime.js as a static file in this repo, in the future it will be generated on demand and ideally be passed to consumer code instead of written to FS

weizman: (i'll rebuild runtime.js after merging this as this is being merged to another PR anyway)

PR #483, opened 2023-04-21
scorchwrap plugin beta

naugtur:
Beta RC for scorchwrap plugin.

 - [x] ~~Rebased on top of #699 - needs that to merge first.~~


kumavis:
impressive!

boneskull:
needs more LoC

packages/scorchwrap/README.md
```diff
@@ -0,0 +1,145 @@
+# ScorchWrap PoC
+
+> It's like stretchwrap, but with lava
```
boneskull: shrinkwrap?

packages/scorchwrap/README.md
```diff
@@ -0,0 +1,145 @@
...
+      policy: { ... LavaMoat policy structure ... }
+      runChecks: true,
+    //   diagnosticsVerbosity: 2,
+    }),
+  ],
+  // ... other webpack configuration properties
+};
+```
+
+One important thing to note when using the ScorchWrap plugin is that it disables the `concatenateModules` optimization in webpack. This is because concatenation won't work with wrapped modules.
```
boneskull: Why not?

packages/scorchwrap/README.md
```diff
@@ -0,0 +1,145 @@
...
+- diagnosticsVerbosity: Optional number property to represent diagnostics output verbosity. A larger number means more overwhelming diagnostics output. Default is 0.  
+  Setting positive verbosity will enable runChecks.
+
+```js
+const ScorchWrapPlugin = require('@lavamoat/scorchwrap');
+
+module.exports = {
+  // ... other webpack configuration properties
+  plugins: [
+    new ScorchWrapPlugin({
```
boneskull: (opinion) export a factory function `scorchWrap(opts)` intended to be the entry point (the class can be exported as well).  from the consumer's perspective, the fact that `ScorchWrapPlugin` is a class is an implementation detail that they needn't care about.  saves a few keystrokes too :P

packages/scorchwrap/README.md
```diff
@@ -0,0 +1,145 @@
...
+      policy: { ... LavaMoat policy structure ... }
+      runChecks: true,
```
boneskull: is this "paranoid" mode?  unclear if this has been implemented based on the language below

packages/scorchwrap/README.md
```diff
@@ -0,0 +1,145 @@
...
+  ],
+  // ... other webpack configuration properties
+};
+```
+
+One important thing to note when using the ScorchWrap plugin is that it disables the `concatenateModules` optimization in webpack. This is because concatenation won't work with wrapped modules.
+
+### Gotchas
+
+- Webpack will gladly add missing dependencies for node builtins like the `events` or `buffer` packages. For LavaMoat policy to handle it, the package needs to be explicitly listed in prod dependencies.
```
boneskull: listed in prod dependencies _where_--`package.json`?  what happens if the user does not explicitly list the package?  if it does anything other than "fails the build," I question the first point of the threat model, "webpack itself is considered trusted".  

IIRC, a webpack user can explicitly disable this behavior; is a plugin able to access this flag?  can `ScorchWrapPlugin` fail if this behavior is not disabled?

packages/scorchwrap/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+# Security Claims
+
+- SES must be added to the page without any bundling or transforming for any security guarantees to be sustained.
```
boneskull: - what does this look like?  another `<script>` tag?
- I am interested in understanding what bundlers/transformers do that breaks whatever they break, and if we have attempted to work around this or otherwise "solve" it

packages/scorchwrap/README.md
```diff
@@ -0,0 +1,145 @@
...
+- Each javascript module resulting from the webpack build is scoped to its package's policy
+
+Threat model
+
+- webpack itself is considered trusted
+- all plugins can bypass LavaMoat protections intentionally
+- It's unlikely but possible that a plugin can bypass LavaMoat protections unintentionally
+- It should not be possible for loaders to bypass LavaMoat protections
+- Some plugins (eg. MiniCssExtractPlugin) execute code from the bundle at build time. To make the plugin work you need to trust it and the modules it runs and add the ScorchWrap.ignore loader for them.
```
boneskull: what if we did some sort of post-bundle analysis?

packages/scorchwrap/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+### policy
+
+There's two ways we could handle policy
+1. inline into each module at compilation time 
+   We're working per-module, so this would produce a lot of duplication but eliminate the need to store entire policy with keys for each package.
+2. add to runtime, keyed  
+   For now holding on to the entire policy seems like a better idea, we could compress the keys at compile time easily.
+   
+Current implementation inlines the policy as-is into the bundle.
```
boneskull: I think I'm missing something about the use-case.  Say you want to prevent bundling up a previously-trusted dep which released a new version that's now malicious. Otherwise, your JS is static and you do no further execution of untrusted code.  

If you run analysis before bundling and the policy ends up denying something, why do we need a runtime component?  Is it because the analysis has limitations?  Is this not something we're solving for?  I can totally see someone grousing about a huge policy file exceeding somebody's perf budget...

packages/scorchwrap/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+
+
+### modes
+
+- default 
+- paranoid - adds extra checks
+
+Things it could check in paranoid mode
+- use other hooks to check if the number of bundle entries equals the number of entries processed by the wrapper
```
boneskull: and that they match, somehow?

packages/scorchwrap/README.md
```diff
@@ -0,0 +1,145 @@
...
+- add a local reference before the with blocks and a function within to look it up and alert if it can be found - should trigger if a transform or plugin action breaks the withproxies. This would be best implemented by the plugin injecting a tiny module with the source to do this.
```
boneskull: is it possible to tell webpack to always run this plugin "last"?  I suppose we'd need to defend against plugins running after.  though I imagine webpack itself does further stuff to the bundle that external/builtin plugins don't (I could be wrong!), so maybe it's unavoidable.

packages/scorchwrap/README.md
```diff
@@ -0,0 +1,145 @@
...
+- allow optionally importing a canary package that tries to escape (kinda overlaps with above, but could use more methods)
+- run an AST on the final bundle to identify the scopes without the `with` blocks (after all minitication etc.)
```
boneskull: this is related to the post-bundling analysis I suggested above

packages/scorchwrap/app/app.js
```diff
@@ -0,0 +1,14 @@
+import leftpad from "leftpad";
```
boneskull: :trollface:

packages/scorchwrap/app/index.html
```diff
@@ -0,0 +1,78 @@
+<!DOCTYPE html>
+<!--[if lt IE 7 ]> <html lang="en" class="ie6" <% if(htmlWebpackPlugin.files.manifest) { %> manifest="<%= htmlWebpackPlugin.files.manifest %>"<% } %>> <![endif]-->
+<!--[if IE 7 ]>    <html lang="en" class="ie7" <% if(htmlWebpackPlugin.files.manifest) { %> manifest="<%= htmlWebpackPlugin.files.manifest %>"<% } %>> <![endif]-->
+<!--[if IE 8 ]>    <html lang="en" class="ie8" <% if(htmlWebpackPlugin.files.manifest) { %> manifest="<%= htmlWebpackPlugin.files.manifest %>"<% } %>> <![endif]-->
+<!--[if IE 9 ]>    <html lang="en" class="ie9" <% if(htmlWebpackPlugin.files.manifest) { %> manifest="<%= htmlWebpackPlugin.files.manifest %>"<% } %>> <![endif]-->
+<!--[if (gt IE 9)|!(IE)]><!--> <html lang="en" class="" <% if(htmlWebpackPlugin.files.manifest) { %> manifest="<%= htmlWebpackPlugin.files.manifest %>"<% } %>> <!--<![endif]-->
```
boneskull: what's all this

packages/scorchwrap/app/index.html
```diff
@@ -0,0 +1,78 @@
...
+<script src="<%= htmlWebpackPlugin.files.chunks[chunk].entry %>"></script>
+<% } %>
+
+<% if (htmlWebpackPlugin.options.devServer) { %>
+<script src="<%= htmlWebpackPlugin.options.devServer%>/webpack-dev-server.js"></script>
+<% } %>
+
+<% if (htmlWebpackPlugin.options.googleAnalytics) { %>
+<script>
+  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
```
boneskull: isn't this illegal in the EU

boneskull (commented):
haven't looked at the code, really; just some questions. will review proper when no longer a draft

packages/scorchwrap/README.md
naugtur: lol, yes.

packages/scorchwrap/README.md
```diff
@@ -0,0 +1,163 @@
...
+ - [ ] ecosystem compatibility quirks we're yet to find
+
+*(for more specific issues to solve see TODO comments in code)*
+
+
+
+## Usage
+
+The ScorchWrap plugin takes an optional options object with the following properties:
+
```
weizman: missing `readableResourceIds`

packages/scorchwrap/README.md
weizman: Bundlers naturally introduce code you did not write which will be executed in addition to your code:

> your code:
```javascript
module.exports = function(arg) { alert(arg) }
```

> bundled code (example taken from GitHub):
Screenshot 2023-08-16 at 14 39 33

When trying to create a project that is sensitive to interceptions, you can't really allow introduction of code you did not audit, because it probably isn't taking such sensitivity into account which might compromise the security of your project.

In the ctx of the example above, no matter how secure the code you wrote is, I can still mess with your program by hooking `Array` prototype, because webpack introduces an unsafe usage of the `push` method:

```javascript
Object.defineProperty(Array.prototype, 'push', {value: () => { throw 'BREAK YOUR PROGRAM' }});
```

Exactly because of that we run our programs on top of SES, to solve just that (SES even hardens `Array` specifically), but it's still better practice to have full knowledge of what ends up running, because if the bundler introduces code that SES doesn't protect against, there's not much you can do against the technique displayed above. 

So to answer your question, to "solve" this, it's best we just implement sensitive code as-is.

packages/scorchwrap/README.md
naugtur: They put multiple modules in named variables within one closure and if the modules are from different packages, we loose the ability to apply policy. It could be possible to wrap them in a way that still works, but I didn't yet figure out if I can do it at all.

packages/scorchwrap/README.md
naugtur: I followed webpack plugin conventions. (to be precise, ChatGPT did and I verified)
I'm all for factory functions, just haven't seen a webpack plugin that does not advertise its use with the `new` keyword.

packages/scorchwrap/README.md
naugtur: not implemented yet. It's an idea for a set of features that may be less performant or theoretically unnecessary

packages/scorchwrap/README.md
naugtur: If they're not in depenedencies in package.json, they don't get taken into account when generating policy and bundling will fail with a message about them missing from policy file.

packages/scorchwrap/README.md
naugtur: historically SES was using rollup to build itself, but after more than one incident where patch version updates introdued changes that made SES no longer secure, they went with their own bundler.

packages/scorchwrap/README.md
naugtur: That's part of the scope for paranoid mode. Please write down your ideas if they're more specific than this.

packages/scorchwrap/README.md
naugtur: We don't want to rely on static analysis, but runtime protections. The policy needs to be there at runtime, so that even if a package contains an RCE (eg. reading from some input and passing to eval) it's still not going to access anything new

packages/scorchwrap/README.md
naugtur: this is a vague idea, but yes, the number of JS assets should equal the number of wrapped modules. Altho it;s not that simple with plugins generating weird stuff (css extract plugin turns css files into js modules and then we have to skip them etc.)

packages/scorchwrap/README.md
naugtur: there's no such thing as "last"" in plugins. Loaders can have order, but what we're doing happens after loaders. For plugins it's a matter of choosing the right hooks and that's about 50% of the work I did here.

packages/scorchwrap/app/app.js
naugtur: :ok_hand:

packages/scorchwrap/app/index.html
naugtur: it's a dump of the default index.html that webpack uses, but with SES script tag added.

packages/scorchwrap/app/index.html
naugtur: see above :D 
I wanted to keep all the mess to be sure we're not incompatible with default webpack. This part can surely be removed.

packages/scorchwrap/src/runtime/runtime.js
```diff
@@ -0,0 +1,159 @@
...
+const LOCKDOWN_ON = typeof lockdown !== "undefined";
+if (LOCKDOWN_ON) {
+  lockdown(LAVAMOAT.options.lockdown);
+} else {
+  console &&
+    console.warn(
+      "LavaMoat: runtime execution started without SES present, switching to no-op."
+    );
+}
+
```
weizman, line 18:
```suggestion
harden(LAVAMOAT)
```
?

packages/scorchwrap/src/runtime/runtime.js
```diff
@@ -0,0 +1,159 @@
...
+/// <reference path="./runtime.d.ts" />
+
+const { create, freeze, assign, defineProperty, entries, fromEntries, getOwnPropertyDescriptors } = Object;
+
+// Avoid running any wrapped code or using compartment if lockdown was not called.
+// This is for when the bundle ends up running despite SES being missing.
+// It was previously useful for sub-compilations running an incomplete bundle as part of the build, but currently that is being skipped. We might go back to it for the sake of build time security if it's deemed worthwihile in absence of lockdown.
+const LOCKDOWN_ON = typeof lockdown !== "undefined";
+if (LOCKDOWN_ON) {
+  lockdown(LAVAMOAT.options.lockdown);
```
weizman:
```suggestion
  lockdown(LAVAMOAT.options.lockdown);
  harden(LAVAMOAT);
```
?

naugtur: Yes. A matter of finding the right time for it though.

packages/scorchwrap/src/plugin.js
```diff
@@ -0,0 +1,558 @@
...
+        });
+      }
+
+      const packageId = getIdentifierForPath(module.resource);
+      if (packageId === undefined) {
+        console.log(module);
+        throw Error(`Failed to find a packageId for ${module.resource}`);
+      }
+
+      let { before, after, source, sourceChanged } = wrapper({
```
weizman:
```suggestion
      const { before, after, source, sourceChanged } = wrapper({
```
?

packages/scorchwrap/src/buildtime/wrapper.js
```diff
@@ -0,0 +1,95 @@
+const { applySourceTransforms } = require("lavamoat-core");
+const diag = require("./diagnostics");
+const fs = require("fs");
+const q = JSON.stringify;
```
weizman: change name?

naugtur: convention stolen from endo - it's a quick quoting function that doesn't fail for objects.

packages/webpack/examples/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "app",
```
naugtur, line 2: (not a concern blocking this PR)
I'm hesitating - to keep the example app or remove it and document using the test fixture a one. 
It's easier to mess around in this app, but it's so much stuff.

Maybe I can clean it up a bit and add a malicious package example to it.

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+      // },
+    }),
+  ],
+  // ... other webpack configuration properties
+};
+```
+
+One important thing to note when using the LavaMoat plugin is that it disables the `concatenateModules` optimization in webpack. This is because concatenation won't work with wrapped modules.
+
+### Avoiding wrapping specific modules
```
```suggestion
### Ignoring Modules
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+> [!WARNING]
+> This is an experimental feature and ignoring may be configured differently in the future if this approach is proven insecure.
```
```suggestion
> This is an experimental feature and the API may change without warning.
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+```
+
+One important thing to note when using the LavaMoat plugin is that it disables the `concatenateModules` optimization in webpack. This is because concatenation won't work with wrapped modules.
+
+### Avoiding wrapping specific modules
+
+> [!WARNING]
+> This is an experimental feature and ignoring may be configured differently in the future if this approach is proven insecure. 
+
+The default way to define specific behaviors for webpack is creating module rules. To ensure ignore rules are applied on the same exact files that match certain rules (the same RegExp may be matched against different things at different times) we're providing the ignore functionality as a loader you can add to the list of existing loaders or use individually.
```
```suggestion
Webpack provides [module rules](https://webpack.js.org/configuration/module/#rule) for defining module-specific behavior. To ensure the rules are applied on the correct files (e.g., the same `RegExp` may be matched against different things at different times), this "ignore" functionality is exposed as a loader.
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+One important thing to note when using the LavaMoat plugin is that it disables the `concatenateModules` optimization in webpack. This is because concatenation won't work with wrapped modules.
+
+### Avoiding wrapping specific modules
+
+> [!WARNING]
+> This is an experimental feature and ignoring may be configured differently in the future if this approach is proven insecure. 
+
+The default way to define specific behaviors for webpack is creating module rules. To ensure ignore rules are applied on the same exact files that match certain rules (the same RegExp may be matched against different things at different times) we're providing the ignore functionality as a loader you can add to the list of existing loaders or use individually.  
+The loader is available as `LavaMoat.ignore` from the default export of the plugin. It doesn't do anything to the code, but its presence is detected and treated as a mark on the file. Any file that's been processed by `LavaMoat.ignore` will not be wrapped in a Compartment.
```
```suggestion
This loader is exposed as the `ignore` property on the plugin's default export.  It doesn't modify code, but instead flags the file as _ignored_. Any file having this flag will _not_ be wrapped in a `Compartment`.

The `ignore` loader will _only_ work when used in a Webpack config. Specifying it inline `require('path/to/ignoreLoader.js!./module.js')` will _not_ result in `module.js` being ignored.  _This is a security feature which prevents your dependencies from declaring they want to be ignored._
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+        use: [
+          MiniCssExtractPlugin.loader,
+          'css-loader',
+          LavaMoat.ignore,
+        ],
+        sideEffects: true,
+      },
+    ],
+  },
+```
```
```suggestion
#### Example: Don't wrap CSS modules

```js
const LavaMoatPlugin = require('@lavamoat/webpack');

module.exports = {
  module: {
    rules: [
      {
        test: /\.css$/,
        use: [
          MiniCssExtractPlugin.loader,
          'css-loader',
          LavaMoatPlugin.ignore,
        ],
        sideEffects: true,
      },
    ],
  },
  // ...
};
```
```

File: packages/webpack/README.md [hunk 26]
```diff
@@ -0,0 +1,229 @@
...
+
+Ignore loader will only work when used in webpack config. Specifying it inline `require('path/to/ignoreLoader.js!./module.js')` will not result in module.js being ignored. (This is a security feature to prevent your dependencies from declaring they want to be ignored.)
```
boneskull: (moved above)

```suggestion
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Gotchas
+
+#### Implicit modules
+- Webpack may include dependencies for node builtins like the `events` or `buffer` packages. In other cases it will ignore the builtins and provide empty modules in their place. (see below)
```
```suggestion
- Webpack may include dependencies polyfilling Node.js built-ins, such as the `events` or `buffer` packages. In other cases, it will ignore the built-ins and provide empty modules in their place (see below).
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+When a dependency (eg. buffer) is provided by webpack and you need to add it explicitly to your dependencies you'll get the following error:
```
```suggestion
When a dependency (eg. `buffer`) is provided by Webpack, and you need to add it explicitly to your dependencies, you'll receive the following error:
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+```
+Error: LavaMoat - Encountered unknown package directory for file "/home/(...)/node_modules/buffer/index.js"
+```
+
+#### Webpack-ignored modules
+Webpack generates this:
+```js
+const nodeCrypto = __webpack_require__(/*! crypto */ "?0b7d");
+```
+when a builtin module is ignored (Node builtins are a good example of that. If you want a builtin to work, you sometimes need to supply a package for it as a dependency yourself)
```
```suggestion
When a built-in Node.js module is ignored, Webpack generates something like this:

```js
const nodeCrypto = __webpack_require__(/*! crypto */ "?0b7d");
```

If you want a built-in to work, you sometimes need to supply your own dependency.
```

File: packages/webpack/README.md [hunk 30]
```diff
@@ -0,0 +1,229 @@
...
+A carveout is necessary in policy enforcement for these modules. 
+Sadly, even treeshaking doesn't eliminate that module. It's left there and failing to work when reached by runtime control flow.
+
+LavaMoat plugin is prepared to skip those and if a package attempts to import a webpack-ignored module an error will not be thrown even if it's not explicitly allowed by policy
```
```suggestion
A carveout is necessary for these modules in LavaMoat's policy enforcement.

Unfortunately, even treeshaking doesn't eliminate it. It's left there, useless, and fails to work if the runtime control flow ever reaches it.

This LavaMoat plugin is prepared to skip such modules, and if a package attempts to import a Webpack-ignored module, you'll be alerted with a thrown exception--even if it's not explicitly allowed by your LavaMoat policy.
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+
+
+
```
```suggestion
```

prettier should clean this up

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+# Security Claims
+
+This is a beta version and it does not provide any guarantees, even those listed below. Use at your own risk.
```
```suggestion
**This is a _beta_ release and does not provide any guarantees; even those listed below. Use at your own risk!**
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+- SES must be added to the page without any bundling or transforming for any security guarantees to be sustained.
+  - The plugin could add it as an asset to the compilation if that's a good Developer Experience. Feedback welcome.
+- Each javascript module resulting from the webpack build is scoped to its package's policy
+
+Threat model
```
```suggestion
## Threat Model
```

Looks like a header to me?

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+- Webpack itself is considered trusted.
+- All plugins can bypass LavaMoat protections intentionally.
+- It's unlikely but possible that a plugin can bypass LavaMoat protections unintentionally.
```
```suggestion
- It's unlikely _but possible_ that a plugin can bypass LavaMoat protections _unintentionally_.
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+- It should not be possible for loaders to bypass LavaMoat protections.
+- Some plugins (eg. MiniCssExtractPlugin) execute code from the bundle at build time. To make the plugin work you need to trust it and the modules it runs and add the LavaMoat.ignore loader for them.
```
```suggestion
- Some third-party plugins (eg. `MiniCssExtractPlugin`) execute code from the bundle at build time. To make such a plugin work, you need to a) trust it, b) trust the modules it runs, and c) add it to the `LavaMoatPlugin.ignore` loader.
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+- Each javascript module resulting from the webpack build is scoped to its package's policy
+
+Threat model
+
+- Webpack itself is considered trusted.
+- All plugins can bypass LavaMoat protections intentionally.
+- It's unlikely but possible that a plugin can bypass LavaMoat protections unintentionally.
+- It should not be possible for loaders to bypass LavaMoat protections.
+- Some plugins (eg. MiniCssExtractPlugin) execute code from the bundle at build time. To make the plugin work you need to trust it and the modules it runs and add the LavaMoat.ignore loader for them. 
+- LavaMoat plugin is not protecting you against malicious packages being executed by other plugins at runtime (but lavamoat cli could)
```
```suggestion
- This Webpack plugin _does not_ protect against malicious execution by other third-party plugins at runtime (use [LavaMoat](https://npm.im/lavamoat) for that).
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+Webpack runtime
```
```suggestion
## Webpack runtime
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+Elements of webpack runtime (`__webpack_require__.*`) are currently mostly left intact. To avoid opening up potential bypasses, some functionality of the webpack runtime will need to be restricted.
```
```suggestion
Elements of the Webpack runtime (e.g., `__webpack_require__.*`) are currently mostly left intact. To avoid opening up potential bypasses, some functionality of the Webpack runtime will need to be restricted.
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+
+
+# Testing 
+
+Run `npm test` to start the automated tests.
+
+## Manual testing
+
+run `npm ci` and `npm run lib:ses` in the packages/webpack folder before you begin
```
```suggestion
To prepare, execute `npm ci` then `npm run -w @lavamoat/webpack lib:ses`.

Then:

- Navigate to `packages/webpack/example/`
- Run `npm ci` and `npm test`
- Open `dist/index.html` in your browser and inspect the console
```

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+`cd example`  
+run `npm ci`  
+run `npm test` to trigger the build  
+open dist/index.html in the browser and look at the console
+
+
+
+# Development
```
boneskull: I feel like this entire section should be put into a different document, or perhaps just make issues for it.

packages/webpack/README.md
```diff
@@ -0,0 +1,229 @@
...
+aa[AA identifiers in policy] --> p[policy processing]
+wp[paths of the bundle] --> waa[paths mapped to AAs]
+waa --> p
+p --> w[replace AAs with numbers from a sequence]
+w --> b[use short identifiers in the bundle and inlined policy]
+
+```
+
+
+# Features
```
boneskull: Is this intended to be part of the `Development` section?

packages/webpack/src/plugin.js
```diff
@@ -0,0 +1,362 @@
+// @ts-check
+
+/** @typedef {import("webpack").Compiler} Compiler */
+/** @typedef {import("webpack").Compilation} Compilation */
+/** @typedef {import("webpack").Generator} Generator */
+/** @typedef {import("./types.js").LavaMoatPluginOptions} LavaMoatPluginOptions */
```
boneskull: I'm pretty well-sold on the idea that creating `@typedef` references like this is a hazard.  Two reasons (maybe others):

1. All typedefs are "exported". This may be undesirable
2. If a type contains type arguments, you either have to copy them outright or risk the reference becoming out-of-sync with the source.

packages/webpack/src/plugin.js
```diff
@@ -0,0 +1,362 @@
...
+    const options = this.options
+    if (typeof options.readableResourceIds === 'undefined') {
+      // default options.readableResourceIds to true if webpack configuration sets development mode
+      options.readableResourceIds = compiler.options.mode !== 'production'
+    }
+
+    let canonicalNameMap
+
+    // Concatenation won't work with wrapped modules. Have to disable it.
+    compiler.options.optimization.concatenateModules = false
```
boneskull, line 93: I believe you.  Why, though?

packages/webpack/src/plugin.js
```diff
@@ -0,0 +1,362 @@
...
+    // compiler.options.optimization.mangleExports = false;
+    // compiler.options.optimization.usedExports = false;
+    // compiler.options.optimization.providedExports = false;
+    // compiler.options.optimization.sideEffects = false;
+    // compiler.options.optimization.moduleIds = "hashed";
+    // compiler.options.optimization.chunkIds = "named";
+
+    // =================================================================
+    // run long asynchronous processing ahead of all compilations
+    compiler.hooks.beforeRun.tapAsync(PLUGIN_NAME, (compilation, callback) =>
```
boneskull, line 106: Is there any disadvantage to kicking off this process first?  I assume the advantage is that it is faster.  How much of a bottleneck is it?  Would it be better done in a worker thread?

packages/webpack/src/plugin.js
```diff
@@ -0,0 +1,362 @@
...
+
+    let mainCompilationWarnings
+
+    compiler.hooks.thisCompilation.tap(
+      PLUGIN_NAME,
+      (compilation, { normalModuleFactory }) => {
+        compilation.hooks.optimizeAssets.tap(PLUGIN_NAME, () => {
+          // By the time assets are being optimized we should have finished.
+          // This will ensure all previous steps have been done.
+          PROGRESS.report('finish')
```
boneskull, line 128: My intuition is that this isn't super-safe to assume.  Is there a better place for it?

packages/webpack/src/plugin.js
```diff
@@ -0,0 +1,362 @@
...
+        })
+
+        // =================================================================
+
+        if (!mainCompilationWarnings) {
+          mainCompilationWarnings = compilation.warnings
+          mainCompilationWarnings.push(
+            new WebpackError(
```
boneskull, line 136: This looks like we're adding a warning to the list of warnings regardless of whether concatenation was asked for or not.  Is there a reason we're always showing it, instead of only when we force a behavior change?

packages/webpack/src/plugin.js
```diff
@@ -0,0 +1,362 @@
...
+                  module.identifierStr.startsWith('ignored')) ||
+                module.resource === undefined // better to explicitly list it as unenforceable than let it fall through the cracks
+              ) {
+                unenforceableModuleIds.push(moduleId)
+              } else {
+                knownPaths.push({ path: module.resource, moduleId }) // typescript is complaining about the use of `resource` here, but it's actually there.
+              }
+            })
+          })
+          diag.rawDebug(4, { knownPaths })
```
boneskull, line 203: builtin webpack diagnostics API?

packages/webpack/src/plugin.js
```diff
@@ -0,0 +1,362 @@
...
+
+          identifierLookup = generateIdentifierLookup({
+            readableResourceIds: options.readableResourceIds,
+            unenforceableModuleIds,
+            paths: knownPaths,
+            policy: options.policy,
+            canonicalNameMap,
+          })
+          if (unenforceableModuleIds.length > 0) {
+            mainCompilationWarnings.push(
```
boneskull, line 214: Is this the same error that the `README` says you'll see?  Why is it a warning and not an error?

packages/webpack/src/plugin.js
```diff
@@ -0,0 +1,362 @@
...
+
+        // Define a handler function to be called for each chunk in the compilation.
+        compilation.hooks.additionalChunkRuntimeRequirements.tap(
+          PLUGIN_NAME + '_runtime',
+          (chunk /*, set*/) => {
+            if (chunk.hasRuntime()) {
+              if (!PROGRESS.done('gneratorCalled')) {
+                mainCompilationWarnings.push(
+                  new WebpackError(
+                    'LavaMoatPlugin: Something was generating runtime before all modules were identified. This might be part of a sub-compilation of a plugin. Please check for any unwanted interference between plugins.',
```
boneskull, line 272: I'm curious if this is only a hazard in our case or in the general case

packages/webpack/src/plugin.js
```diff
@@ -0,0 +1,362 @@
...
+                    data: identifierLookup.unenforceableModuleIds || null,
+                    json: true,
+                  },
+                  { name: 'options', data: runtimeOptions, json: true },
+                  { name: 'policy', data: policyData, json: true },
+                  { name: 'ENUM', file: path.join(__dirname, './ENUM.json'), json: true },
+                  {
+                    name: 'endowmentsToolkit',
+                    shimRequire:
+                      'lavamoat-core/src/endowmentsToolkit.js',
```
boneskull, line 321: is there another way to do this that doesn't require reaching into the source this way?

packages/webpack/src/plugin.js
```diff
@@ -0,0 +1,362 @@
...
+                  },
+                  { name: 'runtime', file: path.join(__dirname, './runtime/runtime.js') },
+                ])
+
+                // If the chunk has already been processed, skip it.
+                if (onceForChunkSet.has(chunk)) {
```
boneskull: why do we do so much work before bailing?

packages/webpack/src/runtime/runtime.d.ts
```diff
@@ -0,0 +1,4 @@
+/// <reference path="../types.js" />
+/// <reference types="ses" />
```
boneskull: one of these doesn't belong

packages/webpack/src/runtime/runtime.js
```diff
@@ -0,0 +1,189 @@
...
+
+    // TODO: figure out what to wrap and what to copy
+    // TODO: most of the work here could be done once instead of for each wrapping
+
+    // Webpack has built-in plugins that add more runtime functions. We might need to support them eventually.
+    // It's a case-by-case basis decision.
+    // TODO: print a warning for other functions on the __webpack_require__ namespace that we're not supporting.
+    //   It's probably best served at build time though - with runtimeRequirements or looking at the items in webpack runtime when adding lavamoat runtime.
+
+    // The following seem harmless and are used by default: ['O', 'n', 'd', 'o', 'r', 's']
```
boneskull, line 137: I'm sorry?

packages/webpack/src/runtime/runtime.js
```diff
@@ -0,0 +1,189 @@
...
+          warn(
+            `A module attempted to read ${prop} directly from webpack's module cache`,
+          )
+          return false
+        },
+      },
+    )
+
+    // override nmd to limit what it can mutate
+    policyRequire.nmd = (moduleReference) => {
```
boneskull, line 159: what is nmd? what is m?

packages/webpack/src/runtime/runtime.js
```diff
@@ -0,0 +1,189 @@
...
+
+  // allow setting, but ignore value for /* module decorator */ module = __webpack_require__.nmd(module)
+  defineProperty(runtimeHandler, 'module', {
+    get: () => module,
+    set: () => {},
+  })
+  freeze(runtimeHandler)
+
+  if (!compartmentMap.has(resourceId)) {
+    // Endow original Math and Date, because SES tames them and we don't need that
```
boneskull, line 177: is this something that could be an option, or is this unsupported in core, too?

packages/webpack/src/runtime/runtime.js
```diff
@@ -0,0 +1,189 @@
...
+    const c = new Compartment({ Math, Date })
+    installGlobalsForPolicy(resourceId, c.globalThis)
+    compartmentMap.set(resourceId, c)
+  }
+
+  return {
+    [NAME_scopeTerminator]: stricterScopeTerminator,
+    [NAME_runtimeHandler]: runtimeHandler,
+    [NAME_globalThis]: compartmentMap.get(resourceId).globalThis,
```
boneskull, line 186: is there any chance that `compartmentMap.get(resourceId)` could be `undefined`?

packages/webpack/src/types.js
```diff
@@ -0,0 +1,51 @@
...
+ * @typedef {Object} LavaMoatPluginOptions
+ * @property {boolean} [runChecks] - check resulting code with wrapping for correctness
+ * @property {boolean} [readableResourceIds] - should resourceIds be readable or turned into numbers - defaults to (mode==='development')
+ * @property {number} [diagnosticsVerbosity] - a number representing diagnostics output verbosity, the larger the more overwhelming
+ * @property {Object} policy - LavaMoat policy
+ * @property {Object} [lockdown] - options to pass to lockdown
+ */
+
+/**
+ * @typedef {function} GetEndowmentsForConfig Creates an object populated with only the deep properties specified in the packagePolicy
```
boneskull:
```suggestion
 * Creates an object populated with only the deep properties specified in the `packagePolicy`
 * @callback GetEndowmentsForConfig 
```

packages/webpack/src/types.js
```diff
@@ -0,0 +1,51 @@
...
+ * @property {Function} makeMinimalViewOfRef - Creates a minimal view of a reference (e.g. a global) that only exposes the properties specified in the packagePolicy
+ * @property {Function} copyValueAtPath - Copies a value at a specific path from the source reference to the target reference
+ * @property {Function} applyGetSetPropDescTransforms 
+ * @property {Function} applyEndowmentPropDescTransforms 
+ * @property {Function} copyWrappedGlobals - Copies wrapped globals from a global reference to a target object with wrapping
+ * @property {Function} createFunctionWrapper - conditionally binds a function - used to ensure functions (like fetch) get called on the this they expect
+ */
+
+/**
+ * @typedef {function} EndowmentsToolkitFactory
```
boneskull:
```suggestion
 * @callback EndowmentsToolkitFactory
```

I actually have types for this stuff in `endo-refactor` branch.  I should probably extract them and send a PR

packages/webpack/test/aa.spec.js
```diff
@@ -0,0 +1,81 @@
...
+
+test('aa - generateIdentifierLookup', (t) => {
+  const lookupTools = generateIdentifierLookup({
+    paths: [
+      '/home/user/app/app.js',
+      '/home/user/app/node_modules/leftpad/index.js',
+      '/home/user/app/node_modules/fast-json-patch/index.mjs',
+      '/home/user/app/node_modules/@ethereumjs/util/dist/index.js',
+      '/home/user/app/node_modules/@ethereumjs/util/node_modules/ethereum-cryptography/dist/index.js',
+      '/home/user/app/node_modules/fast-json-patch/module/duplex.mjs',
```
boneskull, line 12: can any of these be eliminated from the fixture without a drop in coverage?

packages/webpack/test/e2e-hack.spec.js
```diff
@@ -0,0 +1,59 @@
...
+test.before(async (t) => {
+  const webpackConfig = {
+    ...webpackConfigDefault,
+    entry: {
+      hack1: './hack1.js',
+      hack2: './hack2.js',
+      hack3: './hack3.js',
+    },
+  }
+  await t.notThrowsAsync(async () => {
```
boneskull, line 14: I mean, if it failed, the hook would fail.  unsure if wrapping in an assertion is more helpful or harmful

packages/webpack/test/e2e-main.spec.js
```diff
@@ -0,0 +1,75 @@
+const test = require('ava')
+const { scaffold, runScriptWithSES } = require('./scaffold.js')
+const webpackConfigDefault = require('./fixtures/main/webpack.config.js')
+
+test.before(async (t) => {
+  const webpackConfig = { ...webpackConfigDefault }
+  await t.notThrowsAsync(async () => {
+    t.context.build = await scaffold(webpackConfig)
+  }, 'Expected the build to succeed')
+  t.context.bundle = t.context.build.snapshot['/dist/app.js']
```
boneskull, line 10: a reasonable thing instead would be to assert that this value exists

packages/webpack/test/fixtures/main/node_modules/commonjs-package/index.js
boneskull, line 1: (later) too many things (github included) make assumptions about what `node_modules` contains and what put it there.  we should find a different way to accomplish the same thing avoiding the directory name.  trivially, this could be done via `NODE_PATH`

packages/webpack/src/runtime/runtime.js
```diff
@@ -0,0 +1,189 @@
...
+  )
+}
+
+const { getEndowmentsForConfig, copyWrappedGlobals } =
+  LAVAMOAT.endowmentsToolkit()
+
+// These must match assumptions in the wrapper.js
+// sharedKeys are included in the runtime
+
+const { NAME_globalThis, NAME_scopeTerminator, NAME_runtimeHandler } =
```
boneskull, line 33: what is the significance of these variable names?

packages/webpack/src/runtime/runtime.js
```diff
@@ -0,0 +1,189 @@
...
+  freeze,
+  assign,
+  defineProperty,
+  defineProperties,
+  getOwnPropertyDescriptors,
+} = Object
+const warn = typeof console === 'object' ? console.warn : () => {}
+// Avoid running any wrapped code or using compartment if lockdown was not called.
+// This is for when the bundle ends up running despite SES being missing.
+// It was previously useful for sub-compilations running an incomplete bundle as part of the build, but currently that is being skipped. We might go back to it for the sake of build time security if it's deemed worthwihile in absence of lockdown.
```
boneskull, line 17: I am unclear on the tradeoff here vs just aborting instead.  What does "switching to no-op" mean?

boneskull (approved):
The only thing I put effort into was `README.md`.  Everything else I wrote was done half-assed. hopefully there's a good question or two in there, though.

packages/webpack/test/e2e-main.spec.js
boneskull: oh also, see `/scripts/publish.spec.js` for how to properly type a test execution context, if you care

packages/webpack/README.md
naugtur: I like your idea (from meeting) to use the word 'exclude'

packages/webpack/src/runtime/runtime.js
naugtur: this code should not be reachable with a resourceId that's not in the compartmentMap

packages/webpack/test/aa.spec.js
naugtur: this fixture needs a redo, it is not testing some of the behaviors I wish it did.

packages/webpack/test/e2e-hack.spec.js
naugtur: The difference is, without this you get "Rejected promise returned by test" in the result listing which is significantly less helpful

packages/webpack/src/plugin.js
naugtur: because I forgot to move the if :sweat_smile:

packages/webpack/src/runtime/runtime.d.ts
naugtur: why? I thought I'm composing

packages/webpack/src/plugin.js
naugtur: It's gonna be helpful for interactions with come weird plugins when stuff goes sideways. 
Also, once we get a better understanding of how it's used and when, we might actually prevent webpack from running code form the frontend bundle at build time (it seems it's possible for that to happen. Imagine a styled component stealing your github token...)

packages/webpack/src/plugin.js
naugtur: wut?

packages/webpack/src/plugin.js
naugtur: We're always disabling concatenation. If someone disables it manually we could hide the warning, but that's another bit of state to juggle. (I'm lazy)

packages/webpack/README.md
naugtur: is this assuming they're reading the readme but are not in the package's own folder?

packages/webpack/README.md
naugtur: the point is there will NOT be an error.

packages/webpack/README.md
naugtur: I'll let it do that

packages/webpack/README.md
naugtur: sounds good

packages/webpack/README.md
naugtur: yes. these are ideas

packages/webpack/src/runtime/runtime.js
naugtur: discussed in person

packages/webpack/src/runtime/runtime.js
naugtur: discussed in person

packages/webpack/test/fixtures/main/node_modules/commonjs-package/index.js
naugtur: added a draft task to the board for that.

packages/webpack/src/runtime/runtime.js
naugtur: me too :D 

It's what webpack calls its functions

naugtur:
No longer waiting for a review from Gal under current circumstances. Merging soon.

legobeat:
@naugtur merging this unsynced seemed to have broken the lint on `main`: https://github.com/LavaMoat/LavaMoat/actions/runs/6548008016/job/17781690350

PR #487, opened 2023-05-03
ci: test on nodejs v18, v20

legobeat:

PR #491, opened 2023-05-05
Core bug cant read with write perm

kumavis:
`write` is supposed imply `read`, but the read getter reads from the endowments which we were incorrectly not populating

df81c47602ca1ea16196b7e44415d8816f021715 is an unrelated driveby fix

### DIFF
--- a/packages/core/test/scenarios/globalWrites.js
+++ b/packages/core/test/scenarios/globalWrites.js
@@ -45,26 +45,22 @@ module.exports = [
       name: 'globalWrites - and reads',
       defineOne: () => {
         module.exports = {
-          type: typeof globalThis.xyz,
-          value: globalThis.xyz
+          type: typeof globalThis.setTimeout,
         }
       },
       config: {
         resources: {
           one: {
             globals: {
-              xyz: 'write'
+              // using "setTimeout" as an example of an endowment available
+              // in all testing environments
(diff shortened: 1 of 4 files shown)


legobeat:
> df81c47602ca1ea16196b7e44415d8816f021715 is an unrelated driveby fix

Can it be broken out to a separate PR?


Also it seems we have some node version differences - to avoid further regressions as we fix things up, perhaps we can start with #487?

packages/core/test/scenarios/globalWrites.js
```diff
@@ -39,5 +39,30 @@ module.exports = [
...
+            globals: {
+              // using "setTimeout" as an example of an endowment available
+              // in all testing environments
+              setTimeout: 'write'
+            }
+          },
+        }
+      },
+      expectedResult: {
+        type: 'function',
```
legobeat, line 63:
```
✖ runScenarios › Run scenarios with precompiled modules globalWrites - and reads - scenario gives expected result
  ─

  runScenarios › Run scenarios with precompiled modules

  globalWrites - and reads - scenario gives expected result

  Difference:

    {
  -   type: 'undefined',
  +   type: 'function',
    }
```

I guess this is what this is solving for?

https://github.com/LavaMoat/LavaMoat/actions/runs/5013892577/jobs/8987521673?pr=491

PR #492, opened 2023-05-05
Perf: update performance testing

### LINKED ISSUE #489: perf: missing source files
`perf` package is missing source files to be able to run and build: https://github.com/LavaMoat/LavaMoat/blob/f6d0d9653c23f002af1768d581efcb171509a2ff/packages/perf/build.js#L17

kumavis:
this gets the perf testing working again (it was extremely old, predated `lavamoat-node`)
I also added `endo-unsafe` for comparison and there are some interesting results!

**endo performance is much better** than lavamoat in some cases, we should investigate how

Fixes: https://github.com/LavaMoat/LavaMoat/issues/489

stats


kumavis:
accidentally included this change here https://github.com/LavaMoat/LavaMoat/pull/493
lets merge that first

kumavis:
im going to merge it bc its not part of the critical path

PR #493, opened 2023-05-05
core/fix - remove console taming to that errors are logged to the console correctly

kumavis:
I noticed errors were not being logged to the node console correctly, changing this option fixes that


legobeat (approved):
LGTM!

PR #505, opened 2023-05-05
perf: bump trials dependencies

legobeat:
Resolves multiple security advisories.

- contained problematic `scrypt`/`keccak` `node-gyp`.
- https://github.com/LavaMoat/LavaMoat/security/dependabot/365

### DIFF
--- a/packages/perf/trials/increment/package.json
+++ b/packages/perf/trials/increment/package.json
@@ -13,7 +13,7 @@
   },
   "dependencies": {
     "chart-csv": "kumavis/chart-csv#master",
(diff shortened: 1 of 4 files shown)


naugtur (approved):

packages/perf/trials/increment/package.json
```diff
@@ -13,7 +13,7 @@
   },
   "dependencies": {
     "chart-csv": "kumavis/chart-csv#master",
```
naugtur, line 15: I know we trust the guy, but we should strive to not have git dependencies.

PR #514, opened 2023-05-06
deps: bump d3 subdepdendencies

legobeat:
https://github.com/LavaMoat/LavaMoat/security/dependabot/278


boneskull:
landed in #702

PR #517, opened 2023-05-08
perf/deps: bump ses@0.6.4->0.18.4

legobeat:
https://github.com/LavaMoat/LavaMoat/security/dependabot/71


naugtur (approved):

naugtur:
We could look into why the file is there in the first place

PR #550, opened 2023-05-09
GW try help lego with scuttling

weizman:


legobeat:
Cherry-picked into #551

PR #554, opened 2023-05-12
perf,viz: remove subpackage yarn.lock

legobeat:
Where did they come from? What is their fate? Please tune in for next episode.

(Let's remove this maintenance overhead unless there are gotchas? @kumavis )


kumavis:
mistake or pre-dates monorepo? dont recall - seems fine to remove

kumavis (approved):

PR #556, opened 2023-05-12
node/deps: bump json-stable-stringify, node-gyp-build, resolve, yargs

legobeat:
A full dedupe (yarn is lying about doing it) will come in a follow-up.


kumavis:
https://github.com/scinos/yarn-deduplicate
`yarn-deduplicate --strategy highest && yarn`

kumavis (approved):

legobeat:
Just been noticing that the nodejs checks in CI have been finishing faster on lower node versions on average. Will have to follow up locally if there's something there or just GitHub infra stuff.

PR #484, opened 2023-04-25
devDeps: lerna@3.22.1->5.6.2

legobeat:
There are multiple problematic dependencies we are inheriting from lerna and it's high time to update
  - removes devDependency on deprecated, vulnerable `request`
  -    the current version of lerna is 6.6.1
  -    a future effort will have to assess the upgrade to v6

Addresses multiple security advisories
  - https://github.com/LavaMoat/LavaMoat/security/dependabot/276
  - https://github.com/LavaMoat/LavaMoat/security/dependabot/339
  - https://github.com/LavaMoat/LavaMoat/security/dependabot/260
  - https://github.com/LavaMoat/LavaMoat/security/dependabot/213
  - https://github.com/LavaMoat/LavaMoat/security/dependabot/212
  - https://github.com/LavaMoat/LavaMoat/security/dependabot/28
  - https://github.com/LavaMoat/LavaMoat/security/dependabot/27
  - https://github.com/LavaMoat/LavaMoat/security/dependabot/277

PR #566, opened 2023-05-13
devDeps: lerna@^5.6.2->^6.6.2

legobeat:
Follow-up to #484


legobeat:
Been running this when devving the other open PRs. Seems all good. We can still choose to drop lerna for pnpm or yarn3 workspaces but in the meantime this should bring us in line with updates (and brings some performance improvements).

weizman (approved):

naugtur (approved):

PR #567, opened 2023-05-13
core: remove Object.fromEntries polyfill

legobeat:
These were added for node v10 compatibility in https://github.com/LavaMoat/LavaMoat/commit/78aa255024cc3cac611bd02598e785dff8453aa4.
Minimum supported nodejs version is v14 where the polyfilled function is available natively, so these are no longer needed.

### DIFF
--- a/packages/core/src/generatePolicy.js
+++ b/packages/core/src/generatePolicy.js
@@ -1,6 +1,5 @@
 const EventEmitter = require('events')
 const path = require('path')
-const fromEntries = require('fromentries')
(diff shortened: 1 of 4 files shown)


packages/core/src/generatePolicy.js
```diff
@@ -1,6 +1,5 @@
 const EventEmitter = require('events')
 const path = require('path')
-const fromEntries = require('fromentries')
```
naugtur, line 3: (could)
```js
const fromEntries = Object;
```

naugtur (approved):

PR #572, opened 2023-05-13
deps: bump core deps and devDeps

legobeat:
- `core/deps`: `merge-deep`@`^3.0.2`->`^3.0.3`
- `core/deps`: `json-stable-stringify`@`^1.0.1`->`^1.0.2`
- `core/devDeps`: `ses`@`^0.17.0`->`^0.18.4`
  - Note: Prior to this change, users respecting thelockfile actually end upwith two different versions of `ses`, where `core` was on the older `0.17.0`.
- `core/devDeps`: `tmp-promise`@`^3.0.2`->`^3.0.3`

### DIFF
--- a/packages/browserify/test/fixtures/secureBundling/lavamoat/node/policy.json
+++ b/packages/browserify/test/fixtures/secureBundling/lavamoat/node/policy.json
@@ -795,9 +795,14 @@
         "define": true
       },
       "packages": {
-        "json-stable-stringify": true,
+        "lavamoat-core>json-stable-stringify": true,
         "lavamoat-core>lavamoat-tofu": true,
         "lavamoat-core>merge-deep": true
+      }
+    },
+    "lavamoat-core>json-stable-stringify": {
+      "packages": {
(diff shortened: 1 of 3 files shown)


legobeat:
# Poll! :raising_hand: 

How granular do you like our git log to be? Vote by reacting.

- :tada: to keep these commits independent on merge
- :rocket: to squash these into a single commit on merge
- :confused: other. leave comment.
- :smile: i don't care, i just like reacting to things, also hi lego

naugtur (approved):

weizman:
Does [this](https://github.com/LavaMoat/LavaMoat/pull/572#issuecomment-1546534140) mean PR squashing from now on? 😍

legobeat:
@weizman "these commits" here would refer to specifically this situation. It would still be on a case-by-case basis and this is one step on figuring out where to draw the line (in this case I merged them independently but would first squash into 1-3 before doing a rebase-merge in a future identical situation).

PR #576, opened 2023-05-15
core/version - 14.1.0

naugtur:
- change: Policy generation now looks at .cjs, .mjs, .ts and .js files as opposed to only .js files (https://github.com/LavaMoat/LavaMoat/pull/474)
 - change: Deprecate `global`,`self` as aliases for `globalThis` (https://github.com/LavaMoat/LavaMoat/pull/461)
 - fix: Restrict `engines.node` to < 19.0.0 (https://github.com/LavaMoat/LavaMoat/pull/552)
 - fix: Add `WebAssembly` to scuttle globalThis exceptions. This is necessary as of Node.js 18 (https://github.com/LavaMoat/LavaMoat/pull/551)
 - fix: Remove console taming so that errors are logged to the console correctly (https://github.com/LavaMoat/LavaMoat/pull/493)
 - fix: Fix globalThis polyfill (https://github.com/LavaMoat/LavaMoat/pull/567)
 - fix: Handle error when receiver is null in getPropertyDescriptorDeep (https://github.com/LavaMoat/LavaMoat/pull/463)
 - fix: Drop unnecessary additionalOpts arg and migrate back into using scenario object (https://github.com/LavaMoat/LavaMoat/pull/471)
 - fix: Catch errors thrown when calling property getters (https://github.com/LavaMoat/LavaMoat/pull/468)

### DIFF
--- a/packages/browserify/package.json
+++ b/packages/browserify/package.json
@@ -18,7 +18,7 @@
     "convert-source-map": "^1.9.0",
     "duplexify": "^4.1.1",
     "json-stable-stringify": "^1.0.1",
-    "lavamoat-core": "^14.0.0",
+    "lavamoat-core": "^14.1.0",
     "pify": "^4.0.1",
     "readable-stream": "^3.6.0",
     "source-map": "^0.7.4",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "lavamoat-core",
-  "version": "14.0.0",
+  "version": "14.1.0",
   "description": "LavaMoat kernel and utils",
   "main": "src/index.js",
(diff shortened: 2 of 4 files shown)


legobeat (approved):

legobeat (changes requested):
Sibling dependents need to be updated. 

EDIT: Just pushed a commit for that on this branch. Since `viz` is lagging behind, it is bumped separately in #577.

legobeat:
Not a blocker for merge but it would be neat to get #572 in with this release as well - I think the dep bumps are minor enough to not strictly need mention in the release commit message.

weizman (approved):
lgtm

legobeat (approved):

PR #586, opened 2023-05-18
lift <19 nodejs restriction for pkgs not depending on core

legobeat:
~`@lavamoat/aa` is~ Several packages are not actually affected by the issue prompting the restriction in #552, so should be lifted prior to next release of this package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "lerna": "^6.6.2"
   },
   "engines": {
-    "node": ">=14.0.0 <19.0.0"
+    "node": ">=14.0.0"
   },
   "scripts": {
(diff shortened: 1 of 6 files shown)


leotm (approved):

leotm (changes requested):
after further thought, what do we think about expanding this change to include other packages?

nb: tested linting and tests on node v20 locally for aa and thanks for our pair programming call

legobeat:
Changing scope of this PR to lift all relevant packages in monorepo.

naugtur:
+1 to larger PRs covering the same/similar change across many packages

naugtur (approved):

PR #597, opened 2023-05-29
core: internal refactor scuttling options handling

legobeat:
- reduce transient state
- scuttle 
- various internal function renames
- add validation of `scuttleGlobalThis.exceptions`

### DIFF
--- a/packages/core/src/kernelCoreTemplate.js
+++ b/packages/core/src/kernelCoreTemplate.js
@@ -129,14 +129,14 @@
         .flatMap(proto => Object.getOwnPropertyNames(proto))
 
       // support LM, SES exported APIs and polyfills
-      const baseIgnoreProps = ['Compartment', 'Error', 'globalThis']
+      const baseIgnoreProps = [/^Compartment$/, /^Error$/, /^globalThis$/]
       const ignoreProps = Array.from(new Set([
         ...baseIgnoreProps,
         ...extraIgnoreProps,
       ]))
 
       for (const prop of props) {
-        if (ignoreProps.some(p => prop.match(p))) {
+        if (ignoreProps.some(p => p.test(prop))) {
           continue
         }
 
@@ -523,13 +523,16 @@
     }
 
     function decodeRegexp (s) {
-      if (!s.startsWith('/')) {
-        // escape regexp https://stackoverflow.com/questions/3561493/is-there-a-regexp-escape-function-in-javascript
-        return new RegExp(s.replace(/[/\-\\^$*+?.()|[\]{}]/gu, '\\$&'))
+      switch (s.type) {
+        case 'regex':
+          const parts = s.value.split('/')
+          const flags = parts.pop()
+          return new RegExp(parts.join('/'), flags)
+        case 'string':
+          return new RegExp(s.value.replace(/[/\-\\^$*+?.()|[\]{}]/gu, '\\$&'))
+        default:
+          throw new Error(`Invalid Regexp type ${s.type}`)
       }
-      const parts = s.split('/')
-      const flags = parts.pop()
-      return new RegExp(parts.join('/'), flags)
     }
   }
 })()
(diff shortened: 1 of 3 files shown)


packages/core/src/kernelCoreTemplate.js
```diff
@@ -98,82 +97,90 @@
     return kernel
 
     // generate final scuttling options (1) by taking default
-    // options into consideration, (2) turning RE strings into
-    // actual REs and (3) without mutating original opts object
-    function generateScuttleOpts(originalOpts) {
+    // options into consideration, (2) turning regex strings into
+    // actual regexes
+    function decodeScuttleOptions(opts) {
```
legobeat, line 103: `generateScuttleOpts` is now `decodeScuttleOpts`, and has a corresponding `encodeScuttleOpts`.

packages/core/src/kernelCoreTemplate.js
```diff
@@ -98,82 +97,90 @@
...
-          throw new Error(
-            `LavaMoat - 'scuttlerName' function "${opts.scuttlerName}" expected on globalRef.` +
-            'To learn more visit https://github.com/LavaMoat/LavaMoat/pull/462.',
-          )
-        }
-        opts.scuttlerFunc = globalRef[opts.scuttlerName]
-      }
-      return opts
-
-      function toRE(except) {
```
legobeat, line 125: `toRE` is now `decodeRegexp` (which now escapes strings into regexps)

packages/core/src/kernelCoreTemplate.js
```diff
@@ -98,82 +97,90 @@
...
-          return except
-        }
-        const parts = except.split('/')
-        const pattern = parts.slice(1, -1).join('/')
-        const flags = parts[parts.length - 1]
-        return new RegExp(pattern, flags)
-      }
     }
 
-    function performScuttleGlobalThis (globalRef, extraPropsToAvoid = new Array()) {
```
legobeat, line 137: `propsToAvoid` is now `ignoreProps`

packages/core/src/kernelCoreTemplate.js
```diff
@@ -98,82 +97,90 @@
...
-      getPrototypeChain(globalRef)
-        .forEach(proto =>
-          props.push(...Object.getOwnPropertyNames(proto)))
+    function performScuttleGlobalThis (globalRef, extraIgnoreProps = []) {
+      const props = getPrototypeChain(globalRef)
+        .flatMap(proto => Object.getOwnPropertyNames(proto))
 
-      // support LM,SES exported APIs and polyfills
-      const avoidForLavaMoatCompatibility = ['Compartment', 'Error', 'globalThis']
-      const propsToAvoid = new Set([...avoidForLavaMoatCompatibility, ...extraPropsToAvoid])
```
legobeat, line 145: As each entry is iterated over anyway in the typical case, this is now an `Array` instead of a `Set`.

packages/core/src/kernelCoreTemplate.js
```diff
@@ -513,16 +520,10 @@
       return protoChain
     }
 
-    function shouldAvoidProp(propsToAvoid, prop) {
-      for (const avoid of propsToAvoid) {
-        if (avoid instanceof RegExp && avoid.test(prop)) {
-          return true
-        }
-        if (propsToAvoid.has(prop)) {
```
legobeat, line 521: Doing this on each iteration of the inner loop was redundant as it doesn't reference `avoid`.

packages/core/src/kernelCoreTemplate.js
```diff
@@ -98,82 +97,90 @@
...
+        const desc = createDescriptor(globalRef, prop)
+        if (!desc) {
           continue
         }
+
         Object.defineProperty(globalRef, prop, desc)
       }
     }
 
+    function createDescriptor(globalRef, prop) {
```
legobeat, line 152: The name of this function was arbitrarily picked, suggestions welcome.

PR #626, opened 2023-07-20
[DRAFT] experimental new release process: lerna-based

boneskull:
It modifies the Lerna configuration to publish independent packages (without `nx`) using conventional commits.  This will create/update CHANGELOGs and create GitHub Releases for each workspace upon publish.

This is what [Appium](https://github.com/appium/appium) does, and I believe the goals for our release process are nearly identical.  Please look at the releases/tags in that repo for what you get.

To publish, just run `npm run release`.  Note that we may want to add a `preversion` lifecycle script which runs a clean/build.

cc @legobeat


boneskull:
Pinned the dep of `keccak` in `packages/node/test/projects/2` to avoid patch failure, lockfile updates, and ultimately policy file updates.  Don't know if this is desirable.  

Also worth mentioning that, in `projects/1` while Yarn would have copied the files from `fake_node_modules` into `node_modules` (and thus they were under version control...but I'm not sure why), `npm` will create symlinks.  I didn't see any need to keep `projects/1/node_modules` under VCS.  If there's an equivalent to `--check-files` in npm, I'm not sure what it is.

boneskull:
I'm unsure about the failures here.  Could it be due to transitive deps that have changed?

boneskull:
* boneskull/npm
  * **PR #626** <a href="https://app.graphite.dev/github/pr/lavamoat/lavamoat/626" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/lavamoat/lavamoat/626?utm_source=stack-comment).

PR #620, opened 2023-07-13
feat(core): add a schema for policy files

boneskull:
Thought this might come in handy.  Needs titles, descriptions and such.

[This](https://snowplow.io/blog/introducing-schemaver-for-semantic-versioning-of-schemas/) seems like a reasonable idea for schema versioning.

To test in VS Code, you can add this property to any policy file:

```json
"$schema": "https://raw.githubusercontent.com/boneskull/LavaMoat/boneskull/policy-schema/packages/core/schema/lavamoat-policy.v0-0-1.schema.json"
```

PR #617, opened 2023-07-11
feat(allow-scripts): generate & ship typescript declarations

boneskull:
This is a PoC for creating (and shipping) declaration files for LavaMoat packages.

Right now, I've fully typed the sources (not the tests) of `@lavamoat/allow-scripts`, which now validates the types in "strict" mode.  TypeScript is configured for incremental builds (preferable for monorepos), but only builds one module as a PoC.

To generate types, execute `npm run build:types` in the `@lavamoat/allow-scripts` workspace.  They will be output into `packages/allow-scripts/types` (which is not under VCS).

I added available type declarations (from DefinitelyTyped), but I had to write a handful of my own for external modules. These do not appear to be re-exported.  AFAICT (but I haven't tested it!) there aren't any type declarations that are needed as production dependencies (in `@lavamoat/allow-scripts`, anyway).

I took some liberties tweaking existing types (narrowing where appropriate, separating `@typedef`s so that descriptions would be picked up, etc.).  I'd prefer if we used the `T[]` syntax for arrays instead of `Array<T>`, but that's just my personal preference and I stuck to the current convention.

I note that these `.d.ts` files I created have _semicolons_ which is a byproduct of my IDE's default formatting.  If we intend to keep any of these around, I'd recommend pulling in `@typescript-eslint/plugin` and its ilk.

PR #627, opened 2023-07-21
chore: consolidate eslint config

boneskull:
The resulting config is my best guess at the intent of the current configuration.  In addition to linting the repo (the "lint everything" commit), this change:

- Removes all `.eslintignore` , `.eslintrc` files or `package.json`-embedded config and replaces it with a single root `.eslintrc.js`
- Removes all lint-related scripts from all workspaces (including `depcheck`)
- Removes all dev deps on `eslint` and its ilk (and `ava`, since it will come from the workspace root)
- Upgrades `@metamask/eslint-config-nodejs`, `eslint`, `eslint-plugin-ava`, as well as other plugins used by specific packages; moves them all to workspace root
- Removes unused/deprecated `eslint-plugin-node` in lieu of `eslint-plugin-n`, which the new version of `@metamask/eslint-config-nodejs` needs
- Adds a prettier config (in case someone wants to use it; I'd like to add it to our workflow later)

I note that ESLint seemed to be misconfigured before this change; none of the rules specified in the root config were being checked.  This means that there are a whole lot of lint fixes that need to be made.

To lint, run `npm run lint` (which will also run `depcheck`) from the workspace root.  Likewise, `lint:fix` runs a fix (across the whole repo), and `lint:depcheck` will use lerna to run `lint:depcheck` wherever it is defined.


boneskull:
I note that `depcheck` is not monorepo-aware, and depending on what we want from it, could be replaced with [eslint-plugin-import](https://npm.im/eslint-plugin-import).

boneskull:
Worth mentioning there were multiple references to eslint rules from endo's custom config in `core/lib/strict-scope-terminator.js`.  This config doesn't appear anywhere else--I assume it was from a copy/paste--and it wasn't noticed because the aforementioned misconfiguration.

boneskull:
* main
    * **PR #632** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/632" target="_blank">Graphite</a> 
    * **PR #631** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/631" target="_blank">Graphite</a> 
    * **PR #627** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/627" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/627?utm_source=stack-comment).

packages/lavapack/.depcheckrc
```diff
@@ -0,0 +1,3 @@
+ignores:
+  # monorepo deps
+  - "ava"
```
naugtur, line 3: Why is this needed?

boneskull: because `depcheck` is _not_ monorepo-aware, `ava` will be referenced in e.g., `lavapack/test/index.js`, but it will not appear in the `devDependencies` of `lavapack/package.json`.

any dev dep in a workspace which is required from code (unsure if this applies to dev deps only referenced in e.g., `package.json`'s `scripts` prop) but is only in the workspace root's dev deps would need to be in this `ignore` property.

boneskull:
just gonna retarget this one to #629

boneskull:
I'd like to merge #639 into this before merging this into `main`.

.eslintrc.js
```diff
@@ -0,0 +1,116 @@
...
+    // plugin rules
+
+    // all of these override stuff set in @metamask/eslint-config-nodejs.
+    'n/no-extraneous-require': 0,
+    'n/global-require': 0,
+    'n/exports-style': 0,
+    'n/no-process-env': 0,
+    // this rule seems broken
+    'n/no-unpublished-require': 0,
+    // we should probably actually fix these three and disable this override.
```
legobeat: q for later: What's the upside of `n/no-process-exit` here? Agreed on the two other.

legobeat (approved):
Needs a rebase before merge but LGTM otherwise!

.eslintrc.js
boneskull: Not sure exactly what you're asking.  `process.exit()` can be problematic if any of your code is async, because it exits before the loop is clear.  It's safer to just let the event loop die.  Anyway; it's in the codebase, and I didn't want to remove it, so I disabled the rule for now.

legobeat:
5e27065ef8d7f9f2adb718c4dbddd39d3675afda & 85e402f8d98479868fbe97060e08431505fa95bd should be squashed prior to merge

PR #628, opened 2023-07-24
chore: bootstrap releases for path: .

boneskull:
Configuring release-please for path: .


boneskull:
hm, no, this isn't right.

PR #630, opened 2023-07-24
chore(ci): add a release-please configuration

boneskull:
This adds configuration for releasing with the [release-please action](https://github.com/google-github-actions/release-please-action).

The way it works, in a nutshell:

- a commit with a [conventional commits](https://conventionalcommits.org)-valid message gets merged into `main`
- the action runs and creates _one PR per package_, with a new release ready to go, including updates to various CHANGELOGs, etc.
- another commit is merged
- the action either updates its existing PRs or creates new ones
- once a release PR is merged, the github release happens automatically
- the `npm publish` must be run manually by a human

PR #629, opened 2023-07-24
chore: replace yarn with npm

boneskull:
This removes yarn in lieu of npm.

I'm unsure if this affects `@lavamoat/yarn-plugin-allow-scripts` or not.

Two further PRs will build on this.  It is part of the "new release process" effort.

PR #631, opened 2023-07-24
chore: reconfigure lerna for releases

### LINKED ISSUE #227: Changelog
There appears to be no changelog for any of the projects in this repo. A changelog would make it a great deal easier for users of these packages to discover what has changed between releases. The commit history has a lot of noise to sift through that is not relevant for users of the packages.

boneskull:
**This is the same as #626 but the branch is no longer in my fork.**

It modifies the Lerna configuration to publish independent packages (without nx) using conventional commits. This will create/update CHANGELOGs and create GitHub Releases for each workspace upon publish.

This is what Appium does, and I believe the goals for our release process are nearly identical. Please look at the releases/tags in that repo for what you get.

To publish, just run npm run release. Note that we may want to add a preversion lifecycle script which runs a clean/build.

cc @legobeat

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,18 @@
 {
+  "$schema": "https://json.schemastore.org/lerna",
+  "changelogPreset": "conventional-changelog-conventionalcommits",
+  "command": {
+    "publish": {
+      "message": "chore: publish"
+    },
+    "version": {
+      "allowBranch": "main",
+      "conventionalCommits": true,
+      "createRelease": "github"
+    }
+  },
+  "ignoreChanges": ["**/test/**", "**/*.md"],
   "npmClient": "npm",
-  "useWorkspaces": true,
+  "useNx": false,
   "version": "independent"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,11 +13,12 @@
       "devDependencies": {
         "@metamask/eslint-config-nodejs": "^11.0.1",
         "ava": "^5.2.0",
+        "conventional-changelog-conventionalcommits": "^6.1.0",
         "depcheck": "^1.4.0",
         "eslint": "^8.40.0",
         "eslint-plugin-ava": "^13.2.0",
         "eslint-plugin-node": "^11.1.0",
-        "lerna": "^6.6.2"
+        "lerna": "^7.1.4"
       },
       "engines": {
         "node": ">=14.0.0"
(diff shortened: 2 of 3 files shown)


boneskull:
* main
    * **PR #632** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/632" target="_blank">Graphite</a> 
    * **PR #631** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/631" target="_blank">Graphite</a>  👈
    * **PR #627** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/627" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/631?utm_source=stack-comment).

lerna.json
```diff
@@ -1,5 +1,19 @@
...
+      "message": "chore: publish"
+    },
+    "version": {
+      "allowBranch": "main",
+      "conventionalCommits": true,
+      "createRelease": "github"
+    }
+  },
+  "ignoreChanges": ["**/test/**", "**/*.md"],
   "npmClient": "yarn",
```
boneskull:
```suggestion
  "npmClient": "npm",
```

leotm: we should commit this some point right?

if not here then in
- https://github.com/LavaMoat/LavaMoat/pull/629

> so that lerna can adjust some of its internal logic when resolving configuration and packages

lerna.json
```diff
@@ -1,5 +1,19 @@
...
+  "useNx": false,
```
leotm, line 16: nit: looks `false` by default so we could omit this config to keep it simpler/minimal if we want, since explicitly opt-in

leotm (approved):

leotm:
> Note that we may want to add a preversion lifecycle script which runs a clean/build.

captured this follow-up item here
- https://github.com/LavaMoat/LavaMoat/issues/634

leotm:
looks like this fixes couple yo issue
- https://github.com/LavaMoat/LavaMoat/issues/227

lerna.json
```diff
@@ -1,5 +1,19 @@
 {
+  "$schema": "https://json.schemastore.org/lerna",
+  "changelogPreset": "conventional-changelog-conventionalcommits",
```
leotm, line 3: nit: can further customise this later to how we like
- https://github.com/LavaMoat/LavaMoat/issues/636

lerna.json
boneskull: yeah, good catch.

lerna.json
boneskull: I don't trust that it will always be that way.

lerna.json
boneskull: @leotm I'm not sure I understand what your suggestion is

leotm: i remember seeing some prettified changelog in the past so wondered how it's done

till spotted `changelogPreset` also takes an obj!

```ts
// https://github.com/lerna/lerna/blob/main/libs/core/src/lib/conventional-commits/constants.ts
export type ChangelogPresetConfig = string | { name: string; [key: string]: unknown };
export type BaseChangelogOptions = {
  changelogPreset?: ChangelogPresetConfig;
  changelogEntryAdditionalMarkdown?: string;
  // etc
```

pos since 6.2.0 (2022-12-13) looks like

- https://github.com/lerna/lerna/issues/3441
> The current schema only allow for a string based "changelogPreset", but lerna also supports an object syntax for this property.

but yea not important

PR #632, opened 2023-07-25
chore: add commit hooks

boneskull:
This change: 

- adds a pre-commit hook that runs `eslint --fix` on any JS file on the stage.
- adds a commit message check that verifies the commit message is in conventional commits format.
- adds a new CI workflow to check commit messages in pull requests _which target `main`_

We will definitely want this if we want better automation of our releases.  Please see [these docs](https://github.com/conventional-changelog/commitlint/#what-is-commitlint) for more info about [commitlint](https://npm.im/commitlint) and how we can further configure it.  I have created a [configuration](https://github.com/LavaMoat/LavaMoat/blob/boneskull/husky/commitlint.config.js) which adopts the [conventional commits](https;//conventionalcommits.org) format but also relaxes some of the more pedantic default rules.

> Note: I'm unsure if using a `prepare` script is OK here.  This command tells `husky` to install the hooks into the `.git/` directory, and is only run in a development environment upon `npm install`

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -1,0 +1,14 @@
+name: Validate Commit Messages
+on: [pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  commitlint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3
+      - name: Run commitlint
+        uses: wagoid/commitlint-github-action@6319f54d83768b60acd6fd60e61007ccc583e62f #v5
--- a/package.json
+++ b/package.json
@@ -6,6 +6,8 @@
     "util": "^0.12.5"
   },
   "devDependencies": {
+    "@commitlint/cli": "^17.6.7",
+    "@commitlint/config-conventional": "^17.6.7",
     "@metamask/eslint-config-nodejs": "^11.0.1",
     "ava": "^5.2.0",
     "conventional-changelog-conventionalcommits": "^6.1.0",
@@ -13,7 +15,9 @@
     "eslint": "^8.40.0",
     "eslint-plugin-ava": "^13.2.0",
     "eslint-plugin-node": "^11.1.0",
-    "lerna": "^7.1.4"
+    "husky": "^8.0.3",
+    "lerna": "^7.1.4",
+    "lint-staged": "^13.2.3"
   },
   "engines": {
     "node": ">=14.0.0"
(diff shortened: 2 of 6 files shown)


boneskull:
* main
    * **PR #632** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/632" target="_blank">Graphite</a>  👈
    * **PR #631** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/631" target="_blank">Graphite</a> 
    * **PR #627** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/627" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/632?utm_source=stack-comment).

package.json
```diff
@@ -22,12 +26,20 @@
     "test:prep": "lerna run test:prep",
     "lint": "npm --workspaces run lint",
     "lint:fix": "npm --workspaces run lint:fix",
+    "lint:staged": "lint-staged",
+    "lint:commit": "commitlint",
     "preversion": "npm run rebuild",
+    "prepare": "husky install",
```
naugtur: `prepare`  runs on publishing too. Why not `postinstall`?

sidenote: husky is fairly complex. Do we need more than https://www.npmjs.com/package/pre-commit ?
(edit)
Oh, wait - pre-commit won't handle checking the message. nevermind

legobeat: I went for https://github.com/toplenboren/simple-git-hooks last time I had to pick one FWIW. `pre-commit` drags along a bit too much deps too be worth it imo.

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,17 @@
+name: Validate Commit Messages
+on:
+  [pull_request]
```
naugtur: +1 to running it as early as possible, not only for things targeting master. 
This may occasionally be annoying when doing messy experiments, but it's totally worth it.

naugtur (approved):
Please add a section to the README documenting the commit convention matching commitlint.config.js

package.json
naugtur: sounds good. husky is also fine.

Let's address the lifecycle script choice though.

boneskull: ~~I think you may be right about `postinstall` if we want to avoid `husky install` on publish--but only because we're in a workspace root.  `prepare` should probably be calling `npm run rebuild` instead of `preversion`, then.~~

wait, because we have `allow-scripts` configured, it doesn't run automatically anyway.

anyhow, I've changed it to this:

- `postinstall` runs `husky install`
- `prepare` runs `rebuild`, which calls the `rebuild` script in workspaces where it is extant
- `preversion` removed
- add `$root = true` to `allowScripts` config

boneskull:
Here's an example of the commitlint job running: https://github.com/LavaMoat/LavaMoat/actions/runs/5685783177/job/15411381654?pr=632

.github/workflows/commitlint.yml
boneskull: lol, I just reverted this.  I'm not sure it is worth it, honestly.  but I'll put it back.  

note that it does not run on `push` because, well, if you've pushed to `main` it's too late anyway.  maybe we need a branch protection rule (if we don't have one already) that disallows direct pushes to `main`.

PR #651, opened 2023-08-15
chore(node,core): add more test types

boneskull:
Trying to understand better how these tests and scenarios interact.

### DIFF
--- a/packages/core/.depcheckrc.yml
+++ b/packages/core/.depcheckrc.yml
@@ -14,4 +14,5 @@ ignores:
   - 'one'
   # types
   - '@types/json-stable-stringify'
+  - '@types/merge-deep'
   - '@babel/types'
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -43,6 +43,7 @@
   },
   "devDependencies": {
     "@types/json-stable-stringify": "1.0.36",
+    "@types/merge-deep": "3.0.3",
     "ses": "1.1.0",
     "tmp-promise": "3.0.3"
   },
--- a/packages/core/test/scenarios/transforms.js
+++ b/packages/core/test/scenarios/transforms.js
@@ -1,4 +1,4 @@
-const { createScenarioFromScaffold } = require('../util.js')
+const { createScenarioFromScaffold } = require('../util')
 
 module.exports = [
   async () => {
--- a/packages/node/src/defaults.js
+++ b/packages/node/src/defaults.js
@@ -22,6 +22,8 @@ const defaultPaths = getDefaultPaths('node')
  * @property {string} [projectRoot]
  * @property {boolean} [debugMode]
  * @property {boolean} [statsMode]
+ * @property {string} [policy] - Alias for `policyPath`?
+ * @property {string} [policyOverride] - Alias for `policyOverridePath`?
  */
 
 /** @type {LavaMoatOpts} */
--- a/packages/node/src/defaults.js
+++ b/packages/node/src/defaults.js
@@ -11,19 +11,20 @@ const defaultPaths = getDefaultPaths('node')
 
 /**
  * @typedef LavaMoatOpts
- * @property {LavaMoatScuttleOpts} [scuttleGlobalThis]
+ * @property {LavaMoatScuttleOpts} [scuttleGlobalThis] Enable or disable
+ *   scuttling of `globalThis`
  * @property {string[]} [scuttleGlobalThisExceptions]
- * @property {boolean} [writeAutoPolicy]
- * @property {boolean} [writeAutoPolicyDebug]
- * @property {boolean} [writeAutoPolicyAndRun]
- * @property {string} [policyPath]
- * @property {string} [policyDebugPath]
- * @property {string} [policyOverridePath]
- * @property {string} [projectRoot]
- * @property {boolean} [debugMode]
- * @property {boolean} [statsMode]
- * @property {string} [policy] - Alias for `policyPath`?
- * @property {string} [policyOverride] - Alias for `policyOverridePath`?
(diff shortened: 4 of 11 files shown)


boneskull:
* **#651** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/651?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

boneskull:
It looks like some of the types (like "file") aren't test-specific, though.  These are defined in `test/scenarios/scenario.d.ts`: `ScenarioFile`, `NormalizedScenarioFile`, `NormalizedScenarioJSFile`, etc.

packages/core/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "devDependencies": {
     "@types/json-stable-stringify": "^1.0.34",
+    "@types/merge-deep": "^3.0.1",
```
leotm: is there a reason we went for v3.0.1 instead of 3.x or v4

leotm: ah since `lavamoat-core`'s on v3.0.x

```console
➜  LavaMoat git:(boneskull/more-test-types) npm list merge-deep
lavamoat-monorepo@0.0.0 /Users/leo/Documents/GitHub/LavaMoat
└─┬ lavamoat-core@15.0.0 -> ./packages/core
  └── merge-deep@3.0.3
```

leotm: oops was looking at the [wrong](https://github.com/TehShrike/deepmerge) lib earlier (3.x/v4 doesn't exist for ours)

so 3.0.3 looks correct here? to fully align

```suggestion
    "@types/merge-deep": "^3.0.3",
```

edit: ok spotted

> ...only the major and minor release numbers are aligned between library packages and type declaration packages. 

so i guess aligning our patch versions doesn't rly matter 🤷‍♂️

leotm: sidenote Qu: any idea what the DefinitelyTyped [`.9999`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/merge-deep/package.json#L4) patch ver convention is for?

edit: ah

> (The `.9999` is to ensure that local `@types` packages are always newest during local development.)

packages/core/test/scenarios/transforms.js
```diff
@@ -46,6 +48,7 @@ module.exports = [
       name: 'transforms - common pattern "_inheritsLoose" works across package boundaries',
       defineOne: () => {
         /* eslint-disable */
+        // @ts-ignore
```
leotm: which errors are we suppressing btw?
i spotted no errors removing our 3 `// @ts-ignore` comments
and our types still rebuild fine

packages/core/.depcheckrc.yml
```diff
@@ -13,3 +13,4 @@ ignores:
   - 'one'
   # types
   - '@types/json-stable-stringify'
+  - '@types/merge-deep'
```
leotm, line 17: why add it btw when we're not explicitly using it?

leotm: ah nvm we're implicitly providing types (for inference) when requiring (in couple places)

https://github.com/LavaMoat/LavaMoat/blob/f27bf4354c561f66a039921bb2ac6abb1a625819/packages/core/src/mergePolicy.js#L6

https://github.com/LavaMoat/LavaMoat/blob/f27bf4354c561f66a039921bb2ac6abb1a625819/packages/core/test/util.js#L7

packages/core/test/scenarios/transforms.js
boneskull: probably leftover from when I turned on type checks for all JS files and then turned it off after I realized what I'd be in for.

packages/core/.depcheckrc.yml
boneskull: yeah, we're explicitly using `@types/merge-deep` because we have a `require('merge-deep')` in a typechecked file.  unfortunately, I don't see how this depcheck tool could know that, if in the future, we _stop_ using `merge-deep`, then we should also drop `@types/merge-deep`.

boneskull: [dependency-cruiser](https://www.npmjs.com/package/dependency-cruiser) might be a more suitable alternative.  I haven't gotten very deep into it though.

naugtur (commented):
This has a lot of overlap with https://github.com/LavaMoat/LavaMoat/pull/752

I don't get why they're visible in both since they share a merge target of `core-types` https://github.com/LavaMoat/LavaMoat/pull/752

boneskull:
@naugtur I'm not sure what you're seeing here that is overlapping.  Anyway, I retargeted this to #753 since it was struggling with the ESLint settings.

packages/node/src/defaults.js
```diff
@@ -22,6 +22,8 @@ const defaultPaths = getDefaultPaths('node')
  * @property {string} [projectRoot]
  * @property {boolean} [debugMode]
  * @property {boolean} [statsMode]
+ * @property {string} [policy] - Alias for `policyPath`?
+ * @property {string} [policyOverride] - Alias for `policyOverridePath`?
```
leotm: worth trying to clarify what these two are before merge?
(the removing the question marks)

boneskull: these two props were only referenced in [prepareScenarioOnDisk](https://github.com/LavaMoat/LavaMoat/blob/8d866f3af42c79593fcf2201ef66d5b1b5855851/packages/core/test/util.js#L490), but it appears they were not actually used by any scenario.

I've removed them and changed the references to the correct `policyPath` and `policyOverridePath`.

leotm (approved):

PR #654, opened 2023-08-15
fix(lavamoat): fix main entry point

boneskull:
This was pointing to a non-existent file.

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -5,7 +5,7 @@
     "lavamoat-run-command": "src/run-command.js"
   },
   "version": "7.1.0",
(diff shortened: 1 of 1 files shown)


kumavis (approved):

packages/node/package.json
```diff
@@ -5,7 +5,7 @@
     "lavamoat-run-command": "src/run-command.js"
   },
   "version": "7.1.0",
-  "main": "index.js",
+  "main": "src/index.js",
```
naugtur, line 8: I don't know how, but it worked at least once. Maybe by importing via direct path.

PR #668, opened 2023-08-21
chore(workflows): update codeql configuration

naugtur:
Looks like this should fix the codeql workflow.

Also, removed a build step that's not supposed to be necessary for javascript


.github/workflows/codeql-analysis.yml
```diff
@@ -19,6 +19,7 @@ on:
     branches: [ main ]
   schedule:
     - cron: '29 10 * * 0'
+  workflow_dispatch:
```
naugtur, line 22: useful for testing it out

.github/workflows/codeql-analysis.yml
```diff
@@ -38,34 +39,18 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
```
boneskull: (should) use a SHA like so:

```yaml
      uses: actions/checkout@<sha> # tag=v3
```

note: `v3` is a _branch_ (by convention)--not a tag, despite the above comment.  

if/when we switch over to Renovate, it will manage this for us, keeping the SHAs updated.

boneskull (approved):

.github/workflows/codeql-analysis.yml
naugtur: fixed (assuming I found the right commit hashes)

.github/workflows/codeql-analysis.yml
```diff
@@ -38,34 +39,18 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3
```
boneskull: for renovate

```suggestion
      uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # tag=v3
```

.github/workflows/codeql-analysis.yml
```diff
@@ -38,34 +39,18 @@ jobs:
...
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v1
+      uses: github/codeql-action/init@e4262713b504983e61c7728f5452be240d9385a7 # v2
```
boneskull:
```suggestion
      uses: github/codeql-action/init@e4262713b504983e61c7728f5452be240d9385a7 # tag=v2
```

.github/workflows/codeql-analysis.yml
```diff
@@ -38,34 +39,18 @@ jobs:
...
-    #    uses a compiled language
-
-    #- run: |
-    #   make bootstrap
-    #   make release
+        config-file: ./.github/codeql/codeql-config.yml
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v1
+      uses: github/codeql-action/analyze@e4262713b504983e61c7728f5452be240d9385a7 # v2
```
boneskull:
```suggestion
      uses: github/codeql-action/analyze@e4262713b504983e61c7728f5452be240d9385a7 # tag=v2
```

boneskull (approved):

PR #671, opened 2023-08-22
chore(survey/deps): cheerio@^0.22.0->^1.0.0-rc12

legobeat:
CVE-2021-3803 / GHSA-rp65-9cf3-cjxr

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6054,37 +6054,119 @@
       "dev": true,
       "license": "MIT"
     },
-    "node_modules/cheerio": {
-      "version": "0.22.0",
-      "license": "MIT",
+    "node_modules/cheerio-select": {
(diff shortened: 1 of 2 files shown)


boneskull (commented):
IMO this dependency should be dropped entirely.  It was last released a year ago (a prerelease, no less), and has over 600 automated pull requests merged into it since then.  I don't trust it'll _ever_ see another release, but who knows.  Given we don't use it for much of anything, it's likely we can find another suitable replacement.

legobeat:
@boneskull Sounds reasonable but unless we have a patch for that ready I still think it makes sense to resolve the outstanding issues addressed in this update for the time being?

boneskull (approved):

PR #672, opened 2023-08-22
chore(node): add file generated during test

legobeat:


boneskull (approved):

boneskull:
I'd like to understand why this file is getting created, but I don't.

weizman (approved):

PR #673, opened 2023-08-22
fix(core,perf): bump ses to 0.18.7 and reenable lib:ses script

weizman:
ctx #667

### DIFF
--- a/packages/perf/package.json
+++ b/packages/perf/package.json
@@ -26,7 +26,7 @@
     "lavamoat": "^7.0.0",
     "readable-stream": "^3.6.0",
     "@endo/compartment-mapper": "^0.8.4",
-    "ses": "^0.18.4"
+    "ses": "^0.18.7"
   },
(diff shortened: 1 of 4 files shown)


naugtur:
We've merged the commit message conventions already, you're going to need to `rebase -i` to get this to pass

packages/perf/package.json
```diff
@@ -26,7 +26,7 @@
     "lavamoat": "^7.0.0",
     "readable-stream": "^3.6.0",
     "@endo/compartment-mapper": "^0.8.4",
-    "ses": "^0.18.4"
+    "ses": "^0.18.7"
```
naugtur, line 29: (should) get npm to update the top level package-lock.

naugtur (approved):
Looks like the indentation was done on the previous commit with manual code update. I attest when I run lib:ses I get the same result as in this PR.

PR #649, opened 2023-08-11
fix(allow-scripts): update types

boneskull:
This adds `@typescript-eslint/parser` & its ilk; typescript files (just `.d.ts` for now) are now linted.

PR #638, opened 2023-07-31
feat(aa): ship generated types

boneskull:
This ships declaration files for `@lavamoat/aa`.

- added [@types/resolve](https://npm.im/@types/resolve) for types from [resolve](https://npm.im/resolve)
- remove some dead code; `canonicalNameMap` didn't need to be passed to `walkDependencyTreeForBestLogicalPaths`
- maybe fixed a bug in signature for `walkDependencyTreeForBestLogicalPaths` where `performantResolve` was `undefined` (or pulled out of global scope??)

PR #675, opened 2023-08-22
feat(allow-scripts): generate & ship typescript declarations

boneskull:
This is a PoC for creating (and shipping) declaration files for LavaMoat packages.

Right now, I've fully typed the sources (not the tests) of @lavamoat/allow-scripts, which now validates the types in "strict" mode. TypeScript is configured for incremental builds (preferable for monorepos), but only builds one module as a PoC.

To generate types, execute npm run build:types in the @lavamoat/allow-scripts workspace. They will be output into packages/allow-scripts/types (which is not under VCS).

I added available type declarations (from DefinitelyTyped), but I had to write a handful of my own for external modules. These do not appear to be re-exported. AFAICT (but I haven't tested it!) there aren't any type declarations that are needed as production dependencies (in @lavamoat/allow-scripts, anyway).

I took some liberties tweaking existing types (narrowing where appropriate, separating @typedefs so that descriptions would be picked up, etc.). I'd prefer if we used the T[] syntax for arrays instead of Array<T>, but that's just my personal preference and I stuck to the current convention.

I note that these .d.ts files I created have semicolons which is a byproduct of my IDE's default formatting. If we intend to keep any of these around, I'd recommend pulling in @typescript-eslint/plugin and its ilk.

### DIFF
--- a/config/tsconfig.base.json
+++ b/config/tsconfig.base.json
@@ -1,0 +1,13 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "allowSyntheticDefaultImports": true,
+    "checkJs": true,
+    "declaration": true,
+    "declarationMap": true,
+    "emitDeclarationOnly": true,
+    "composite": true,
+    "types": ["node"]
+  },
+  "extends": "@tsconfig/node14/tsconfig.json"
+}
--- a/config/tsconfig.base.json
+++ b/config/tsconfig.base.json
@@ -9,5 +9,5 @@
     "composite": true,
     "types": ["node"]
   },
-  "extends": "@tsconfig/node14/tsconfig.json"
+  "extends": "@tsconfig/node16/tsconfig.json"
 }
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "devDependencies": {
     "@commitlint/cli": "^17.6.7",
     "@commitlint/config-conventional": "^17.6.7",
-    "@tsconfig/node14": "^14.1.0",
+    "@tsconfig/node16": "^16.1.1",
     "@types/node": "^20.5.9",
     "@types/yargs": "^17.0.24",
     "ava": "^5.3.1",
@@ -23,9 +23,9 @@
     "glob": "^10.3.4",
     "husky": "^8.0.3",
     "lint-staged": "^13.2.3",
-    "typescript": "^5.2.2",
     "memfs": "^4.3.0",
-    "prettier": "^3.0.2"
+    "prettier": "^3.0.2",
+    "typescript": "^5.2.2"
   },
   "engines": {
     "node": ">=14.0.0 <=20.x"
--- a/package.json
+++ b/package.json
@@ -8,6 +8,9 @@
   "devDependencies": {
     "@commitlint/cli": "^17.6.7",
     "@commitlint/config-conventional": "^17.6.7",
+    "@tsconfig/node14": "^14.1.0",
+    "@types/node": "^20.5.9",
+    "@types/yargs": "^17.0.24",
     "ava": "^5.3.1",
     "conventional-changelog-conventionalcommits": "^6.1.0",
     "depcheck": "^1.4.0",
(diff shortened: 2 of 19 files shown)


boneskull:
* main
  * **PR #694** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/694" target="_blank">Graphite</a> 
    * **PR #688** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/688" target="_blank">Graphite</a> 
    * **PR #675** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/675" target="_blank">Graphite</a>  👈
      * **PR #649** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/649" target="_blank">Graphite</a> 
      * **PR #638** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/638" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/675?utm_source=stack-comment).

boneskull:
This is the same as #617, which was closed due to Reasons.

packages/allow-scripts/src/types/bin-links/shim-bin..d.ts
legobeat, line 1: Extraneous `.` in filename

leotm (approved):
perhaps a cheeky prettier write before merging? since we merged https://github.com/LavaMoat/LavaMoat/pull/725

Screenshot 2023-10-05 at 3 01 18 pm

package.json
```diff
@@ -8,6 +8,9 @@
   "devDependencies": {
     "@commitlint/cli": "^17.6.7",
     "@commitlint/config-conventional": "^17.6.7",
+    "@tsconfig/node16": "^16.1.1",
+    "@types/node": "^20.5.9",
```
leotm, line 12: afterthought: since we support v20

https://github.com/LavaMoat/LavaMoat/blob/b026356079a4bfbd0f39edd578373c07d100453c/package.json#L26-L28

and now v20 types

not worth supporting `@tsconfig/node20` in our tsconfig.base too?

config/tsconfig.base.json
```diff
@@ -0,0 +1,13 @@
...
+    "allowJs": true,
+    "allowSyntheticDefaultImports": true,
+    "checkJs": true,
+    "declaration": true,
+    "declarationMap": true,
+    "emitDeclarationOnly": true,
+    "composite": true,
+    "types": ["node"]
+  },
+  "extends": "@tsconfig/node16/tsconfig.json"
```
leotm, line 12: see https://github.com/LavaMoat/LavaMoat/pull/675/files#r1347518873

package.json
leotm: noting

https://raw.githubusercontent.com/tsconfig/bases/main/bases/node16.json
- compilerOptions
  - lib: es2021
  - target: es2021

https://github.com/tsconfig/bases/blob/main/bases/node20.json
- compilerOptions
  - lib: es2023
  - target: es2022

config/tsconfig.base.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
```
leotm, line 2:
```suggestion
  "compilerOptions": {
    "module": "nodenext",
```

https://www.typescriptlang.org/tsconfig#module
> You very likely want "nodenext" for modern node projects.

leotm: nvm looks like node16 anyway (and should do this for the `moduleResolution` too)

package.json
leotm: kinda off-topic but i'll share anyway
https://github.com/tsconfig/bases/blob/main/bases/node12.json
https://github.com/tsconfig/bases/blob/main/bases/node14.json
seemed strange at first seeing their module systems/resolutions in the future (node16) :suspect:
but makes sense since node10 module res only supports cjs, but node16 supports esm too
so the change was done across the board [4mo ago](https://github.com/tsconfig/bases/commit/a8c454f4a1caf4511b1e58a6b71ef4529264f628)

boneskull: it's a minimum compat thing, so since our minimum is node 16 we should use the node16 config.

Issue #451, opened 2023-01-12
Refactor lavapack runtime builder behavior


weizman:
### Current Situation 

- As of today, the 'runtime.js' and the 'runtime-cjs.js' files are generated files under 'lavapack/src/' directory which are derived from 'runtime-template.js' and 'runtime-cjs-template.js' accordingly. 
- The lavapack package exports logic to regenerate those ever since scuttling feature was introduced, so MM could generate the runtime kernel with the appropriate and desired scuttling configuration. 
- browserify package also runs on the generated runtime file, but as of now it doesn't generate it when is called, but counts on it to already exist in the lavapack package src directory. 

### Problems

- Generated files should not be included in the repository as a rule of thumb. 
- Counting on a file that is generated to a constant location won't allow multiple parallel calls to lavapack (e.g. a future need in MM to bundle multiple versions with different configurations in parallel)

### Desired Situation 

- Instead of writing to a constant location, make the builder return the built string to the caller for them to decide where to write it to. 
- Modify MM and browserify to deal with that update.
- Also make sure to update bad names `buildRuntime[CJS,ES,UMD]` as part of the process.

PR #676, opened 2023-08-23
chore(lavapack): ensure artifacts get built upon publish

boneskull:
This change:

- removes `packages/lavapack/src/runtime.js` and `packages/lavapack/src/runtime-cjs.js` from version control
- adds both to `packages/lavapack/.gitignore`
- removes `cross-env`; it is not needed to set `NODE_ENV=production`; this is only used in `viz` and webpack's default behavior is to use production mode _anyway_
- adds explicit build and test scripts to the `publish` script, since `pre-` and `post-` scripts cannot run.
- many updates to `RELEASE.md`


boneskull:
@naugtur I figured out how to do away with `cross-env`.

legobeat:
Could we also move it out from `./src` (into for example a new `./gen`) as part of this? OtoH the (great!) `cross-env` removal could be broken out separately

Also https://github.com/LavaMoat/LavaMoat/actions/runs/5957331050/job/16159825857?pr=676#step:7:1679

boneskull:
Looks like `pretest` isn't getting run, possibly....hm.

naugtur:
> Could we also move it out from ./src

Unfortunately that would be a breaking change - `src/` has become a public API surface and is depended upon by both MM-extension and snaps runtime environment. 

We'd have to do it with a major release (preferably containing a bigger refactor along with it)

boneskull:
> Unfortunately that would be a breaking change - src/ has become a public API surface and is depended upon by both MM-extension and snaps runtime environment.

We could fake it using an export map...

package.json
```diff
@@ -35,8 +34,8 @@
     "rebuild": "npm -w @lavamoat/lavapack -w @lavamoat/viz -w @lavamoat/yarn-plugin-allow-scripts run rebuild",
     "lint:staged": "lint-staged",
     "lint:commit": "commitlint",
-    "postinstall": "husky install",
-    "prepublishOnly": "cross-env NODE_ENV=production npm run rebuild && npm run test:prep && npm run test",
+    "postinstall": "husky install; npm run rebuild",
+    "prepublishOnly": "npm run --omit=dev rebuild && npm run test:prep && npm run test",
```
leotm: > * removes `cross-env`; it is not needed to set `NODE_ENV=production` when using `npm run` since `npm run --production` _sets it for you_

do we not want to add the `--production` flag here?

to preserve setting `NODE_ENV=production`

boneskull: `--production` is deprecated. `--omit=dev` is how it's expressed now. 

(note that this is not a choice I would have made.)

naugtur:
> We could fake it using an export map...

I'm not entirely sure they get resolved using contents of package.json in all cases, but it's not a documented feature so we can definitely move them around and replace their usage elsewhere.  
Preferably not in this PR blocking the release.

weizman:
Here's an old issue on the topic #451, which according to, it might make sense to not include the generated file in the published repo at all, and not even save the generated file to disk, but instead return the output to the caller to decide what to do with it.

Do we not want to consider that perhaps? The fact both internal and external code depend on those files to simply exist isn't ideal, because ever since scuttling, their content is destined to be configured differently (depends on who needs it), so what's the use of even having some version of it in the published version?

For example:

Under #676, running `yarn build` inside `packages/lavapack` outputs:

```javascript
...
// debug options are hard-coded at build time
    const {
      debugMode,
    } = {"debugMode":false}
    // security options are hard-coded at build time
    const {
      scuttleGlobalThis,
    } = __lavamoatSecurityOptions__
...
```

Where `__lavamoatSecurityOptions__` makes this files useless, and requires it to go through extra processing for whoever wants to use it. In MetaMask, we override and rebuild this file by calling `lavapack.buildRuntime(...)` which then overrides the file with:

```javascript
...
// debug options are hard-coded at build time
    const {
      debugMode,
    } = {"debugMode":false}
    // security options are hard-coded at build time
    const {
      scuttleGlobalThis,
    } = {scuttleGlobalThis: true, exceptions: ['fetch']}
...
```

Bottom line - we might not benefit a great deal from having the runtime file a part of the published package at all?

boneskull:
@weizman I think that should probably be a separate discussion--but if we envision non-programmatic usage by current or future consumers, we should definitely keep the file. 😄 

The aim of this change is to better ensure we're publishing what we currently intend to publish.

boneskull:
* main
  * **PR #676** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/676" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/676?utm_source=stack-comment).

naugtur:
@weizman I'm hoping to eventually get to a point where lavapack package no longer exists and everything is either in core or in packages using it.

boneskull (commented):
reorganize the manual publishing steps so that they a) come first and b) start at the "merge release PR" step

RELEASE.md
```diff
@@ -2,31 +2,53 @@
...
-## Automated Release Process
+## How to Publish to npm
 
-The [Release Please][] GitHub Action automates all parts of the release process _except_ the publish to the public npm registry.
+1. **Review, approve, then merge** a currently-open [Release Please][] pull request.
+   > See [How Releases Work](#how-releases-work) for more information.
+2. _Immediately after_ merging, **checkout** the `main` branch and **pull** the changes from [LavaMoat/LavaMoat][] into your working copy.
+   > The remote name depends on your working copy configuration; it may be `origin` or `upstream`—just make sure it's not pulling from a fork.
+   >
+   > This _should_ be a fast-forward operation; if it isn't, reset your working copy's `main` branch to this repo's `main` branch.
```
naugtur:
```suggestion
2. _Immediately after_ merging, **checkout** the `main` branch and **pull** the changes from [LavaMoat/LavaMoat][] into your working copy.
   > It is recommended you make a clean clone or use a separate local copy for publishing. 
   > Never publish from a fork.
   > Pulling `main` _should_ be a fast-forward operation; if it isn't, please start from a clean clone to be on the safe side.
```

RELEASE.md
```diff
@@ -2,31 +2,53 @@
...
+3. **Assert** you are logged in to the npm registry; use `npm login` and/or `npm whoami`. _Do not do this on an untrusted machine._
+4. **Assert** your working copy has no modifications (e.g., via `git status`).
+5. **Execute** `npm ci` ("Clean Install"; not "Continuous Integration").
+   > This operation will obliterate any `node_modules` and `packages/*/node_modules` directories before installation.
+6. Be ready with your 2FA passcode, then **execute** `npm run publish`. Artifacts will be rebuilt and the default set of tests will run.
```
naugtur:
```suggestion
6. Be ready with your 2FA passcode, then **execute** `npm run release`. Artifacts will be rebuilt and the default set of tests will run.
```

package.json
```diff
@@ -25,20 +24,22 @@
...
-    "lint:commit": "commitlint",
-    "postinstall": "husky install",
-    "prepublishOnly": "cross-env NODE_ENV=production npm run rebuild && npm run test:prep && npm run test",
-    "publish": "lerna publish",
-    "setup": "allow-scripts"
+    "lint": "npm run lint:eslint && npm run lint:deps",
+    "postinstall": "husky install; npm run rebuild",
+    "publish:lerna": "lerna publish",
+    "publish:rebuild": "npm run rebuild && npm run test:prep && npm run test:workspaces",
+    "publish": "npm run publish:rebuild && npm run publish:lerna",
```
naugtur:
```suggestion
    "release:lerna": "lerna publish",
    "release:rebuild": "npm run rebuild && npm run test:prep && npm run test:workspaces",
    "release": "npm run publish:rebuild && npm run publish:lerna",
```

naugtur (approved):
Approve + suggestion to rename the `publish` script to `release` to avoid accidents/confusion

boneskull:
@naugtur I've committed your suggestions as I also saw potential for confusion there.

FWIW I chose `publish` because when I write about a "release" in the documentation, I'm always talking about a _GitHub release_--which, by the time a contributor runs the publish script--has already happened.

boneskull:
squashing the committed suggestions

PR #677, opened 2023-08-23
chore(ci): do not use npm ci

boneskull:
`npm ci` rimrafs `node_modules` then installs everything in `package-lock.json`.  However, it is highly likely that the result of running `npm install` in a development environment will result in a very different tree, which updates `package-lock.json`.  If one of the dependencies installed ships a poorly-built shrinkwrap file, the updated `package-lock.json` will contain potentially incompatible software from the shrinkwrap. This will be manifested _upon the next `npm ci`_ because it _strictly_ installs what's in `package-lock.json`.

TL;DR: `npm ci` is fast, but using it exclusively in CI can be dangerous.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -20,15 +20,20 @@ jobs:
         uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3
         with:
           node-version: ${{ matrix.node-version }}
-          cache: "npm"
+          cache: 'npm'
       - name: Upgrade npm # for workspace support
         run: npm i -g npm
       - name: Install Dependencies
         uses: bahmutov/npm-install@1a235c31658a322a3b024444759650ee6345c26d # tag=v1
         with:
           useRollingCache: true
-          install-command: npm ci --foreground-scripts
+          install-command: npm install --foreground-scripts
       - name: Lint
         run: npm run lint
       - name: Test
         run: npm test
+
+      - # if `npm install` updated `package-lock.json`, this ensures
+        # a clean install works as expected
+        name: Check Lockfile
+        run: npm ci


boneskull:
This may or may not be a bug in `npm`, but the better solution is for projects to _not_ use shrinkwrap files.

legobeat:
I don't suppose you have a link or pointer to an example of this?

naugtur:
This is surprising to me. I was more leaning towards using `npm ci` exclusively and only running install when adding new dependencies or manually updating some.
We should install everything according to what's written in package-lock and care about it being updated in meaningful ways as we go (in PR reviews) 

In the interest of being clear - I think not installing packages according to package-lock is a much bigger issue than the possibility of a dependency making a mess with versions and nesting of packages if it uses shrinkwrap.  
As a result I would rather not introduce this change.

.github/workflows/nodejs.yml
```diff
@@ -20,15 +20,20 @@ jobs:
...
         with:
           useRollingCache: true
-          install-command: npm ci --foreground-scripts
+          install-command: npm install --foreground-scripts
       - name: Lint
         run: npm run lint
       - name: Test
         run: npm test
+
+      - # if `npm install` updated `package-lock.json`, this ensures
```
naugtur, line 36: This is no longer the list of packages known to us before. We should not have a workflow run with versions updated to newer than the ones we used locally.

boneskull:
@legobeat https://github.com/DavidAnson/markdownlint-cli2/issues/198

...but I'll close this, given @naugtur's comment.  Things can get very ugly if a package starts shipping a shrinkwrap file and a dev doesn't notice; but I agree it's likely a rare occurrence.  Socket will catch this sort of thing, but only if you click thru to the details about the package change.

PR #678, opened 2023-08-25
fix(tofu): parse private names

### LINKED ISSUE #642: lavamoat-tofu Error: unknown ast node type when trying to get name: "PrivateName"
happens when going through @fastify/http-proxy

Error: unknown ast node type when trying to get name: "PrivateName"
    at getNameFromNode (/app/node_modules/lavamoat-tofu/src/util.js:46:11)

boneskull:
This updates `getNameFromNode()` to understand what to do with private names, and also updates `findGlobals()` to ignore private names.

Closes #642

I probably want @kumavis to look at this, because I'm not sure of the intent here.  I'm fairly confident my change in `findGlobals()` is correct, but--what is the name returned by  `getNameForNode()` used for?

### DIFF
--- a/packages/tofu/src/util.js
+++ b/packages/tofu/src/util.js
@@ -42,7 +42,9 @@ function getNameFromNode (node) {
     return node.name
   } else if (node.type === 'ThisExpression') {
     return 'this'
-  } else {
+  } else if (node.type === 'PrivateName') {
+    return `#${node.id.name}`
+  }else {
     throw new Error(`unknown ast node type when trying to get name: "${node.type}"`)
   }
 }
@@ -139,7 +141,7 @@ function objToMap (obj) {
 function mapToObj (map) {
   const obj = {}
   map.forEach((value, key) => {
-    obj[key] = value 
+    obj[key] = value
   })
   return obj
 }
--- a/packages/tofu/src/findGlobals.js
+++ b/packages/tofu/src/findGlobals.js
@@ -55,6 +55,11 @@ function findGlobals (ast) {
         }
       }
 
+      // private names are not globals
+      if (parentType === 'PrivateName' && path.parent.id === path.node) {
+        return
+      }
+
       // save global
       saveGlobal(path)
     },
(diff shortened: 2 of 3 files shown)


packages/tofu/src/util.js
```diff
@@ -42,7 +42,9 @@ function getNameFromNode (node) {
     return node.name
   } else if (node.type === 'ThisExpression') {
     return 'this'
-  } else {
+  } else if (node.type === 'PrivateName') {
+    return `#${node.id.name}`
+  }else {
```
legobeat, line 47:
```suggestion
  } else {
```
lint

boneskull: ESLint doesn't enforce this.  If we want to worry about it, we should adopt `prettier`.

boneskull:
* main
  * **PR #678** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/678" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/678?utm_source=stack-comment).

packages/tofu/src/util.js
legobeat: There is tons of code-style we should try to enforce despite tooling not doing so. Descriptive variable- and function names come to mind.

Tooling like eslint and prettier is valuable but complimentary - not authoritative.

> If we want to worry about it

I don't.

boneskull: I'd advocate for them being authoritative--with the goal being eliminating all mention of style issues in PRs.

naugtur: I agree with both of you :sweat_smile: 
- machines should do the formatting
- we should do the right thing even if it's possible not to
- some level of automation is great. It stops being great when it gets in the way (eg. prettier demanding some variable names are not in use)
- style comments in PRs are a waste of time
- the reviewer's time is wasted but there's not much the reviewer can do to prevent it

I say we use autoformatting locally even if not enforced. It's what I've been doing and somehow it picked up on the existing config and formatted consistently with the existing code.
Would be great to not have to rely on being lucky like that tho. :) 

Let's close this conversation regardless of outcomes for the PR. I see a branch introducing prettier. Any style conversations can continue there.

naugtur, line 47:
```suggestion
  } else {
```

naugtur (approved):

PR #681, opened 2023-08-28
ci: Add alpine-based Dockerfile

legobeat:
- Adds an integration test running ~all~ basic package scripts inside an Alpine Linux Docker container to cross-check for any implicit runtime dependencies on the `ubuntu-latest` environment in GitHub Actions.
- Adds `Development` section to `CONTRIBUTING.md` with information on supported development environments.
  - Took the conservative stance here and only including tested environments (ubuntu+alpine) - considering those two, Debian seems safe enough as well.

The `node:18-alpine` base image does not have `glibc` and many other other packages common in other distros like Ubuntu, and comes with a different default shell.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,15 +10,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871  # v4
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@4c0219f9ac95b02789c1075625400b2acbff50b1 # v2
+        uses: docker/setup-buildx-action@c47758b77c9736f4b2ef4073d4d51994fabfe349  # v3
         with:
           driver: docker-container
           driver-opts: |
-            image=docker.io/moby/buildkit:master@sha256:1b6abfd9bbfbf129245dc1f288aa38f064dfee8d608b8be9086936f5580c78f2
+            image=docker.io/moby/buildkit:master@sha256:c8029e1a3e837c6e3b3c7dc97114385e902a2a321306bd6aa5c86da3ab6414db
       - name: Build OCI image
-        uses: docker/build-push-action@4a13e500e55cf31b7a5d59a38ab2040ab0f42f56 #v5
+        uses: docker/build-push-action@4f58ea79222b3b9dc2c8bbdd6debcef730109a75  # v6
         with:
           # Use local cache dir when running using Act; gha cache otherwise
           cache-from: type=${{ env.ACT && 'local,src=/tmp/act-cache' || 'gha' }}
@@ -27,39 +27,21 @@ jobs:
           file: ./Dockerfile
           load: true
           push: false
-          tags: dev:latest
-      - name: lint
-        uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185 # v3
-        with:
-          image: dev:latest
-          run: 'npm run lint && npm run lint:fix'
-      - name: publish-dry-run
-        uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185 # v3
-        with:
-          image: dev:latest
-          run: |
-            # note: package scripts only seem to support maximum a single unpublished package present in the repository
-            newpkg_path=$(jq -r 'to_entries|map(select(.value=="0.0.0"))[0].key' .release-please-manifest.json)
-            export npm_config_newpkg=$(jq -r .name "${newpkg_path}/package.json")
-            echo "New package: ${npm_config_newpkg}"
-            # these two commands should be identical
-            npm run release:dry-run
-            npm run release:publish -- --dryRun
+          tags: 'dev:latest'
       - name: rebuild
-        uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185 # v3
+        uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185  # v3
         with:
-          image: dev:latest
-          run: npm run setup && npm run rebuild
-      - name: lint:git
-        uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185 # v3
+          image: 'dev:latest'
+          run: |
+            npm run setup && npm run rebuild
+      - name: lint
+        uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185  # v3
         with:
-          image: dev:latest
-          run: >
-            git init -b main &&
-              git add README.md && npm run lint:staged &&
-              git commit -m 'docs: Add README.md' && npm run lint:commit -- -e
+          image: 'dev:latest'
+          run: 'npm run lint'
       - name: test
-        uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185 # v3
+        uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185  # v3
         with:
-          image: dev:latest
-          run: npm run test:prep && npm run test
+          image: 'dev:latest'
+          run: |
+            npm run test:prep && npm run test
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@c47758b77c9736f4b2ef4073d4d51994fabfe349  # v3
         with:
-          driver: docker-container
+          driver: docker-container # for advanced features like build caching when running Act locally
           driver-opts: |
             image=docker.io/moby/buildkit:master@sha256:c8029e1a3e837c6e3b3c7dc97114385e902a2a321306bd6aa5c86da3ab6414db
       - name: Build OCI image
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,65 @@
+name: CI-Docker
+on: [push, pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  test-dev-alpine:
+    name: test-dev-alpine
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@4c0219f9ac95b02789c1075625400b2acbff50b1 # v2
+        with:
+          driver: docker-container
+          driver-opts: |
+            image=docker.io/moby/buildkit:master@sha256:1b6abfd9bbfbf129245dc1f288aa38f064dfee8d608b8be9086936f5580c78f2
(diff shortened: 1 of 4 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,76 @@
...
+        with:
+          image: dev:latest
+          run: 'npm run lint && npm run lint:fix'
+      - name: publish
+        uses: addnab/docker-run-action@v3
+        env:
+          VERDACCIO_STORAGE_PATH: /app/verdaccio/storage
+        with:
+          image: dev:latest
+          run: >
```
legobeat: This whole block is quite ugly and due to the fact that lerna does not have an equivalent of `--dry-run` for publish - therefore we spin up an ephemeral registry here. Ideally this wouldn't be needed.

https://github.com/lerna/lerna/issues/51

.github/workflows/docker.yml
```diff
@@ -0,0 +1,68 @@
...
+  test-dev-alpine:
+    name: test-dev-alpine
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@4c0219f9ac95b02789c1075625400b2acbff50b1 # v2
+        with:
+          driver: docker-container
```
legobeat: This is necessary to support the `cache-from`/`cache-to` in the next step.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,68 @@
...
+          VERDACCIO_STORAGE_PATH: /app/verdaccio/storage
+        with:
+          image: dev:latest
+          run: >
+            git init -b main &&
+              git add package.json packages/*/package.json &&
+              git commit -m 'feat!: Add manifests' &&
+              npm ci &&
+              (verdaccio &) &&
+              sleep 7 &&
```
legobeat: If this proves flakey it could be replaced with something waits for the port to be open. Seems overkill at this point, though.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,68 @@
...
+              npm set registry http://localhost:4873 &&
+              sh -c '(echo user; sleep 0.5; echo password; sleep 0.5; echo email@example.com) | npm --registry=http://localhost:4873 adduser --auth-type legacy' &&
```
legobeat: giphy

boneskull (commented):
I guess I'm not entirely convinced the overhead of maintaining the container and the extra CI workflow is worth proving that we don't need anything special installed to build LavaMoat on Linux.  The main reason being that if there's something missing in whatever Linux distro someone's trying to use to build LavaMoat, I think we can safely assume they could install it themselves.  That said, it'd be nice to know what those things _are_--if they are.

What's `verdaccio` doing here?

legobeat:
@boneskull I could be convinced to drop the testing of the `publish` part as it's the more unwieldy one. But I don't think we can officially support things we don't test for.

> What's verdaccio doing here?

It acts as registry that the `publish` command publishes to (there are review comments as to why)

boneskull:
> But I don't think we can officially support things we don't test for.

Things aren't really that black-and-white.  We cannot possibly test on every Linux distro and version thereof, nor with every version of Node.js, nor macOS, nor Windows, etc., etc.  

So... I don't understand why we need the container.  Is this replicating an existing use-case?  In that case, I'd be more inclined to support it.  If this was implied, I missed it. 😄 

I'm also not sure what we're trying to test with the publishing bit.  IMO the more important thing is to test what installing what's in a tarball, what happens when you install it, and if it actually _works_ (which is what [midnight-smoker](/boneskull/midnight-smoker) is for).

legobeat:
Mostly the motivation is to cross-check on *something* else. Alpine is minimal and provides fair enough variation from Ubuntu that I think we can claim support beyond "runs on GitHub Actions ubuntu-latest image" (like in the new docs-section). 

I do think Windows and macOS tests should be added if Lavamoat is to support them.

This already caught https://github.com/LavaMoat/LavaMoat/issues/682, so I'd say it's useful.

From experience, integration-testing publish-scripts can prevent unpleasant surprises when you least need them for packages which are published infrequently or irregularly.

boneskull:
@legobeat From your experience with failed publishing, what was the failure mode?

legobeat:
- Rebased on `main`
- Removed verdaccio and simplified publish-testing workflow (since lerna is no longer used)
- Added support to run the GHA workflow locally under `act`

.github/workflows/docker.yml
naugtur: This needs documenting - I don't think any of us in the team would know why this is the way it is.

naugtur:
> Adds Development section to README.md with information on supported development environments.
That commit seems missing.

boneskull (commented):
I don't understand what much of anything in `docker.yml` does, and I'm loathe to have to maintain it.  If there was an easier way to run from a custom image, I'd be more amenable to it.

Overall, I'm not convinced the juice is worth the squeeze.  In the example given (#682), this caught an implicit dependency in the _test environment_.  I am not particularly worried about this sort of thing; I'm more worried about broken publishes and production bugs--and this addition does little to assuage that.

I'm not going to block this, but I don't believe it necessary to add.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,64 @@
...
+          run: npm run setup && npm run rebuild
+      - name: lint:git
+        uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185 # v3
+        with:
+          image: dev:latest
+          run: >
+            git init -b main &&
+              git add README.md && npm run lint:staged &&
+              git commit -m 'docs: Add README.md' && npm run lint:commit -- -e
+      - name: test
```
naugtur, line 42: I suggest we limit ourselves to only running this one.
It's the only user-facing test we need.

naugtur:
Let's revive this.

- the description says there should be a readme change
- I'd be willing to approve the PR with the docker-related complexity if we get rid of the part that tests our dev environment and stick to testing the software.

legobeat:
> Let's revive this.
> 
>     * the description says there should be a readme change
> 
>     * I'd be willing to approve the PR with the docker-related complexity if we get rid of the part that tests our dev environment and stick to testing the software.

- docs & description updated 
- bumped base image & actions
- limits CI docker actions to build/lint/test
- moved the containerfile from `test/integration/Dockerfile-alpine` to `Dockerfile`

boneskull:
@legobeat It looks like you might have removed the publishing tests?

legobeat:
> @legobeat It looks like you might have removed the publishing tests?

@boneskull yes, as mentioned in https://github.com/LavaMoat/LavaMoat/pull/681#issuecomment-2406493362.
I'd personally prefer to keep them but removal was requested in https://github.com/LavaMoat/LavaMoat/pull/681#issuecomment-2328766307

boneskull:
@legobeat OK, thanks.

I am going to abstain from reviewing, since I'm ambivalent about adding it.  I just don't know how to maintain this if it breaks and I worry about bus factor.

Dockerfile
```diff
@@ -0,0 +1,24 @@
+ARG NODE_VERSION=18
+ARG BASE_IMAGE_DIGEST=sha256:ea8e360a721d870337fe899c70ea7def62f2a72cf1b6f7beb8a3ccaac8b6049c
```
leotm: which version is this? could we add a comment (like we do in our .yml's)

.github/workflows/docker.yml
```diff
@@ -0,0 +1,47 @@
...
+  contents: read
+
+jobs:
+  test-dev-alpine:
+    name: test-dev-alpine
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871  # v4
+      - name: Set up Docker Buildx
```
leotm, line 14:
```suggestion
      - name: Set up Docker Build
```

leotm: ah nvm it's not a typo, it's in the name of the action it uses

.github/workflows/docker.yml
leotm:
```suggestion
          driver: docker-container # for advanced features like build caching when running Act locally
```

Dockerfile
legobeat: `node:18-alpine` - and sure, I added a comment!

leotm: 🤦‍♂️

CONTRIBUTING.md
```diff
@@ -93,15 +93,30 @@ This section describes how to configure your development environment to hack on
...
 
 #### Node.js
 
-LavaMoat is distributed as a [Node.js][node] command-line tool. You'll want to install the latest LTS version (see the [Node.js release schedule][release-schedule] for more information).
+LavaMoat is distributed as a [Node.js][node] command-line tool. You'll want to install the active LTS version (see the [Node.js release schedule][release-schedule] for more information).
 
 :::tip
 
-Don't have Node.js installed? You can [download and install Node.js from the official site][node-download], but we recommend using a Node.js version manager (e.g., [nvm][]).
+Don't have Node.js installed? You can [download and install Node.js from the official site][node-download].
```
leotm, line 119: why we switching from recommending a node version manager to not anymore?

leotm: noting we're introducing

https://github.com/LavaMoat/LavaMoat/blob/98bced7fe99dbf005455c2c0d50632614c9f91d3/Dockerfile#L6

with our current CI matrix range from 16 <--> 22

legobeat: Personally not a fan of the approach as a general recommendation - there are saner ways to manage Node.js runtimes and a specific recommendation feels out of place here. If anyone relies on containers for development, NVM will also most likely be redundant and (in best case) unused.

leotm: by node version manager i meant any _brand_, not `nvm` specifically (;
by saner d'you mean other version managers? or other approaches in mind

leotm: > If anyone relies on containers for development, NVM will also most likely be redundant and (in best case) unused.

true ^

i suppose recommending just use latest official version of node (like [RN on Linux](https://reactnative.dev/docs/set-up-your-environment?os=linux) setup) makes setup easier from the get-go too

CONTRIBUTING.md
```diff
@@ -93,15 +93,30 @@ This section describes how to configure your development environment to hack on
...
+
+- Latest Ubuntu LTS [`x86_64`]
+- Debian Stable [`x86_64`]
+- Latest Alpine Linux Stable [`x86_64`]
+
+On a best-effort basis, development should also be expected to work on:
+
+- macOS 13.5+ [`x86_64` / `aarch64`]
+- Windows Subsystem for Linux 1 [`x86_64`]
+- Windows Subsystem for Linux 2 [`x86_64` / `aarch64`]
```
leotm, line 106: didn't know about [WSL 2](https://learn.microsoft.com/en-us/windows/wsl/about#what-is-wsl-2) ^ nice

should we not expect WSL to work on `aarch64` too?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,47 @@
+name: CI-Docker
```
leotm, line 1: initially thought let's keep this labelled `CI` with job name `docker-test-dev-alpine`
but in hindsight current naming makes sense being our custom image

leotm (approved):

CONTRIBUTING.md
legobeat: > should we not expect WSL to work on `aarch64` too?

I guess but it also seems super niche and would be a hassle to debug for?

PR #685, opened 2023-08-28
examples: use workspace versions of lavamoat packages

legobeat:
example packages are currently not using repository versions of lavamoat packages, but fetching published versions from `npmjs.org` like with any other dep.

This changes them to use the workspace-internal packages.


weizman (approved):
lgtm (thank you)

boneskull:
An alternative to this may be just adding each example dir in the list of workspaces in the root `package.json`, which would create the links automatically and _might_ receive version bump updates via `release-please` (note that we'll also want to set each package to `private`).

legobeat:
> An alternative to this may be just adding each example dir in the list of workspaces in the root `package.json`, which would create the links automatically and _might_ receive version bump updates via `release-please` (note that we'll also want to set each package to `private`).

Hmm, would you be able to elaborate / do you have pointers/docs/references?  Don't think I follow exactly but may be due to new developments in npm I'm less familiar with.

boneskull:
@legobeat When you run `npm install` and there's stuff in `package.json`'s `workspaces`, `npm` will install dependencies for every package directory and it will automatically create symlinks in the root `node_modules` (or the workspace's `node_modules`, depending) for intra-project workspaces.  So if `packages/a` depends on `packages/b`, `npm` creates `node_modules/b` which is symlinked to `packages/b`; this way, `packages/a` can resolve it.

So if we add those example dirs as workspaces, the same thing will happen.  No `file:` dependencies or separate `npm install`s needed.  The examples' `package.json`s will just be the usual version, and whoever wants to try the example will have an easier time installing the dependencies.

`release-please` (if it works the way I think it should) will then ensure that the versions are kept up-to-date as dependencies get bumped; once `packages/b` gets bumped, `packages/a/package.json` will have its `dependencies.b` field updated with the new version.

boneskull:
IDK; you could try it and see how it works.  Be sure to add `private: true` to each `package.json`, though.

legobeat:
Ah, right. Do you think we can treat that as a separate improvement that can be done at a later stage? Your proposal sounds overall better but also comes with a few "if"s.

boneskull:
Depends on the urgency of this change, I suppose.  When are we installing the examples?  Are we using them as test fixtures? And if they are supposed to be examples, _shouldn't_ they pull from the registry?  I mean: whoever wants to try an example will now need to fix it before installing anyway, unless they clone the whole repo.

legobeat:
> Depends on the urgency of this change, I suppose. When are we installing the examples? Are we using them as test fixtures? And if they are supposed to be examples, _shouldn't_ they pull from the registry? I mean: whoever wants to try an example will now need to fix it before installing anyway, unless they clone the whole repo.

I believe we get a catch-22 in case of some breaking changes that affect the examples since they are used in the integration tests. It's the only mention left I see of `<19` here: https://github.com/legobeat/LavaMoat/actions/runs/5998297018/job/16266300490?pr=3#step:7:175

naugtur (approved):
I think this is the right way to go about it now. 
+1 to marking them as private

legobeat:
Unfortunately overlooked commitlint error https://github.com/LavaMoat/LavaMoat/actions/runs/6009764943/job/16299831782

PR #688, opened 2023-08-28
chore: add prettier

boneskull:
This adds `prettier` to our workflow.

Formatting happens as a pre-commit hook on relevant files _after_ ESLint has performed any automatic fixes successfully.

Rules were removed from our ESLint config which conflict with Prettier.

* * *

If this PR is left as-is, ESLint won't report any errors for "incorrectly" formatted code.

We _could_ add [eslint-plugin-prettier](https://npm.im/eslint-plugin-prettier) which treats _all_ incorrect formatting as an ESLint error.  This would obviate the need to run `prettier` directly, but it's probably going to be too noisy & verbose.


boneskull:
* main
  * **PR #694** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/694" target="_blank">Graphite</a> 
    * **PR #688** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/688" target="_blank">Graphite</a>  👈
    * **PR #675** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/675" target="_blank">Graphite</a> 
      * **PR #649** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/649" target="_blank">Graphite</a> 
      * **PR #638** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/638" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/688?utm_source=stack-comment).

.eslintignore
```diff
@@ -0,0 +1,14 @@
...
+packages/*/test/**/fixtures/**/*
+packages/*/test/projects/**/*
+packages/core/lib/*.umd.js
+packages/lavapack/bundle.js
+packages/lavapack/src/runtime-cjs.js
+packages/lavapack/src/runtime.js
+packages/perf/trials/**/*
+packages/survey/mitm/**/*
+packages/viz/dist/**/*
+packages/viz/src/example-policies/**/*
```
naugtur, line 14: I'm assuming this is extracted so that other tools pick it up. Otherwise we'd keep it in one file to minimize the number of top-level files.
Correct me if I'm wrong.

.vscode/settings.json
```diff
@@ -1,3 +1,7 @@
 {
-    "editor.tabSize": 2
-}
\ No newline at end of file
+  "editor.tabSize": 2,
+  "prettier.semi": false,
+  "prettier.singleQuote": true,
+  "prettier.trailingComma": "all",
+  "editor.defaultFormatter": "esbenp.prettier-vscode"
```
naugtur, line 3: Why wouldn't vscode's prettier know that from our config?

naugtur:
What's the scale of changes this triggers in the repository if we'd run prettier on everything now? 
It's be best if they were tiny. :crossed_fingers:

.eslintignore
boneskull: Yes.

.vscode/settings.json
boneskull: mm, it should know the settings, but it doesn't know what the default formatter is.  I'll trim that stuff out we don't need

boneskull:
> What's the scale of changes this triggers in the repository if we'd run prettier on everything now?
> It's be best if they were tiny. 🤞

Currently it creates a diff that is 26317 lines long.

boneskull:
generally with this stuff I think it's better to just do the reformat in one swoop, but it will create conflicts in every open PR.  should probably wait until we get a release out, at least, and I can send another PR that does it.  

the command to do this is:

```bash
npm exec prettier --write . 
```

boneskull:
I created #725 if we want to format everything in one swoop.

kumavis (approved):

PR #691, opened 2023-08-30
ci: run tests on windows and macos

### LINKED ISSUE #679: add windows to CI matrix
I guess it sounds so simple when you put it that way.

Anyway: do this and fix everything that breaks.

legobeat:
This runs unit tests on `macOS-latest` and `windows-latest`, in addition to existing `ubuntu-latest`.

Only runs on node version from `.nvmrc` (currently v18).

#### Related

- Resolves #679
- WIP for actually making tests pass on Windows: https://github.com/legobeat/LavaMoat/pull/7

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -59,3 +59,51 @@ jobs:
         run: npm run lint:deps
       - name: Validate Lockfile
         run: npm ls
+
+  compatibility-test-windows:
+    name: Test Windows compatibility
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+    env:
+      FORCE_COLOR: 1
+    steps:
+      - name: Checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@b39b52d1213e96004bfcb1c61a8a6fa8ab84f3e8 # v4
+        with:
+          node-version-file: .nvmrc
+          cache: npm
+      - name: Install Dependencies
+        uses: bahmutov/npm-install@d476752204653fb5cce6c09db0eaf220761f5d9e # v1.8.37
+        with:
+          useRollingCache: true
+          install-command: npm ci --foreground-scripts
+      # Limit testing to packages supported to run under windows
+      - name: Test
+        run: |
+          npm run build
+          npm run --if-present --workspace=packages/aa --workspace=packages/allow-scripts --workspace=packages/browserify --workspace=packages/core --workspace=packages/lavapack --workspace=packages/node --workspace=packages/preinstall-always-fail --workspace=packages/tofu --workspace=packages/webpack --workspace=packages/yarn-plugin-allow-scripts test:prep
+          npm run --if-present --workspace=packages/aa --workspace=packages/allow-scripts --workspace=packages/browserify --workspace=packages/core --workspace=packages/lavapack --workspace=packages/node --workspace=packages/preinstall-always-fail --workspace=packages/tofu --workspace=packages/webpack --workspace=packages/yarn-plugin-allow-scripts test
+
+  compatibility-test-macos:
+    name: Test macOS compatibility
+    runs-on: macOS-latest
+    env:
+      FORCE_COLOR: 1
+    steps:
+      - name: Checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@b39b52d1213e96004bfcb1c61a8a6fa8ab84f3e8 # v4
+        with:
+          node-version-file: .nvmrc
+          cache: npm
+      - name: Install Dependencies
+        uses: bahmutov/npm-install@d476752204653fb5cce6c09db0eaf220761f5d9e # v1.8.37
+        with:
+          useRollingCache: true
+          install-command: npm ci --foreground-scripts
+      - name: Test
+        run: npm test
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -67,6 +67,9 @@ jobs:
       fail-fast: false
     env:
       FORCE_COLOR: 1
+      TEST_WORKSPACES: >-
+        --workspace=packages/lavapack
+        --workspace=packages/preinstall-always-fail
     steps:
       - name: Checkout
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
@@ -83,9 +86,10 @@ jobs:
       # Limit testing to packages supported to run under windows
       - name: Test
         run: |
-          npm run build
-          npm run --if-present --workspace=packages/aa --workspace=packages/allow-scripts --workspace=packages/browserify --workspace=packages/core --workspace=packages/lavapack --workspace=packages/node --workspace=packages/preinstall-always-fail --workspace=packages/tofu --workspace=packages/webpack --workspace=packages/yarn-plugin-allow-scripts test:prep
-          npm run --if-present --workspace=packages/aa --workspace=packages/allow-scripts --workspace=packages/browserify --workspace=packages/core --workspace=packages/lavapack --workspace=packages/node --workspace=packages/preinstall-always-fail --workspace=packages/tofu --workspace=packages/webpack --workspace=packages/yarn-plugin-allow-scripts test
+          npm run --if-present ${{ env.TEST_WORKSPACES }} build
+          npm run build:types
+          npm run --if-present ${{ env.TEST_WORKSPACES }} test:prep
+          npm run --if-present ${{ env.TEST_WORKSPACES }} test
 
   compatibility-test-macos:
     name: Test macOS compatibility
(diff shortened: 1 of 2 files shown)


legobeat:
#692 discovered through this.

.github/workflows/nodejs.yml
```diff
@@ -32,3 +32,28 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macOS-latest, windows-latest]
+    env:
+      FORCE_COLOR: 1
+    steps:
+      - name: Checkout
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3
```
boneskull:
```suggestion
        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # tag=v3
```

for renovate

.github/workflows/nodejs.yml
```diff
@@ -32,3 +32,28 @@ jobs:
...
+      - name: Use Node.js
+        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3
```
boneskull:
```suggestion
        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # tag=v3
```

.github/workflows/nodejs.yml
```diff
@@ -32,3 +32,28 @@ jobs:
...
+        with:
+          node-version-file: .nvmrc
```
boneskull, line 105: If we're going to keep this as its own job, can we just use a static version here?  `latest` or `lts` or whatever the default is.  I don't really want to add a build dependency upon `.nvmrc`.  I have mixed feelings about that file. :P

.github/workflows/nodejs.yml
```diff
@@ -32,3 +32,28 @@ jobs:
         run: npm run lint
       - name: Test
         run: npm test
+
+  platform-compatibility-test:
+    name: Test platform compatibility
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
```
boneskull, line 67: I guess I'd lean more towards just adding these to the main build matrix.  Why not?

boneskull:
@legobeat I think we need to allow scripts for `nx` so it can grab/build its native module(s).

legobeat:
> @legobeat I think we need to allow scripts for `nx` so it can grab/build its native module(s).

Hmm, if you're thinking about this CI error, I think that may be due to npm cache reuse from the ubuntu install.

```
> npm run build && npm run test:prep && npm run test:workspaces


> lavamoat-monorepo@0.0.0 build
> lerna run build

lerna notice cli v7.1.4
lerna ERR! Error: Cannot find module '@nx/nx-win32-x64-msvc'
lerna ERR! Require stack:
lerna ERR! - D:\a\LavaMoat\LavaMoat\node_modules\nx\src\native\index.js
lerna ERR! - D:\a\LavaMoat\LavaMoat\node_modules\nx\src\project-graph\utils\retrieve-workspace-files.js
lerna ERR! - D:\a\LavaMoat\LavaMoat\node_modules\nx\src\utils\nx-plugin.js
lerna ERR! - D:\a\LavaMoat\LavaMoat\node_modules\nx\src\config\workspaces.js
lerna ERR! - D:\a\LavaMoat\LavaMoat\node_modules\nx\src\config\configuration.js
lerna ERR! - D:\a\LavaMoat\LavaMoat\node_modules\nx\src\daemon\client\client.js
lerna ERR! - D:\a\LavaMoat\LavaMoat\node_modules\lerna\dist\index.js
lerna ERR! - D:\a\LavaMoat\LavaMoat\node_modules\lerna\dist\cli.js
lerna ERR!     at Module._resolveFilename (node:internal/modules/cjs/loader:1077:15)
lerna ERR!     at Module._load (node:internal/modules/cjs/loader:922:27)
lerna ERR!     at Module.require (node:internal/modules/cjs/loader:1143:19)
lerna ERR!     at require (node:internal/modules/cjs/helpers:121:18)
lerna ERR!     at Object.<anonymous> (D:\a\LavaMoat\LavaMoat\node_modules\nx\src\native\index.js:66:29)
lerna ERR!     at Module._compile (node:internal/modules/cjs/loader:1256:14)
lerna ERR!     at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
lerna ERR!     at Module.load (node:internal/modules/cjs/loader:1119:32)
lerna ERR!     at Module._load (node:internal/modules/cjs/loader:960:12)
lerna ERR! lerna Cannot find module '@nx/nx-win32-x64-msvc'
lerna ERR! lerna Require stack:
lerna ERR! lerna - D:\a\LavaMoat\LavaMoat\node_modules\nx\src\native\index.js
lerna ERR! lerna - D:\a\LavaMoat\LavaMoat\node_modules\nx\src\project-graph\utils\retrieve-workspace-files.js
lerna ERR! lerna - D:\a\LavaMoat\LavaMoat\node_modules\nx\src\utils\nx-plugin.js
lerna ERR! lerna - D:\a\LavaMoat\LavaMoat\node_modules\nx\src\config\workspaces.js
lerna ERR! lerna - D:\a\LavaMoat\LavaMoat\node_modules\nx\src\config\configuration.js
lerna ERR! lerna - D:\a\LavaMoat\LavaMoat\node_modules\nx\src\daemon\client\client.js
lerna ERR! lerna - D:\a\LavaMoat\LavaMoat\node_modules\lerna\dist\index.js
lerna ERR! lerna - D:\a\LavaMoat\LavaMoat\node_modules\lerna\dist\cli.js
Error: Process completed with exit code 1.

```


**EDIT**: Hmm, no, I think you're right: https://github.com/actions/setup-node#caching-global-packages-data

> Note: The action does not cache node_modules

legobeat:
https://github.com/nrwl/nx/issues/17767
https://github.com/nrwl/nx/issues/18453

boneskull:
@legobeat looks like I fixed mac anyway. we expect windows to fail, of course

boneskull:
See #706 ; we don't need to worry about `nx` and its ilk after that gets merged.

.github/workflows/nodejs.yml
legobeat: >  I don't really want to add a build dependency upon .nvmrc. I have mixed feelings about that file.

Feeling aside, why specify on two places when one place do trick?

boneskull:
[shx](https://npm.im/shx) might be helpful for this sort of thing if you want to avoid writing more custom scripts in js.

.github/workflows/nodejs.yml
legobeat: I don't have a straightforward method to limit cross-platform tests to single nodejs-version while it seems redundant to run every version on every platform. Worth it to refactor and deepen the matrix into a dict one level deeper?

legobeat:
- WIP for actually making tests pass on Windows: https://github.com/legobeat/LavaMoat/pull/7

legobeat:
This now selectively only run Windows tests for packages which are known to pass. This to allow merging this and addressing resolving outstanding packages independently.

#### Related
- #1002
- #1003
- #1004

.github/workflows/nodejs.yml
```diff
@@ -48,3 +48,63 @@ jobs:
...
+        --workspace=packages/${{
+          join(
+            fromJson('
+              [
+                "lavapack",
+                "preinstall-always-fail"
+              ]'
+            ), ' --workspace=packages/'
+          )
+        }}
```
boneskull: this needs a comment, whatever it is

.github/workflows/nodejs.yml
boneskull: [this is how you do it](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixexclude)

basically, add `exclude` prop to `matrix`, and specify the OS/node-version pairs we don't want to run.

.github/workflows/nodejs.yml
boneskull: good god what is that

packages/browserify/package.json
```diff
@@ -57,7 +57,7 @@
     "files": [
       "test/*.spec.js"
     ],
-    "timeout": "2m",
+    "timeout": "4m",
```
boneskull, line 60: another way, probably not worth doing:

1. remove the `ava` config from `browserify/package.json`
2. add a `ava.config.mjs` which reproduces everything but contains a dynamic timeout based on platform
😄

boneskull (approved):
this is fine by me, but I'd rather see if we can combine the jobs for DRY reasons

packages/browserify/package.json
legobeat: FTR, this is because browserify tests for macos often time out otherwise, occasionally still had it happen at `3m`.

.github/workflows/nodejs.yml
legobeat: Building flags for specifying which workspaces to run, from a json array..
Was thinking the config itself should be broken out to a package. But I've changed it now to just declare the actual string inline, seeing the reception :sweat_smile:

legobeat:
> this is fine by me, but I'd rather see if we can combine the jobs for DRY reasons

Agree that would be nice, but merging this with separate and open for follow-up merging them

PR #696, opened 2023-09-02
chore: add new commitlint types deps,devDeps

legobeat:
Nice to be able to label dependency/devDependency-only changes as such.

Based on https://github.com/conventional-changelog/commitlint/blob/v17.7.1/%40commitlint/config-conventional/index.js.


boneskull (commented):
I'm 👎 on this due to added overhead when automating merges and releases.  

Renovate is configured to use `fix` for `deps`, and `chore` for `devDeps`, which means a production upgrade will cause a version bump.  There's going to be added overhead in making `deps` trigger a version bump.

Even if certain "types" have no semantic meaning in the conventional commit specification , automated release tools like Release Please, `semantic-release`, `standard-version`, etc., _do_ attach meaning to them.  If we want those tools to change their behavior based on any custom type, we will be adding more overhead & maintenance.

IMO, this is what we should be using:

- If the dep change is in the workspace root, the `type(scope)` should _always_ be `chore(deps)` because the workspace root should have no production deps.
- If the dep change is in a workspace, the `type(scope)` should be `fix(<workspace-name>)` if production, and `chore(<workspace-name>)` if a dev dep.

This is how Renovate is configured to work, anyhow, and I think it's sane.  Remember, we can put whatever we want into the `scope`.

legobeat:
Fair enough, let's go with that for now.

PR #697, opened 2023-09-06
[WIP] Boneskull/policy converter

naugtur:


packages/endomoat/src/index.js
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * @param {Record<string, boolean>} item - A value in `ResourcePolicy`
+ */
+function toEndoRsrcPkgsPolicyPkgs(item) {
+  /** @type {Exclude<import('@endo/compartment-mapper').PackagePolicy['packages'], undefined>} */
+  const policyItem = {}
+  for (const [key, value] of Object.entries(item)) {
+    if (key === LAVAMOAT_PKG_POLICY_ROOT) {
+      // unsure what to do here
```
naugtur, line 58: lavamoat policy should not specify anything about the root package as it's allowed all by default. We need to generate the item in endo policy that defines our expected behavior for root.

We'll likely need an attenuator there when we get to supporting `"write"` policy values.

packages/endomoat/test/index.spec.js
```diff
@@ -0,0 +1,51 @@
...
+    entry: {
+      globals: [ROOT_POLICY],
+      noGlobalFreeze: true,
+      packages: ANY_POLICY,
+      builtins: ANY_POLICY,
+    },
+    resources: {
+      a: {
+        packages: {b: true},
+        globals: {console: true},
```
naugtur, line 39: I think globals should always be rewritten to an array of default attenuator params, even if they're not nested. It's 2 extra characters and we get more consistency (and control)

packages/endomoat/test/index.spec.js
```diff
@@ -0,0 +1,51 @@
...
+      a: {
+        packages: {
+          'b': true,
+        },
+        globals: {
+          'console': true,
+        },
+        builtins: {
+          'fs.readFile': true,
+        },
```
naugtur, line 21: note: there's also a key called "native" and it's rarely used so I don't yet know how we're going to support it.

packages/endomoat/src/index.js
```diff
@@ -0,0 +1,145 @@
...
+const PROPERTY_ATTENUATOR = '@lavamoat/endomoat/attenuator/property'
+
+/**
+ * @param {Record<string, boolean>} item - A value in `ResourcePolicy`
+ */
+function toEndoRsrcPkgsPolicyBuiltins(item) {
+  /** @type {Exclude<import('@endo/compartment-mapper').PackagePolicy['builtins'], undefined>} */
+  const policyItem = {}
+  for (const [key, value] of Object.entries(item)) {
+    if (key.includes('.')) {
```
naugtur, line 27: I like this selective attenuator setup. 

Tip: you can use the default attenuator for both globals and modules, because these are two separate exports. (deliberately designed that way)

packages/endomoat/src/index.js
```diff
@@ -0,0 +1,145 @@
...
+  }
+  return policyItem
+}
+
+/**
+ *
+ * @param {import('lavamoat-core/schema').ResourcePolicy} resources
+ * @returns {import('@endo/compartment-mapper').PackagePolicy<RootPolicy>}
+ */
+function toEndoRsrcPkgsPolicy(resources)  {
```
naugtur, line 81:
```suggestion
function toEndoRsrcPkgsPolicy(resourcePolicy)  {
```

packages/endomoat/src/index.js
```diff
@@ -0,0 +1,145 @@
...
+      [RSRC_POLICY_PKGS]: ENDO_WILDCARD_POLICY,
+      [RSRC_POLICY_BUILTINS]: ENDO_WILDCARD_POLICY,
+      noGlobalFreeze: true,
+    },
+    resources: {},
+  }
+
+  if (lmPolicy.resources) {
+    for (const [rsrcName, rsrcPolicy] of Object.entries(lmPolicy.resources)) {
+      if (rsrcName === LAVAMOAT_ROOT_PKG) {
```
naugtur: I don't think we support setting anything on root at this time. Let me know where you've found examples of this :sweat_smile:

boneskull:
closed in lieu of #772

PR #683, opened 2023-08-28
fix(browserify/test): Don't let `yarn link` clobber global directory

legobeat:
Fixes: #682 

Unblocks: #681

PR #698, opened 2023-09-07
chore: use npm for browserify tests

### LINKED ISSUE #682: browserify tests attempt to modify system environment
As can be observed [here](https://github.com/LavaMoat/LavaMoat/actions/runs/5996407503/job/16260892432?pr=681#step:10:171) (#681), `yarn link` invocation in `lavamoat-browserify` tests can trigger attempts at linking global bins.

Fix: Use temporary directory instead, as already done with projectDir.

boneskull:
This changes the `lavamoat-browserify` tests to use `npm` instead of `yarn`, and installs via absolute path instead of `npm link`; this does not modify the global `node_modules` folder.

Note: a test fixture policy file is updated.

This supercedes #683.

Closes #682

### DIFF
--- a/packages/browserify/test/fixtures/secureBundling/lavamoat/node/policy.json
+++ b/packages/browserify/test/fixtures/secureBundling/lavamoat/node/policy.json
@@ -81,6 +81,7 @@
       "globals": {
         "__dirname": true,
         "__filename.slice": true,
+        "console.error": true,
         "console.warn": true,
         "process.cwd": true,
         "setTimeout": true
@@ -91,6 +92,7 @@
         "@lavamoat/lavapack>through2": true,
         "@lavamoat/lavapack>umd": true,
         "browserify>JSONStream": true,
+        "external:../../node_modules/espree/dist/espree.cjs": true,
         "json-stable-stringify": true,
         "lavamoat-core": true,
         "readable-stream": true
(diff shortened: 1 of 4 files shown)


boneskull:
* main
  * **PR #698** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/698" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/698?utm_source=stack-comment).

packages/browserify/test/util.js
```diff
@@ -174,17 +156,16 @@ async function prepareBrowserifyScenarioOnDisk ({ scenario }) {
     )
     runBrowserifyPath = `${__dirname}/fixtures/runBrowserifyBundleFactor.js`
   }
-  // yarn link workspace dependencies
+
+  // install must happen before link, otherwise npm will remove any linked packages upon install
+  const installDevDepsResult = spawnSync('npm', ['install', '--no-save', ...depsToInstall], { cwd: projectDir })
   overrideDepsWithLocalPackages(projectDir)
-  // limit concurrency so that yarn v1 doesnt break its own cache
-  const installDevDepsResult = await limitConcurrency(async function () {
```
boneskull, line 163: unless we're running multiple processes against this code, the concurrency-limit thing _shouldn't_ be needed because a) `spawnSync` is synchronous and b) afaik `npm` does not have this problem.

packages/browserify/test/fixtures/secureBundling/lavamoat/node/policy.json
```diff
@@ -790,6 +792,25 @@
         "duplexify>end-of-stream>once>wrappy": true
       }
     },
+    "external:../../node_modules/acorn-jsx/index.js": {
```
naugtur, line 795: BTW Note: I don't think we have support for this in endo

naugtur:
When I try to run the tests locally, I get 23 failures because 
```
Error: ENOENT: no such file or directory, open '.../LavaMoat/packages/lavapack/src/runtime.js'
```
I think it only works when all workspace tests run. We'd need a pre-test or something... 
I guess that's one of the reasons why the file was committed before.

we need to ensure `npm run build` runs in lavapack before these tests. Not sure how yet.

boneskull:
@naugtur Aren't we supposed to be running `npm run setup` after install?  :D  

Anyhow, the only way to avoid this given our configuration is to ensure lavapack gets built first.  We can certainly invoke it from the `test` script in browserify...

naugtur:
oh, right `setup`

naugtur (approved):

PR #699, opened 2023-09-08
refactor(core): consolidate endowments logic

naugtur:
This is not the major refactoring of core that we were sometimes discussing.
I'm refactoring the endowments bit so that they're more ready to use externally as a toolkit for putting the globals in place.


boneskull (approved):

naugtur:
waiting to merge after release today

Issue #693, opened 2023-08-31
Allow setting lockdown options for lavapack/lavamoat-browserify


naugtur:
The way core kernel is put together it's not accepting lockdown options.
https://github.com/LavaMoat/LavaMoat/blob/main/packages/core/src/kernelTemplate.js#L55

It'd be great if lockdown could be configured

notes on how to implement:
- threading the options from the top is one way to do it
- Now that we've split lockdown into 'repairIntrinsics' and 'hardenIntrinsics', we can allow 'repairIntrinsics' to be called externally. If it's called again, it will produce a specific error that we can silence. We should only silence this particular error, not any others. After that, we can proceed with 'hardenIntrinsics'.

PR #701, opened 2023-09-12
Allow specifying `lockdownOptions` in `lavamoat-browserify`

FrederikBolding:
Allow specifying `lockdownOptions` in `lavamoat-browserify` for more customizability.

Proposed fix for https://github.com/LavaMoat/LavaMoat/issues/693

Feedback appreciated!

### DIFF
--- a/packages/core/src/kernelTemplate.js
+++ b/packages/core/src/kernelTemplate.js
@@ -19,6 +19,7 @@
     // security options are hard-coded at build time
     const {
       scuttleGlobalThis,
+      lockdownOptions,
     } = __lavamoatSecurityOptions__
 
     function getGlobalRef () {
@@ -51,7 +52,7 @@
     // eslint-disable-next-line no-extra-semi
     ;templateRequire('ses')
 
-    const lockdownOptions = {
+    const DEFAULT_LOCKDOWN_OPTIONS = {
       // gives a semi-high resolution timer
       dateTaming: 'unsafe',
       // this is introduces non-determinism, but is otherwise safe
@@ -62,7 +63,7 @@
       stackFiltering: 'verbose',
     }
 
-    lockdown(lockdownOptions)
+    lockdown(Object.assign({}, DEFAULT_LOCKDOWN_OPTIONS, lockdownOptions ?? {}))
 
     // initialize the kernel
     const createKernelCore = __createKernelCore__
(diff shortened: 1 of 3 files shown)


packages/core/src/kernelTemplate.js
```diff
@@ -62,7 +63,7 @@
       stackFiltering: 'verbose',
     }
 
-    lockdown(lockdownOptions)
+    lockdown(Object.assign({}, DEFAULT_LOCKDOWN_OPTIONS, lockdownOptions ?? {}))
```
boneskull, line 66: I think @kumavis might need to speak to this, but:

1. Are we allowed to use `Object.assign` here--does `assign` need to be destructured?
2. Why not `{...DEFAULT_LOCKDOWN_OPTIONS, ...lockdownOptions}`?

boneskull (commented):
There should be a test for this

boneskull:
@FrederikBolding Commit messages now must adhere to the [conventional commits spec](https://conventionalcommits.org).  This would be something like `feat(browserify): add lockdownOptions to options`

naugtur:
I'll drop the feedback I gave in person:
Looks too easy :smile: 
I'm gonna have to think why I got pushback earlier (1+yr ago) when I brought up introducing it. 

Agree a test would increase confidence. Which also means this depends on https://github.com/LavaMoat/LavaMoat/pull/698 I guess

packages/core/src/kernelTemplate.js
FrederikBolding: It seems that the pattern elsewhere in LM was using `Object.assign` over spread, happy to do either, just trying to stick to the existing pattern.

FrederikBolding:
> I'll drop the feedback I gave in person: Looks too easy 😄 I'm gonna have to think why I got pushback earlier (1+yr ago) when I brought up introducing it.
> 
> Agree a test would increase confidence. Which also means this depends on #698 I guess

Just to clarify I have tested this locally with the snaps environment, but I agree that it should be tested.

I will wait for further feedback on the implementation before spending time figuring out the testing though.

packages/core/src/kernelTemplate.js
boneskull: I'm curious about that pattern myself.  could be a holdover from earlier versions of Node.js or not

naugtur: kernel is being bundled into browserify bundles so it was likely for the browser support reasons.

PR #709, opened 2023-09-13
move assemble function out of plugin to new file

weizman:


packages/scorchwrap/src/plugin.js
```diff
@@ -8,18 +8,13 @@
...
-  RuntimeModule,
-  // RuntimeGlobals,
-  // ModuleDependency,
-} = require('webpack')
+const { WebpackError } = require('webpack')
 const { wrapper } = require('./buildtime/wrapper')
 const { generateIdentifierLookup } = require('./buildtime/aa')
 const diag = require('./buildtime/diagnostics')
 const progress = require('./buildtime/progress')
+const { assembleRuntime } = require('./buildtime/assemble')
```
naugtur, line 16: (nit) I'd name the file assembleRuntime or runtimeBuilder for immediate clarity that it's not involved in processing the app sources in any way. But I'm not sure if a person who knows what runtime means in the contect of webpack would not consider it obvious by that point.

packages/scorchwrap/src/buildtime/assemble.js
```diff
@@ -0,0 +1,64 @@
...
+        LAVAMOAT['${name}'] = module.exports;
+      })();`
+    }
+    assembly += `\n;/*${name}*/;\n${sourceString}`
+  })
+  assembly += `;
+  __webpack_require__.${KEY} = LAVAMOAT.defaultExport;
+  (typeof harden !== 'undefined') && harden(__webpack_require__.${KEY});` // The harden line is likely unnecessary as the handler is being frozen anyway
+  return {
+    addTo({ compilation, chunk }) {
```
naugtur: (could) we could refactor this part while we're at it and make this module return a VirtualRuntimeModule that then gets put where it belongs by the code in plugin.js
I regret creating the addTo function now. I used to add multiple runtime modules but that's no longer the case

packages/scorchwrap/src/plugin.js
weizman: how about "build"? I like keeping files to a single word when possible, not by any means of course

packages/scorchwrap/src/buildtime/assemble.js
weizman: maybe something like 6fa51b506?

naugtur (approved):

naugtur:
Please clean up history before merging. I've rebased my branch to have clean commits there which makes this more difficult. But if you intended to squash, it should be fairly easy to rebase or cherry-pick on top.

PR #714, opened 2023-09-13
fix: normalize pkg versions

boneskull:
Due to the errant release, "newer" versions were published of all of these packages, but VCS did not reflect those new versions.  This updates the versions for each package to the "newest" version and changes the release-please manifest to point to those versions, so that next release will bump them.

### DIFF
--- a/packages/browserify/package.json
+++ b/packages/browserify/package.json
@@ -1,6 +1,6 @@
 {
   "name": "lavamoat-browserify",
-  "version": "15.7.4",
+  "version": "15.9.0",
   "description": "browserify plugin for sandboxing dependencies with LavaMoat",
   "main": "src/index.js",
   "directories": {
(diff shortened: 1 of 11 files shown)


boneskull:
Note that all our workspaces depending on `lavamoat-tofu` ask for `^v6.0.3` where `v6.2.0` is in range, so they are all using the updated version in production.  In a development environment, they use the workspace in `packages/tofu`, which is v6.0.3. So this needs fixing.

leotm (approved):

naugtur:
Same thing happened to lavapack - there's a 5.4.0 in the registry.

packages/browserify/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "lavamoat-browserify",
-  "version": "15.7.4",
+  "version": "16.0.0",
```
naugtur: Why would we bump major?

naugtur (approved):

PR #724, opened 2023-09-18
test(scorchwrap): e2e test suite for the webpack plugin

naugtur:
The tests are scaffolded to actually run webpack and produce a bundle.
Node's `vm` module is used to run the bundle in a new context and observe results.
The fixture is all co-located with fake modules it uses and therefore an npm install is not necessary.

main - it's the happy-path test that asserts for basic functionality and correctness of bundling.
hack - it's a (incomplete) set of hacks that we want to defend against.

Note 2 of the 3 commits contain fixes for issues I found while testing :)


packages/scorchwrap/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "",
   "main": "src/plugin.js",
   "scripts": {
-    "test": "ava --timeout=30s test/*.js",
+    "test": "ava --timeout=300s test/*.spec.js",
```
naugtur: this may be unnecessary - without the real dependency tree these tests are quite fast.

packages/scorchwrap/src/plugin.js
```diff
@@ -127,7 +127,7 @@ const wrapGeneratorMaker = ({
 
       // skip doing anything if marked as ignored by the ignoreLoader,
       // only accept ignore loader from config, not inline
```
weizman: maybe an explanation on the `!`?

packages/scorchwrap/package.json
weizman: 300 is a lot

packages/scorchwrap/src/runtime/runtime.js
```diff
@@ -57,10 +57,7 @@ const enforcePolicy = (requestedResourceId, referrerResourceId) => {
   ) {
     return
   }
-  const myPolicy = LAVAMOAT.policy.resources[referrerResourceId]
-  if (!myPolicy) {
-    throw Error('Policy missing for ' + referrerResourceId)
```
weizman, line 62: shouldn't we strictly expect policies to always exist?

packages/scorchwrap/test/aa.spec.js
```diff
@@ -59,10 +59,24 @@ test('aa - generateIdentifierLookup', (t) => {
...
+        '@ethereumjs/util>ethereum-cryptography',
+    })),
+    unenforceableModuleIds: [],
+    readableResourceIds: true,
+  })
 
   const translatedPolicy = lookupTools.getTranslatedPolicy()
-  console.log(translatedPolicy.resources)
-  t.is(Object.keys(translatedPolicy.resources).length, 3)
+  // console.log(translatedPolicy.resources)
```
weizman: remove

packages/scorchwrap/src/runtime/runtime.js
naugtur: That was my initial implementation and it's not what the rest of LavaMot is doing. Policy generation skips all packages that don't use any powers. It's necessary for eliminating a lot of spam from the policy definition.

packages/scorchwrap/test/aa.spec.js
naugtur: oops

packages/scorchwrap/src/plugin.js
naugtur: done

weizman (approved):
lgtm

Issue #704, opened 2023-09-12
ditch lerna


boneskull:
I think we can ditch lerna for publishing.  `npm --workspaces publish` seems to work pretty well. Update docs and scripts.

boneskull:
Closed via #727

PR #727, opened 2023-09-19
chore: add custom wrapper for publishing

boneskull:
The aim here is to wholly replace Lerna, since we're no longer using it for much. 
Unfortunately, `npm publish` doesn't have any sort of flag to ignore already-published packages, so here we are.

The script:

1. Finds all workspace directories (public packages only)
2. Performs some minimal validation of each workspace's `package.json`
3. Checks to see if the version in the workspace's `package.json` has already been published to the npm registry
4. If the package doesn't exist in the registry, fail _unless_ the `--newPkg <pkg-name>` option was provided to the script
5. Otherwise, if the package version has already been published, skip it
6. Execute `npm publish` once with arg `--workspace <pkg-name>` for every package that needs publishing. Output is piped thru, so you can examine the results of `npm pack` before they go out

There's a decent test suite here, too.

The script has some basic `--help` output, and supports the following options:

- `--root <path>` - Specify the workspace root directory. Default is just the parent dir of the script's dir
- `--dryRun` - Run `npm publish` in "dry run" mode
- `--newPkg <name>` - Workspace package with name `name` has not yet been published to the registry


boneskull:
* main
  * **PR #727** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/727" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/727?utm_source=stack-comment).

scripts/publish.js
```diff
@@ -0,0 +1,118 @@
...
+
+// @ts-check
+
+const { workspaces } = require('../package.json')
+const { glob } = require('glob')
+const path = require('node:path')
+const fs = require('node:fs/promises')
+const util = require('node:util')
+const exec = util.promisify(require('node:child_process').exec)
+const { spawn } = require('node:child_process')
```
boneskull: exec, because it's easy to promisify and we read the output as json
spawn, because we want to print the npm output to console (shows what was packed into the tarball)

scripts/publish.js
```diff
@@ -0,0 +1,118 @@
...
+      args.push('--dry-run')
+      console.info('*** DRY RUN *** DRY RUN *** DRY RUN *** DRY RUN ***')
+    }
+
+    console.info(`Running \`npm ${args.join(' ')}\``)
+
+    spawn('npm', args, {
+      cwd: ROOT,
+      stdio: 'inherit',
+      shell: true,
```
boneskull: don't love this, but we need it for nvm.  otherwise we will need to use something like `which` to find it

scripts/publish.js
```diff
@@ -49,13 +49,28 @@ async function publish(pkgs, dryRun = true) {
...
+ * @returns {Promise<any>}
+ */
+const chillJSONParse = async (promise) =>
+  promise
+    .catch((err) => {
+      console.error(err.message)
+      return '{}'
+    })
+    .then(JSON.parse)
+
```
boneskull: @naugtur So if this parsing fails, either a) `package.json` is invalid, or b) `npm show --json` is invalid, and we should continue _but_ with a message to the user that they may need to publish this package manually. 

I don't know why we'd want to continue normally if a `package.json` is invalid though.

scripts/publish.js
```diff
@@ -66,29 +81,18 @@ async function main() {
...
-                  dirent.name,
-                  'package.json'
-                ),
-                'utf8'
-              )
+            const { name, version, private } = await chillJSONParse(
+              fs.readFile(path.join(dirent.fullpath(), 'package.json'), 'utf8')
             )
-            if (private) {
+            if (private || !name) {
```
boneskull: if `name` is undefined, that's an invalid `package.json` file and we should probably say something

scripts/publish.js
```diff
@@ -66,29 +81,18 @@ async function main() {
...
               return
             }
 
-            const { stdout } = await exec(`npm view ${name} versions --json`, {
-              cwd: ROOT,
-            })
-
-            const versions = JSON.parse(stdout)
+            const versions = await chillJSONParse(
```
boneskull: why mix await and then/catch?  these are all wrapped in a `Promise.all()`, so async operations should not happen in serial anyhow

scripts/publish.js
naugtur: If package.json is not found or invalid, the result of parsing is an empty object, the error was printed and the package was skipped.
I introduced it meaning to complete publishing what makes sense and let you fix and rerun to publish the thing that was broken before. 
But I think I was making a wrong assumption - if we throw an error from parsing package.json - the whole publish is not happening and the release can be fixed before another attempt.  
I've lost confidence in my alleged improvement :sweat_smile:

scripts/publish.js
naugtur: it was for readability. the function passed to map is async, so this will not turn anything into a sequence.

scripts/publish.js
naugtur: Let's drop the chillJSONParse function and use regular JSON.parse in a big try-catch in the map callback to add a message explaining the package from the folder was not fit to publish and then adding the rest of the error for debugging.

scripts/publish.js
```diff
@@ -0,0 +1,442 @@
...
+
+/**
+ * Minimal `package.json` definition.
+ *
+ * @todo If we ever pull in `type-fest` for whatever reason, use its `PackageJson` instead.
+ * @typedef PackageJson
+ * @property {boolean} [private]
+ * @property {string} name
+ * @property {string} version
+ */
```
naugtur, line 118: (could) This might work better if put in a d.ts file

naugtur (approved):
This is solid. We're gonna have to publish it to the ecosystem after we use it for a little while.

scripts/publish.js
boneskull: will look at that if/when we end up transitioning this to a package.

boneskull:
Added some checks around the `npm publish` invocation.

boneskull:
Closes #704

PR #725, opened 2023-09-18
chore: format all files

boneskull:
This just formats everything all at once, if we want to do that.

PR #728, opened 2023-09-21
chore(vscode): auto-run fix and format on save

boneskull:
This runs `prettier` and `eslint --fix` upon save of any fixable, formattable, and non-ignored file in the repo.

The same thing happens as a pre-commit hook.

Some people may not like this idea, because there are reasons to dislike it.  I prefer to have my code formatted as I write it, so this makes me more nimble.  However, it can be a problem if you edit a file then decide not to commit your changes -- you'll need to `git checkout -- <file>` to clean your workspace, because any save via the editor will reformat again.

Anyway: if this is not popular with all of the VSCode users here, I'll close it.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,8 @@
 {
-  "editor.tabSize": 2,
-  "editor.defaultFormatter": "esbenp.prettier-vscode"
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.codeActionsOnSave": {
+    "source.fixAll": true,
+    "source.formatDocument": true
+  },
+  "prettier.useEditorConfig": true
 }


boneskull:
That said, if we merge the reformat all at once (#725) then this will likely have much less impact

.vscode/settings.json
```diff
@@ -1,4 +1,8 @@
 {
-  "editor.tabSize": 2,
-  "editor.defaultFormatter": "esbenp.prettier-vscode"
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.codeActionsOnSave": {
```
naugtur, line 3: I'm not a huge fan of that, but maybe if it's limited to the repo and not messing with other people's projects I open it's enjoyable.

naugtur (approved):

PR #730, opened 2023-09-26
feat(core): add overrideTaming: 'severe' 

naugtur:
for improved developer experince under lavamoat


weizman (approved):

packages/core/src/kernelTemplate.js
```diff
@@ -60,6 +60,8 @@
       errorTaming: 'unsafe',
       // shows the full call stack
       stackFiltering: 'verbose',
+      // prevents most common override mistake cases from triping up users
```
legobeat:
```suggestion
      // prevents most common override mistake cases from tripping up users
```
nit

PR #732, opened 2023-10-05
feat(core): ship some types

boneskull:
This adds _some_ types to `lavamoat-core`.

Interesting stuff includes the `LavaMoatPolicy` type and the types for `endowmentsToolkit.js`

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -12,9 +12,11 @@
   "dependencies": {
     "json-stable-stringify": "^1.0.2",
     "lavamoat-tofu": "^7.0.0",
-    "merge-deep": "^3.0.3"
+    "merge-deep": "^3.0.3",
+    "type-fest": "^4.7.1"
   },
   "devDependencies": {
+    "@types/json-stable-stringify": "^1.0.34",
     "ses": "^0.18.8",
     "tmp-promise": "^3.0.3"
   },
@@ -39,5 +41,13 @@
       "test/*.spec.js"
     ],
     "timeout": "30s"
-  }
+  },
+  "types": "./types/src/types.d.ts",
+  "files": [
+    "lib",
+    "src",
+    "types",
+    "!*.tsbuildinfo",
+    "CHANGELOG.md"
+  ]
 }
--- a/packages/core/src/endowmentsToolkit.js
+++ b/packages/core/src/endowmentsToolkit.js
@@ -1,12 +1,26 @@
-// the contents of this file will be copied into the prelude template
-// this module has been written so that it required directly or copied and added to the template with a small wrapper
+// @ts-check
+
+/**
+ * Utilities for generating the endowments object based on a `globalRef` and a
+ * {@link LMPolicy.PackagePolicy}.
+ *
+ * The contents of this file will be copied into the prelude template this
+ * module has been written so that it required directly or copied and added to
+ * the template with a small wrapper.
+ *
+ * The `PackagePolicy` uses a period-deliminated path notation to pull out deep
+ * values from objects. These utilities help create an object populated with
+ * only the deep properties specified in the `PackagePolicy`.
+ *
+ * @packageDocumentation
+ */
+
 module.exports = endowmentsToolkit
 
-// utilities for generating the endowments object based on a globalRef and a packagePolicy
-
-// The packagePolicy uses a period-deliminated path notation to pull out deep values from objects
-// These utilities help create an object populated with only the deep properties specified in the packagePolicy
-
+/**
+ * @param {object} opts
+ * @param {DefaultWrapperFn} [opts.createFunctionWrapper]
+ */
 function endowmentsToolkit({
   createFunctionWrapper = defaultCreateFunctionWrapper,
 } = {}) {
@@ -21,13 +35,12 @@ function endowmentsToolkit({
   }
 
   /**
-   * @function getEndowmentsForConfig Creates an object populated with only the deep properties specified in the packagePolicy
+   * Creates an object populated with only the deep properties specified in the packagePolicy
    * @param {object} sourceRef - Object from which to copy properties
-   * @param {object} packagePolicy - LavaMoat policy item representing a package
+   * @param {LMPolicy.PackagePolicy} packagePolicy - LavaMoat policy item representing a package
    * @param {object} unwrapTo - For getters and setters, when the this-value is unwrapFrom, is replaced as unwrapTo
    * @param {object} unwrapFrom - For getters and setters, the this-value to replace (default: targetRef)
-   * @return {object} - The targetRef
-   *
+   * @returns {object} - The targetRef
    */
   function getEndowmentsForConfig(
     sourceRef,
@@ -39,7 +52,9 @@ function endowmentsToolkit({
       return {}
     }
     // validate read access from packagePolicy
+    /** @type {string[]} */
     const whitelistedReads = []
+    /** @type {string[]} */
     const explicitlyBanned = []
     Object.entries(packagePolicy.globals).forEach(
       ([path, packagePolicyValue]) => {
@@ -79,6 +94,15 @@ function endowmentsToolkit({
     )
   }
 
+  /**
+   *
+   * @param {object} sourceRef
+   * @param {string[]} paths
+   * @param {object} unwrapTo
+   * @param {object} unwrapFrom
+   * @param {string[]} explicitlyBanned
+   * @returns {object}
+   */
   function makeMinimalViewOfRef(
     sourceRef,
     paths,
@@ -86,6 +110,7 @@ function endowmentsToolkit({
     unwrapFrom,
     explicitlyBanned = []
   ) {
+    /** @type {object} */
     const targetRef = {}
     paths.forEach((path) => {
       copyValueAtPath(
@@ -101,13 +126,37 @@ function endowmentsToolkit({
     return targetRef
   }
 
+  /**
+   * @param {string} visited
+   * @param {string} next
+   */
   function extendPath(visited, next) {
+    // FIXME: second part of this conditional should be unnecessary
     if (!visited || visited.length === 0) {
       return next
     }
     return `${visited}.${next}`
   }
 
+  /**
+   * @template T
+   * @param {T|null} value
+   * @returns {value is null}
+   */
+  function isEmpty(value) {
+    return !value
+  }
+
+  /**
+   *
+   * @param {string} visitedPath
+   * @param {string[]} pathParts
+   * @param {string[]} explicitlyBanned
+   * @param {object} sourceRef
+   * @param {object} targetRef
+   * @param {object} unwrapTo
+   * @param {object} unwrapFrom
+   */
   function copyValueAtPath(
     visitedPath,
     pathParts,
@@ -129,7 +178,7 @@ function endowmentsToolkit({
     )
 
     // if source missing the value to copy, just skip it
-    if (!sourcePropDesc) {
+    if (isEmpty(sourcePropDesc)) {
       return
     }
 
@@ -154,7 +203,7 @@ function endowmentsToolkit({
 
     // if this is not the last path in the assignment, walk into the containing reference
     if (remainingParts.length > 0) {
-      const { sourceValue, sourceWritable } = getSourceValue()
+      const { sourceValue, sourceWritable } = getSourceValue(sourcePropDesc)
       const nextSourceRef = sourceValue
       let nextTargetRef
       // check if value exists on target and does not need selective treatment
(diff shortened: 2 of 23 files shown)


boneskull:
* **#732** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/732?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3  dependent PRs ([#733](https://github.com/LavaMoat/LavaMoat/pull/733) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/733?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#752](https://github.com/LavaMoat/LavaMoat/pull/752) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/752?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#755](https://github.com/LavaMoat/LavaMoat/pull/755) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/755?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈
* `main`

<h2></h2>

boneskull:
This includes some fixes to the eslint configuration, as it's presently confused about whether or not it should lint generated types.  It shouldn't, and it doesn't in CI, but it does locally.

packages/core/package.json
```diff
@@ -12,9 +12,11 @@
   "dependencies": {
     "json-stable-stringify": "^1.0.2",
     "lavamoat-tofu": "^6.2.1",
-    "merge-deep": "^3.0.3"
+    "merge-deep": "^3.0.3",
+    "type-fest": "^4.3.3"
```
legobeat: Can we achieve this without needing a new runtime dependency? Seems like this belongs in devDependencies.

packages/core/src/endowmentsToolkit.js
```diff
@@ -328,12 +418,19 @@ function endowmentsToolkit({
...
+   * @returns {{prop: PropertyDescriptor|null, receiver: object|null}}
+   */
   function getPropertyDescriptorDeep(target, key) {
+    /** @type {object|null} */
     let receiver = target
     // eslint-disable-next-line no-constant-condition
     while (true) {
       // abort if this is the end of the prototype chain.
-      if (!receiver) {
+      if (receiver === null) {
```
legobeat: Potential behavior change, I think? Can those be broken out as separate PR(s) and merged independently to shipping types?

packages/core/package.json
boneskull: It's not a dev dep because we export a type that uses it (policy schema).

packages/core/src/endowmentsToolkit.js
boneskull: I've reverted this

packages/core/src/endowmentsToolkit.js
```diff
@@ -1,12 +1,26 @@
-// the contents of this file will be copied into the prelude template
-// this module has been written so that it required directly or copied and added to the template with a small wrapper
```
naugtur, line 2: We should keep this comment, unless it's no longer true.

packages/core/src/generatePolicy.js
```diff
@@ -126,7 +170,7 @@ function createModuleInspector(opts = {}) {
         allowReturnOutsideFunction: true,
         errorRecovery: true,
       })
-    if (includeDebugInfo && ast.errors && ast.errors.length) {
+    if (includeDebugInfo && isParsedAST(ast) && ast.errors.length) {
```
naugtur, line 173: This is clever. And I don't even mind the runtime overhead vs clarity in this case.

packages/core/src/generatePolicy.js
```diff
@@ -246,16 +316,28 @@ function createModuleInspector(opts = {}) {
...
+    packageBuiltins = [...(packageBuiltins ?? []), ...moduleBuiltins]
     packageToBuiltinImports.set(packageName, packageBuiltins)
   }
 
-  function generatePolicy({ policyOverride, includeDebugInfo } = {}) {
+  /**
+   * @type {GeneratePolicyFn}
+   */
+  function generatePolicy({ policyOverride, includeDebugInfo = false }) {
+    /** @type {import('./schema').Resources} */
```
naugtur, line 327: we're in a closure so if I understood correctly, you can do a typedef import and get ResourcePolicy etc. directly without reexporting it. Was that the case that only top level typedef needs to be avoided because it's reexported?

boneskull: Yes, it could be used here without worrying about an export (I think!), but if `Resources` becomes generic in the future, we would need to re-implement those  type args in `@template` tags.  That said, you could probably get away with it here--kind of like you can get away with writing JS without semicolons.

_Personally,_ I don't want to think too hard about whether or not it's safe to create the `@typedef` reference, so I choose not to and make that tradeoff.  Similarly, I don't like thinking about whether or not ASI is going to work either, so I always use semicolons. :smile:

packages/core/src/generatePolicy.js
boneskull: `ast.errors` doesn't mean very much to TS' type inference, but a proper type guard like `isParsedAST()` does.

packages/core/src/endowmentsToolkit.js
boneskull: it's there; it's just moved.

packages/core/src/endowmentsToolkit.js
```diff
@@ -379,7 +484,7 @@ function endowmentsToolkit({
     endowmentSources.forEach((source) => {
       const descriptors = Object.getOwnPropertyDescriptors(source)
       Object.values(descriptors).forEach((desc) => {
-        if ('get' in desc) {
+        if ('get' in desc && desc.get) {
```
kumavis, line 495: I usually do `desc.get !== undefined`

packages/core/src/generatePolicy.js
```diff
@@ -160,6 +213,7 @@ function createModuleInspector(opts = {}) {
     if (includeDebugInfo) {
       const moduleDebug = debugInfo[moduleRecord.specifier]
       moduleDebug.sesCompat = {
+        // XXX: I don't think this is needed, since it appears we overwrite all properties
```
kumavis: this is just for the debug file. what do you mean about overwriting all props?

packages/core/src/endowmentsToolkit.js
boneskull: i want it to be truthy

packages/core/src/generatePolicy.js
boneskull: `compatWarnings` has props `primordialMutations`, `strictModeViolations` and `dynamicRequires`, all of which are overwritten below.  I don't think there are other props.

boneskull: (meaning that `...compatWarnings` is unnecessary. at least, I think)

packages/core/src/endowmentsToolkit.js
boneskull: fwiw `'foo' in bar` is the TypeScript-Approved:tm: way to check if a prop "has" something; skipping right to `foo.bar` will typically be an error if it's unclear from type inference

naugtur: in the case of a descriptor there's no difference between it being truthy or not-undefined because it can only be a function or missing. 
Is this for typescript only?

packages/core/src/generatePolicy.js
naugtur: As the best practices around this are far from set in stone, I'm trying to explore options in case we come up with something that's more convenient. 

Also note that making this pattern more popular makes SES slower, because it relies on a transform to replace each `import(` with something else

boneskull: The transform in question doesn't happen in SES, fwiw.  It means we have to apply it ourselves; hence https://github.com/endojs/endo/pull/1812

packages/core/src/endowmentsToolkit.js
boneskull: According to TS, `get` can be `null`.  And `get: null` would be true for `'get' in desc`; hence the truthiness check.  TS' ECMAScript libs aren't the spec, but I trust them.

boneskull:
I'm going to split this up into two or three separate PRs.

packages/core/src/endowmentsToolkit.js
naugtur: If we want to support the folks using typescript directly on platforms we don't yet foresee (poor souls) this seems like the right thing to do.

naugtur:
> I'm going to split this up into two or three separate PRs.

I was ready to say we approve, merge and see what happens, but a split might serve us well. That is, if your tooling can figure out the tree of dependent items.

packages/core/src/endowmentsToolkit.js
```diff
@@ -263,20 +329,29 @@ function endowmentsToolkit({
...
+      const getter = sourcePropDesc.get
       wrappedPropDesc.get = function () {
+        // eslint-disable-next-line @typescript-eslint/no-this-alias
         const receiver = this
         // replace the "receiver" value if it points to fake parent
         const receiverRef =
           receiver === unwrapFromGlobalThis ? unwrapToGlobalThis : receiver
         // sometimes getters replace themselves with static properties, as seen wih the FireFox runtime
-        const result = Reflect.apply(sourcePropDesc.get, receiverRef, [])
+        const result = Reflect.apply(getter, receiverRef, [])
```
naugtur: (must) Let's make sure this is not breaking the code. I'm pretty sure the comment above is explaining why we're not using a local copy but referencing the getter from the original source and this change would break it.

naugtur (changes requested):
I've spotted a new bug being introduced, I think: 
https://github.com/LavaMoat/LavaMoat/pull/732/files#diff-5ed0581fdb5b3ec8105f7f2e6773d6f5db45dbe74dccf2a921cc9c0e40ecda60R354

packages/core/src/endowmentsToolkit.js
legobeat: Regardless of outcome:  This seems like a non-type-related refactor that could/should be broken out as separate change? Or maybe I'm missing some context on why this change is here?

boneskull: Sorry this wasn't explained; TS' type inference doesn't know that `sourcePropDesc.get` is not `undefined` at this point, because it doesn't know when `wrappedPropDesc.get()` is called, so it cannot be sure, despite the conditional on L345 (or L346, depending).

I've changed it to just use a type assertion instead.

To be clear, I didn't make any code changes in this PR that _weren't_ prompted by type checking.

boneskull: By leaving it as-is, we're saying: _`sourcePropDesc.get` is allowed to be `undefined` here, and if so, allow `Reflect.apply()` to throw an exception_.  Just want to confirm that's the intent.

packages/core/src/endowmentsToolkit.js
```diff
@@ -295,17 +379,26 @@ function endowmentsToolkit({
...
     if (sourcePropDesc.set) {
+      const setter = sourcePropDesc.set
       wrappedPropDesc.set = function (value) {
         // replace the "receiver" value if it points to fake parent
+        // eslint-disable-next-line @typescript-eslint/no-this-alias
         const receiver = this
         const receiverRef =
           receiver === unwrapFromGlobalThis ? unwrapToGlobalThis : receiver
-        return Reflect.apply(sourcePropDesc.set, receiverRef, [value])
+        return Reflect.apply(setter, receiverRef, [value])
```
naugtur: (must) same issue as with https://github.com/LavaMoat/LavaMoat/pull/732/files#diff-5ed0581fdb5b3ec8105f7f2e6773d6f5db45dbe74dccf2a921cc9c0e40ecda60R354

Sorry I didn't point it out here. I did in a different PR that was showing it. 

GH will not let me add a suggestion that fixes it...

naugtur:
BTW, trees managed by Graphite are cool, but they encourage amending the existing commit with review fixes and it makes it harder for the reviewer to see what changes were made since they last saw it. 
Still net positive, but I thought I'd point it out ;) 
Back to reviewing the whole thing top to bottom.

packages/core/src/endowmentsToolkit.js
naugtur: We could doublecheck before applying. The breaking change was that we cached the getter instead of using the fresh one that could have been updated.

boneskull:
@naugtur No, that's just me squashing commits.  nothing to do with graphite

packages/core/src/endowmentsToolkit.js
boneskull: done

packages/core/src/endowmentsToolkit.js
boneskull: I'm not sure what sort of action would be reasonable here _other_ than just letting it throw.  In other words, I don't know how to improve on that.

naugtur:
> @naugtur No, that's just me squashing commits.  nothing to do with graphite

Ok, in that case, I request that whenever possible/convenient commits remain visible for easier review and you squash them later.

boneskull:
@naugtur Will do.  Ball still in yr court though

naugtur (approved):

PR #737, opened 2023-10-17
chore: enable releases of webpack plugin

naugtur:


release-please-config.json
```diff
@@ -4,6 +4,7 @@
   "packages": {
     "packages/aa": {},
     "packages/allow-scripts": {},
+    "packages/webpack": {},
```
leotm, line 7: we might need to add this to `.release-please-manifest.json` like:

`"packages/webpack": "1.0.0-beta.1",` cc @boneskull

naugtur: AFAIK the `.release-please-manifest.json` entry is optional and it defines the starting version before which stuff should be ignored.

naugtur: hmm. the docs sound like you're right :) updating

leotm (approved):

PR #739, opened 2023-10-18
feat(webpack): add emitting lockdown.js to dist

naugtur:
branched from the release setup. I need to learn how to use the tool


leotm (approved):
asset `lockdown.js` emitted to `/dist` ✅ 
running `npm test` in `/examples`

packages/webpack/src/buildtime/emitSes.js
```diff
@@ -0,0 +1,41 @@
...
+
+      compilation.emitAsset('lockdown.js', asset)
+
+      if (HtmlWebpackPluginInUse && HtmlWebpackPluginInterop) {
+        HtmlWebpackPluginInUse.constructor
+          .getHooks(compilation)
+          .beforeEmit.tapAsync('LavaMoatWebpackPlugin-lockdown', (data, cb) => {
+            const scriptTag = '<script src="./lockdown.js"></script>'
+            const headTagRegex = /<head[^>]*>/i
+            const scriptTagRegex = /<script/i
```
legobeat:
```suggestion
            const headTagRegex = /<head[^>]*>/iu
            const scriptTagRegex = /<script/iu
```
maybe?

legobeat: nit: This will incorrectly match `<scriptWhateverMaybeNotEvenAClosedTag` including occurrences in comments (and only the first occurrence), so the script will only seem to get injected but be non-operational.

Maybe correctness in that sense is not a concern at this point, just noting that this doesn't _really_ hold on untrusted input

boneskull (commented):
Curious why we need to consume `lockdown` this way?

packages/webpack/src/buildtime/emitSes.js
naugtur: Agree on the regex+html being impossible to fix. There's a long tail of cases where it could break. 
I'm willing to accept that option since it's a convenience helper and the default remains that you include lockdown at your own accord.

naugtur:
> Curious why we need to consume `lockdown` this way?

`lockdown` must remain unprocessed and run before anything else has a chance to run. It's easier to assert that by forcing the developer to include it first than to add it in front of the resulting dist file(s) as in any sufficiently complex webpack setup we might fail to predict which of the output files loads first and which of them load separately on independent pages. Until we come up with a decent default, this is the most forward-compatible way of adding it I could think of.

legobeat (approved):

PR #745, opened 2023-10-25
chore: do not call allow-scripts in workspace root

### LINKED ISSUE #707: root allow-scripts config is of dubious value
In the root `package.json` we have a `postinstall` script which does not automatically run due to `ignore-scripts = true` in our `.npmrc`.

There's also an `allow-scripts` config that _only_ allows scripts from the root `package.json` to run.  This is invoked via `npm run setup`.

It occurs to me that if the `setup` script was just changed to whatever `postinstall` is doing, we could remove both `postinstall` and the `allow-scripts` config.

NiteshSamuel:
packge.json file is updated. setup is updated with postinstall value.


package.json
```diff
@@ -8,6 +8,7 @@
   "devDependencies": {
     "@commitlint/cli": "^17.6.7",
     "@commitlint/config-conventional": "^17.6.7",
+    "@lavamoat/allow-scripts": "^3.0.0",
```
boneskull: this is not necessary, because `@lavamoat/allow-scripts` is in the `packages/allow-scripts` directory.

```suggestion
```

boneskull (changes requested):
Thanks for the PR!  I added a requested change.  There are three (3) additional issues, though.

1. Remove the `allow-scripts` property from the root `package.json`.
2. Your commit message must conform to the [conventional commits](https://conventionalcommits.org) spec.  In this case, your commit message should look something like this:

    ```
    chore: do not call allow-scripts in workspace root
    
    Closes #707
    ```
3. Merge commits are disallowed.  Rebase your branch onto `main` instead, and force-push your changes.

NiteshSamuel:
Hi @boneskull, thanks for the knowledge I appreciate it.  I did what you suggested. Have a look, and thanks again.

boneskull (approved):
Thank you!

boneskull:
Note to whoever merges: squash this one

Issue #721, opened 2023-09-14
ci: fail if working tree is dirty


boneskull:
This should be a new step.  This can happen if e.g., a dependency is updated in `package.json` but `package-lock.json` didn't get updated.

naugtur:
I'd be more specific - after the `npm ci` step it's possible that package-lock has changed - that'd be caused by versions from local package.json files being different than the noes preserved in package-lock. npm will not update the dependencies when `npm ci` runs, but it will fix/reformat the lockfile and update the version field at the top.

A test that's failing if package-lock was changed in CI would indicate "you forgot to refresh the lockfiles"

boneskull:
IMO it should be if _any_ file under VCS changes. Cause that may affect a release regardless of what it is.

PR #751, opened 2023-11-07
chore: normalize repository & engines fields

boneskull:
This ensures that the `repository` props in each workspace's `package.json` is in the object format and references its directory within this monorepo.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,7 +7,6 @@
     "": {
       "name": "lavamoat-monorepo",
       "version": "0.0.0",
-      "hasInstallScript": true,
       "workspaces": [
         "packages/*"
       ],
@@ -18575,6 +18574,10 @@
         "ses": "^0.18.5",
         "typescript": "^5.0.2",
         "webpack": "^5.84.1"
+      },
+      "engines": {
+        "node": "^16.20.0 || ^18.0.0 || ^20.0.0",
+        "npm": ">=7.0.0"
       },
       "peerDependencies": {
         "webpack": "^5.80.2"
--- a/packages/preinstall-always-fail/package.json
+++ b/packages/preinstall-always-fail/package.json
@@ -2,7 +2,7 @@
   "name": "@lavamoat/preinstall-always-fail",
   "version": "2.0.0",
   "description": "",
-  "main": "index.js",
+  "main": "package.json",
   "scripts": {
     "test": "exit 0",
     "preinstall": "echo \"Don't run npm lifecycle scripts by default! Create a .yarnrc or .npmrc and set enableScripts: false. Then, whitelist them with @lavamoat/allow-scripts\" && exit 1"
--- a/packages/preinstall-always-fail/package.json
+++ b/packages/preinstall-always-fail/package.json
@@ -2,7 +2,6 @@
   "name": "@lavamoat/preinstall-always-fail",
   "version": "2.0.0",
   "description": "",
-  "main": "package.json",
   "scripts": {
     "test": "exit 0",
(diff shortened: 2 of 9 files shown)


boneskull:
* main
  * **PR #751** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/751" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/751?utm_source=stack-comment).

packages/preinstall-always-fail/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@lavamoat/preinstall-always-fail",
   "version": "2.0.0",
   "description": "",
-  "main": "index.js",
+  "main": "package.json",
```
legobeat: Why should the entrypoint be `package.json` here?

leotm (commented):
we could also add `"directory": "packages/lavapack"` to lavapack

https://github.com/LavaMoat/LavaMoat/blob/f27bf4354c561f66a039921bb2ac6abb1a625819/packages/lavapack/package.json#L39-L42

package-lock.json
```diff
@@ -7,7 +7,6 @@
     "": {
       "name": "lavamoat-monorepo",
       "version": "0.0.0",
-      "hasInstallScript": true,
```
leotm, line 10: i spotted this change too running `npm i` in `main`

think we forgot to commit this in
- https://github.com/LavaMoat/LavaMoat/pull/745

after removing `postinstall` and moving it to our `setup` script

boneskull: yeah.  I didn't realize it would need to be updated.  TIL!

packages/preinstall-always-fail/package.json
boneskull: there is no `index.js` in this package, but there is a `package.json`. 😄

boneskull:
looks like I missed lavapack

package-lock.json
boneskull: relevant: https://github.com/LavaMoat/LavaMoat/issues/721

packages/preinstall-always-fail/package.json
legobeat: Could we just omit the `main` field alltogether since there is no entry-point?

legobeat:
```suggestion
```

leotm (approved):

packages/preinstall-always-fail/package.json
leotm: would agree on omitting [based](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#main) on
> if your package is named `foo`, and a user installs it, and then does `require("foo")`, then your main module's exports object will be returned.

since `package.json` doesn't export an obj, _but_
> If `main` is not set, it defaults to `index.js` in the package's root folder.

which is dubious since `index.js` doesn't exist...

so guess it's one of those pick your poison decisions 🐍

legobeat (approved):

packages/preinstall-always-fail/package.json
boneskull: `require('some.json')` just gives you an object, so `require('preinstall-always-fail')` will now give you an object instead of throwing an exception.  It doesn't really matter, given that this package isn't intended to do anything except _be_.

boneskull: I'll just remove it.  I don't think it matters.

legobeat (approved):

PR #742, opened 2023-10-20
feat(endomoat): placeholder PR

boneskull:
This is a WIP branch focused on rebuilding `lavamoat` on top of `@endo/compartment-mapper`.

## Setup

> [!NOTE]
> Updated _Nov 22 2023_

1. Clone https://github.com/boneskull/endo or add it as a remote to an existing working copy of https://github.com/endojs/endo.
2. Checkout branch https://github.com/boneskull/endo/tree/boneskull/for-endomoat (`boneskull/for-endomoat`)
3. Run `yarn install`
4. Since `npm link` doesn't work well in workspaces _and_ Yarn is involved, we should just create tarballs of our `endojs/endo` branch. 
    ```bash
    rm -rf /tmp/endo && mkdir -p /tmp/endo && npm pack -w @endo/compartment-mapper \
      -w @endo/evasive-transform -w ses --pack-destination=/tmp/endo
    ```
6. Navigate into your `lavamoat/lavamoat` working copy and checkout `endo-refactor`
7. Run `npm install`
8. Overwrite deps from our tarballs:
    ```bash
    npm install --no-save -w @lavamoat/endomoat /tmp/endo/*.tgz
    ```
9. Run `npm ls @endo/compartment-mapper` to validate you've done it correctly. You should see two instances; one for `@lavamoat/endomoat` and another for `@lavamoat/perf`, but they should _not_ mention `deduped`. For example:
    ```
    lavamoat-monorepo@0.0.0 /Users/boneskull/projects/lavamoat/lavamoat
    ├─┬ @lavamoat/endomoat@2.3.1 -> ./packages/endomoat
    │ └── @endo/compartment-mapper@0.9.2
    └─┬ lavamoat-perf@0.0.0 -> ./packages/perf
      └── @endo/compartment-mapper@0.9.2
    ```
10. If you're feeling lucky, run `npm run rebuild:types`, which should only succeed if the above steps were taken successfully.  _This does not mean it will succeed_, but it absolutely won't otherwise!

PR #733, opened 2023-10-06
feat(core): add a schema for policy files

boneskull:
Continuation of #620.

[This](https://snowplow.io/blog/introducing-schemaver-for-semantic-versioning-of-schemas/) seems like a reasonable idea for schema versioning.

To test in VS Code, you can add this property to any policy file:

```json
"$schema": "https://raw.githubusercontent.com/boneskull/LavaMoat/boneskull/policy-schema/packages/core/schema/lavamoat-policy.v0-0-1.schema.json"
```

PR #752, opened 2023-11-07
chore(types): normalize typescript configuration locations

boneskull:
@leotm rightly had some questions about the `@typescript-eslint` configuration in a previous PR, and I've attempted to answer those here.

I had experimented with stuffing `tsconfig.json` files into subdirectories (e.g., `config`) to keep them out of the sources. This becomes problematic due to the way that `@typescript-eslint` reads `tsconfig.json` files; it wants to read the closest ancestor `tsconfig.json` (using its setting `project: true`). This is the _same_ strategy ESLint uses to resolve ESLint configs, so it kind of makes sense.

Specifying a list of `tsconfig.json` files for the `project` setting picks the first one it can find relative to the file under lint. I _think_ that might make sense if we have multiple `tsconfig.json` files where the sources (its `includes` prop) have overlap. e.g., a `tsconfig.json` for builds and another for linting. That's not our use case, though--any given file has no more than one associated `tsconfig.json`.

To that end, any `tsconfig.json` must live at the root of its sources--its `rootDir`.


boneskull:
* **#752** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/752?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3  dependent PRs ([#742](https://github.com/LavaMoat/LavaMoat/pull/742) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/742?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#761](https://github.com/LavaMoat/LavaMoat/pull/761) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/761?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#770](https://github.com/LavaMoat/LavaMoat/pull/770) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/770?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈
* **#732** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/732?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#733](https://github.com/LavaMoat/LavaMoat/pull/733) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/733?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#755](https://github.com/LavaMoat/LavaMoat/pull/755) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/755?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

packages/core/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "../../config/tsconfig.workspace.json",
   "files": [],
   "references": [
-    { "path": "./config/tsconfig.sources.json" },
-    { "path": "./config/tsconfig.test.json" }
+    { "path": "./src/tsconfig.json" },
+    { "path": "./test/tsconfig.json" }
```
leotm: or do we prefer the shorthand

```suggestion
"references": [{ "path": "./src" }, { "path": "./test" }]
```

like we do with [`aa`](https://github.com/LavaMoat/LavaMoat/pull/752/files#diff-5bafdc438b51c268487cdf7d7c9625b4a3e8dde54aaa2742c13e4065bc12b06fR4) and [`allow-scripts`](https://github.com/LavaMoat/LavaMoat/pull/752/files#diff-eb3256b16392cdc709989b19046f3f164c77ed1cf459cc473b1f99b8804b8153R4)?

leotm: <details><summary>minor consideration pos worth noting RE dev experience</summary>

---

the longhand traverses IDE paths nicely

Screenshot 2023-11-09 at 4 29 12 pm

Screenshot 2023-11-09 at 4 30 04 pm

---

whereas shorthand breaks it

Screenshot 2023-11-09 at 4 30 33 pm

Screenshot 2023-11-09 at 4 31 05 pm

</details>

tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "@tsconfig/node16/tsconfig.json",
```
leotm, line 2: i know unrelated to this pr, but could pos simplify further extending our base instead?

```suggestion
  "extends": "./config/tsconfig.base.json",
```

so no longer extending `@tsconfig/node16/tsconfig.json` directly in two places

---

or can get this merged first and i'll follow-up

leotm (commented):
looks like we may need to add

```json
"composite": true
```

to few of our referenced package tsconfigs?

---

edit: ah nvm, we're already doing so by extending our `../../config/tsconfig.workspace.json` for each pkg

tsconfig.json
```diff
@@ -2,7 +2,8 @@
...
   "files": [],
   "compilerOptions": {
-    "noEmit": true
+    "noEmit": true,
+    "allowJs": true
```
leotm, line 5: was wondering if this is needed? since the types still rebuild fine without it

leotm: slightly more unrelated to this pr, noticed

```json
{
  "extends": "@tsconfig/node16/tsconfig.json",
  "files": []
}
```

is all we need for types to still rebuild successfully

so can we get rid of the remaining clutter? 

---

again could get this merged first and i could follow-up

packages/core/tsconfig.json
boneskull: I think I prefer to be explicit about the `tsconfig.json` file -- simply because cmd-clicking on the string in VSCode will navigate to the file, whereas omitting `tsconfig.json` will not.

Also it's required if your file _isn't_ called `tsconfig.json` for whatever reason.

So I'll normalize that, too.

tsconfig.json
boneskull: Yeah, maybe we should do this.  I wasn't really sure.  AFAICT, the workspace root configuration will be used in VS Code when a file should be typechecked (e.g., with a `// @ts-check` directive) but is otherwise not included in a `tsconfig.json`.  Think something like `scripts/foo.js` where the file should be typechecked, but it isn't shipped in the tarball nor are declarations generated for it.  But I'm not entirely sure--it might just use the closest ancestor `tsconfig.json`.   

The `composite: true` `compilerOption` in `tsconfig.base.json` doesn't make sense for the root workspace, either.  I don't think that has any negative consequences, but again, I'm not sure. It would be allowed to be referenced in as part of an incremental build setup in some `tsconfig.json` which is _not_ part of this repo?  That _might_ be useful, for somebody, somehow??

tsconfig.json
boneskull: Yeah, I can't remember what I was thinking regarding `noEmit`.  It seemed like a good idea to default to "don't emit anything", but given it'd only apply to files which aren't part of the build _anyway_, there's no point.

As I mentioned above, these config files affect more than just the build, and this root one serves as sort of "default settings" for the entire repo.  It also comes into play when using `@typescript-eslint/parser` (and Wallaby, but that's a tool only I use, afaik); there are files that are linted with this parser but are otherwise not part of the build.  

There's also some interplay with VSCode settings such as "implicit type-checking in JavaScript files". 

I'm not going to be able to verify that `@typescript-eslint/parser` actually needs that `allowJs: true` in _this_ PR (will need to do that in #753), but using a sane default like `@tsconfig/node16` seems reasonable, at minimum.

Are you saying that we can remove the references and `tsc` will find all of the `tsconfig.json` files in the repo and build them??

tsconfig.json
leotm: > Yeah, maybe we should do this. I wasn't really sure. AFAICT, the workspace root configuration will be used in VS Code when a file should be typechecked (e.g., with a `// @ts-check` directive) but is otherwise not included in a `tsconfig.json`. Think something like `scripts/foo.js` where the file should be typechecked, but it isn't shipped in the tarball nor are declarations generated for it. But I'm not entirely sure--it might just use the closest ancestor `tsconfig.json`.

yeah initially hard to tell 😅 since alot of our opts influence the build output
so briefly tested our current setup to shed some light
on `packages/aa/test/index.spec.js` adding `// @ts-check` and `const a = 5`
then adding to relevant tsconfig's `"compilerOptions": { "noUnusedLocals": true }`
- `tsconfig.json` - no effect
- `config/tsconfig.base.json` - TS6133
  - `config/tsconfig.test.json` - TS6133
    - `packages/aa/test/tsconfig.json` - TS6133
  - `config/tsconfig.workspace.json` - no effect
    - `packages/aa/tsconfig.json` - no effect

(then ofc `packages/aa/src/tsconfig.json` and `config/tsconfig.emit-declarations.json` not relevant here)

so looks like root `tsconfig.json` ignored
and yeah closest ancestor used

> The `composite: true` `compilerOption` in `tsconfig.base.json` doesn't make sense for the root workspace, either. I don't think that has any negative consequences, but again, I'm not sure. It would be allowed to be referenced in as part of an incremental build setup in some `tsconfig.json` which is _not_ part of this repo? That _might_ be useful, for somebody, somehow??

i'm not sure either 😅

config/tsconfig.test.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "composite": true,
+    "checkJs": true,
+    "emitDeclarationOnly": false,
```
leotm: `"emitDeclarationOnly": false,` worth keeping here? since it default's to `false`

config/tsconfig.test.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "composite": true,
```
leotm: can maybe leave out `"composite": true,` since we already got it from our tsconfig.base?

config/tsconfig.workspace.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "composite": true,
```
leotm: can maybe leave out `"composite": true,` since we already got it from our tsconfig.base?

config/tsconfig.workspace.json
```diff
@@ -0,0 +1,8 @@
...
+    "allowJs": true,
+    "checkJs": true
```
leotm: spotted our tsconfig.base already has `"allowJs": true,` which we could leave out here
(but i understand might be worth being explicit, since `checkJs` works in tandem w `allowJs`)

leotm: or we could add `checkJs` to our tsconfig.base
then remove both `allowJs` and `checkJs` from here

packages/allow-scripts/tsconfig.json
```diff
@@ -1,13 +1,8 @@
...
-      "@lavamoat/aa": ["../aa"]
-    }
-  },
-  "references": [{ "path": "../aa" }]
+  "extends": "../../config/tsconfig.workspace.json",
+  "files": [],
+  "references": [
+    { "path": "./src/tsconfig.json" },
+    { "path": "./test/tsconfig.json" }
+  ]
```
leotm: how come no more path/ref to aa btw?

leotm: ah it's been moved/split into `./src/tsconfig.json` nice

config/tsconfig.emit-declarations.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./tsconfig.workspace.json",
+  "compilerOptions": {
+    "declaration": true,
+    "declarationMap": true,
+    "emitDeclarationOnly": true,
+    "noEmit": false
+  }
```
leotm, line 8: worth adding our `outDir` and `rootDir` here?

```suggestion
  "compilerOptions": {
    "declaration": true,
    "declarationMap": true,
    "emitDeclarationOnly": true,
    "noEmit": false
  }
```

so simplify our packages extending it

leotm: worth adding our `outDir` and `rootDir` here?

```suggestion
  "compilerOptions": {
    "outDir": "../types",
    "rootDir": "."
    "declaration": true,
    "declarationMap": true,
    "emitDeclarationOnly": true,
    "noEmit": false
  }
```

to simplify our packages extending this (via their `src/tsconfig.json`)

tsconfig.json
leotm: > Yeah, I can't remember what I was thinking regarding `noEmit`. It seemed like a good idea to default to "don't emit anything", but given it'd only apply to files which aren't part of the build _anyway_, there's no point.
> 
> As I mentioned above, these config files affect more than just the build, and this root one serves as sort of "default settings" for the entire repo. It also comes into play when using `@typescript-eslint/parser` (and Wallaby, but that's a tool only I use, afaik); there are files that are linted with this parser but are otherwise not part of the build.
> 
> There's also some interplay with VSCode settings such as "implicit type-checking in JavaScript files".
> 
> I'm not going to be able to verify that `@typescript-eslint/parser` actually needs that `allowJs: true` in _this_ PR (will need to do that in #753), but using a sane default like `@tsconfig/node16` seems reasonable, at minimum.

👍 i bought Quokka with Wallaby discounted in their bundle (love a good deal), but barely touched Wallaby 😅

> Are you saying that we can remove the references and `tsc` will find all of the `tsconfig.json` files in the repo and build them??

i tested removing the references then rebuilding types, but not fully inspecting the outputs are what/how we want oops, after some more reading can see why they're needed

leotm (approved):
🚀

(noting remaining CI lint addressed in https://github.com/LavaMoat/LavaMoat/pull/753)

tsconfig.json
boneskull: I think I'd prefer to leave it as-is.  My _intent_ for `tsconfig.base.json` is that _non-root_ `tsconfig.json` files extend it.  Since I'd rather not add comments to `tsconfig.base.json` (while allowed by TS, but possibly impedes integration with other tools), I'll create `config/README.md` explaining myself.

config/tsconfig.test.json
boneskull: If `noEmit` is `false` then `emitDeclarationOnly` cannot be `true` (TS will error and abort).  At some point, the config this one extended had `emitDeclarationOnly: true` in it, which I needed to override with `false`.  That could be removed now.

config/tsconfig.workspace.json
boneskull: I think we can actually just eliminate one of these config files altogether.

config/tsconfig.emit-declarations.json
boneskull: unsure how relative paths are determined. it might be from the extended config or from the extendee config.  will give it a shot.

boneskull: yeah, it works!

boneskull: Well.  OK, so .. it works for _some_ properties.  It works for `outDir`.  It doesn't work for `rootDir`.  It doesn't work for `references`, either.  

`rootDir` only needs setting if it isn't `.`, anyway.

boneskull: No, it doesn't work.  I had confused myself somehow.  All paths are relative to the path of the config file.

.config/README.md
```diff
@@ -0,0 +1,70 @@
+# Some Config Files Are Here
+
+In here, config files some.
```
naugtur, line 3: Wut?

Also, this suggests we could move more here?

.config/README.md
```diff
@@ -0,0 +1,70 @@
...
+   {
+     "extends": "../../.config/tsconfig.workspace.json",
+     "references": [
+       { "path": "./src/tsconfig.json" },
+       { "path": "./test/tsconfig.json" }
+     ]
+   }
+   ```
+
+   If you don't want to typecheck your tests, omit the second object in the `references` array.
```
naugtur, line 33: I do have mixed feelings about typechecking tests. Mostly because I worry it might get in the way. Happy to give it a spin and push back if it does.

.config/README.md
```diff
@@ -0,0 +1,70 @@
...
+   {
+     "extends": "../../../.config/tsconfig.src.json",
+     "compilerOptions": {
+       "outDir": "../types"
+     },
+     "include": ["."]
+   }
+   ```
+
+   This file controls how and where `tsc` emits declarations; our convention is to dump them all in `types/`--which Git ignores--but `npm` will publish otherwise (see step 6).
```
naugtur:
```suggestion
   This file controls how and where `tsc` emits declarations; our convention is to dump them all in `types/`. We also want to publish these types along with our packages (see step 6).
```

.config/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+   `tsconfig.src.json` contains `checkJs: true`; `tsc` will expect _all_ `.js` files to be well-typed. If that's a problem (it probably is), override the value to `false` and add `// @ts-check` to the top of each file needing typechecking.
```
naugtur:
```suggestion
   `tsconfig.src.json` contains `checkJs: true`; `tsc` will expect _all_ `.js` files to be well-typed. If that's not applicable to the package you're setting up, override the value to `false` and add `// @ts-check` to the top of each file needing typechecking. 
```

.config/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+   If you are a glutton for punishment, add `strict: true` to this config file.
```
naugtur:
```suggestion
   If you are a glutton for punishment, add `strict: true` to this config file. It's not the default for this repository tho.
```

.config/README.md
```diff
@@ -0,0 +1,70 @@
...
+     "references": [{ "path": "../src/tsconfig.json" }]
+   }
+   ```
+
+   You may need to omit e.g., test fixtures; you can use the `exclude` prop for this. `tsc` will not build declarations from files in `test`.
+
+4. Add a `types` prop in your workspace's `package.json` pointing to the declaration file that corresponds to your entry point. Generally this is `types/index.d.ts`.
+   > If your entry point is not in `src/`, you will need to add `"rootDir": ".."` to `src/tsconfig.json`'s `compilerOptions`, then add `../index.js` (or whatever it is) to the `includes` property.
+5. _If you have an `exports` field_, [read this](https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports) and do that.
+6. Add `types`, `!tsconfig.json` and `!*.tsbuildinfo` to the `files` prop of your `package.json`. If it doesn't exist, create it, and add whatever other files you need to put in the published package. Check your work with `npm pack --dry-run`.
```
naugtur, line 70: what are *.tsbuildinfo and why do we generate them?

If our tsconfig was not extending a bunch of stuff that won't end up in the tarball anyway, I'd be voting for keeping it in the tarball so people can check which typescript we've been using out of the many possible permutations. (is the community at large getting any closer to settling on one?)

packages/allow-scripts/src/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../../.config/tsconfig.src.json",
+  "compilerOptions": {
+    "outDir": "../types",
+    "paths": {
+      "@lavamoat/aa": ["../../aa"]
+    }
+  },
+  "include": ["."],
+  "references": [{ "path": "../../aa/tsconfig.json" }]
```
naugtur, line 10: Why do we need references? I thought I know what they're for, but then Leo told me everything works the same with them removed...

packages/core/src/endowmentsToolkit.js
```diff
@@ -1,12 +1,26 @@
-// the contents of this file will be copied into the prelude template
-// this module has been written so that it required directly or copied and added to the template with a small wrapper
```
naugtur: (could) I see this was moved into the comment below, but I'd leave at least a short warning at the top

(yes, I know we've had that conversation before, but then I didn't even notice the comment was moved - hence the suggestion to keep a tiny warning)

packages/core/src/endowmentsToolkit.js
```diff
@@ -263,20 +329,29 @@ function endowmentsToolkit({
...
+      const getter = sourcePropDesc.get
       wrappedPropDesc.get = function () {
+        // eslint-disable-next-line @typescript-eslint/no-this-alias
         const receiver = this
         // replace the "receiver" value if it points to fake parent
         const receiverRef =
           receiver === unwrapFromGlobalThis ? unwrapToGlobalThis : receiver
         // sometimes getters replace themselves with static properties, as seen wih the FireFox runtime
-        const result = Reflect.apply(sourcePropDesc.get, receiverRef, [])
+        const result = Reflect.apply(getter, receiverRef, [])
```
naugtur: (must) Let's make sure this is not breaking the code. I'm pretty sure the comment above is explaining why we're not using a local copy but referencing the getter from the original source and this change would break it.

packages/core/src/endowmentsToolkit.js
```diff
@@ -295,17 +374,26 @@ function endowmentsToolkit({
...
     if (sourcePropDesc.set) {
+      const setter = sourcePropDesc.set
       wrappedPropDesc.set = function (value) {
         // replace the "receiver" value if it points to fake parent
+        // eslint-disable-next-line @typescript-eslint/no-this-alias
         const receiver = this
         const receiverRef =
           receiver === unwrapFromGlobalThis ? unwrapToGlobalThis : receiver
-        return Reflect.apply(sourcePropDesc.set, receiverRef, [value])
+        return Reflect.apply(setter, receiverRef, [value])
```
naugtur: (must) same as above

packages/core/src/generatePolicy.js
```diff
@@ -160,6 +213,7 @@ function createModuleInspector(opts = {}) {
     if (includeDebugInfo) {
       const moduleDebug = debugInfo[moduleRecord.specifier]
       moduleDebug.sesCompat = {
+        // XXX: I don't think this is needed, since it appears we overwrite all properties
```
naugtur: (nit) Let's use one of the TODO: or FIXME: if this comment is here to stay and not just meant for the review. 

I don' know what's the full shape of sesCompat. This might be bikeshedding future-proofing

naugtur (changes requested):
All comments optional except one where I think the change is introducing a bug
https://github.com/LavaMoat/LavaMoat/pull/752/files#diff-5ed0581fdb5b3ec8105f7f2e6773d6f5db45dbe74dccf2a921cc9c0e40ecda60R354

packages/core/src/endowmentsToolkit.js
naugtur: Also, this change is already present in the branch which is set as a merge target for this PR. Something broke with the hierarchy there.

.config/README.md
boneskull: It can get in the way, especially when straying off of the happy path.  It may or may not be worth it.  Generally, I think it is.

.config/README.md
boneskull: Yes. In fact, I do in #753.

that sentence was bad on purpose. I'll make it less bad.

.config/README.md
boneskull: `tsbuildinfo` files are for TS' incremental builds.  When you have a monorepo with a bunch of `tsconfig.json`'s laying about, and they have interdependencies, TS creates these files so it knows what it does/doesn't need to recompile.

boneskull: re: shipping `tsconfig.json`, I'm not sure how that would help consumers.  It doesn't have any impact on how TS consumes our packages.  Further, since we're depending on e.g., `@tsconfig/node16`, it wouldn't work anyway w/o adding that package as a prod dep.

packages/allow-scripts/src/tsconfig.json
boneskull: It might work the same if you're lucky.  The references basically tell `tsc`'s incremental builder "compile x before y".  This is especially important when running `tsc --watch`, but if you omit them, you can get into weird situations where some workspace either hasn't been compiled yet or it's providing old types to its consumer.  I have suffered.

packages/core/src/endowmentsToolkit.js
boneskull: Whatever the problem here was, it's resolved now.  I'm not sure what happened.

packages/core/src/generatePolicy.js
boneskull: hmm, I'll apply this suggestion to #732

packages/allow-scripts/src/tsconfig.json
boneskull: cc @leotm

boneskull:
@naugtur This is blocked on you

legobeat (approved):
lgtm

naugtur (approved):

PR #753, opened 2023-11-07
chore: reconfigure eslint

boneskull:
The aim here is:

- to avoid excess churn in the root `.eslintrc.js` file
- to provide some reusable config for workspaces
- allow execution of ESLint from the workspace root _and_ any workspace directory while retaining proper ignore patterns
- validate the shape of configs via TS
- maximize compat with `@typescript-eslint/parser`, which should be used whenever we have a reasonable amount of typed JS
- re-enable some rules; create exceptions at the file level
- enable warnings for path concatenation

This may make `.eslintignore` redundant again.

## Reviewers:

- This PR contains a change which applies lint fixes to the codebase.  It is here because CI would fail otherwise.
- To review this, you'll want to look at changes in **`.config/`**, any **`.eslintrc.js`** and any **`.depcheckrc`** files. You can ignore anything else unless you want to look at what actually changes.

### DIFF
--- a/.config/README.md
+++ b/.config/README.md
@@ -68,3 +68,57 @@ In addition the root `tsconfig.json` serves as the "default" for files in the re
    > If your entry point is not in `src/`, you will need to add `"rootDir": ".."` to `src/tsconfig.json`'s `compilerOptions`, then add `../index.js` (or whatever it is) to the `includes` property.
 5. _If you have an `exports` field_, [read this](https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports) and do that.
 6. Add `types`, `!tsconfig.json` and `!*.tsbuildinfo` to the `files` prop of your `package.json`. If it doesn't exist, create it, and add whatever other files you need to put in the published package. Check your work with `npm pack --dry-run`.
+
+## ESLint Configuration
+
+This directory contains the following configuration(s):
+
+- [`eslintrc.typed-workspace.js`](./eslintrc.typed-workspace.js) - to be extended in the `.eslintrc.js` of a TypeScript-enabled workspace.
+
+  This config uses a type-aware parser (`@typescript-eslint/parser`) to provide more accurate linting, which works on both TypeScript sources and JavaScript sources (those which are type-checked, anyhow).
+
+### How to Configure a Workspace for ESLint
+
+All workspaces will inherit config from the [root ESLint config][].
+
+Please use a CJS file named `.eslintrc.js` (for consistency).
+
+#### TypeScript-Enabled Workspaces
+
+For TS-enabled workspaces, **an `.eslintrc.js` file is required**.
+
+Scaffold your `.eslintrc.js` like so:
+
+```js
+// @ts-check
+
+/**
+ * @type {import('eslint').Linter.Config}
+ */
+module.exports = {
+  extends: '../../.config/eslintrc.typed-workspace',
+}
+```
+
+#### Everything Else
+
+For other workspaces, **`.eslintrc.js` is optional**. It is only necessary if the workspace needs no changes from the [root ESLint config][].
+
+It's recommended to use this as a starting point, as it will give you in-editor feedback if your config is invalid:
+
+```js
+// @ts-check
+
+/**
+ * @type {import('eslint').Linter.Config}
+ */
+module.exports = {
+  // go to town here
+}
+```
+
+### Resources
+
+- [ESLint Configuration Docs](https://eslint.org/docs/user-guide/configuring)
+
+[root ESLint config]: ../.eslintrc.js
--- a/.config/README.md
+++ b/.config/README.md
@@ -1,16 +1,16 @@
 # Some Config Files Are Here
 
-In here, config files some.
+> All config files in this directory should be documented in this `README`.
 
 ## TypeScript Configuration
 
 We have three (3) TypeScript configuration files here:
 
-- `tsconfig.workspace.json` - "Base" config for workspaces in this repo.
-- `tsconfig.src.json` - Config for files within a workspace's `src/` directory (sources).
-- `tsconfig.test.json` Config for files within a workspace's `test/` directory (tests).
+- [`tsconfig.workspace.json`][] - "Base" config for workspaces in this repo.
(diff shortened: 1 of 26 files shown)


boneskull:
* **#753** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/753?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#831** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/831?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#809](https://github.com/LavaMoat/LavaMoat/pull/809) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/809?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#820](https://github.com/LavaMoat/LavaMoat/pull/820) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/820?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

boneskull:
Just wanted to note this, as I found it kind of quirky.

From the `ignorePatterns` [documentation](https://eslint.org/docs/latest/use/configure/ignore#ignorepatterns-in-config-files)

> If a glob pattern starts with `/`, the pattern is relative to the base directory of the config file. For example, `/foo.js` in `lib/.eslintrc.json` matches to `lib/foo.js` but not `lib/subdir/foo.js`.

> If a config is provided via the `--config` CLI option, the ignore patterns that start with `/` in the config are relative to the current working directory rather than the base directory of the given config. For example, if `--config configs/.eslintrc.json` is present, the ignore patterns in the config are relative to `.` rather than `./configs`.

boneskull:
Updated `.config/README.md` with information about configuring ESLint.

.eslintrc.js
```diff
@@ -1,22 +1,62 @@
...
+ * - Note: glob patterns in this file should _not_ be relative.
+ *
+ * @packageDocumentation
+ */
+
+/**
+ * The minimum EcmaScript runtime environment to support
+ * @see {@link https://eslint.org/docs/latest/use/configure/language-options#specifying-environments}
+ */
+const ECMASCRIPT_ENV = 'es2021'
```
legobeat, line 21: Please break out the bump from es2020 to es2021 in a follow-up to the refactoring (which otherwise looks great!)?

naugtur: There's one area in which we cannot go beyond es2020. I'd have to dig out why, but I remember adding a test to ensure we produce a dist file that's es2020

Now that we moved away from old Node versions this might be irrelevant, but I'm trying to recall how we found out about that issue when we broke it back then.

boneskull: If we dropped Node.js v14, then our minimum version is ES2021 unless it's a browser thing.

boneskull: @naugtur The ES2020 check arrived in https://github.com/LavaMoat/LavaMoat/pull/467, but there's little context there as to _why_, though I imagine it's because of whatever browsers MM supports.

Based on the versions mentioned in [this comment](https://github.com/MetaMask/metamask-extension/issues/6805#issuecomment-1398953144) have full support for ES2021, so we should be able to use it if Node.js v16 has full support (which it does).

Thoughts?  I'm not sure where to get more information.

cc @legobeat

legobeat: I think that dropping es2020 support at this point should be reasonable but regardless: Still reinforcing that in any case, it should be done in a separate changeset from a PR such as this - doing that is orthogonal to the aims of this PR.

Dropping support for a major ES version is not inside the scope of reconfiguring linting.

.config/README.md
```diff
@@ -62,9 +60,213 @@ In addition the root `tsconfig.json` serves as the "default" for files in the re
...
+   }
+   ```
+
+> [!TIP] Use Strict
+>
+> Use `{"strict": true}` in the `compilerOptions` of your `src/tsconfig.json` for maximum safety against `null` and `undefined` values.
+>
+> This is less important for tests.
+
+### TypeScript-Related Miscellany
```
leotm, line 90:
```suggestion
### TypeScript-related Miscellany
```

.eslintrc.js
naugtur: The ES2020 test was introduced because lavapack runtime is inlined directly into the final bundle, so it dictates what the bundles built with lavamoat support. And MM is supporting (or was at the time) browsers that were missing a feature form 2021. 
If we want to introduce it, it's a big breaking change with a big red warning for users.

This setting is fine for all code other than the runtime files that th test is covering, so it's a matter of skipping linting them which we might be already doing.

boneskull: @legobeat Based on that information, are we OK with this?  the runtime files in lavapack [are _not_ linted](https://github.com/LavaMoat/LavaMoat/blob/9e9a15f29fbdc758fa45e9c8370fa5abe864c7e6/.eslintignore#L8-L10)

boneskull: @naugtur but, would it not make sense to lint those files loosely but restrict the language level to es2020?  I'm guessing that's `lavapack/src/runtime-cjs.js` and `lavapack/src/runtime.js`?  are they generated?

legobeat: Proposed alternative (with same end-result):
- #1022
  - #1023

leotm: since `lavapack/src/builder-runtime.js` asserts `ECMA_VERSION_2020`
before generating `lavapack/src/runtime-cjs.js` and `lavapack/src/runtime.js`
from `lavapack/src/runtime-cjs-template.js` and `lavapack/src/runtime-template.js`

think suggested PR split makes sense? so we bump `lavapack/src/builder-runtime.js` from `ECMA_VERSION_2020` to `ECMA_VERSION_2021` before
- https://github.com/LavaMoat/LavaMoat/pull/1023

---

edit: nvm, after discussing, we want to stick to `lavapack/src/builder-runtime.js` asserting `ECMA_VERSION_2020` for browser compat, unaffected by eslint ecma bump

leotm: and prob no need to loosely lint our target generated `lavapack/src/runtime-cjs.js` and `lavapack/src/runtime.js` files? since their src templates already linted

PR #757, opened 2023-11-13
chore: fix blame display to stop attributing everything to Chris

naugtur:


boneskull (approved):
Didn't know you could do this!  Cool.  TIL

PR #758, opened 2023-11-13
refactor(core): extract scuttling from the kernel

### LINKED ISSUE #731: refactor scuttling in core for consumption by the webpack plugin and lavamoat-endo
- Scuttling is currently integrated deeply into the kernel. 
- Webpack plugin (and future lavamoat-endo) is not using the kernel
- We still want scuttling
- We need standalone scuttling in core that can be imported as a single module and used with a call of a function (or a bunch)

naugtur:
Closes #731

- [x] extract scuttling to a separate file
- [ ] ~~make it an official export (preferably along with endowmentsToolkit)~~ (decided to perform in a separate PR)
- [x] add a self-contained test

@legobeat I'm not sure how this relates to the planned refactoring of scuttling, but this is what Chris and I need to use it elsewhere.


boneskull:
@naugtur Do you want me to review this while it's still in draft?

naugtur:
@boneskull yes, I want you to take a look and holler at me if anything looks like it'd not be usable in Endomoat if extracted this way.
I did my best for it to be usable for our new stuff.

packages/core/src/generateKernel.js
```diff
@@ -22,6 +22,7 @@ const endowmentsToolkitSrc = fs.readFileSync(
   path.join(__dirname, '/endowmentsToolkit.js'),
   'utf-8'
 )
+const scuttleSrc = fs.readFileSync(path.join(__dirname, '/scuttle.js'), 'utf-8')
```
boneskull:
```suggestion
const scuttleSrc = fs.readFileSync(path.join(__dirname, 'scuttle.js'), 'utf-8')
```

packages/core/src/kernelCoreTemplate.js
```diff
@@ -69,22 +69,16 @@
...
     const rootPackageCompartment = createRootPackageCompartment(globalRef)
 
-    // scuttle globalThis right after we used it to create the root package compartment
-    if (scuttleOpts.enabled) {
-      if (!Array.isArray(scuttleOpts.exceptions)) {
-        throw new Error(`LavaMoat - scuttleGlobalThis.exceptions must be an array, got "${typeof scuttleOpts.exceptions}"`)
-      }
-      scuttleOpts.scuttlerFunc(globalRef, realm => performScuttleGlobalThis(realm, scuttleOpts.exceptions))
-    }
+    applyDefaultScuttling(globalRef, scuttleGlobalThis)
```
boneskull: Default as opposed to what?

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,185 @@
...
+    current = Reflect.getPrototypeOf(current)
+  }
+  return protoChain
+}
+
+/**
+ * @type {object}
+ * @property {Function} performScuttleGlobalThis - Scuttle the globalThis object once.
+ * @property {Function} applyDefaultScuttling - Apply default scuttling from options.
+ */
```
boneskull:
```suggestion
```

Use of `Function` is a) discouraged by TS and b) causes the information about the function's parameters and return value to be lost.  It's also an object literal, which will be inferred by TS.

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,185 @@
...
+        `LavaMoat - property "${prop}" of globalThis cannot be set under scuttling mode. ` +
+          'To learn more visit https://github.com/LavaMoat/LavaMoat/pull/360.'
+      )
+    }
+    // eslint-disable-next-line no-inner-declarations
+    function get() {
+      throw new Error(
+        `LavaMoat - property "${prop}" of globalThis is inaccessible under scuttling mode. ` +
+          'To learn more visit https://github.com/LavaMoat/LavaMoat/pull/360.'
+      )
```
boneskull: Our errors should probably subclass `Error` and have error codes, but idk if you want to visit that here.

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,185 @@
...
+ * @typedef {object} ScuttleOpts
+ * @property {boolean} enabled - Whether scuttling is enabled or not.
+ * @property {Array<string|RegExp>} exceptions - List of properties to exclude from scuttling.
+ * @property {string} scuttlerName - Name of the scuttler function to use.
+ * @property {(globalRef: object, scuttle: Function) => void} scuttlerFunc - Scuttler function to use.
+ */
+
+/**
+ * @typedef {object} GlobalRef
+ * @property {object} [globalThis] - Reference to the global object.
```
boneskull: If `globalThis` is intended to be read-only for consumers of this code, then the type `object` is OK enough. If we intend to write to this property, then it should be `Record<string, any>`

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,185 @@
...
+} = Object
+
+/**
+ * Applies scuttling, with the default set of options, including using Snow if passed in as scuttlerFunc.
+ * cSuttle globalThis right after we used it to create the root package compartment
+ *
+ * @param {GlobalRef} globalRef - Reference to the global object.
+ * @param {ScuttleOpts} opts - Scuttling options.
+ */
+function applyDefaultScuttling(globalRef, opts) {
```
boneskull: If we are willing to change the signature here, I like it when "options" are all optional. 😄

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,185 @@
+/**
+ * @typedef {object} ScuttleOpts
+ * @property {boolean} enabled - Whether scuttling is enabled or not.
+ * @property {Array<string|RegExp>} exceptions - List of properties to exclude from scuttling.
+ * @property {string} scuttlerName - Name of the scuttler function to use.
+ * @property {(globalRef: object, scuttle: Function) => void} scuttlerFunc - Scuttler function to use.
```
boneskull: Whatever `scuttle` is, we should define it with a `@callback` and use that instead.

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,185 @@
...
+function shouldAvoidProp(propsToAvoid, prop) {
+  for (const avoid of propsToAvoid) {
+    if (avoid instanceof RegExp && avoid.test(prop)) {
+      return true
+    }
+    if (propsToAvoid.has(prop)) {
+      return true
+    }
+  }
+  return false
```
boneskull: The `Set` is useful to avoid duplicate `string`s, but has no upside if all of the members are `RegExp` objects.  Two separate `RegExp` objects are obviously not guaranteed to not test `true` for the same string.  But the function is kind of difficult to understand as written.  This one is shorter and probably slower: 

```suggestion
  return [...propsToAvoid].some(avoid => new RegExp(avoid).test(prop))
```

don't really care what you use instead as long as it's easier to understand.

boneskull (commented):
There's nothing here that's going to be obviously problematic for endomoat

legobeat:
> @legobeat I'm not sure how this relates to the planned refactoring of scuttling, but this is what Chris and I need to use it elsewhere.

Overall direction looks good so far - isolating the scuttling logic into its own module should make that more straight-forward :+1:

boneskull:
Does it make sense to pull this entirely out of `core` and into its own package?

naugtur:
> Does it make sense to pull this entirely out of `core` and into its own package?

Definitely not as long as it's part of the core kernel, where it gets inlined via templateRequire.

packages/core/src/kernelCoreTemplate.js
naugtur: default as in "how it used to work" with all the config and the usage of Snow (passed as scuttlerFunc)

The non-default way is to take scuttling and apply it once - via the other exported function.
We may want to use it that way in endomoat as scuttling other realms makes little sense in Node.

packages/core/src/scuttle.js
naugtur: It's meant to have some of its properties deleted, so I should probably switch to a Record.

packages/core/src/scuttle.js
naugtur: I think backwards-compatibility prevents us from changing it, but I didn't do proper research. I don't want to change too much TBH, just get this working thing to continue working but from a separate file.

packages/core/src/scuttle.js
naugtur: TBF I find the original code easier to understand than the suggestion. And I wanted to focus on refactoring this. Would you mind if I left it as-is?

weizman: But would that work for when `avoid` is a string rather than an RE? Maybe:

`Array.from(propsToAvoid).some(avoid => avoid === prop || new RegExp(avoid).test(prop))`

(we prefer avoiding `...` because transpilers sometime rewrite it and we prefer to remain in control as much as possible of what bundling outputs)

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,180 @@
...
+ * @param {Set<string|RegExp>} propsToAvoid - List of properties to exclude from scuttling.
+ * @param {string} prop - Property to check.
+ * @returns {boolean} - Whether the property should be avoided or not.
+ */
+function shouldAvoidProp(propsToAvoid, prop) {
+  for (const avoid of propsToAvoid) {
+    if (avoid instanceof RegExp && avoid.test(prop)) {
+      return true
+    }
+    if (propsToAvoid.has(prop)) {
```
weizman: I agree with @boneskull 's [comment](https://github.com/LavaMoat/LavaMoat/pull/758/files#r1391763925), but if is disregarded, this line must be fixed as it is probably a mistake (as it was originally before this PR)

weizman (commented):
lgtm minor fixed though

packages/core/src/scuttle.js
naugtur: I'm extracting, trying to keep it mostly unchanged. Would prefer to start elsewhere.

packages/core/src/scuttle.js
naugtur: I now see I partially misunderstood the original suggestion. 
@weizman Would you like to take over this PR as you seem more ready to jump right in than I am. Let's bring that to our next pairing session with any changes you see fit.

.eslintrc.js
```diff
@@ -107,6 +107,8 @@ module.exports = {
     {
       files: ['packages/*/src/**/*.js'],
       rules: {
+        // Array constructor allows us to leverage defensive coding
+        '@typescript-eslint/no-array-constructor': 'off',
```
boneskull: how?

packages/core/src/generateKernel.js
```diff
@@ -22,6 +22,7 @@ const endowmentsToolkitSrc = fs.readFileSync(
   path.join(__dirname, '/endowmentsToolkit.js'),
   'utf-8'
 )
+const scuttleSrc = fs.readFileSync(path.join(__dirname, 'scuttle.js'), 'utf-8')
```
boneskull:
```suggestion
const scuttleSrc = fs.readFileSync(require.resolve('./scuttle.js'), 'utf-8')
```

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,181 @@
+/**
+ * @typedef {object} ScuttleOpts
+ * @property {boolean} enabled - Whether scuttling is enabled or not.
+ * @property {Array<string|RegExp>} exceptions - List of properties to exclude from scuttling.
+ * @property {string} scuttlerName - Name of the scuttler function to use.
+ * @property {(globalRef: object, scuttle: Function) => void} scuttlerFunc - Scuttler function to use.
+ */
```
boneskull:
```suggestion
 * @property {(globalRef: GlobalRef, scuttle: ScuttleFn) => void} scuttlerFunc - Scuttler function to use.
 */
 
 /**
  * @callback ScuttleFn
  * @param {GlobalRef} globalRef
  * @returns {void}
  */
```

...if I'm understanding the code right, anyway.

Although, I should mention I don't understand this API.  AFAICT, the API asks the consumer for a `GlobalRef` and a `ScuttleFn`.  It then calls the `ScuttleFn` with the `GlobalRef`.  What if `ScuttleOpts` contained something like this instead?

```js
/**
 * @typedef ScuttleOpts
 * @property {boolean} enabled - Whether scuttling is enabled or not.
 * @property {Array<string|RegExp>} exceptions - List of properties to exclude from scuttling.
 * @property {string} scuttlerName - Name of the scuttler function to use.
 * @property {ScuttleFn} scuttle
 * @property {GlobalRef} globalRef
 */
```

Also:

- re: `exceptions` - are there any properties that _should not ever_ be excluded?
- re: `scuttlerName` - what's the purpose of this?
- re: `enabled` - why would we call the function if we don't intend to use it...?

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,181 @@
...
+  getOwnPropertyDescriptor,
+  create,
+  defineProperty,
+} = Object
+
+const { isArray, from } = Array
+
+const { getPrototypeOf } = Reflect
+
+const { warn } = console
```
boneskull, line 29: I can't recall, but the `console` methods don't need the `console` context, do they?

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,181 @@
...
+    }
+    opts.scuttlerFunc = globalRef[opts.scuttlerName]
+  }
+  return opts
+
+  /**
+   * @param {string|RegExp} except - Exception to convert to RegExp.
+   * @returns {string|RegExp} - Converted exception.
+   */
+  function toRE(except) {
```
boneskull, line 107: consider using [escape-string-regexp](https://www.npmjs.com/package/escape-string-regexp)?

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * Runs scuttling on the globalRef. Use applyDefaultScuttling for full scope of options.
+ *
+ * @param {GlobalRef} globalRef - Reference to the global object.
+ * @param {Array<string|RegExp>} extraPropsToAvoid - List of additional properties to exclude from scuttling beyond the default ones.
+ */
+function performScuttleGlobalThis(globalRef, extraPropsToAvoid = new Array()) {
+  const props = new Array()
+  getPrototypeChain(globalRef).forEach((proto) =>
```
boneskull, line 127: our iteration is inconsistent.  why not a `.map()`?  why use a `for...of` loop below?

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,181 @@
...
+  const obj = create(null)
+  for (const prop of props) {
+    // eslint-disable-next-line no-inner-declarations
+    function set() {
+      warn(
+        `LavaMoat - property "${prop}" of globalThis cannot be set under scuttling mode. ` +
+          'To learn more visit https://github.com/LavaMoat/LavaMoat/pull/360.'
+      )
+    }
+    // eslint-disable-next-line no-inner-declarations
```
boneskull: why?

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,181 @@
...
+      continue
+    }
+    let desc = getOwnPropertyDescriptor(globalRef, prop)
+    if (desc?.configurable === true) {
+      desc = { configurable: false, set, get }
+    } else if (desc?.writable === true) {
+      const p = new Proxy(obj, { getPrototypeOf: get, get, set })
+      desc = { configurable: false, writable: false, value: p }
+    } else {
+      continue
```
boneskull: what is this case?

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,181 @@
...
+}
+
+/**
+ * @param {Set<string|RegExp>} propsToAvoid - List of properties to exclude from scuttling.
+ * @param {string} prop - Property to check.
+ * @returns {boolean} - Whether the property should be avoided or not.
+ */
+const shouldAvoidProp = (propsToAvoid, prop) =>
+  from(propsToAvoid).some(
+    (avoid) => avoid === prop || new RegExp(avoid).test(prop)
```
boneskull: does `avoid` need escaping?

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,181 @@
...
+function getPrototypeChain(value) {
+  const protoChain = []
+  let current = value
+  while (current) {
+    if (typeof current !== 'object' && typeof current !== 'function') {
+      break
+    }
+    protoChain.push(current)
+    current = getPrototypeOf(current)
+  }
```
boneskull, line 182:
```suggestion
  while (current && (typeof current === 'object' || typeof current === 'function')) {
    protoChain.push(current)
    current = getPrototypeOf(current)
  }
```

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,126 @@
...
+const { evaluateWithSourceUrl } = require('./util')
+const { scuttle } = require('../src/scuttle')
+
+const SCUTTLER_NAME_ERROR = {
+  message:
+    'LavaMoat - \'scuttlerName\' function "SCUTTLER" expected on globalRef.' +
+    'To learn more visit https://github.com/LavaMoat/LavaMoat/pull/462.',
+}
+
+const err = (intrinsic) =>
```
boneskull, line 24: (style) guess I would go for templates here instead of string concatenation

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,126 @@
...
+  intrinsic +
+  '" of globalThis is inaccessible under ' +
+  'scuttling mode. To learn more visit https://github.com/LavaMoat/LavaMoat/pull/360.'
+
+function testScuttling(
+  globalRef,
+  intrinsics = ['Array', 'Int8Array', 'RegExp']
+) {
+  return intrinsics.some((intrinsic) => {
+    try {
```
boneskull: why do we look for an exception here? the purpose of this function isn't very clear to me

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,126 @@
...
+    'expect intrinsics to not be accessible'
+  )
+})
+
+test('scuttle - exceptions', (t) => {
+  const { vmGlobalThis } = evaluateWithSourceUrl('some-code', ';', {})
+  t.is(testScuttling(vmGlobalThis), false, 'expect intrinsics to be accessible')
+  scuttle(vmGlobalThis, {
+    enabled: true,
+    exceptions: ['/[a-zA-Z0-8]*Array/', 'RegExp'],
```
boneskull: seems like we could avoid requiring the forward-slashes?

why do we need to support strings anyhow?

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,126 @@
...
+test('scuttle - scuttle func', (t) => {
+  const { vmGlobalThis } = evaluateWithSourceUrl('some-code', ';', {})
+  t.is(testScuttling(vmGlobalThis), false, 'expect intrinsics to be accessible')
+  let globalRef
+  const cb = (realm, scuttle) => ((globalRef = realm), scuttle(realm))
+  t.throws(
+    (_) => scuttle(vmGlobalThis, { enabled: true, scuttlerName: 'SCUTTLER' }),
+    SCUTTLER_NAME_ERROR
+  )
+  Object.defineProperty(vmGlobalThis, 'SCUTTLER', { value: cb })
```
boneskull: wait, so, the scuttler, in addition to a function passed to `scuttle`, can just be a function on the `GlobalRef`?

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,126 @@
...
+    'expect global reference to be the same as provided by the scuttler function'
+  )
+  t.is(
+    testScuttling(vmGlobalThis),
+    true,
+    'expect intrinsics to not be accessible'
+  )
+})
+
+test('scuttle - resilient', (t) => {
```
boneskull, line 126: (style) I try to do a single assertion per test so that every assertion is guaranteed to be executed

boneskull (commented):
I have questions

packages/core/src/scuttle.js
weizman: Screenshot 2024-01-04 at 16 26 18

packages/core/src/scuttle.js
weizman: I don't think it provides the service I'm looking for here..
Also I think it is preferred we don't rely on third party services for production code unless we really have to.
Apparently that isn't the case entirely in LavaMoat atm, but still

packages/core/src/scuttle.js
weizman: b7f6a206a

weizman: fixed (moved to outer function to avoid pattern) b7f6a206a

packages/core/src/scuttle.js
weizman: When the property is both non configurable nor writeable (e.g. `globalThis.document`)

packages/core/src/scuttle.js
weizman: Can't see why, these props are expected to be props that are foundable on the global object, quite a finite well known group

packages/core/src/scuttle.js
weizman: Reason? Or just style?

packages/core/test/scuttle.spec.js
weizman: I want lines of code to remain short, this practice is similar in the source code too

packages/core/test/scuttle.spec.js
weizman: Scuttling means you replace the getter of a property on the global object with a getter that throws instead, so that access attempt to a forbidden property will throw an exception for users to easily tell what went wrong.
So to test that, we need to prove expected exceptions are in fact thrown.
If however the exception caught in `testScuttling` is not what we expect, I want the test to break.
`testScuttling`'s purpose is to test whether a prop is scuttled or not.

packages/core/test/scuttle.spec.js
weizman: These have to be primitive values, because in the browser version, this gets bundled

packages/core/test/scuttle.spec.js
weizman: Not exactly, opts do not respect non "stringifyable", because in the browser version, this gets bundled, so you can't really pass a function.
`scuttlerName` expects a string to represent a property expected to be found on the global object, which its value should be the `scuttlerFunc` function.
Not great, but this is a clear limitation we have to respect if we want LavaMoat to work for browser envs.
We can consider other ways to do this, but the limitation itself isn't something we can play around I think.

packages/core/test/scuttle.spec.js
weizman: I think that's a fair point. I'll try to follow that guideline. 
I do however think it's slightly more difficult here since I test scenarios and these scenarios naturally have multiple steps I wish to assert..
Do you see a way for me to adopt your suggestion without losing that?

packages/core/src/scuttle.js
weizman: fair enough a3d5942e7

packages/core/src/scuttle.js
weizman: This seems to be a mistake, `ScuttleOpts` does not accept `ScuttleFn`, it accepts either a `bool` or `{enabled, exceptions, scuttlerName}` (not sure where `@property ScuttleFn` came from).
I now better understand the confusion [here](https://github.com/LavaMoat/LavaMoat/pull/758#discussion_r1441915767).

weizman: fixed 4864384b1

weizman: * re: `exceptions` - not that I can think of
* re: `scuttlerName` - see [here](https://github.com/LavaMoat/LavaMoat/pull/758#discussion_r1441915767)
* re: `enabled` - when you put it like that, you're right, evolution-wise it (kinda) [made sense](https://github.com/LavaMoat/LavaMoat/pull/462)(?). It was about leaving this experimental and opt-in explicit.

.eslintrc.js
weizman: I take it back. It doesn't. fixed: 4d8e36de4

weizman:
Thanks for the review @boneskull , mind having another look?

packages/core/test/scuttle.spec.js
boneskull: Can you give an example?  What's a "step"?

packages/core/src/scuttle.js
boneskull: If you're converting an arbitrary string to a regexp, you will need to escape it properly.  If there's a discrete list of strings we expect, we could keep a `Map<string, RegExp>` or something. if we don't know what those strings are, then we have more work to do.

I didn't realize how simple that package was.  but it seems like we could adapt the regex for our special case here (it looks like there's some sort of convention in use??)

```js
	return string
		.replace(/[|\\{}()[\]^$+*?.]/g, '\\$&')
		.replace(/-/g, '\\x2d');
```

packages/core/src/scuttle.js
boneskull: if there is any possibility that there are any special characters in `avoid`, we should escape them and/or use a mapping, similar to my above comment

packages/core/src/scuttle.js
boneskull: style, I guess.  I guess I like to avoid `break` statements if not absolutely needed

packages/core/test/scuttle.spec.js
boneskull: something to be said for consistency, I suppose!

another style issue would be use a template on L7 above such that we do not need to escape anything

packages/core/test/scuttle.spec.js
boneskull: > Scuttling means you replace the getter of a property on the global object with a getter that throws instead,

I am curious about why this is, when we use no-ops elsewhere, right?

boneskull: > If however the exception caught in testScuttling is not what we expect, I want the test to break.

Why don't we use `ava`'s own facilities to do this?  `t.throws`

packages/core/test/scuttle.spec.js
boneskull: I don't understand why bundling would mean we cannot use a `RegExp`.

Why is the character class `[a-zA-Z0-8]` and not `[a-zA-Z0-9]`?

(I'm asking all these questions because I don't know anything about any of this)

packages/core/test/scuttle.spec.js
boneskull: IMO these guardrails need to be more clearly-defined via comments and/or types and/or exceptions. Whether that should be part of this PR is up to you

By "this gets bundled" you mean how the kernel uses templates to import source code as strings, or something else?  Trying to follow

packages/core/test/scuttle.spec.js
naugtur: `0-8` looks like a typo

packages/core/test/scuttle.spec.js
naugtur: > By "this gets bundled" you mean how the kernel uses templates to import source code as strings, or something else? Trying to follow

yes

packages/core/test/scuttle.spec.js
naugtur: I think 1 assertion per test is a nice idea to strive for, but I've seen it become counterproductive if applied despite difficulties. So let's try, but if there's a price to pay, consider if it's worth it.

packages/core/test/scuttle.spec.js
naugtur: length of lines should be a concern for machines, not humans ;) 
If linter permits it, use a template string.
On the other hand, if linter permits it, use whatever string concatenation you want.

packages/core/test/scuttle.spec.js
boneskull: yes, certainly if the test has an expensive setup, there's little upside in repeatedly spending it

packages/core/src/scuttle.js
weizman: Can you give me an example of why might this be needed?
While we don't know the strings we might receive, we know they're not going to include any special chars other than `abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_` because they represent names of props we expect to find on the global object.
Is it possible potentially to find a "🐴" in a property name of the global object? Yes.
Will it ever happen though?

weizman: Either way, fixed aba162b32

weizman: Also, [it](https://github.com/sindresorhus/escape-string-regexp/blob/main/index.js) doesn't seem to do what I would expect it to:

Screenshot 2024-01-29 at 14 37 01

In addition to escaping special chars, it escapes chars I wouldn't expect it to as well, turning `[a-z]` into `'\\[a\\x2dz\\]'` which breaks the RE

weizman: Diving into this (or finding some other thing that cuts it for us) is possible, but I'm not sure I see the benefit

packages/core/src/scuttle.js
weizman: Actually, funny enough, your [original comment](https://github.com/LavaMoat/LavaMoat/pull/758#discussion_r1391763925) starting a chain reaction where I introduced a second redundant string-to-regex operation on top of an earlier former one (usage of `toRE`). 

Therefore, this `new RegExp` call here is reverted back to how this function behaved originally (fixed a21f75349)

packages/core/src/scuttle.js
weizman: I personally find it less convenient to both read and debug `while` loops with complicated conditions, I prefer test those within the loop itself (personally)

packages/core/test/scuttle.spec.js
weizman: You just bombed attackers search for LavaMoat-scuttling protected programs with `window['999']` exposed API sir (fixed 8a71f8464)

weizman: For the browser version of LavaMoat, where the program gets bundled along with our protection (e.g. lavamoat-browserify), this means configurable arguments that need to be introduced into the output must be primitive values, otherwise how are you going to ship them into the bundled output?

For example, if it was:

```javascript
{
    enabled: true,
    exceptions: [new RegExp('/[a-zA-Z0-8]*Array/'), new RegExp('RegExp')],
}
```

You'd end up with an output similar to this:

```javascript
// lavamoat
const args = {"enabled":true,"exceptions":[{},{}]};
function scuttle() {
...

// program
alert('this program is protected by lavamoat');
```

if that makes sense?

weizman: UNLESS - this is already clear to you and what you meant is that we can accept from the user RegExp instances, convert them ourselves to string before placing them into the bundle, and then have the bundle revert them back to RegExp instances at runtime?

I'm open to that if you think that's important

packages/core/test/scuttle.spec.js
weizman: like this 0291d093c?

packages/core/test/scuttle.spec.js
weizman: > I am curious about why this is, when we use no-ops elsewhere, right?
https://github.com/LavaMoat/LavaMoat/pull/360#discussion_r946406834

weizman: > Why don't we use ava's own facilities to do this? t.throws

fixed 5e094809ebbc75a92fbed5201555006545b1437c

packages/core/test/scuttle.spec.js
weizman: fixed (addressed by https://github.com/LavaMoat/LavaMoat/pull/758#discussion_r1469769937)

weizman:
Thanks for the 2nd review @boneskull , mind having another look?

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,186 @@
...
+ * @typedef {object} ScuttleOpts
+ * @property {boolean} enabled - Whether scuttling is enabled or not.
+ * @property {Array<string|RegExp>} exceptions - List of properties to exclude from scuttling.
+ * @property {string} scuttlerName - Name of the scuttler function to use which is expected to be found as a
+ * property on the global object (e.g. if scuttlerName is 'x', scuttler function is obtained from globalThis['x']).
+ */
+
+/**
+ * @typedef {object} GlobalRef
+ * @property {object} [globalThis] - Reference to the global object.
```
boneskull:
```suggestion
 * @property {Record<string, any>} [globalThis] - Reference to the global object.
```

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,186 @@
...
+ * @param {GlobalRef} globalRef - Reference to the global object.
+ * @param {ScuttleOpts} opts - Scuttling options.
+ */
+function scuttle(globalRef, opts) {
+  const scuttleOpts = generateScuttleOpts(globalRef, opts)
+
+  if (scuttleOpts.enabled) {
+    if (!isArray(scuttleOpts.exceptions)) {
+      throw new Error(
+        `LavaMoat - scuttleGlobalThis.exceptions must be an array, got "${typeof scuttleOpts.exceptions}"`
```
boneskull:
```suggestion
        `LavaMoat - scuttleGlobalThis.exceptions must be an array, got ${typeof scuttleOpts.exceptions}`
```

Also it's unclear what `scuttleGlobalThis` is referring to, exactly.  Unless we're eating the stack trace, I don't see why we'd need to mention _which_ function was called with invalid args.

packages/core/src/scuttle.js
boneskull: I didn't test the thing I pasted. 🙂	

A property name can be literally any string in JS, afaik.  Some of those strings contain characters that have special meaning to regular expressions.  That's my concern, but as long as it's not exposing some sort of ReDoS-style flaw, then I'm not sure we'd need to worry too much about it (until we do).

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,145 @@
...
+    !configurables.includes(p)
+      ? vmGlobalThis[p]
+      : t.throws(() => vmGlobalThis[p], { message: err(p) })
+  )
+})
+
+test('scuttle - resilient', (t) => {
+  const { vmGlobalThis } = evaluateWithSourceUrl('some-code', ';', {})
+  const { all, configurables } = getPropsGroups(vmGlobalThis)
+  all.map((p) => vmGlobalThis[p])
```
boneskull:
```suggestion
  all.forEach((p) => vmGlobalThis[p])
```

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,145 @@
...
+  scuttle(vmGlobalThis, { enabled: true })
+  all.map((p) =>
```
boneskull:
```suggestion
  all.forEach((p) =>
```

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,145 @@
...
+    !configurables.includes(p)
+      ? vmGlobalThis[p]
+      : t.throws(() => vmGlobalThis[p], { message: err(p) })
+  )
+  t.throws(() => Object.defineProperty(vmGlobalThis, 'Array', { value: 111 }), {
+    message: 'Cannot redefine property: Array',
+  })
+  vmGlobalThis.Array = 1
+  all.map((p) =>
```
boneskull:
```suggestion
  all.forEach((p) =>
```

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,145 @@
...
+  const opts = { enabled: true, scuttlerName: 'SCUTTLER' }
+  t.throws(() => scuttle(vmGlobalThis, opts), SCUTTLER_NAME_ERROR)
+  Object.defineProperty(vmGlobalThis, opts.scuttlerName, { value: cb })
+  scuttle(vmGlobalThis, opts)
+  t.is(
+    globalRef,
+    vmGlobalThis,
+    'expect global reference to be the same as provided by the scuttler function'
+  )
+  all.map((p) =>
```
boneskull:
```suggestion
  all.forEach((p) =>
```

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,145 @@
...
+      (p) =>
+        configurables.includes(p) &&
+        t.throws(() => vmGlobalThis[p], { message: err(p) })
+    )
+})
+
+test('scuttle - scuttle func', (t) => {
+  const { vmGlobalThis } = evaluateWithSourceUrl('some-code', ';', {})
+  const { all, configurables } = getPropsGroups(vmGlobalThis)
+  all.map((p) => vmGlobalThis[p])
```
boneskull:
```suggestion
  all.forEach((p) => vmGlobalThis[p])
```

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,145 @@
...
+    'Uint32Array',
+    'Int32Array',
+    'Float32Array',
+    'Float64Array',
+    'Uint8ClampedArray',
+    'BigUint64Array',
+    'BigInt64Array',
+    'SharedArrayBuffer',
+  ]
+  exceptions.map((p) => vmGlobalThis[p])
```
boneskull:
```suggestion
  exceptions.forEach((p) => vmGlobalThis[p])
```

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,145 @@
...
+  all
+    .filter((p) => !exceptions.includes(p))
+    .map(
```
boneskull:
```suggestion
    .forEach(
```

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,145 @@
...
+    !configurables.includes(p)
+      ? vmGlobalThis[p]
+      : t.throws(() => vmGlobalThis[p], { message: err(p) })
+  )
+})
+
+test('scuttle - exceptions', (t) => {
+  const { vmGlobalThis } = evaluateWithSourceUrl('some-code', ';', {})
+  const { all, configurables } = getPropsGroups(vmGlobalThis)
+  all.map((p) => vmGlobalThis[p])
```
boneskull:
```suggestion
  all.forEach((p) => vmGlobalThis[p])
```

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,145 @@
...
+test('scuttle - opts as object', (t) => {
+  const { vmGlobalThis } = evaluateWithSourceUrl('some-code', ';', {})
+  const { all, configurables } = getPropsGroups(vmGlobalThis)
+  all.map((p) => vmGlobalThis[p])
+  scuttle(vmGlobalThis, {})
+  all.map((p) => vmGlobalThis[p])
+  scuttle(vmGlobalThis, { enabled: false })
+  all.map((p) => vmGlobalThis[p])
+  scuttle(vmGlobalThis, { enabled: true })
+  all.map((p) =>
```
boneskull:
```suggestion
  all.forEach((p) =>
```

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,145 @@
...
+})
+
+test('scuttle - opts as object', (t) => {
+  const { vmGlobalThis } = evaluateWithSourceUrl('some-code', ';', {})
+  const { all, configurables } = getPropsGroups(vmGlobalThis)
+  all.map((p) => vmGlobalThis[p])
+  scuttle(vmGlobalThis, {})
+  all.map((p) => vmGlobalThis[p])
+  scuttle(vmGlobalThis, { enabled: false })
+  all.map((p) => vmGlobalThis[p])
```
boneskull:
```suggestion
  all.forEach((p) => vmGlobalThis[p])
```

boneskull (approved):
nothing blocking.  I started adding suggestions for replacing `map` with `forEach` but stopped once I realized how many there were.  my complaint is that it's semantically incorrect usage of `map` and obscures intent. that's it!

packages/core/src/scuttle.js
weizman: fixed 1db9dd352

packages/core/src/scuttle.js
legobeat: functions and objects are always truthy, no? Could be simplified further:

```suggestion
  while (typeof current === 'object' || typeof current === 'function') {
    protoChain.push(current)
    current = getPrototypeOf(current)
  }
```

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,145 @@
...
+  )
+  t.throws(() => Object.defineProperty(vmGlobalThis, 'Array', { value: 111 }), {
+    message: 'Cannot redefine property: Array',
+  })
+  vmGlobalThis.Array = 1
+  all.map((p) =>
+    !configurables.includes(p)
+      ? vmGlobalThis[p]
+      : t.throws(() => vmGlobalThis[p], { message: err(p) })
+  )
```
legobeat:
```suggestion
  all.forEach((p) =>
    configurables.includes(p)
      ? t.throws(() => vmGlobalThis[p], { message: err(p) })
      : vmGlobalThis[p]
  )
```

Unless there's some subtle reason why this ternary condition should be inverted, in which case an inline comment is warranted.

packages/core/test/scuttle.spec.js
```diff
@@ -0,0 +1,145 @@
...
+test('scuttle - resilient', (t) => {
+  const { vmGlobalThis } = evaluateWithSourceUrl('some-code', ';', {})
+  const { all, configurables } = getPropsGroups(vmGlobalThis)
+  all.map((p) => vmGlobalThis[p])
+  scuttle(vmGlobalThis, { enabled: true })
+  all.map((p) =>
+    !configurables.includes(p)
+      ? vmGlobalThis[p]
+      : t.throws(() => vmGlobalThis[p], { message: err(p) })
+  )
```
legobeat:
```suggestion
  all.forEach((p) =>
    configurables.includes(p)
      ? t.throws(() => vmGlobalThis[p], { message: err(p) })
      : vmGlobalThis[p]
    )
```
Unless there's some subtle reason why this ternary condition should be inverted, in which case an inline comment is warranted.

packages/core/src/scuttle.js
legobeat: I might also question if it's sustainable with an API where whether the input is interpreted as a regex or not is determined by the presence of a leading `/`. I'd prefer either of:
1. Toggling this behavior for all exceptions via a new boolean flag
2. Always treat them all as string (do we actually have or foresee real-world use-cases where regexes are required?)
3. Always treat them all as regex (prob anchoring them and require user to explicitly `.*`, by prepending `^` and appending `$`)
4. Use separate fields for regex-exceptions and string-exceptions

But I think semantics like that shouldn't be changed as part of this refactoring PR - can be done in a follow-up if desired.

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,186 @@
...
+    }
+    scuttleOpts.scuttlerFunc(globalRef, (realm) =>
+      performScuttleGlobalThis(realm, scuttleOpts.exceptions)
+    )
+  }
+}
+
+/**
+ * @param {GlobalRef} globalRef - Reference to the global object.
+ * @param {ScuttleOpts|boolean} originalOpts - Scuttling options. Accepts `true` for backwards compatibility.
```
legobeat, line 71: Can we add a note on deprecation of the boolean form? Would be nice to shed the backwards-compat sooner than later - would make this function less complex and more readable.

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,186 @@
...
+ */
+function generateScuttleOpts(globalRef, originalOpts = create(null)) {
+  const defaultOpts = {
+    enabled: true,
+    exceptions: [],
+    scuttlerName: '',
+  }
+  const opts = assign(
+    {},
+    originalOpts === true ? { ...defaultOpts } : { ...originalOpts },
```
legobeat: For my education: What is the significance of the spread here? This would seem equivalent to me but I guess there may be cases where it's not?
```suggestion
    originalOpts === true ? defaultOpts : originalOpts,
```
, as `assign` already limits its behavior to enumerable own props and only the value in the first parameter is modified.

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,186 @@
...
+    { scuttlerFunc: (globalRef, scuttle) => scuttle(globalRef) },
+    {
+      exceptions: (originalOpts.exceptions || defaultOpts.exceptions).map((e) =>
+        toRE(e)
+      ),
+    }
```
legobeat:
```suggestion
    {
      scuttlerFunc: (globalRef, scuttle) => scuttle(globalRef) },
      exceptions: (originalOpts?.exceptions || defaultOpts.exceptions).map((e) =>
        toRE(e)
      ),
    }
```

packages/core/src/scuttle.js
```diff
@@ -0,0 +1,186 @@
...
+      performScuttleGlobalThis(realm, scuttleOpts.exceptions)
+    )
+  }
+}
+
+/**
+ * @param {GlobalRef} globalRef - Reference to the global object.
+ * @param {ScuttleOpts|boolean} originalOpts - Scuttling options. Accepts `true` for backwards compatibility.
+ * @returns {ScuttleOpts} - Final scuttling options.
+ */
```
boneskull, line 73:
```suggestion
/**
 * @overload
 * @param {GlobalRef} globalRef - Reference to the global object.
 * @param {ScuttleOpts} originalOpts - Scuttling options
 * @returns {ScuttleOpts} - Final scuttling options.
 */
 
/**
 * @overload
 * @deprecated
 * @param {GlobalRef} globalRef - Reference to the global object.
 * @param {boolean} originalOpts - Deprecated; use options instead
 * @returns {ScuttleOpts} - Final scuttling options.
 */
 
/**
 * @param {GlobalRef} globalRef - Reference to the global object.
 * @param {ScuttleOpts|boolean} originalOpts - Scuttling options
 * @returns {ScuttleOpts} - Final scuttling options.
 */
```

To @legobeat's comment, try this. I'm not sure if it'll mark the entire function deprecated, or just that one usage of it.  But if it's the latter then 👍 .  add a `// @ts-check` at the top of the file to see what it thinks (you can remove later if you want)

you can test if it works via something like:

```js
generateScuttleOpts({}, {});

generateScuttleOpts({}, true);
```

If the second one only is ~~struck through~~ then it works.

packages/core/src/scuttle.js
boneskull: yeah, seems like you could just skip the spreads since the `defaultOpts` are not created outside of the function scope, so you don't need to worry about mutating it

packages/core/src/scuttle.js
boneskull: @weizman do you have prettier setup to format as a pre-commit hook, or on save?  run `npm run setup` to configure the hooks

packages/core/src/scuttle.js
boneskull: @legobeat that's what I wrote lol

packages/core/test/scuttle.spec.js
boneskull: 👍  to avoiding unnecessary negation in ternaries or if/else conditions

packages/core/test/scuttle.spec.js
weizman: a22476635

packages/core/test/scuttle.spec.js
weizman: a22476635

packages/core/test/scuttle.spec.js
weizman: a22476635 (applies to the rest as well)

packages/core/src/scuttle.js
weizman: 086ed7ac0

packages/core/src/scuttle.js
weizman: Yea I do, it always fixes stuff but didn't fix this one.
anyway 6e1b49290790ccd8fa9c221340ebc1fbec8b218e

packages/core/src/scuttle.js
weizman: 🥲 
Screenshot 2024-01-31 at 12 28 31

packages/core/src/scuttle.js
weizman: moved to https://github.com/LavaMoat/LavaMoat/pull/758#discussion_r1471797705

packages/core/src/scuttle.js
weizman: I wasn't able to learn online how to achieve this.. thoughts? @boneskull

weizman: Decided with @boneskull that this might not be easy to achieve, we can circle back to this in a different PR

PR #755, opened 2023-11-09
chore(deps): upgrade type-fest across entire repo

boneskull:
`lavamoat-core`'s `type-fest` is out-of-sync with the rest; now they are all the same.

PR #761, opened 2023-11-15
feat: ship types for tofu

boneskull:
This adds and ships types for `lavamoat-tofu`.

There's also a fix to types in `lavamoat-core` that I had noticed while I was working on this.

### DIFF
--- a/packages/core/src/schema/lavamoat-policy.v0-0-1.schema.ts
+++ b/packages/core/src/schema/lavamoat-policy.v0-0-1.schema.ts
@@ -62,6 +62,14 @@ export interface NodeLocation {
 }
 
 /**
+ * @deprecated - Use `true` instead
+ */
+export type GlobalPolicyRead = 'read'
+export type GlobalPolicyWrite = 'write'
+
+export type GlobalPolicyValue = GlobalPolicyRead | GlobalPolicyWrite | boolean
+
+/**
  * Describe the resources available to your application and direct dependencies
  */
 export interface Resources {
@@ -83,7 +91,7 @@ export interface ResourcePolicy {
  * `true` to allow and `false` to deny
  */
 export interface GlobalPolicy {
-  [k: string]: boolean
+  [k: string]: GlobalPolicyValue
 }
 
 /**
--- a/packages/tofu/package.json
+++ b/packages/tofu/package.json
@@ -26,15 +26,20 @@
   },
   "dependencies": {
     "@babel/parser": "7.23.6",
-    "@babel/traverse": "7.23.7"
+    "@babel/traverse": "7.23.7",
+    "@babel/types": "7.23.6",
+    "@types/babel__traverse": "7.20.5",
+    "type-fest": "4.7.1"
   },
   "devDependencies": {
-    "deep-equal": "2.2.3"
+    "deep-equal": "2.2.3",
+    "lavamoat-core": "^15.0.0"
   },
   "ava": {
     "files": [
       "test/*.spec.js"
     ],
     "timeout": "30s"
-  }
+  },
+  "types": "./types/index.d.ts"
 }
--- a/packages/tofu/src/codeSampleFromAstNode.js
+++ b/packages/tofu/src/codeSampleFromAstNode.js
@@ -1,5 +1,11 @@
 module.exports = { codeSampleFromAstNode }
 
+/**
+ *
+ * @param {import('@babel/types').Node & {loc: import('@babel/types').SourceLocation}} node
+ * @param {import('lavamoat-core').LavamoatModuleRecord} moduleRecord
+ * @returns
+ */
 function codeSampleFromAstNode(node, moduleRecord) {
   const result = {}
   const { content } = moduleRecord
@@ -9,7 +15,7 @@ function codeSampleFromAstNode(node, moduleRecord) {
     result.range += `,${end.line}:${end.column}`
   }
   // prepare sample
-  const lines = content.split('\n')
+  const lines = /** @type {string} */ (content).split('\n')
   const startLine = lines[start.line - 1]
   const sample = startLine.slice(start.column, start.column + 70)
   result.sample = sample
--- a/packages/tofu/src/findGlobals.js
+++ b/packages/tofu/src/findGlobals.js
@@ -17,7 +17,17 @@ const nonReferenceIdentifiers = [
 
 module.exports = { findGlobals }
 
+/**
+ * @typedef {import('@babel/traverse').NodePath<import('@babel/types').Identifier|import('@babel/types').ThisExpression>} IdentifierOrThisExpressionNodePath
+ */
+
+/**
+ *
+ * @param {import('@babel/types').Node} ast
+ * @returns {Map<string, IdentifierOrThisExpressionNodePath[]>}
+ */
 function findGlobals(ast) {
+  /** @type {ReturnType<typeof findGlobals>} */
   const globals = new Map()
   traverse(ast, {
     // ReferencedIdentifier
@@ -76,13 +86,21 @@ function findGlobals(ast) {
 
   return globals
 
+  /**
+   *
+   * @param {IdentifierOrThisExpressionNodePath} path
+   * @param {string} [name]
+   */
+  // @ts-ignore - FIXME needs logic changes for type safety
   function saveGlobal(path, name = path.node.name) {
     // init entry if needed
     if (!globals.has(name)) {
       globals.set(name, [])
     }
     // append ref
-    const refsForName = globals.get(name)
+    const refsForName = /** @type {IdentifierOrThisExpressionNodePath[]} */ (
+      globals.get(name)
+    )
     refsForName.push(path)
   }
 }
--- a/packages/tofu/src/inspectPrimordialAssignments.js
+++ b/packages/tofu/src/inspectPrimordialAssignments.js
@@ -6,10 +6,32 @@ const { isMemberLikeExpression } = require('./util.js')
 
 module.exports = { inspectPrimordialAssignments }
 
+/**
+ * @typedef PrimordialAssignment
+ * @property {import('@babel/traverse').Node} node
+ * @property {import('@babel/traverse').NodePath} path
+ * @property {string[]} memberPath
+ */
+
+/**
+ * @typedef {import('@babel/types').MemberExpression | import('@babel/types').OptionalMemberExpression} MemberLikeExpression
+ */
+
+/**
+ * @typedef {MemberLikeExpression & Omit<MemberLikeExpression, 'computed'> & {computed: false}} NonComputedMemberLikeExpression
+ */
+
+/**
+ *
+ * @param {import('@babel/types').Node} ast
+ * @param {readonly string[]} namedIntrinsics
+ * @returns {PrimordialAssignment[]}
+ */
 function inspectPrimordialAssignments(
   ast,
   namedIntrinsics = defaultNamedIntrinsics
 ) {
+  /** @type {PrimordialAssignment[]} */
   const results = []
   traverse(ast, {
     AssignmentExpression: function (path) {
@@ -47,15 +69,24 @@ function inspectPrimordialAssignments(
   return results
 }
 
+/**
+ *
+ * @param {MemberLikeExpression} node
+ * @returns {string[]}
+ */
 function memberExpressionChainToPath(node) {
+  /** @type {string[]} */
   const path = []
   // walk down property chain
   while (isMemberLikeExpression(node.object)) {
+    // @ts-ignore - FIXME needs logic changes for type safety
     path.push(node.property.name)
     node = node.object
   }
   // reached the end
+  // @ts-ignore - FIXME needs logic changes for type safety
   path.push(node.property.name)
+  // @ts-ignore - FIXME needs logic changes for type safety
   path.push(node.object.name)
   // fix order
   path.reverse()
--- a/packages/tofu/src/inspectSesCompat.js
+++ b/packages/tofu/src/inspectSesCompat.js
@@ -9,15 +9,44 @@ const { inspectDynamicRequires } = require('./inspectSource.js')
 
 module.exports = { inspectSesCompat }
 
-const strictModeViolationErrorCues = [
+const strictModeViolationErrorCues = /** @type {const} */ ([
   'Unexpected reserved word',
   'Legacy octal literals are not allowed in strict mode',
   'Expecting Unicode escape sequence',
   "'with' in strict mode",
   'Deleting local variable in strict mode',
-]
+])
 
+/**
+ * Babel doesn't type `ParseError` very well.
+ * @typedef {import('@babel/parser').ParseError & Error & {loc: {line: number, column: number, index: number}, pos: number}} ParseError
+ */
+
+/**
+ * @typedef {Omit<import('@babel/parser').ParseResult<import('@babel/types').File>, 'errors'> & {errors?: Array<import('@babel/parser').ParseError & Error & {loc: {line: number, column: number, index: number}, pos: number}>}} ParseResult
+ */
+
+/**
+ * @typedef StrictModeViolation
+ * @property {ParseError} error
+ * @property {ParseError['loc']} loc
+ * @property {ParseError['loc']['index']} pos
+ */
+
+/**
+ * @typedef InspectSesCompatResult
+ * @property {import('./inspectPrimordialAssignments.js').PrimordialAssignment[]} primordialMutations
+ * @property {StrictModeViolation[]} strictModeViolations
+ * @property {import('./inspectSource.js').RequireCallResult[]} dynamicRequires
+ */
+
+/**
+ *
+ * @param {ParseResult} ast
+ * @returns {InspectSesCompatResult}
+ */
 function inspectSesCompat(ast) {
+  /** @type {InspectSesCompatResult} */
   const results = {
     primordialMutations: [],
     strictModeViolations: [],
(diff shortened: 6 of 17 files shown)


boneskull:
* **#761** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/761?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#752** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/752?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#742](https://github.com/LavaMoat/LavaMoat/pull/742) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/742?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#770](https://github.com/LavaMoat/LavaMoat/pull/770) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/770?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#732** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/732?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#733](https://github.com/LavaMoat/LavaMoat/pull/733) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/733?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#755](https://github.com/LavaMoat/LavaMoat/pull/755) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/755?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

packages/tofu/src/findGlobals.js
```diff
@@ -76,13 +82,26 @@ function findGlobals(ast) {
...
-  function saveGlobal(path, name = path.node.name) {
-    // init entry if needed
-    if (!globals.has(name)) {
-      globals.set(name, [])
+  /**
+   *
+   * @param {import('@babel/traverse').NodePath<import('@babel/types').Identifier|import('@babel/types').ThisExpression>} path
+   * @param {string} [name]
+   */
+  function saveGlobal(path, name) {
```
legobeat: I do think these logic changes that go beyond simple accommodating refactors ought to be split out from shipping types. And/or could `@ts-ignore` to sidestep for now?

packages/core/src/schema/lavamoat-policy.v0-0-1.schema.ts
```diff
@@ -61,6 +61,11 @@ export interface NodeLocation {
   line: number
 }
 
+export type GlobalPolicyRead = 'read'
+export type GlobalPolicyWrite = 'write'
+
+export type GlobalPolicyValue = GlobalPolicyRead | GlobalPolicyWrite | boolean
```
naugtur, line 70: We should use this opportunity to get rid of the `'read'` value. It's never used, `true` is used everywhere

naugtur: @kumavis I recall you saying you don't want it anymore ;)

packages/tofu/package.json
```diff
@@ -5,10 +5,14 @@
   "license": "MIT",
   "dependencies": {
     "@babel/parser": "^7.23.0",
-    "@babel/traverse": "^7.23.2"
+    "@babel/traverse": "^7.23.2",
+    "@babel/types": "^7.23.3",
+    "@types/babel__traverse": "^7.20.4",
```
naugtur: so we need both? won't `@babel/types` define that already?

packages/tofu/src/codeSampleFromAstNode.js
```diff
@@ -1,5 +1,11 @@
 module.exports = { codeSampleFromAstNode }
 
+/**
+ *
+ * @param {import('type-fest').Merge<import('@babel/types').Node, {loc: import('@babel/types').SourceLocation}>} node
```
naugtur: Can we hide it somewhere? Put it in a types file and reference here by name or something? It's scary.

packages/tofu/src/findGlobals.js
```diff
@@ -76,13 +82,26 @@ function findGlobals(ast) {
...
+      name = path.node.name
+    }
+    if (name) {
+      // init entry if needed
+      if (!globals.has(name)) {
+        globals.set(name, [])
+      }
+      // append ref
+      const refsForName =
+        /** @type {import('@babel/traverse').NodePath<import('@babel/types').Identifier|import('@babel/types').ThisExpression>[]} */ (
```
naugtur: This type is being reused here a bunch, I'd be happy if we could name it something. The annoyance caused by repeating it is growing beyond the potential annoyance of it being exported.

packages/tofu/src/findGlobals.js
naugtur: The only behavioral change I see here is if `path.node` doesn't have `name` but I think it'll end up returning undefined anyway. Can you point to a case where something changes?

packages/tofu/src/inspectPrimordialAssignments.js
```diff
@@ -47,16 +69,30 @@ function inspectPrimordialAssignments(
...
   // walk down property chain
   while (isMemberLikeExpression(node.object)) {
-    path.push(node.property.name)
-    node = node.object
+    if ('name' in node.property) {
+      path.push(node.property.name)
+      node = node.object
+    } else {
+      break
+    }
```
naugtur: This may be a logic change. `node.object` is assigned to `node` 
Are we sure it's not possible for a name being absent from property (or property not being there) while the object field exists?

packages/tofu/src/findGlobals.js
legobeat: My comment was ambiguously phrased I guess, it was referring to the internal logic. In either case it seems like it's a non-trivial change enough to belong in a separate change.

packages/tofu/src/inspectSesCompat.js
```diff
@@ -37,9 +66,15 @@ function inspectSesCompat(ast) {
...
-    (k) => k in global && typeof sesAllowlist[k] === 'object'
-  )
+  const sesNamedIntrinsics = Reflect.ownKeys(sesAllowlist)
+    .filter(
+      (k) =>
+        k in global &&
+        // @ts-expect-error - filter loses specificity
+        typeof sesAllowlist[k] === 'object'
+    )
+    .map(String)
```
naugtur, line 76: Can we force-cast instead of looping again?

packages/tofu/src/inspectSource.js
```diff
@@ -173,7 +173,17 @@ function inspectEsmImports(ast) {
   return { esmImports }
 }
 
+/**
+ * @typedef {import('@babel/traverse').NodePath<import('@babel/types').CallExpression>} RequireCallResult
```
naugtur, line 227: I'd like to do this with `import('@babel/traverse').NodePath<import('@babel/types').Identifier|import('@babel/types').ThisExpression>` above

packages/tofu/src/inspectSource.js
```diff
@@ -309,46 +341,77 @@ function inspectPatternElementForDeclarations(node, keyPath = []) {
...
   }
-  // expand each property into a path, recursively
+  /** @type {Declaration[]} */
   let results = []
+  // expand each property into a path, recursively
   node.elements.forEach((child, propName) => {
-    results = results.concat(
-      inspectPatternElementForDeclarations(child, [...keyPath, propName])
-    )
+    if (child) {
```
naugtur, line 449: how could an element be falsy? was TS complaining?

packages/tofu/src/referenceTracker.js
```diff
@@ -162,55 +270,98 @@ const usageDetectors = {
...
   MemberExpression: (targetPath) => {
     const leftSide = targetPath.node.object
     if (leftSide.type === 'Identifier') {
       const binding = targetPath.scope.getBinding(leftSide.name)
       if (binding) {
-        return new ReferenceLinkage({
-          path: binding.path,
-          label: 'parent of member',
-        })
+        return [
```
naugtur: why is this an array now?

packages/tofu/src/ses-whitelist.js
```diff
@@ -1,3 +1,5 @@
+// @ts-nocheck
+
```
naugtur, line 2: (could) document why the bailout

naugtur (approved):
All comments optional, LGTM
[edit] changed my mind on one.

packages/core/src/schema/lavamoat-policy.v0-0-1.schema.ts
boneskull: Sounds like a breaking change...? I could flag it as `@deprecated`.

packages/tofu/package.json
boneskull: `@types/babel__traverse` depends upon `@babel/types`, but they aren't the same.

packages/tofu/src/codeSampleFromAstNode.js
boneskull: I can create a `@typedef` for it, I suppose?

packages/tofu/src/inspectPrimordialAssignments.js
boneskull: I'll take another look at this and explain myself

packages/tofu/src/inspectSesCompat.js
boneskull: I could make it a `reduce`, but `Reflect.ownKeys()` returns _all_ non-private keys in an object, so for our purposes, we will want to cast to `string`.

packages/tofu/src/inspectSource.js
boneskull: I only changed stuff if it was.  `ArrayPattern.elements` is of type `Array<null | PatternLike | LVal>`, so, yes.

packages/tofu/src/referenceTracker.js
boneskull: because a `Detector`--defined as per the usage--expects an array.  the result is returned by [`tracePathTousages()`](https://github.com/LavaMoat/LavaMoat/pull/761/files/b09ea4b4854b71b820c3ed201d174ccf105d31fa#diff-45f2498bbc6fc41108f02cdfc08366bdec96fb18e677cb875b6acdfede5cdc0fR343) and [`walkUsage()`](https://github.com/LavaMoat/LavaMoat/pull/761/files/b09ea4b4854b71b820c3ed201d174ccf105d31fa#diff-45f2498bbc6fc41108f02cdfc08366bdec96fb18e677cb875b6acdfede5cdc0fR108) iterates over it.

Looks like a bug to me, but I could be missing something.

packages/tofu/src/ses-whitelist.js
boneskull: I'd be even happier if we, say, moved this into a `vendor/` dir and kept it up-to-date with `ses`.

packages/tofu/src/util.js
```diff
@@ -58,34 +85,62 @@ function getNameFromNode(node) {
...
+ * @param {import("./inspectPrimordialAssignments").MemberLikeExpression} node
+ * @returns {node is import("./inspectPrimordialAssignments").NonComputedMemberLikeExpression}
+ */
 function isNonComputedMemberLikeExpression(node) {
   return !node.computed && isMemberLikeExpression(node)
 }
 
+/**
+ *
+ * @param {import("@babel/types").LVal} identifierNode
```
boneskull, line 107: Note: I guess I'd expect `prettier` to make the quotes consistent in these types, but maybe it doesn't do that out-of-the-box.  [prettier-plugin-jsdoc](https://npm.im/prettier-plugin-jsdoc) seems like it could be helpful.

packages/tofu/src/findGlobals.js
boneskull: @legobeat , @naugtur tbh, this seems pretty trivial to me, especially if we consider what this code is actually doing. though...

image

packages/tofu/src/inspectSesCompat.js
```diff
@@ -37,9 +66,15 @@ function inspectSesCompat(ast) {
...
   })
   // check for mutations to named intrinsics
-  const sesNamedIntrinsics = Reflect.ownKeys(sesAllowlist).filter(
-    (k) => k in global && typeof sesAllowlist[k] === 'object'
-  )
+  const sesNamedIntrinsics = Reflect.ownKeys(sesAllowlist)
+    .filter(
+      (k) =>
+        k in global &&
+        // @ts-expect-error - filter loses specificity
```
boneskull: alternatively, I could make a type assertion, e.g., `/** @type {keyof sesAllowList} */(k)` or `if (k in sesAllowList)`.  the former is appropriate, I think.

packages/tofu/src/inspectSesCompat.js
```diff
@@ -54,8 +89,13 @@ function inspectSesCompat(ast) {
...
 }
 
+/**
+ *
+ * @param {string[]} memberPath
+ * @returns {boolean}
+ */
 function hasSetterInWhitelist(memberPath) {
-  let allowListTarget = sesAllowlist
+  let allowListTarget = /** @type {Record<string, any>} */ (sesAllowlist)
```
boneskull, line 98: there's another way to do this, but don't know if it's worth bothering

packages/tofu/src/inspectSource.js
```diff
@@ -309,46 +341,77 @@ function inspectPatternElementForDeclarations(node, keyPath = []) {
...
-  // expand each property into a path, recursively
+  /** @type {Declaration[]} */
   let results = []
-  node.properties.forEach((prop) => {
-    const propName = prop.key.name
-    const child = prop.value
-    results = results.concat(
-      inspectPatternElementForDeclarations(child, [...keyPath, propName])
-    )
+  const properties = /** @type {import('@babel/types').ObjectProperty[]} */ (
```
boneskull: note to self: go back and take another look at why the type assertion is needed.

packages/core/src/schema/lavamoat-policy.v0-0-1.schema.ts
naugtur: Works for me.

packages/tofu/src/findGlobals.js
naugtur: I'd love a testcase. Still prefer it in a separate commit, in a PR that gets merged before or after this one (depending on which way it's less hassle) 
I can promise a fast turnaround on reviewing the extracted bit.  
The idea being, we have two separate commits and PRs for change management - one with logic changes and the other with just types. 
That's also what Lego was after.

packages/tofu/src/inspectSesCompat.js
naugtur: [edit - I deleted what I initially wrote]

I'll need to take a second look here, because maybe we do want to allow Symbols too, not just string keys, even if the current implementation was not ready for that.

packages/tofu/src/inspectSource.js
naugtur: that sucks. Whose fault is that, ours or babel's?

packages/tofu/src/referenceTracker.js
naugtur: I'd defer merging until we know for sure. I was expecting a more "you missed that obvious thing there" answer.

packages/tofu/src/ses-whitelist.js
naugtur: Not in this PR, but yes. Let's even look into exporting it from `ses`

naugtur (changes requested):
Blocking on https://github.com/LavaMoat/LavaMoat/pull/761/files#diff-45f2498bbc6fc41108f02cdfc08366bdec96fb18e677cb875b6acdfede5cdc0fR310 because it made me anxious

packages/tofu/src/inspectSesCompat.js
naugtur: You know I'm not the best person to ask if typing something properly is worth bothering, but this here is not really propagating anywhere we'd visit often.

packages/tofu/src/findGlobals.js
boneskull: OK.  I can `@ts-ignore` what's broken for now.  I'll use that policy going forward.

packages/tofu/src/inspectPrimordialAssignments.js
boneskull: I've yanked this out and will put the logic changes in another PR

packages/tofu/src/inspectSource.js
boneskull: babel.

packages/tofu/src/ses-whitelist.js
boneskull: created #814

packages/tofu/src/inspectSesCompat.js
boneskull: The reason this happens is because `Reflect.ownKeys()` just returns `Array<string|symbol>` instead of the discrete list of keys in `sesAllowList`. So all we know is we have some `string`s and `symbol`s and do not know that they have anything to do with `sesAllowList`.

boneskull: To that end, I'm curious why we need to use `Reflect.ownKeys()` here?  @kumavis

boneskull:
@naugtur I don't understand what this is blocked on.  I removed the logic changes necessary to satisfy the type checker.

naugtur (approved):

boneskull:
Going to attempt to squash some commits in a logically-consistent manner, then will merge.

PR #765, opened 2023-11-22
[experimental] policy generation for ESM

boneskull:
- The build is likely to fail.
- There is no explicit test for this use-case; testing done manually. 
- There _is_ a test in `tofu` for its portion.
- The commits are a mess.
- This is _not_ targeting `boneskull/core-types`, though it probably should.  There may be some duplication of types between this PR and that PR, as I added a few in the places I was stomping around.
- The large number of files changed is mainly due to to the addition of a large test fixture copied and minimally modified from `@endo/compartment-mapper`.

cc @naugtur who wanted to see. careful what you wish for


packages/tofu/src/util.js
```diff
@@ -178,10 +185,20 @@ function getParents(nodePath) {
...
+function isInFunctionDeclaration(nodePath) {
+  let target = nodePath.parentPath
+  while (target) {
+    // if function declaration found, short-circuit instead of continuing
+    if (target.isFunctionDeclaration() || target.isFunctionExpression()) {
+      return true
+    }
+    target = target.parentPath
+  }
+  return false
```
boneskull, line 203: This is opportunistic and I changed it while debugging.  This is more performant than the previous implementation, but not by an order of magnitude.

packages/tofu/src/inspectSource.js
```diff
@@ -133,44 +133,38 @@ function inspectGlobals(
...
-function inspectEsmImports(ast) {
-  const esmImports = []
+/**
+ * This finds all modules `import`ed into the AST, as well as any re-exported modules.
+ *
+ * @param {import('@babel/types').Node} ast
+ * @param {string[]} [packagesToInspect]
+ * @returns {{ esmImports: string[] }}
+ */
+function inspectEsmImports(ast, packagesToInspect) {
```
boneskull, line 143: The previous implementation of this function returned "nested" imports where appropriate, e.g., `import {foo} from 'bar'` results in `bar.foo`.  This does not happen in CJS packages, so I figured it best to maintain parity. 

 I'm unsure of any drawback to retaining the old behavior except that it will likely explode the size of a generated policy file.

boneskull: Also, we don't actually need the `switch` statement nor the extra conditionals--these look to have been purely defensive checks--based on the documented type of the `NodePath` as provided to the callbacks to `traverse()`

packages/tofu/test/inspectEsmImports.spec.js
```diff
@@ -29,22 +29,26 @@ testInspect(
   import defaultExport3, * as name2 from "package09";
   import "package10";
   var promise = import("package11");
+  export * as name3 from "package12";
+  export { name4 } from "package13";
+  export { default as name5 } from "package14";
+  export const foo = "bar";
```
boneskull, line 35: a) Treating re-exports as imports may be semantically incorrect and b) there may be some other reason to not do this.  

Respective to a) I don't know why this would be true, and b) I don't know what that would be.

packages/tofu/src/findGlobals.js
```diff
@@ -15,6 +15,17 @@ const nonReferenceIdentifiers = [
...
 ]
 
+const importExportSpecifierTypes = new Set(
+  /** @type {const} */ ([
+    'ImportSpecifier',
+    'ImportDefaultSpecifier',
+    'ImportNamespaceSpecifier',
+    'ExportSpecifier',
+    'ExportDefaultSpecifier',
+    'MetaProperty',
```
boneskull, line 25: this is `import.meta`, which we need to account for insofar as we need to tell tofu it isn't a "global"

packages/node/src/parseForPolicy.js
```diff
@@ -19,88 +18,81 @@ const {
...
+ * @returns {string|null}
+ */
+
 // file extension omitted can be omitted, eg https://npmfs.com/package/yargs/17.0.1/yargs
-const commonjsExtensions = ['', '.js', '.cjs']
-const resolutionOmittedExtensions = ['.js', '.json']
+const jsExtensions = new Set(
+  /** @type {const} */ (['.cjs', '.mjs', '.js', '.json'])
+)
+const cjsResolutionOmittedExtensions = /** @type {const} */ (['.js', '.json'])
```
boneskull, line 36: This list is a) incomplete (`.node`) and b) does not apply to ES modules.  Should probably investigate its use further and prevent it from being used an an ESM context, if feasible.

packages/node/src/parseForPolicy.js
```diff
@@ -19,88 +18,81 @@ const {
...
+  new Set(
+    /** @type {const} */ ([
+      ...builtinModules,
+      ...builtinModules.map((id) => `node:${id}`),
+    ])
+  )
+)
+
+function isBuiltin(name) {
+  return BUILTIN_MODULES.has(name)
```
boneskull, line 51: we could use `require('node:module').isBuiltin(name)`, which works with both `node:`-prefixed paths and without.  I added this myself!!!

packages/node/src/parseForPolicy.js
```diff
@@ -19,88 +18,81 @@ const {
...
     resolveHook,
     canonicalNameMap,
   })
-  const moduleSpecifier = resolveHook(
-    entryId,
-    path.join(projectRoot, 'package.json')
-  )
+
+  const moduleSpecifier = resolveHook(String(entryId).trim(), pkgJsonPath)
+  console.debug('moduleSpecifier = %s', moduleSpecifier)
```
boneskull, line 95: errant debug

packages/node/src/parseForPolicy.js
```diff
@@ -114,54 +106,100 @@ async function parseForPolicy({
...
 
-function makeResolveHook({ projectRoot, resolutions = {}, canonicalNameMap }) {
+/**
+ *
+ * @param {string} rootDir
+ * @param {Record<string, string>} canonicalNameMap
+ * @param {Record<string, string>} [resolutions]
+ * @returns {ResolveHook}
+ */
+function makeResolutionResolveHook(rootDir, canonicalNameMap, resolutions) {
```
boneskull, line 116: I extracted this because I thought I would want two separate `makeResolveHook` implementations, but that turned out not to be true.  I left it as its own function just for the sake of splitting up a big function.

packages/node/src/parseForPolicy.js
```diff
@@ -114,54 +106,100 @@ async function parseForPolicy({
...
-    if (result) {
-      // if path is a relative path, it should be relative to the projectRoot
-      if (path.isAbsolute(result)) {
-        requestedName = result
+
+    if (isBuiltin(requestedName)) {
+      return requestedName
+    }
+
+    requestedName = requestedName || '.'
```
boneskull, line 163: I should probably add a comment why this was needed.  something about the entry point being monkeyed with earlier in the control flow results in an empty `requestedName` if the entry point is equal to the cwd (which it probably should be if we're in an ESM package)

packages/node/src/parseForPolicy.js
```diff
@@ -114,54 +106,100 @@ async function parseForPolicy({
...
-        requestedName = path.resolve(projectRoot, result)
+        requestedName = resolutionOverride
       }
     }
-    // utilize node's internal resolution algo
-    const { resolve } = createRequire(new URL(`file://${referrer}`))
-    /* eslint-disable no-useless-catch */
-    let resolved
+
+    const { resolve } = Module.createRequire(referrer)
```
boneskull, line 173: if you didn't catch it elsewhere: we cannot use the `resolve` package for ESM because it doesn't understand `exports` nor `imports`. it is still used as a fallback elsewhere where we _really_ need the `package.json` and the package doesn't want to give it to us.

packages/node/src/parseForPolicy.js
```diff
@@ -171,8 +209,9 @@ function makeImportHook({
     const extension = path.extname(filename)
     const packageName = getPackageNameForModulePath(canonicalNameMap, filename)
 
-    if (commonjsExtensions.includes(extension)) {
-      return makeJsModuleRecord(specifier, filename, packageName)
+    if (jsExtensions.has(extension)) {
+      const moduleRecord = makeJsModuleRecord(specifier, filename, packageName)
+      return moduleRecord
```
boneskull, line 214: combine

packages/node/src/parseForPolicy.js
```diff
@@ -211,24 +264,39 @@ function makeImportHook({
...
     // parse
     const ast = parseModule(content, specifier)
     // get imports
+    const { esmImports } = inspectEsmImports(ast, null, false)
     const { cjsImports } = inspectImports(ast, null, false)
+
+    // FIXME: if the package is CJS, and the extension is not `.mjs`, then the
+    // file cannot contain `import` statements. this will require reading the
+    // local `package.json` to determine, however.
+    const imports = [...new Set([...esmImports, ...cjsImports])]
```
boneskull, line 287: I think we end up reading every `package.json` one way or another.  So it'd be cool to have perhaps a LRU cache of them that we could pull from.  Such an API, if used here, would need to accept an arbitrary file and return its closest ancestor `package.json` (which may or may not be a cache hit).  

That is--_if_ we care to do anything about this.  A CJS script containing an `import` statement will fail with a `SyntaxError` anyway.

But, I don't think Endo/SES would dig on `const require = Module.createRequire(path)` _anyway_, so maybe it's moot.

packages/tofu/src/findGlobals.js
```diff
@@ -27,6 +38,12 @@ function findGlobals(ast) {
       if (nonReferenceIdentifiers.includes(parentType)) {
         return
       }
+
```
boneskull, line 41: also I really wanted to refactor this to use the builtin type guards in `@babel/traverse`'s `NodePath` objects.  unclear to me if it's any better--type prop comparison would probably be enough since it's likely a discriminated union--but it avoids us needing to do string comparisons on `.type` props (typo-prone).

e.g., `path.node.parent.type === 'VariableDeclarator'` would be rewritten as `path.parentPath.isVariableDeclarator()`

packages/tofu/src/util.js
boneskull: cc @kumavis

boneskull:
* `main`
  * **PR #765** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/765?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/765?utm_source=stack-comment).

packages/node/src/parseForPolicy.js
naugtur: I think this list only covers for `require('./filename')` where `filename.js` exists. It has no reason to be used for esm and it's possible it already is not.

packages/node/src/parseForPolicy.js
naugtur: sounds good. It should be available in node versions we care about (found a 16.9.0 changelog with it)

packages/node/src/parseForPolicy.js
naugtur: AFAIR none of our tools (endo, lavamoat) polyfill the createRequire to be controlled by us. we could do that in the import hook we pass to importLocation

packages/node/src/parseForPolicy.js
boneskull: yeah, I'm not sure how this is used offhand; it just seems to be exported.

packages/tofu/src/inspectSource.js
```diff
@@ -133,44 +133,38 @@ function inspectGlobals(
...
-            return
-          }
-          default: {
-            throw new Error(
-              `inspectEsmImports - unknown import specifier type "${spec.type}"`
-            )
-          }
-        }
-      })
+    ExportNamedDeclaration: (path) => {
```
naugtur, line 161: this looks too simple :D

boneskull:
Split up into #808, #809 , #810, and #811. Closing

PR #766, opened 2023-11-30
Update allow scripts (rebased)

legobeat:
- #478 rebased on `main`

### DIFF
--- a/packages/allow-scripts/src/yarn-berry-plugin.cjs
+++ b/packages/allow-scripts/src/yarn-berry-plugin.cjs
@@ -1,0 +1,22 @@
+// @ts-check
+
+const { execute } = require('@yarnpkg/shell')
+
+/**
+ * @import {Plugin} from '@yarnpkg/core'
+ */
+
+module.exports = /** @type {Plugin} */ {
+  hooks: {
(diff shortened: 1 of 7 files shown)


boneskull:
I've salvaged this.  But someone else should decide what to do with it, as I'm not sure of the intention here.

cc @naugtur @leotm

naugtur (approved):

packages/allow-scripts/src/yarn-berry-plugin.cjs
```diff
@@ -0,0 +1,22 @@
+// @ts-check
```
naugtur, line 1: We already have this as a typescript thing in the packages. Do we want to replace with a no-build option?

naugtur:
These updates generally make sense but are minor. 
If we want to have a pure-js version of the yarn plugin, we should think where to put it.

packages/allow-scripts/src/yarn-berry-plugin.cjs
boneskull: I didn't want to throw a tsconfig in there

boneskull:
I don't know what I did to hose this thing, but whatever.

PR #769, opened 2023-11-30
Fix a quirk in webpack plugin's cjs support

naugtur:
The test causes the following error:
```
'Cannot assign to read only property \'exports\' of object \'[object Object]\'',
```

The fix: runtimeHandler was frozen, `exports` was not a local variable like in regular module closures, so it had to be made writable. Having a getter there is a bit wasteful, but alternatives would mean a larger bundle or riskier implementation

### DIFF
--- a/packages/webpack/test/fixtures/main/main.js
+++ b/packages/webpack/test/fixtures/main/main.js
@@ -1,4 +1,5 @@
 import { hello as helloCommonJS } from 'commonjs-package'
+import { re } from 'commonjs-quirks'
 import { hello as helloES6 } from 'es6-module-package'
 import 'side-effects-package/styles.css'
 import { hello as helloTypeScript } from 'typescript-package'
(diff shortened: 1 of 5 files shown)


packages/webpack/test/fixtures/main/main.js
```diff
@@ -1,4 +1,5 @@
 import { hello as helloCommonJS } from 'commonjs-package'
+import { re } from 'commonjs-quirks'
```
boneskull, line 2: what's this for?

boneskull (approved):

packages/webpack/test/fixtures/main/main.js
naugtur: To assert that we not only don't throw on the weird behavior of semver pkg, but also properly accept its export. I kept the original name for no reason other than it being searchable in node_modules

PR #772, opened 2023-12-06
feat: create @lavamoat/node

boneskull:
Continuation of #742

This is a prototype re-implementation of `lavamoat` based on `@endo/compartment-mapper`.  There's still work to be done here, but it's now ready for review.  We'd like to get it merged into `main`, but it needn't be published yet (imo).

Ref: https://github.com/endojs/endo/issues/423

* * *

<details>
<summary>UPDATE Jun 7 2024</summary>

This PR has been folded into a single branch as there were too many to keep track of.  That, of course, makes this PR large.  But _don't worry_ -- the high file count is due to the number of test fixtures!

It will fail checks until necessary changes are merged in Endo (and released).  Outstanding changes include: 

- [x] endojs/endo#2310

Ref: #1168 

Note: There are some changes to `core` in here which could probably be extracted.

</details>

<details>

<summary>UPDATE Sep 12 2024</summary>

### Building

Unfortunately, linking two monorepos together can be difficult. It is somewhat complicated by the fact that Endo is a Yarn project.

For this branch to work properly, you will need to link it against the `boneskull/dynamic` branch in endojs/endo/#2310 (if you have [gh](https://cli.github.com) installed, try `gh pr checkout 2310` in your `endojs/endo` working copy).  After you've checked out the branch, run `yarn` to install the deps.

_If you're looking to hack on this, you probably want option 1. If you are just trying to run it, use option 2._

#### Option 1: `npm link`

- **Advantage:** Works once it's setup
- **Disadvantage:** Hard to setup

1.  In `endojs/endo`, create global symlinks to the listed packages:

    ```sh
    npm link -w @endo/compartment-mapper \
      -w ses \
      -w @endo/env-options \
      -w @endo/cjs-module-analyzer \
      -w @endo/static-module-record \
      -w @endo/zip \
      -w @endo/evasive-transform
    ```
    
    All of the above packages are necessary. When you link a workspace, you get a symlink to the workspace directory. But the _dependencies_ may have been hoisted or deduped to the `node_modules` off of the workspace root. The workspace dir will likely have its own `node_modules`, and hoisted deps won't be there!
    
    This could theoretically be solved by telling Yarn _not_ to hoist or dedupe. This optional exercise is for the reader.
2. Display the versions of the packages which `@lavamoat/endomoat` depends on:

   ```sh
   npm pkg get version -w ses -w @endo/evasive-transform -w @endo/compartment-mapper
   ```
   
   You should see something like:
   
   ```json
   {
     "ses": "x.x.x",
     "@endo/evasive-transform": "y.y.y",
     "@endo/compartment-mapper": "z.z.z"
   }
   ```
   
   If the above versions do not match the versions of these dependencies in `lavamoat/lavamoat`'s `./packages/endomoat/package.json`, **this will not work**.  Keep the  information handy.
3. Cleanup any built artifacts:

    ```sh
    npm run postpack -w @endo/compartment-mapper \
      -w @endo/env-options \
      -w @endo/cjs-module-analyzer \
      -w @endo/static-module-record \
      -w @endo/zip \
      -w @endo/evasive-transform && \
    npm run clean -w ses
    ```
4. Build the types:

    ```sh
    npm run prepack -w @endo/compartment-mapper \
      -w @endo/env-options \
      -w @endo/cjs-module-analyzer \
      -w @endo/static-module-record \
      -w @endo/zip \
      -w @endo/evasive-transform && \
    npm run prepare -w ses
    ```

    > Unfortunately, this will create a whole lot of `.d.ts` files which _aren't_ in `.gitignore`. To clean them up later, re-run step 3.
5. Navigate to your `lavamoat/lavamoat` working copy.
6. Checkout this PR (`gh pr checkout 772`)
7. Run `npm ci` or equivalent 
8. Run `npm ls` and ensure there are no errors
9. Look at `./packages/endomoat/package.json`; ensure the versions match the output from step  2. If they don't, **you must upgrade them**; this will necessarily change `./package-lock.json and ./packages/endomoat/package.json`.
10. Run the following to complete the link (again in the working copy root):

    ```sh
    npm link @endo/compartment-mapper \
      ses \
      @endo/env-options \
      @endo/cjs-module-analyzer \
      @endo/static-module-record \
      @endo/zip \
      @endo/evasive-transform
    ```
11. Run `npm ls '@endo/*' ses`. You should see a bunch of errors. This is expected. Even so, _ensure the dep versions branching from `@lavamoat/endomoat` all point to the deduped packages in the root workspace._
12. Run `npm run rebuild` which will build `@lavamoat/endomoat`'s declaration files. You shouldn't see more than a handful of errors. If this is off by one or more orders of magnitude, something is wrong.

#### Option 2: tarballs

- **Advantage:** Pretty easy
- **Disadvantage:** You have to repeat it every time you change something in Endo

1. Pack the tarballs:

    ```sh
    npm pack -w @endo/compartment-mapper \
      -w ses \
      -w @endo/env-options \
      -w @endo/cjs-module-analyzer \
      -w @endo/static-module-record \
      -w @endo/zip \
      -w @endo/evasive-transform
    ```
    
    This will dump a bunch of `.tgz` files in the working copy root.
2. Navigate to your `lavamoat/lavamoat` working copy.
3. Run `npm install "/path/to/endojs/endo/*.tgz" -w @lavamoat/endomoat --no-save`
4. Run `npm run rebuild` which will build `@lavamoat/endomoat`'s declaration files.  You shouldn't see more than a handful of errors. If this is off by one or more orders of magnitude, something is wrong.

</details>

<details>
**UPDATE Oct 8 2024**

This now targets ~~#1395~~ #1397 for as long as that PR exists. 
</details>

<details>
**UPDATE Nov 14 2024**

This should now "work" with the current version of `@endo/compartment-mapper` and should require no linking.
</details>

**UPDATE Jan 9 2025**

This PR has been squashed into (3) commits; the original-ish set of commits is archived in the [`archive/lavamoat-node-endo-rewrite`](https://github.com/LavaMoat/LavaMoat/tree/archive/lavamoat-node-endo-rewrite) branch.


boneskull:
* **#772** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/772?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#969](https://github.com/LavaMoat/LavaMoat/pull/969) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/969?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1349](https://github.com/LavaMoat/LavaMoat/pull/1349) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1349?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/772?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1397** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1397?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1442** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1442?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/endomoat/src/cli.js
```diff
@@ -0,0 +1,130 @@
...
+          },
+          'policy-override': {
+            alias: ['o', 'override'],
+            describe:
+              'Filepath to the policy override file, relative to the application directory',
+            type: 'string',
+            normalize: true,
+            default: constants.DEFAULT_POLICY_OVERRIDE_PATH,
+          },
+          'endo-policy': {
```
boneskull: hide this

boneskull:
I think I can retarget this to `main`; I'm not sure why the rebase gnomes made it target `boneskull/aa-custom-resolver`.  Just need `@endo/compartment-mapper` to cut a release

.gitignore
```diff
@@ -121,4 +121,4 @@ packages/*/types
 # VScode
 .vscode
 
-!**/fixtures/main/node_modules
+!packages/webpack/test/fixtures/main/node_modules
```
boneskull: I'm unsure why I narrowed this, but I did.

packages/endomoat/test/helpers.js
```diff
@@ -0,0 +1,161 @@
...
+    fs.promises.readFile
+  )
+
+  /** @type {string} */
+  let projectDir
+  /** @type {string} */
+  let policyDir
+
+  // for eslint
+  await Promise.resolve()
```
boneskull: The rule getting tripped comes from the jessie.js eslint plugin (IIRC), which demands that the first `await` within an `async` function is not within a block.  

It doesn't _seem_ like `try` is the block type the rule should be concerned about; it makes more sense if that block is a conditional. But maybe there's a good, non-obvious reason?

packages/endomoat/test/scenarios.spec.js
```diff
@@ -0,0 +1,65 @@
...
+    GLOBAL_WRITE_REGEX.test(JSON.stringify(scenario.configOverride))
+  )
+}
+
+const SKIP_SCENARIOS = new Set([
+  'globalRef - check default containment',
+  'globalRef - Webpack code in the wild works',
+])
+
+test('Run scenarios', async (t) => {
```
boneskull: this runs all the scenarios from `core` which haven't been explicitly skipped due to something-not-yet-implemented or other reasons.  the stuff with `???` is skipped in `lavamoat-node` as well.

packages/endomoat/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "type": "git",
+    "url": "https://github.com/LavaMoat/LavaMoat.git",
+    "directory": "packages/endomoat"
+  },
+  "homepage": "https://github.com/LavaMoat/LavaMoat/tree/main/packages/endomoat",
+  "author": "Lavamoat Project",
+  "license": "MIT",
+  "private": true,
+  "workspaces": [
+    "test/fixtures/main"
```
naugtur: this is odd. why do we have it?

packages/endomoat/src/cli.js
```diff
@@ -0,0 +1,129 @@
...
+        ),
+    /**
+     * Default command handler.
+     *
+     * @remarks
+     * (That's "handler for the default command"--_not_ "default handler for a
+     * command").
+     */
+    async (argv) => {
+      const { run } = await import('./index.js')
```
naugtur: Why do we need dynamic imports here? It's fine for them to be in a node specific cli, but we tended to avoid them in code because none of our tools support them.

packages/endomoat/package.json
boneskull: this is a (possibly better) way to prepare the test fixtures; I wanted to see if it'd work.  we could use this elsewhere and eliminate some prep scripts, it looks like.

packages/endomoat/src/cli.js
boneskull: premature optimization, I suppose.  there's no need to load `index.js` if the arg parsing fails because a user provided a bad option.

packages/endomoat/package.json
boneskull: the question to ask, imo, is "why _not_ do this?" and right now I don't have an answer.

legobeat (dismissed):
Do we need to also update bundled ses for this?

```
$ npm run test:ses
> lavamoat-core@15.2.1 test:ses
> diff -wq ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js

Files ../../node_modules/ses/dist/lockdown.umd.js and ./lib/lockdown.umd.js differ
npm ERR! Lifecycle script `test:ses` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: lavamoat-core@15.2.1 
npm ERR!   at location: /home/runner/work/LavaMoat/LavaMoat/packages/core 
npm ERR! Lifecycle script `test` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: lavamoat-core@15.2.1 
npm ERR!   at location: /home/runner/work/LavaMoat/LavaMoat/packages/core 
```

boneskull:
> Do we need to also update bundled ses for this?

This is done. Also see #1078

boneskull:
CLI is now:

```
endomoat <entrypoint>

Run an application

Positionals:
  entrypoint  Path to the application entry point            [string] [required]

Options:
      --help                         Show help                         [boolean]
      --version                      Show version number               [boolean]
  -p, --policy                       Filepath to a policy file
                                 [string] [default: "lavamoat/node/policy.json"]
  -o, --policy-override, --override  Filepath to a policy override file
                        [string] [default: "lavamoat/node/policy-override.json"]
      --policy-debug, --pd           Filepath to a policy debug file
                           [string] [default: "lavamoat/node/policy-debug.json"]
      --cwd                          Path to application root directory
                                         [string] [default: (current directory)]
```

See #969 for how policy gen fits in

naugtur:
I think it's time to merge this one. I'll dig deeper into some details here and there before we do, but I think the only change I'd want to see in this PR or the 969 one is giving it its final name. (that we're yet to choose)

packages/core/package.json
```diff
@@ -39,6 +39,7 @@
     "json-stable-stringify": "1.1.1",
     "lavamoat-tofu": "^7.3.0",
     "merge-deep": "3.0.3",
+    "ses": "1.4.1",
```
legobeat: core already has `ses` vendored, as well as as a `devDependency`.

This adds a conflicting version (`1.4.1` vs `1.5.0`) to the package, adding to further redundancy.

I think adding it as runtime dependency would make a lot of sense, if it means removing the checked-in version and the devDependency at the same time. But it doesn't seem reasonable to have all three (two at runtime) simultaneously.

That refactor would stand well on its own, as a preceding change to this PR, I think.

legobeat:
@boneskull I broke out the new feature in core into #1266 .

.vscode/settings.json
```diff
@@ -6,5 +6,5 @@
     "source.formatDocument": "explicit"
   },
   "prettier.useEditorConfig": true,
-  "files.insertFinalNewline": false
+  "editor.formatOnSave": true
```
legobeat: - #1267

packages/core/src/schema/lavamoat-policy.v0-0-1.schema.ts
```diff
@@ -84,6 +84,10 @@ export interface ResourcePolicy {
    * Allow native modules
    */
   native?: boolean
+  /**
+   * Allow dynamic requires
+   */
+  dynamic?: boolean
```
legobeat: https://github.com/LavaMoat/LavaMoat/pull/1266/files#r1701203126

packages/endomoat/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "url": "https://github.com/LavaMoat/LavaMoat.git",
+    "directory": "packages/endomoat"
+  },
+  "homepage": "https://github.com/LavaMoat/LavaMoat/tree/main/packages/endomoat",
+  "author": "Lavamoat Project",
+  "license": "MIT",
+  "private": true,
+  "engines": {
+    "node": "^16.20.0 || ^18.0.0 || ^20.0.0",
+    "npm": ">=7.0.0"
```
legobeat:
```suggestion
    "node": "^18.0.0 || ^20.0.0 <20.16.1 || ^22.5.1",
    "npm": ">=9.0.0"
```

packages/endomoat/package.json
```diff
@@ -0,0 +1,67 @@
...
+  "scripts": {
+    "lint:deps": "depcheck",
+    "test": "npm run test:run",
+    "test:run": "ava",
+    "update-policy": "node ./scripts/update-own-policy.js"
+  },
+  "dependencies": {
+    "@endo/compartment-mapper": "1.1.5",
+    "@endo/evasive-transform": "1.1.2",
+    "@types/node": "18.19.28",
```
legobeat: This should be in `devDependencies`

packages/endomoat/src/types.ts
legobeat, line 1: This could perhaps be renamed to `types.d.ts` after #1264

packages/endomoat/src/policy-gen/policy-generator.js
```diff
@@ -0,0 +1,310 @@
...
+import { LMRCache } from './lmr-cache.js'
+import { PolicyGeneratorContext } from './policy-generator-context.js'
+
+const { entries, freeze } = Object
+
+/**
+ * Service which generates a LavaMoat policy from compartment map descriptors
+ * and sources
+ */
+export class PolicyGenerator {
```
legobeat: It seems like we don't have an explicit typedef for a corresponding interface? 
To the extent that it is exposed in public API, I think it would be very beneficial to have that be as an interface (which this class would implement).

WDYT?

boneskull: It's not a public API

packages/core/src/schema/lavamoat-policy.v0-0-1.schema.ts
naugtur: After more recent work this field should go away AFAIR.

packages/node/src/policy-gen/greedy-compartment.js
```diff
@@ -0,0 +1,161 @@
+/**
+ * Provides {@link GreedyCompartment}, which you can just pretend doesn't exist.
```
naugtur: a suggestion for its purpose would fit nicely in here

packages/node/src/policy-gen/greedy-compartment.js
```diff
@@ -0,0 +1,161 @@
...
+  return class GreedyCompartment extends Compartment {
+    /**
+     * Creates a "greedy" import hook if we can.
+     *
+     * @param {CompartmentOptions} [options]
+     */
+    constructor(options = {}) {
+      const { importHook, name } = options
+
+      if (!importHook || !name || name === ATTENUATORS_COMPARTMENT) {
```
naugtur: - I don't think we need to care about attenuators compartment here (now that we won't even use it)
- Endo should not make us care about the attenuators compartment here. I'm not fully content with how I designed the attenuators compartment's relationship with the rest of the app.

no actions here, unless we wanna remove the `ATTENUATORS_COMPARTMENT` reference.

packages/node/src/policy-gen/greedy-compartment.js
```diff
@@ -0,0 +1,161 @@
...
+
+          for (const [
+            moduleDescriptorName,
+            { compartment: moduleDescriptorCompartmentName },
+          ] of entries(compartmentDescriptor.modules)) {
+            // ignore self
+            if (
+              moduleDescriptorCompartmentName &&
+              compartmentMap.compartments[moduleDescriptorCompartmentName] !==
+                compartmentDescriptor
```
naugtur: (could) 
This would be easier to process as an early `continue`

packages/node/src/policy-gen/greedy-compartment.js
```diff
@@ -0,0 +1,161 @@
...
+                  compartmentDescriptor.compartments.has(
+                    moduleDescriptorCompartmentName
+                  )
+                ) {
+                  // this supports the "deprecated" RecordModuleDescriptor
+                  // format and the new SourceModuleDescriptor format. we
+                  // are futurists here at LavaMoat.
+                  //
+                  // note: `record` and `source` are non-extensible, which
+                  // is why the shallow clones are needed
```
naugtur: (must) use a WeakMap to ensure that if we get the same reference again, we return the same (clone) reference again. This is mandatory unless we can prove importHook can never return the same reference twice. A lot in SES and Endo depend on reference identity tho, so we might loose some guarantees if we don't maintain that.

packages/node/src/policy-gen/greedy-compartment.js
```diff
@@ -0,0 +1,161 @@
...
+              if (modCompartmentDescriptor) {
+                const canonicalName = getCanonicalName(
+                  modCompartmentDescriptor,
+                  compartmentMap.entry.compartment ===
+                    modCompartmentDescriptor.name
+                )
+                if (
+                  canonicalName &&
+                  canonicalName in (policyOverride?.resources ?? {}) &&
+                  compartmentDescriptor.compartments.has(
```
naugtur: I don't understand why this 3rd condition is here

packages/node/src/policy-gen/greedy-compartment.js
```diff
@@ -0,0 +1,161 @@
...
+    constructor(options = {}) {
+      const { importHook, name } = options
+
+      if (!importHook || !name || name === ATTENUATORS_COMPARTMENT) {
+        super(options)
+        return
+      }
+
+      /** @type {ImportHook} */
+      const importHookWrapper = async (specifier) => {
```
naugtur: (nit) If it's not accepting an importHook reference as arg, should it be called `wrappedImportHook` instead of `importHookWrapper`?

packages/node/src/policy-gen/greedy-compartment.js
```diff
@@ -0,0 +1,161 @@
...
+                  // note: `record` and `source` are non-extensible, which
+                  // is why the shallow clones are needed
+                  if (isRecordModuleDescriptor(moduleDescriptor)) {
+                    if (
+                      !moduleDescriptor.record.imports.includes(
+                        moduleDescriptorName
+                      )
+                    ) {
+                      moduleDescriptor.record = {
+                        ...moduleDescriptor.record,
```
naugtur: Are we making a new copy for each field being added? I mean, the likelihood of this happening is not great and most often we'll have just one item to add, but switching to WeakMap (above) seems to be a good opportunity to copy once and extend later. 
Then we might also want to make it non-extensible at the end of the loop.

packages/node/src/constants.js
```diff
@@ -8,27 +8,54 @@
...
+ * could pass `@lavamoat/endomoat/attenuator` as default attenuator and have it
+ * loaded statelessly. We're using an impossible specifier to match with an
+ * external module instead.
  */
-export const DEFAULT_POLICY_OVERRIDE_FILENAME = 'policy-override.json'
+export const DEFAULT_ATTENUATOR = '@attenuator@'
 
 /**
- * Default policy debug filename
+ * Default policy filename
```
naugtur: mistake?

packages/node/src/default-attenuator.js
```diff
@@ -47,24 +50,24 @@ export const attenuateModule = (params, originalObject) => {
...
-      if (resource.globals && typeof resource.globals === 'object') {
-        entries(resource.globals).forEach(([key, value]) => {
-          if (value === POLICY_ITEM_WRITE) {
-            knownWritableFields.add(key)
-          }
-        })
-      }
-    })
-  }
+  values(lavaMoatPolicy?.resources ?? {}).forEach((resource) => {
```
naugtur: (nit) 
I liked my explicit if that didn't require concluding you'll skip on iterating if policy is empty. 
Also helpful in avoiding mistakes if we ever want to add some work other than looping over resources

packages/node/src/default-attenuator.js
```diff
@@ -96,8 +99,22 @@ export const makeGlobalsAttenuator = ({ policy } = { policy: undefined }) => {
...
+         * the same value from the original LavaMoat policy!
+         *
+         * `getEndowmentsForConfig()` accepts a LavaMoat `ResourcePolicy` having
+         * a `globals` prop (a `GlobalPolicy`). Since we control the types of
+         * the parameters provided to the attenuator (`GlobalAttenuatorParams`),
+         * we've made those parameters (the first of which is `policy`)
+         * structurally compatible with a `GlobalPolicy`. Below, we're creating
+         * a phony `ResourcePolicy` to make it fit.
+         */
+        { globals: policy },
```
naugtur: The comment is not entirely correct.

The attenuator is not getting Endo policy, it's getting whatever we put as attenuator argument in the endo policy resulting from our translation, which is the literal content of globals field from LavaMoat policy in our case.

boneskull: - [ ] rewrite this

packages/node/src/policy-gen/greedy-compartment.js
boneskull: - ~~[ ] `WeakMap` the `ModuleSource`~~

packages/node/src/policy-gen/greedy-compartment.js
boneskull: yeah, I don't think `name === ATTENUATORS_COMPARTMENT` will ever be true

packages/node/src/policy-gen/greedy-compartment.js
boneskull: @naugtur

I tried this, but then realized if we're keeping this cache for `ModuleSource`, we should also keep once for `ModuleSource.imports` since that prop must also become a new array.  then I realized we don't need the cache at all, and can instead _mutate_ objects instead of overwriting them _if_ they are extensible.  

initially, they will be frozen, and we must reassign those properties to extensible shallow clones.  if we encounter them again, they will be extensible, so we can mutate freely.

at any rate, after consulting with Kris, it does not seem like we should worry about it too much. we also do not need to re-freeze them.

packages/node/src/policy-gen/greedy-compartment.js
boneskull: yes, we didn't need it.

packages/core/src/makePrepareRealmGlobalFromConfig.js
```diff
@@ -14,23 +14,31 @@ function makePrepareRealmGlobalFromConfig({ createFunctionWrapper }) {
     getTopLevelWriteAccessFromPackageConfig,
   }
 
+  /**
+   * Should never return an array containing `'Function'`
```
weizman, line 18: What's this about?

package.json
```diff
@@ -68,6 +70,7 @@
     "typescript": "5.6.3"
   },
   "overrides": {
+    "@endo/compartment-mapper": "$@endo/compartment-mapper",
```
naugtur, line 73: what's this doing?

packages/core/src/makePrepareRealmGlobalFromConfig.js
naugtur: How about a note why it shouldn't

packages/node/package.json
```diff
@@ -0,0 +1,67 @@
...
+  "scripts": {
+    "lint:deps": "depcheck",
+    "test": "npm run test:run",
+    "test:run": "ava"
+  },
+  "dependencies": {
+    "@endo/compartment-mapper": "1.4.0",
+    "@endo/evasive-transform": "1.3.0",
+    "@types/node": "18.19.67",
+    "chalk": "4.1.2",
```
naugtur, line 53: (could) there's so many alternatives to chalk nowadays and people keep recommending them as more lightweight. And I bet Sindre is tired of holding on to that one. (maybe not in this PR tho)

packages/node/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "json-stable-stringify": "1.1.1",
+    "lavamoat-core": "^16.0.0",
+    "loggerr": "4.2.0",
```
naugtur, line 55: Why do we need this instead of console.log? Did it come with the nice output for incompatibilities we find in policy gen?

packages/node/src/cli.js
```diff
@@ -0,0 +1,381 @@
...
+        describe: 'Filepath to a policy file',
+        type: 'string',
+        normalize: true,
+        default: constants.DEFAULT_POLICY_PATH,
+        nargs: 1,
+        requiresArg: true,
+        global: true,
+      },
+      'policy-override': {
+        alias: ['o', 'override'],
```
naugtur: Do we want to support the old names for these too?

packages/node/src/cli.js
```diff
@@ -0,0 +1,381 @@
...
+      'policy-debug': {
+        alias: ['pd'],
+        describe: 'Filepath to a policy debug file',
+        default: constants.DEFAULT_POLICY_DEBUG_PATH,
+        nargs: 1,
+        type: 'string',
+        requiresArg: true,
+        global: true,
+      },
+      cwd: {
```
naugtur: this used to be called `--projectRoot` and I think it was a slightly better name

packages/node/src/cli.js
```diff
@@ -0,0 +1,381 @@
...
+        type: 'string',
+        nargs: 1,
+        requiresArg: true,
+        normalize: true,
+        default: process.cwd(),
+        defaultDescription: '(current directory)',
+        coerce: path.resolve,
+        global: true,
+      },
+      dev: {
```
naugtur, line 186: (should) I think we used to include dev dependencies by default and for our main usecase - running builds in CI, it would be a good choice to keep it that way.

packages/node/src/compartment-map.js
```diff
@@ -0,0 +1,165 @@
+import { captureFromMap } from '@endo/compartment-mapper/capture-lite.js'
+import { defaultParserForLanguage } from '@endo/compartment-mapper/import-parsers.js'
+import { importLocation } from '@endo/compartment-mapper/import.js'
+import { mapNodeModules } from '@endo/compartment-mapper/node-modules.js'
```
naugtur: Could we do stuff via the official compartment-mapper API? Maybe it needs to be exposed differently so that we can. We're the second most prominent consumer of the package afterall ;)

packages/node/src/compartment-map.js
```diff
@@ -0,0 +1,165 @@
...
+    captureSources: sources,
+    compartmentRenames,
+  } = await captureFromMap(readPowers, nodeCompartmentMap, {
+    ...captureOpts,
+    ...ENDO_OPTIONS,
+    Compartment: LavaMoatCompartment,
+    modules: {
+      [DEFAULT_ATTENUATOR]: {
+        // by passing policyOverride, we set any writable globals in the output policy
+        attenuateGlobals: makeGlobalsAttenuator({ policy: policyOverride }),
```
naugtur: ~~Why is this only referring to policyOverride? Is that just naming, or are we not passing the main policy to the attenuator? Generated policy could contain writable items too.~~

After more reading it seems this is used in policy generation. In that case, why do we have the attenuator at all?

packages/node/src/policy-gen/index.js
```diff
@@ -0,0 +1,200 @@
...
+ * Generates a LavaMoat policy or debug policy from a given entry point using
+ * `@endo/compartment-mapper`
+ *
+ * @param {string | URL} entrypointPath
+ * @param {GeneratePolicyOptions} [opts]
+ * @returns {Promise<LavaMoatPolicy>}
+ * @public
+ */
+export const generatePolicy = async (
+  entrypointPath,
```
naugtur: (nit) I don't think turning a URL into a path should be a responsibility of this function.

packages/node/src/policy-gen/policy-gen-compartment.js
```diff
@@ -0,0 +1,222 @@
...
+ */
+const updateModuleSource = (moduleDescriptor, canonicalName) => {
+  let moduleSource = isRecordModuleDescriptor(moduleDescriptor)
+    ? moduleDescriptor.record
+    : isSourceModuleDescriptor(moduleDescriptor)
+      ? moduleDescriptor.source
+      : undefined
+
+  if (!moduleSource) {
+    throw new TypeError(`Unsupported module descriptor type; this is a bug`)
```
naugtur: (should)
`this is a bug` implies we want it reported, but there's no specifics in the error message. These error messages need more context in them to help us figure out what happened based on a bug report with just the message.

packages/node/src/log.js
```diff
@@ -0,0 +1,38 @@
...
+ * `debug` and `info` levels. Output originating in `@lavamoat/node` is
+ * generally warnings or information about the policy generation and conversion
+ * process. The "output" of `@lavamoat/node` should be considered the output of
+ * the program being executed; thus `@lavamoat/node` should not use `stdout` for
+ * its own output.
+ *
+ * Because the value is handled as a tuple, we cannot generate it via
+ * {@link Array.fill}
+ */
+export const log = new Loggerr({
```
naugtur, line 25: What's the benefit of using this over console.error that makes up for using a dependency?

On the flip side, I see it's actively maintained by Wes, so must be real good.

packages/core/src/makePrepareRealmGlobalFromConfig.js
boneskull: We don't allow write access to `Function` and read access is implied.

boneskull: We don't allow write access to `Function` and read access is implied.

package.json
```diff
@@ -40,7 +40,9 @@
   "devDependencies": {
     "@commitlint/cli": "19.6.0",
     "@commitlint/config-conventional": "19.6.0",
+    "@endo/compartment-mapper": "1.4.0",
```
boneskull, line 43: Like other packages, we are pulling this in only to set the [override](https://github.com/LavaMoat/LavaMoat/pull/772/files/eb70b9cf1070a47baa2ae5209fac1c858e34a08b#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R73) below. This ensures `@endo/compartment-mapper` is the same version across all workspaces.

packages/node/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "directory": "packages/node"
+  },
+  "homepage": "https://lavamoat.github.io",
+  "bugs": {
+    "url": "https://github.com/LavaMoat/LavaMoat/issues"
+  },
+  "author": "LavaMoat Project",
+  "license": "MIT",
+  "engines": {
+    "node": "^16.20.0 || ^18.0.0 || ^20.0.0 || ^22.0.0",
```
boneskull: - [x] Set `engines` to minimum 18

packages/node/README.md
```diff
@@ -0,0 +1,153 @@
...
+npm install @lavamoat/node
+```
+
+## Usage
+
+> [!IMPORTANT]
+>
+> Before proceeding, it's recommended to [check out LavaMoat's docs][docs].
+> Understanding LavaMoat's concepts will help you cross the LavaMoat Drawbridge
+> of Success™.
```
naugtur, line 55: awesome

packages/node/src/import-hook.js
```diff
@@ -0,0 +1,59 @@
...
+  const ns = require(specifier)
+  return freeze(
+    /** @type {ThirdPartyStaticModuleInterface} */ ({
+      imports: [],
+      exports: keys(ns),
+      execute: (moduleExports) => {
+        moduleExports.default = ns
+        assign(moduleExports, ns)
+      },
+    })
```
naugtur: (could) an added benefit of DRYing up this bit would be the execute function not having a reference to require in its context.

naugtur (commented):
submitting  comments from first round of reading the cleaned up sources

packages/node/src/compartment-map.js
boneskull: - [ ] send PR to export the crap we need to use

packages/node/src/policy-gen/index.js
boneskull: So I was thinking about this and since it's a public API, I prefer to be liberal in what we accept.  The lower-level API is _illiberal_ and works with strings instead of URLs, so I think the conversion is appropriate here.

packages/node/package.json
boneskull: Noted that `chalk` is a dep of `loggerr` so it's already there

packages/node/package.json
boneskull: Noted that it provides log levels and some nice  formatting, which we would otherwise have to implement ourselves.

packages/node/src/policy-gen/policy-gen-compartment.js
boneskull: @naugtur What would you suggest?

packages/node/src/policy-gen/policy-gen-compartment.js
boneskull, line 1: @naugtur I've updated the algorithm in here; please re-review it

package.json
boneskull: ensuring all `@endo/compartment-mapper` is at the version defined in this file

packages/node/README.md
```diff
@@ -0,0 +1,153 @@
...
+    native modules) is controlled by user-defined policy
+- Secures your supply chain by generating and maintaining (with your help) policies
+
+LavaMoat is built on top of [SES][].
+
+## Supply Chain Security
+
+Over time, your application's dependencies will change. When they do, LavaMoat detects any changes to resource access. For example, if an upgraded transitive dependency now uses the `left-pad` package, this will be evident from an auto-generated policy file. You can then choose to allow or deny this access (or stop using `left-pad`).
+
+In this way, LavaMoat is a little like [Socket][], which alerts you to dubious changes in a package's behavior (among other things). LavaMoat, however, surfaces _all_ such changes; it's _up to you_ to decide whether or not they're reasonable.
```
boneskull, line 25: @naugtur rewrite

naugtur (approved):
We can rewrite the readme later, this is very clean now in comparison. :ship: it

PR #787, opened 2023-12-14
feat(webpack): wip handle globals and resolution of node builtins

naugtur:
- added handling of `global` via webpack
- testing and exploring the way node builtins are aliased and how that affects policy enforcement
- looking into workspace level dependencies handling (they're identified as `external:../../` which is not ideal) - probably to be fixed in another PR targeting `aa`


naugtur:
distributed most fixes from here to other PRs.

PR #792, opened 2023-12-15
fix(lavapack): espree should be a dep, not a dev dep

boneskull:
It's referenced in `src/builder-runtime.js`

cc @kumavis


kumavis (approved):

boneskull:
How do we prevent this going forward?  I think that [depcheck](https://npm.im/depcheck) is probably just insufficient for our purposes.  I've been using [dependency-cruiser](https://npm.im/dependency-cruiser) on a personal project, which I've found to be far more powerful and monorepo-aware.  This is not the only alternative, though.  I'll create an issue for this.

PR #803, opened 2023-12-17
docs(aa): use an ʻokina instead of misusing an apostrophe

davidmurdoch:
giphy

### DIFF
--- a/packages/aa/README.md
+++ b/packages/aa/README.md
@@ -2,7 +2,7 @@
 
 LavaMoat's canonical package name convention.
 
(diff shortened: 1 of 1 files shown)


kumavis (commented):
> Unicode

> In the Unicode standard, the ʻokina is encoded as U+02BB ʻ MODIFIER LETTER TURNED COMMA, which can be rendered in HTML by the entity &#699; (or in hexadecimal form &#x02BB;).[1]

kumavis (approved):

PR #808, opened 2023-12-21
feat(aa): add custom resolver option

boneskull:
`loadCanonicalNameMap` now supports passing in a custom resolver, which is needed for actually resolving ESM packages, due to the inability of the [`resolve`](https://npm.im/resolve) package to do so.  

Implementing the custom resolver (from #811) would break our custom "performant resolver"; we may need to abandon it anyway, because `package.json` _must_ be read in order to resolve any package having an `exports` field.  

An alternative implementation would _require_ a custom resolver, which might make more separation-of-concerns sense.

@naugtur I'm not sure if the webpack plugin would find passing a custom resolver into `@lavamoat/aa` useful?

### DIFF
--- a/packages/aa/src/index.js
+++ b/packages/aa/src/index.js
@@ -38,9 +38,6 @@ const performantResolve = createPerformantResolve()
  * @see {@link https://nodejs.org/api/packages.html#subpath-exports}
  */
 function createPerformantResolve() {
-  if (performantResolve) {
-    return performantResolve
-  }
   /**
    * @param {string} filepath
    */
@@ -117,11 +114,14 @@ function wrappedResolveSync(resolve, depName, basedir) {
   } catch (e) {
     const err = /** @type {Error} */ (e)
     if (
-      ('code' in err && err.code !== 'MODULE_NOT_FOUND') ||
-      !err.message?.startsWith('Cannot find module')
+      err &&
+      typeof err === 'object' &&
+      (('code' in err && err.code === 'MODULE_NOT_FOUND') ||
+        err.message?.startsWith('Cannot find module'))
     ) {
-      throw err
+      return
     }
+    throw err
   }
 }
 
--- a/packages/aa/src/index.js
+++ b/packages/aa/src/index.js
@@ -30,7 +30,12 @@ const performantResolve = createPerformantResolve()
  * resolved, offering 2x performance improvement compared to using original
  * resolve
  *
+ * This resolver, using the `resolve` package, is incompatible with subpath
+ * exports.
+ *
  * @returns {Resolver}
+ * @see {@link https://npm.im/resolve}
+ * @see {@link https://nodejs.org/api/packages.html#subpath-exports}
  */
 function createPerformantResolve() {
   if (performantResolve) {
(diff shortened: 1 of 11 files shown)


boneskull:
* `main`
  * **PR #808** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/808?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #811** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/811?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/808?utm_source=stack-comment).

naugtur:
> @naugtur I'm not sure if the webpack plugin would find passing a custom resolver into `@lavamoat/aa` useful?

Initially, yes. I think we already experimented with it and the current state of passing it in seemed sufficient.
Where is this needed?

For webpack plugin, eventually it's likely gonna need to own the traversal of dependencies and therefore use aa for path->aa transform or not at all.

packages/aa/src/index.js
```diff
@@ -12,10 +12,19 @@ module.exports = {
   createPerformantResolve,
 }
 
+/** @type {Resolver|undefined} */
+let performantResolve
```
legobeat:
```suggestion
/** @type {Resolver|undefined} */
let performantResolve
```
I'm guessing this global state acts as a memoization-optimization? Could we avoid making it mutable by instantiating it immediately?
```suggestion
/** @type {Resolver} */
const performantResolve = createPerformantResolve()
```

boneskull: yeah, seems good

boneskull:
@LavaMoat/devs - should we deprecate or remove `createPerformantResolve` from the public API?  I don't see it used anywhere.

naugtur:
> @LavaMoat/devs - should we deprecate or remove `createPerformantResolve` from the public API? I don't see it used anywhere.

It being part of the public API is not intentional. A bug. I'd not oppose hiding it in a patch release.

packages/aa/src/index.js
```diff
@@ -83,12 +102,24 @@ function wrappedResolveSync(resolve, depName, packageDir) {
...
   try {
     return resolve.sync(depRelativePackageJsonPath, { basedir: packageDir })
-  } catch (err) {
-    if (!(/** @type {Error} */ (err).message.includes('Cannot find module'))) {
-      throw err
+  } catch (e) {
+    const err = /** @type {Error} */ (e)
+    if (
+      err.message?.includes('Cannot find module') &&
+      resolve !== performantResolve
```
naugtur: I'm pretty sure I don't want this feature. 

If someone's supplying a specific resolver, they should be limited to that resolver's knowledge. Us finding the module would likely be invalid anyway.

naugtur (changes requested):

naugtur:
Dirty check in tests is failing on a missing policy update for the switch to `node:` prefixed package names.

packages/aa/src/index.js
boneskull: let me make sure I'm not actually relying on this behavior...

boneskull:
@naugtur I've removed the misfeature and updated the policy file

packages/aa/src/index.js
```diff
@@ -74,22 +98,16 @@ async function loadCanonicalNameMap({ rootDir, includeDevDeps, resolve }) {
...
+ * @param {string} basedir - Dir to resolve from
  * @returns {string | undefined}
  */
-function wrappedResolveSync(resolve, depName, packageDir) {
+function wrappedResolveSync(resolve, depName, basedir) {
   const depRelativePackageJsonPath = path.join(depName, 'package.json')
-  try {
-    return resolve.sync(depRelativePackageJsonPath, { basedir: packageDir })
-  } catch (err) {
-    if (!(/** @type {Error} */ (err).message.includes('Cannot find module'))) {
```
naugtur, line 87: I think we were silencing the module not found errors to handle that more gracefully elsewhere.

But you can see that from the tests already.

naugtur (approved):

packages/aa/src/index.js
legobeat: I guess this would also be more robust using `message.startsWith` as opposed to `message.includes`.

naugtur (approved):

boneskull:
needs policy file update.  going to open a PR for a solution to avoid this before hitting CI

PR #809, opened 2023-12-21
feat(tofu): add ESM support

boneskull:
This makes `tofu` ESM-aware.

@kumavis originally implemented most of this.  I changed it such that:

- It uses the same granularity as the CJS import detection (module level)
- It considers `export ... from` syntax (re-export)

### DIFF
--- a/packages/tofu/src/findGlobals.js
+++ b/packages/tofu/src/findGlobals.js
@@ -15,6 +15,17 @@ const nonReferenceIdentifiers = [
   'RestElement',
 ]
 
+const importExportSpecifierTypes = new Set(
+  /** @type {const} */ ([
+    'ImportSpecifier',
+    'ImportDefaultSpecifier',
+    'ImportNamespaceSpecifier',
+    'ExportSpecifier',
+    'ExportDefaultSpecifier',
+    'MetaProperty',
+  ])
+)
+
 module.exports = { findGlobals }
 
 /**
@@ -37,6 +48,12 @@ function findGlobals(ast) {
       if (nonReferenceIdentifiers.includes(parentType)) {
         return
       }
+
+      // toss out esm imports/exports
+      if (importExportSpecifierTypes.has(parentType)) {
+        return
+      }
+
       if (parentType === 'VariableDeclarator' && path.parent.id === path.node) {
         return
       }
--- a/packages/tofu/src/findGlobals.js
+++ b/packages/tofu/src/findGlobals.js
@@ -3,37 +3,36 @@
 const { default: traverse } = require('@babel/traverse')
 const { isInFunctionDeclaration, isMemberLikeExpression } = require('./util')
 
-const nonReferenceIdentifiers = [
+/**
+ * Types which are not references to globals
+ */
+const nonReferenceIdentifiers = new Set([
+  'ArrayPatten',
+  'BreakStatement',
+  'CatchClause',
+  'ClassMethod',
+  'ContinueStatement',
+  'ExportDefaultSpecifier',
+  'ExportSpecifier',
   'FunctionDeclaration',
   'FunctionExpression',
-  'ClassMethod',
+  'ImportDefaultSpecifier',
+  'ImportNamespaceSpecifier',
+  'ImportSpecifier',
   'LabeledStatement',
-  'BreakStatement',
-  'ContinueStatement',
-  'CatchClause',
-  'ArrayPatten',
+  'MetaProperty',
   'RestElement',
-]
-
-const importExportSpecifierTypes = new Set(
-  /** @type {const} */ ([
-    'ImportSpecifier',
-    'ImportDefaultSpecifier',
-    'ImportNamespaceSpecifier',
-    'ExportSpecifier',
-    'ExportDefaultSpecifier',
-    'MetaProperty',
-  ])
-)
+])
 
 module.exports = { findGlobals }
 
 /**
- * @typedef {import('@babel/traverse').NodePath<import('@babel/types').Identifier|import('@babel/types').ThisExpression>} IdentifierOrThisExpressionNodePath
+ * @typedef {import('@babel/traverse').NodePath<
+ *   import('@babel/types').Identifier | import('@babel/types').ThisExpression
+ * >} IdentifierOrThisExpressionNodePath
  */
 
 /**
- *
  * @param {import('@babel/types').Node} ast
  * @returns {Map<string, IdentifierOrThisExpressionNodePath[]>}
  */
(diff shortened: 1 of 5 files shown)


boneskull:
* **#811** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/811?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#810** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/810?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#809** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/809?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#831** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/831?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#753](https://github.com/LavaMoat/LavaMoat/pull/753) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/753?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#820](https://github.com/LavaMoat/LavaMoat/pull/820) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/820?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

packages/tofu/src/inspectSource.js
```diff
@@ -133,44 +133,38 @@ function inspectGlobals(
...
+  /** @type {Set<string>} */
+  const esmImports = new Set()
+
+  const handleNodePath = (path) => {
+    const importSource = path.node.source?.value
+
+    if (
+      importSource &&
+      ((packagesToInspect && pkgsToInspect.has(importSource)) ||
+        !packagesToInspect)
```
legobeat: Nit: I feel the added parentheses in the second clause don't help and this rewrite makes it more obvious. WDYT?
```suggestion
      !packagesToInspect
      || pkgsToInspect.has(importSource) && importSource
```

boneskull: prettier does this. it does it because not everyone always remembers order of operations

leotm (commented):
how do we feel about normalising the naming to

```
  inspectRequires,
  inspectImports,
  inspectDynamicRequires,
  inspectDynamicImports, // in future
```

or

```
  inspectCjsRequires,
  inspectEsmImports,
  inspectDynamicCjsRequires,
  inspectDynamicEsmImports, // in future
```

(and presuming dynamic esm imports out of scope for now)

boneskull:
@leotm I think I'll punt on renaming anything further for now.

- `require()` and dynamic require calls can occur in _both_ CJS and ESM modules (though it takes more effort to do so in the latter, and maybe only in Node.JS)
- `import()` calls can occur in _both_ CJS and ESM modules
- `import` and `export` statements can _only_ occur in ESM modules

`lavamoat-tofu` doesn't know what context it's running in (e.g., it doesn't know if `require` can actually be used, because it surely  cannot within an unbundled web-facing ES module), which may be less-than-ideal.  but sounds like we will make some decisions around this soonish, even if that is just maintaining the status quo.

also I'm unsure if `lavamoat-tofu` can accurately understand `require()` as created by Node's `Module.createRequire()`...

packages/tofu/src/findGlobals.js
```diff
@@ -3,26 +3,36 @@
...
+  'ImportDefaultSpecifier',
+  'ImportNamespaceSpecifier',
+  'ImportSpecifier',
   'LabeledStatement',
-  'BreakStatement',
-  'ContinueStatement',
-  'CatchClause',
-  'ArrayPatten',
+  'MetaProperty',
   'RestElement',
```
naugtur, line 24: I hope that's just a reorder. Don't have the patience to verify :facepalm:

boneskull: yes, but `MetaProperty` is new, and refers to `import.meta`

packages/tofu/src/inspectSource.js
naugtur: This is confusing to me for other reasons. Since we're turning packagesToInspect into the pkgsToInspect set, could we limit ourselves to using just one? I'd prefer if packagesToInspect were not uttered below the first line of the function. 

And since the empty set is not falsy, but doesn't contain anything, `packagesToInspect && pkgsToInspect.has(importSource)) ||       !packagesToInspect` becomes `pkgsToInspect.has(importSource)) || pkgsToInspect.size===0`. Amirite?

packages/tofu/src/inspectSource.js
```diff
@@ -175,60 +172,55 @@ function inspectGlobals(
...
-                /** @type {import('@babel/types').Node} */ (spec).type
-              }"`
-            )
-          }
-        }
-      })
+    ExportNamedDeclaration: (path) => {
+      handleNodePath(path)
     },
+    ImportDeclaration: handleNodePath,
```
naugtur, line 211: Why is it wrapped in one case and not the other?

packages/tofu/src/inspectSource.js
```diff
@@ -175,60 +172,55 @@ function inspectGlobals(
...
   traverse(ast, {
-    ImportDeclaration: (path) => {
-      const { node } = path
-      const { specifiers, source } = node
-      // not sure if this is ever not a StringLiteral, but just in case
-      if (source.type !== 'StringLiteral') {
-        return
-      }
-      const importSource = source.value
-      specifiers.forEach((spec) => {
```
naugtur, line 194: How is all this no longer needed? I'm gonna need to dig into this some more. For now what I'm seeing here is we add stuff to imports regardless of all the specifiers processing that was happening here and the case where importName is modified is no longer handled. Not sure why.

packages/tofu/test/inspectEsmImports.spec.js
```diff
@@ -29,58 +26,40 @@ testInspect(
...
+    'package05',
+    'package06/path/to/specific/un-exported/file',
+    'package07',
+    'package08',
+    'package09',
+    'package10',
+    'package12',
+    'package13',
+    'package14',
+  ]
```
naugtur, line 48: This doesn't seem equivalent. Are we dropping information that's actually unnecessary?

boneskull: It's not equivalent with the previous behavior.  We're dropping the granularity down to module level.  Previously it was granular to named exports.  This is why e.g., `package06/path/to/specific/un-exported/file` only appears once.

packages/tofu/src/inspectSource.js
boneskull: `ImportDeclaration` handles everything we care about

naugtur: You sound very certain. I have no reason to disbelieve that until we see stuff missing from policies.

packages/tofu/test/inspectEsmImports.spec.js
naugtur: sounds like a big improvement.

naugtur (approved):

naugtur:
This PR lived long enough, I'm gonna merge it now and ask forgiveness later.

PR #810, opened 2023-12-21
feat(core): ESM support in policy generation

boneskull:
This adds ESM support to the policy generator.

## TODO

- [x] add ~~scenarios~~ test project(s). only policy gen will be tested since the runtime cannot execute ESM
- [x] handle CJS/ESM disagreements about allowed global variable names

### DIFF
--- a/packages/core/src/generatePolicy.js
+++ b/packages/core/src/generatePolicy.js
@@ -14,6 +14,7 @@ const {
     mapToObj,
     reduceToTopmostApiCallsFromStrings,
   },
+  inspectEsmImports,
   // @ts-expect-error no types yet
 } = require('lavamoat-tofu')
 const { mergePolicy } = require('./mergePolicy')
@@ -38,13 +39,16 @@ function createModuleInspector(opts) {
   const debugInfo = {}
 
   /** @type {ModuleInspector} */
-  const inspector = /** @type {any} */ (new EventEmitter())
-  inspector.inspectModule = (moduleRecord, opts2 = {}) => {
-    inspectModule(moduleRecord, { ...opts, ...opts2 })
-  }
-  inspector.generatePolicy = (opts2) => {
-    return generatePolicy({ ...opts, ...opts2 })
-  }
+  const inspector = Object.assign(new EventEmitter(), {
+    /** @type {InspectModuleFn} */
+    inspectModule: (moduleRecord, opts2) => {
+      inspectModule(moduleRecord, { ...opts, ...opts2 })
+    },
+    /** @type {GeneratePolicyFn} */
+    generatePolicy: (opts2) => {
+      return generatePolicy({ ...opts, ...opts2 })
+    },
+  })
 
   return inspector
 
@@ -251,7 +255,14 @@ function createModuleInspector(opts) {
    * @param {boolean} includeDebugInfo
    */
   function inspectForGlobals(ast, moduleRecord, packageName, includeDebugInfo) {
-    const commonJsRefs = ['require', 'module', 'exports', 'arguments']
+    const commonJsRefs = [
+      'require',
+      'module',
+      'exports',
+      'arguments',
+      'import',
+      'export',
+    ]
     const globalObjPrototypeRefs = Object.getOwnPropertyNames(Object.prototype)
     const foundGlobals = inspectGlobals(ast, {
       // browserify commonjs scope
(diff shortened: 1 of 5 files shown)


boneskull:
* **#811** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/811?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#810** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/810?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#809** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/809?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#831** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/831?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#753](https://github.com/LavaMoat/LavaMoat/pull/753) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/753?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#820](https://github.com/LavaMoat/LavaMoat/pull/820) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/820?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

packages/core/src/generatePolicy.js
```diff
@@ -251,7 +255,14 @@ function createModuleInspector(opts) {
...
    */
   function inspectForGlobals(ast, moduleRecord, packageName, includeDebugInfo) {
-    const commonJsRefs = ['require', 'module', 'exports', 'arguments']
+    const commonJsRefs = [
+      'require',
+      'module',
+      'exports',
+      'arguments',
+      'import',
+      'export',
```
naugtur: this seems incorrect. I can definitely create a function named `import` or use a global named `import` in a cjs module.

naugtur:
git got confused, this needs rebasing.

packages/core/src/generatePolicy.js
boneskull: Indeed!  This will have to make a distinction between CJS and ESM earlier; this part of the code will need to know what module system is in play.  I don't know any way to do that other than to first check file extension, and if it's `.js`, read `package.json` (or use a lib to do something like this).

boneskull: Also, I think we shouldn't do that in `core` at all; the consuming module (lavamoat, endomoat, webpack, etc) will need to tell it (via callback or otherwise) what module system to use.

(why wasn't `import` a reserved keyword? urgh)

boneskull: since this wants a `LavamoatModuleRecord`, it may make sense to stuff the information in there, wouldn't it?

now I am wondering why `ast` is its own param and not `moduleRecord.ast`

boneskull: @naugtur I'm going to sit on this until you respond; I don't want to implement this in e.g., `lavamoat-node` or elsewhere if it's not the right direction.

naugtur: Let's discuss this synchronously.
If we don't know what's esm or cjs in here, how do we inspectESMImports?
Endo doesn't support dynamic imports anyway, so maybe that changes our assumptions?
Does the AST specify the difference between dynamic import and a function named import?

boneskull:
@naugtur I've made the fixes to the global detection, as we discussed a week or two ago.

packages/core/src/generatePolicy.js
```diff
@@ -162,7 +174,7 @@ function createModuleInspector(opts) {
       moduleRecord.ast ||
       parse(moduleRecord.content, {
         // esm support
-        sourceType: 'module',
+        sourceType: 'unambiguous',
```
naugtur, line 177: what does this mean? if it's not ambiguous, why doesn't it have a different value?

naugtur (approved):

naugtur:
I'm not entirely sure how sourceType gets populated, but I guess I'll see it in action.

packages/core/src/generatePolicy.js
boneskull: it means, "babel, please determine which module type this is".  I am not sure why it's called "unambiguous", though; it's a confusing property.

boneskull:
It just kind of...does.  AFAICT it always does.

PR #825, opened 2023-12-22
chore(ci): fail if any VCS files changed

### LINKED ISSUE #721: ci: fail if working tree is dirty

boneskull:
This adds a script to determine if any files under version control changed, which should be run post-test-suites.

In the check summary, it will display the file(s) changed and provide references to the lines changed. Unfortunately, it will _not_ tell you _why_ the lines changed; you'll have to put on your Sherlock Holmes hat and figure it out.

image

Closes #721

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -15,8 +15,17 @@ jobs:
       matrix:
         node-version: [16.x, 18.x, 20.x]
     steps:
-      - name: Checkout
+      - name: Checkout (PR)
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
+
+      # fetch-depth: 0 looks like it has a perf impact, so let's not do that if we don't have to.
+      # see https://github.com/actions/checkout/blob/main/README.md#usage
+      - name: Checkout (Push)
+        if: github.event_name != 'pull_request'
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
+        with:
+          fetch-depth: 0 # needed for dirty check
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@b39b52d1213e96004bfcb1c61a8a6fa8ab84f3e8 # v4
         with:
@@ -31,6 +40,8 @@ jobs:
           install-command: npm ci --foreground-scripts
       - name: Test
         run: npm test
+      - name: Dirty Check
+        run: node ./scripts/dirty-check.js
 
   lint:
     runs-on: ubuntu-latest
--- a/scripts/dirty-check.js
+++ b/scripts/dirty-check.js
@@ -1,0 +1,75 @@
+// @ts-check
+
+/**
+ * This script is intended to be run in a GitHub Actions workflow to check if
+ * test execution changed any files under version control. If so, the script
+ * will exit with a non-zero exit code and print an "error" to the GitHub
+ * Actions log for each changed file.
+ *
+ * This is not a shell script because portability.
+ *
+ * @example
+ *
+ * ```yaml
+ * on: [push, pull_request]
+ * jobs:
+ *   test:
+ *     runs-on: ubuntu-latest
+ *   steps:
+ *     - run: your stuff
+ *     - run: node ./scripts/dirty-check.js
+ * ```
+ *
+ * @packageDocumentation
+ * @see
+ * {@link https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions}
+ */
+
+const { promisify } = require('node:util')
+const execFile = promisify(require('node:child_process').execFile)
+
+/**
+ * Asks `git` if any files changed and exits with a non-zero exit code if so.
+ *
+ * @returns {Promise<void>}
+ * @todo We could determine what lines changed in what file so the file would be
+ *   annotated with the error, but it requires some extra parsing of `git diff`
+ *   output.
+ */
+async function main() {
+  await execFile('git', ['add', '.'])
+  const diffResultPromise = execFile(
+    'git',
+    ['diff', '--staged', '-b', '--quiet'],
+    { encoding: 'utf8' },
+  )
+
+  try {
+    await diffResultPromise
+    console.log('::notice::No dirty files detected')
+  } catch (e) {
+    if (e.code !== 1 || e.killed || e.stderr !== '') {
+      throw e
+    }
+    // something changed, and we should fail the build
+    process.exitCode = 1
+
+    console.log(
+      '::error,title=Dirty File::Working area contains unexpected changes; see job log for diff'
+    )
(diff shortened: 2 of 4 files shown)


boneskull:
* `main`
  * **PR #825** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/825?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/825?utm_source=stack-comment).

kumavis (approved):

scripts/dirty-check.js
```diff
@@ -0,0 +1,133 @@
...
+  /** @type {[file: string, lines: [line: number, endLine: number][]][]} */
+  const changeEntries = []
+
+  /** @type {string | undefined} */
+  let currentFile
+
+  /** @type {[line: number, endLine: number][]} */
+  let currentLines = []
+
+  // console.error(stdout)
```
leotm: can we remove this

boneskull:
smth messed up.  at minimum, we shouldn't show multiple annotations per file.

boneskull:
See #859 which this rule surfaced.

boneskull:
The diff script now ignores whitespace changes at EOL because I'm seeing that happen in `packages/node/test/projects/2/atob.policy.json` and I don't know how to fix it and don't want to spend more time figuring it out.

boneskull:
afaict the difference in that `atob/policy.json` is unique to linux, so I'll need to run this in a container for debugging, which I'm doing now.

boneskull:
I'm hitting https://github.com/docker/for-mac/issues/6998 when attempting to run via CI via `act`, so we'll have to sit on this for now.

scripts/dirty-check.js
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Asks `git` if any files changed and exits with a non-zero exit code if so.
+ *
+ * @returns {Promise<void>}
+ * @todo We could determine what lines changed in what file so the file would be
+ *   annotated with the error, but it requires some extra parsing of `git diff`
+ *   output.
+ */
+async function main() {
+  const { stdout } = await execFile('git', ['diff', '-b', '--name-only'], {
```
legobeat:
```suggestion
  const { stdout } = await execFile('git', ['diff', '-w', '-b', '--name-only'], {
```

if the remaining issue is simply whitespace-related, ignoring whitespace-only changes could be a temporary workaround?

i guess a proper solution involves updating `.gitattributes`

boneskull: This is one of the problems, yes. 

I was having a tough time reproducing this locally, since it seems to be linux-only, and Docker Desktop for M2 is broken, so I can't run `act` to quickly iterate.

scripts/dirty-check.js
```diff
@@ -0,0 +1,70 @@
...
+  if (changedFiles.length) {
+    process.exitCode = 1
+
+    for (const file of changedFiles) {
+      console.log(
+        '::error file=%s,title=Dirty File::%s changed unexpectedly; see job log for diff',
+        file,
+        file
+      )
+      const { stdout } = await execFile('git', ['diff', '--', file])
```
legobeat:
```suggestion
      const { stdout } = await execFile('git', ['diff', '-w', '--', file])
```

.github/workflows/nodejs.yml
```diff
@@ -15,8 +15,17 @@ jobs:
...
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
+
+      # fetch-depth: 0 looks like it has a perf impact, so let's not do that if we don't have to.
+      # see https://github.com/actions/checkout/blob/main/README.md#usage
+      - name: Checkout (Push)
+        if: github.event_name != 'pull_request'
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
+        with:
+          fetch-depth: 0 # needed for dirty check
```
legobeat, line 28:
```suggestion
          fetch-depth: 2 # needed for dirty check
```
Curious why this is not sufficient :thinking:

legobeat:
@boneskull:

- #988  (https://github.com/LavaMoat/LavaMoat/actions/runs/7733360337/job/21085285858)
- #987
- #990

.github/workflows/nodejs.yml
boneskull: me too

scripts/dirty-check.js
boneskull: You'd think that would help... I'll try it, but I thought I did already.

legobeat: The `--name-only` flag apparently changes the diff-algorithm and makes it ignore both `-b` and `-w.
- #990

scripts/dirty-check.js
```diff
@@ -0,0 +1,75 @@
...
+  await execFile('git', ['add', '.'])
+  const diffResultPromise = execFile(
+    'git',
+    ['diff', '--staged', '-b', '--quiet'],
+    { encoding: 'utf8' },
+  )
+
+  try {
+    await diffResultPromise
+    console.log('::notice::No dirty files detected')
```
legobeat, line 49: (could): Nice to have a standard of splitting log messages from parsable output to allow for easy parsing of result. For example for this script: logging to stderr, diff to stdout.
```suggestion
    console.warn('::notice::No dirty files detected')
```
`console.error` and `console.warn` output to stderr whereas `console.log` outputs to stdout.

scripts/dirty-check.js
```diff
@@ -0,0 +1,75 @@
...
+  } catch (e) {
+    if (e.code !== 1 || e.killed || e.stderr !== '') {
+      throw e
+    }
+    // something changed, and we should fail the build
+    process.exitCode = 1
+
+    console.log(
```
legobeat, line 57:
```suggestion
    console.warn(
```

scripts/dirty-check.js
```diff
@@ -0,0 +1,75 @@
...
+      { encoding: 'utf8' },
+    )
+    console.log(stdout)
+  }
+}
+
+if (require.main === module) {
+  main().catch((err) => {
+    // last resort
+    console.log(`::error::${err}`)
```
legobeat, line 72:
```suggestion
    console.error(`::error::${err}`)
```

legobeat (approved):

scripts/dirty-check.js
boneskull: From what I can tell, GitHub only listens on stdout for [commands](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-a-workflow-command), so this is intentional

boneskull:
I am going to squash parts of this (not all of it)

PR #835, opened 2024-01-09
fix(tofu): make lavamoat-core a peer dep

boneskull:
`lavamoat-core` was a dev dep of `lavamoat-tofu`, but the former depends directly on the latter.  this creates a cycle, which is bad.

This makes `lavamoat-core` a peer dep of `lavamoat-tofu`, which avoids the cycle.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22787,11 +22787,13 @@
         "type-fest": "4.7.1"
       },
       "devDependencies": {
-        "deep-equal": "2.2.3",
-        "lavamoat-core": "^15.0.0"
+        "deep-equal": "2.2.3"
       },
       "engines": {
         "node": "^16.20.0 || ^18.0.0 || ^20.0.0"
+      },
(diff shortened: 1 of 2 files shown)


boneskull:
* `main`
  * **PR #835** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/835?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/835?utm_source=stack-comment).

naugtur (approved):
For the record: I think I understand what this does but I didn't check if my assumptions are correct.

naugtur:
My assumptions were correct, but I didn't take into account that it would not be updated by release-please and a major release creates a peer dependency problem now.

PR #836, opened 2024-01-10
feat(laverna): create laverna

boneskull:
# Laverna

 image

> Publish multiple workspaces (that's all)

If you're familiar with [lerna](https://lerna.js.org), **Laverna** does this: `lerna publish from-package`.

If you're unfamiliar with it: **Laverna** publishes all workspaces that haven't yet been published.

## Features

**Laverna** is a thin wrapper around `npm publish` which:

- Invokes `npm publish` for all workspaces _that haven't yet been published_
- Handles new packages (via a flag)
- Ignores private packages
- Requires a confirmation (by default)
- Provides a "dry-run" mode
- Prints all output from `npm publish`, including the output of `npm pack`, to enable review

## Non-Features

Perhaps more importantly, **Laverna**:

- Builds nothing
- Bumps no versions
- Does not write to `package.json`
- Does not interact with `git` (no tags, no commits, no pushes)
- Does not interact with GitHub (no releases)
- Does not cache anything

## Supported Environments

- Node.js v18.13.0+
- `npm` v8.19.3+

## Install

It's recommended to install as a dev dependency:

```shell
npm install @lavamoat/laverna -D
```

## Usage Summary

```plain
laverna [options..]

Publish multiple workspaces (that's all)

Options:

 --dryRun        - Enable dry-run mode
 --root=<path>   - Path to workspace root (default: current working dir)
 --newPkg=<name> - Workspace <name> has never been published (repeatable)
 --yes/-y        - Skip confirmation prompt

Problems? Visit https://github.com/LavaMoat/lavamoat/issues
```

### Using Release Please or Similar

For a typical release workflow, you might:

1. Bump workspace versions (e.g. via [release-please-action](https://github.com/google-github-actions/release-please-action))
2. In your updated working copy's workspace root, run `npm exec laverna` (or `npm exec laverna -- --dry-run` first)
3. Bask in glory

### Publishing a New Package

If you're publishing a package in a new workspace, you might:

1. Set the initial version of the new package (call it `foo`) in its `package.json`.
2. Run `npm exec laverna -- --newPkg=foo` (or `npm exec laverna -- --newPkg=foo --dry-run` first)
3. Profit

### Automating Publishes

You'll need to figure out how to add `--newPkg` to your CI/CD pipeline. Otherwise, use the `--yes` flag when invoking `laverna`.

## API

**Laverna** provides a programmable JavaScript API (for what it's worth).

Refer to the TypeScript definitions for details.

## Notes

[LavaMoat](https://github.com/LavaMoat/LavaMoat) needed something that did this, but didn't need Lerna for anything else. So, here we are.

## License

Copyright © 2023 Consensys, Inc. Licensed MIT


boneskull:
* **#903** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/903?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#836** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/836?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/836?utm_source=stack-comment).

boneskull:
Think I want to throw a few more E2E tests in here.

boneskull:
Note that this requires a minimum version of Node.js v18.13.0.  The entire test suite cannot be run in an older version of Node.js.  It's probably too painful to exempt testing _just this package_ from the GHA matrix, so I've done this via a dynamic Ava config.

The reasoning is twofold:

1. We should probably not publish new packages which support on an EOL version of Node.js; that's unnecessary tech debt
2. I found myself re-implementing spies/mocks. But `node:test` exports a `mock` namespace-like object which provides exactly what I needed.

boneskull:
Oh, and the reason I'm doing this _in the first place_ is that it's absolutely a thing that other projects will find useful  (including my own).  So... yes, selfish reasons.

boneskull:
**REVIEWER(S)**

This is mostly just a refactored `scripts/publish.js` which landed in #727.  I wrote some more tests and cleaned it up for public consumption.

packages/laverna/README.md
```diff
@@ -0,0 +1,95 @@
...
+## Usage Summary
+
+```plain
+laverna [options..]
+
+Publish multiple workspaces (that's all)
+
+Options:
+
+ --dryRun        - Enable dry-run mode
```
legobeat, line 53: While using `--camelCasedFlags` is consistent with the rest of lavamoat cli tooling, recognizing `--dry-run` and `--new-pkg` would make transitioning more seamless. 

LGTM otherwise!

packages/laverna/README.md
```diff
@@ -0,0 +1,95 @@
...
+1. Bump workspace versions (e.g. via [release-please-action](https://github.com/google-github-actions/release-please-action))
+2. In your updated working copy's workspace root, run `npm exec laverna` (or `npm exec laverna -- --dry-run` first)
+3. Bask in glory
+
+### Publishing a New Package
+
+If you're publishing a package in a new workspace, you might:
+
+1. Set the initial version of the new package (call it `foo`) in its `package.json`.
+2. Run `npm exec laverna -- --newPkg=foo` (or `npm exec laverna -- --newPkg=foo --dry-run` first)
```
legobeat: `--dry-run` results in `ERR_PARSE_ARGS_UNKNOWN_OPTION`. I think it should support the flag in this format (either in addition to or instead of the current one). I'd certainly keep mixing it up, for one.

packages/laverna/README.md
boneskull: the old publish script also uses camelCased flags, so this wouldn't be an impediment there.

I _would_ like to use kebab-cased flags, but I'm using Node.js' parsing facilities, and if I want an object that gives me both, I'd need to code it myself.

legobeat: In any case, the examples below still give the non-recognized form as examples, so one of them needs to change

https://github.com/LavaMoat/LavaMoat/pull/836#discussion_r1457660222

legobeat (changes requested):

boneskull:
@legobeat I've added support for kebab-case.

packages/laverna/src/cli.js
```diff
@@ -0,0 +1,110 @@
...
+    // refuse to run with both camelCase and kebab-case flags
+    if ((opts.dryRun && opts['dry-run']) || (opts.newPkg && opts['new-pkg'])) {
+      console.error(ERR, 'Use camelCase or kebab-case flags; not both')
+      process.exitCode = 1
+      return
+    }
+
+    // prefer camelCase
+    opts.dryRun = opts.dryRun ?? opts['dry-run']
+    opts.newPkg = opts.newPkg ?? opts['new-pkg']
```
legobeat:
```suggestion
    // prefer camelCase
    opts.dryRun = opts.dryRun || opts['dry-run']
    opts.newPkg = new Set([..opts.newPkg, ...opts['new-pkg'])
```

- `--newPkg` / `--new-pkg` is repeatable. While a bit silly, I don't see why `--newPkg=foo --new-pkg=bar` should not be considered valid?

- being a boolean flag defaulting to `false` and not accepting passed values/negation, `--dryRun`, `--dry-run`, and `--dryRun --dry-run` should all be redundant/equivalent



So in both cases we don't really have notion of conflicts and need for precedence, so could prob skip these checks alltogether?

packages/laverna/src/cli.js
```diff
@@ -0,0 +1,110 @@
...
+    type: 'boolean',
+  },
+  newPkg: {
+    type: 'string',
+    multiple: true,
+  },
+  'new-pkg': {
+    type: 'string',
+    multiple: true,
+  },
```
legobeat:
```suggestion
  newPkg: {
    type: 'string',
    multiple: true,
    default: [],
  },
  'new-pkg': {
    type: 'string',
    multiple: true,
    default: [],
  },
```

packages/laverna/src/defaults.js
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Default options for Laverna
+ */
+const DEFAULT_OPTS = Object.freeze(
+  /** @type {import('.').AllLavernaOptions} */ ({
+    dryRun: false,
+    yes: false,
+    root: DEFAULT_ROOT,
+    newPkg: [],
```
legobeat, line 64:
```suggestion
    newPkg: new Set(),
```

packages/laverna/src/laverna.js
```diff
@@ -0,0 +1,415 @@
...
+        ['view', name, 'versions', '--json'],
+        {
+          cwd,
+        }
+      ).then(({ stdout }) => stdout.trim())
+    } catch (e) {
+      if (/** @type {NodeJS.ErrnoException} */ (e).code === 'ENOENT') {
+        throw new Error(`Could not find npm: ${e}`)
+      }
+      if (newPkg.includes(name)) {
```
legobeat, line 130:
```suggestion
      if (newPkg.has(name)) {
```

legobeat (dismissed):
TS errors in tests:

```
Error: packages/laverna/test/laverna.spec.js(24,44): error TS2694: Namespace '"node:test"' has no exported member 'Mock'.
Error: packages/laverna/test/laverna.spec.js(25,35): error TS2694: Namespace '"node:test"' has no exported member 'Mock'.
Error: packages/laverna/test/laverna.spec.js(419,6): error TS7006: Parameter 'call' implicitly has an 'any' type.
Error: packages/laverna/test/laverna.spec.js(422,16): error TS7006: Parameter 'arg' implicitly has an 'any' type.
Error: packages/laverna/test/laverna.spec.js(506,6): error TS7006: Parameter 'call' implicitly has an 'any' type.
Error: packages/laverna/test/laverna.spec.js(508,21): error TS7006: Parameter 'arg' implicitly has an 'any' type.
Error: packages/laverna/test/laverna.spec.js(530,6): error TS7006: Parameter 'call' implicitly has an 'any' type.
Error: packages/laverna/test/laverna.spec.js(532,21): error TS7006: Parameter 'arg' implicitly has an 'any' type.
Error: packages/laverna/test/laverna.spec.js(693,6): error TS7006: Parameter 'call' implicitly has an 'any' type.
Error: packages/laverna/test/laverna.spec.js(696,16): error TS7006: Parameter 'arg' implicitly has an 'any' type.
Error: packages/laverna/test/laverna.spec.js(756,6): error TS7006: Parameter 'call' implicitly has an 'any' type.
Error: packages/laverna/test/laverna.spec.js(758,21): error TS7006: Parameter 'arg' implicitly has an 'any' type.
Error: packages/laverna/test/laverna.spec.js(788,6): error TS7006: Parameter 'call' implicitly has an 'any' type.
Error: packages/laverna/test/laverna.spec.js(791,16): error TS7006: Parameter 'arg' implicitly has an 'any' type.
Error: Process completed with exit code 2.
```

packages/laverna/src/cli.js
boneskull: I don't want to write code to merge and dedupe the arrays generated by passing `--newPkg` and `--new-pkg`.  I also didn't want inconsistent behavior around `--dryRun`.  

I didn't really want to write _any_ code to handle both kebab-case and camelCase.  If we need stuff like that, I'd rather just pull in `yargs`.

I'm inclined to rip the kebab-case stuff out.

boneskull: Another way to look at it: we can add all the bells & whistles later, if desired.  While I absolutely understand the conventions and UX around CLI apps, it's not my intent to give heavy weight usability (for now)--especially since this is something nobody else asked for.

legobeat: > I don't want to write code to merge and dedupe the arrays generated by passing --newPkg and --new-pkg


Well, it's there in the suggestion now; I don't think you'll need to write any new code for that? 


> I didn't really want to write any code to handle both kebab-case and camelCase

The suggestion here is indeed to reduce the special-handling - as it doesn't seem necessary.

>  I also didn't want inconsistent behavior around `--dryRun`.

What's the inconsistent behavior?

packages/laverna/src/defaults.js
boneskull: since `parseArgs` won't give me a `Set` back, I'll need to create more types. Given the expected size of the `newPkg` array, this seems like premature optimization--and it really does not matter if the same package occurs multiple times in the array, since we do not iterate over it.

packages/laverna/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@lavamoat/laverna",
+  "version": "0.0.0",
+  "description": "Publish multiple workspaces (that's all)",
```
legobeat, line 4:
```suggestion
  "description": "Publish multiple NPM workspaces (that's all)",
```

boneskull:
@legobeat I've made the changes we discussed

package.json
```diff
@@ -30,9 +30,8 @@
     "release:publish": "node scripts/publish.js --newPkg=${npm_config_newpkg}",
     "release:rebuild": "npm run rebuild && npm run test:prep && npm run test:workspaces",
     "setup": "husky install; npm run rebuild",
-    "test": "npm run build && npm run test:prep && npm run test:workspaces && npm run test:scripts",
```
legobeat: The `scripts/publish.js` is still used by the LavaMoat project for package publishing - it looks premature to remove tests scripts and disable tests in CI as long as that is the case?

Seems like this should be part of #903

boneskull: yes, that looks like I missed it

boneskull:
@legobeat I'm not planning on changing the build matrix so that we can skip Node.js v16 on a particular repo, since I can't think of an elegant solution (after admittedly not trying very hard).  Two reasons:

1. I don't want to have to manually maintain a list of which workspaces get tested against what versions of Node.js
2. It's not possible to simply skip a job step based on the contents of the matrix (e.g., the node version).  

    > [see docs](https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability); we need the `matrix` context to do this, and `job[job_id].if` does not have access to the `matrix` context

That leaves me something like:

- write a bespoke test script to check the `engines` field of a workspace and decide whether to invoke `ava` or not ||
- (set some CI env var and change `laverna`'s `test` script to skip if present &&
- create a new job that invokes running on node.js v18/v20 only and on `laverna`)

As I said, not so elegant, though that bespoke script probably could handle everything.

boneskull:
@legobeat ping

boneskull:
@legobeat I think I've got everything here...

packages/laverna/ava.config.js
```diff
@@ -0,0 +1,13 @@
+// if `mock` is not present in the `node:test` package, `laverna.spec.js` should be skipped
+
+const config = {
+  files: ['test/cli.spec.js'],
+}
+
+try {
+  if (require('node:test').mock) {
+    config.files.push('test/laverna.spec.js')
+  }
```
naugtur, line 12: I like this solution. Would be nice to make this more explicit.
```suggestion
  } else {
    console.error('[SKIP] Laverna tests skipped - too old Node.js version')
  }
```

naugtur:
While we still want to avoid breaking older Node.js support for our users, none of us are using it. 
We are the users of laverna. It's ok to skip testing it in old Node.js versions. 

With the script testing cleanup moved to the other PR, I think this is ready to merge.

naugtur (approved):

package.json
```diff
@@ -96,6 +96,7 @@
   "prettier": {
     "jsdocCommentLineStrategy": "keep",
     "jsdocPreferCodeFences": true,
+    "jsdocTagsOrder": "{\"template\": 24.5}",
```
naugtur, line 99: I don't understand this and therefore fear it. I'm a simple man. Please help me understand why we want this here.

scripts/publish.js
```diff
@@ -315,7 +339,13 @@ const Publish = {
...
                     // this could also be handled in a catch() chained to the `exec` promise
                     if ('stdout' in err) {
-                      /** @type {{error: {code: string, summary: string, detail: string}}} */
+                      /** @type {{
+  error: {
+    code: string
+    summary: string
+    detail: string
+  }
+}} */
```
naugtur, line 348: prettier went bonkers?

packages/laverna/test/fixture/new-version/packages/lavamoat/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "lavamoat",
+  "version": "31337.420.69"
```
naugtur, line 3: this random number generator, so mainstream

naugtur:
> TS errors in tests:
> 
> ```
> Error: packages/laverna/test/laverna.spec.js(24,44): error TS2694: Namespace '"node:test"' has no exported member 'Mock'.
> Error: packages/laverna/test/laverna.spec.js(25,35): error TS2694: Namespace '"node:test"' has no exported member 'Mock'.
> Error: packages/laverna/test/laverna.spec.js(419,6): error TS7006: Parameter 'call' implicitly has an 'any' type.
> Error: packages/laverna/test/laverna.spec.js(422,16): error TS7006: Parameter 'arg' implicitly has an 'any' type.
> Error: packages/laverna/test/laverna.spec.js(506,6): error TS7006: Parameter 'call' implicitly has an 'any' type.
> Error: packages/laverna/test/laverna.spec.js(508,21): error TS7006: Parameter 'arg' implicitly has an 'any' type.
> Error: packages/laverna/test/laverna.spec.js(530,6): error TS7006: Parameter 'call' implicitly has an 'any' type.
> Error: packages/laverna/test/laverna.spec.js(532,21): error TS7006: Parameter 'arg' implicitly has an 'any' type.
> Error: packages/laverna/test/laverna.spec.js(693,6): error TS7006: Parameter 'call' implicitly has an 'any' type.
> Error: packages/laverna/test/laverna.spec.js(696,16): error TS7006: Parameter 'arg' implicitly has an 'any' type.
> Error: packages/laverna/test/laverna.spec.js(756,6): error TS7006: Parameter 'call' implicitly has an 'any' type.
> Error: packages/laverna/test/laverna.spec.js(758,21): error TS7006: Parameter 'arg' implicitly has an 'any' type.
> Error: packages/laverna/test/laverna.spec.js(788,6): error TS7006: Parameter 'call' implicitly has an 'any' type.
> Error: packages/laverna/test/laverna.spec.js(791,16): error TS7006: Parameter 'arg' implicitly has an 'any' type.
> Error: Process completed with exit code 2.
> ```

This is marked as the change request blocking the merge and also as resolved. Should we consider it done?

package.json
boneskull: heh, well, I would have explained myself, but it's `package.json`, so.

https://github.com/hosseinmd/prettier-plugin-jsdoc/issues/226

scripts/publish.js
boneskull: hmm, changes to this file should not be in this PR.  I'll pull that out

PR #866, opened 2024-01-17
chore(ci): update labelbot docs glob

leotm:
Since root repo docs only living in root dir

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -88,7 +88,7 @@
   - changed-files:
       - any-glob-to-any-file:
           - '**/*.md'
-          - '**/docs/**'
+          - 'docs/**'
           - '**/example?(s)/**'
       - all-globs-to-all-files:
           - '!**/snapshots/**/*.md'


.github/labeler.yml
```diff
@@ -88,7 +88,7 @@
   - changed-files:
       - any-glob-to-any-file:
           - '**/*.md'
-          - '**/docs/**'
+          - 'docs/**'
```
leotm, line 91: or remove it and let all .md's be picked up from

https://github.com/LavaMoat/LavaMoat/blob/716ced82813c94afdd255d7db443475959488203/.github/labeler.yml#L90

boneskull: The intent here was to pick up _any_ markdown file (except snapshots) and any present or future docs.

leotm: agreed ^ i think `docs/**` still makes sense for our repo (like in the [examples](https://github.com/actions/labeler/blob/main/README.md))

Screenshot 2024-01-18 at 3 51 12 pm

but if we plan on recursive docs folders in future (e.g. `packages/aa/docs/newDoc.md`) then i'll go ahead and close this pr

boneskull: yeah, it's just so we don't need to change it later, I guess.  tbh I don't see any harm in leaving it as-is

PR #867, opened 2024-01-17
fix: cache regular expressions to avoid OOM error

### LINKED ISSUE #21006: Build error: RegExpCompiler Allocation failed - process out of memory
We have had some CI failures as follows:

Screenshot from 2023-09-22 10-56-36

@Gudahtt commented:

> This comment suggests that there is a size limit in v8 for RegExp code objects, of 128 MB: https://bugs.chromium.org/p/v8/issues/detail?id=10441#c5
> Apparently this error can be encountered when your application uses RegExps that build up in memory without getting garbage collected. Solutions are to use smaller RegExps or take precautions to ensure they can be garbage collected sooner. The garbage collection explains why it's intermittent.

We should investigate and fix at some point, because this hurts development velocity

davidmurdoch:
Fixes https://github.com/MetaMask/metamask-extension/issues/21006 

Well, maybe. Kind of.

It's probably just a sloppy workaround. I suspect there is a memory leak somewhere in LavaMoat, or maybe just in the way MM uses LavaMoat.

(needs tests)


packages/core/src/sourceTransforms.js
```diff
@@ -12,9 +12,11 @@ function applySourceTransforms(source) {
   ])
 }
 
+// create the RegExp once to avoid creating it over and over again
+// helps work around https://github.com/MetaMask/metamask-extension/issues/21006
+const someDirectEvalPattern = new RegExp('\\beval(\\s*\\()', 'g')
```
boneskull: this looks like it should fail linting because L19 would need to move here as well.

but! why do we need to use a regex literal here? @kumavis ?

naugtur: - I don't see why we'd need a literal
- the only concern with a `/g` regex is it maintians an offset marker, so there's one pattern of usage that will not work if the regex is cached. 

Tests might not be complex enough to surface the issue. I'd rather go and refresh my memory of the behavior from documentation and see if the usage pattern is somewhere here below.

If we want to cover for it with blunt force, ` regexp.lastIndex = 0` on line 17 would do it I think.

packages/core/src/kernelCoreTemplate.js
```diff
@@ -70,6 +70,8 @@
     const { prepareCompartmentGlobalFromConfig } = templateRequire('makePrepareRealmGlobalFromConfig')({ createFunctionWrapper })
     const { strictScopeTerminator } = templateRequire('strict-scope-terminator')
 
+    // cache regular expressions to work around https://github.com/MetaMask/metamask-extension/issues/21006
+    const regexCache = new Map()
```
legobeat, line 74: caching could potentially have the opposite effect than intended if it prevents GC by holding on to references. So I guess we want some profiling and/or benchmarks confirming this cache as at least not detrimental first?

The change in `sourceTransforms.js` is more straight-forward and shouldn't have any potential for adverse side-effects.

packages/core/src/sourceTransforms.js
legobeat: Moving out `someDirectEvalPattern` and `replaceFn` to the outer-scope (so L21 should move up, too) seems straight-forward to me and should be a strict improvement with minimal risk. Looks like L19 can be removed at this point given the absence of any linting error here?

fwiw i also considered if we should make the regex unicode-aware as well but neither `/u` nor `/v` make a difference wrt how `\s` is interpreted so I don't think that is relevant here.

Doing further improvements to how this is done can be done separately from such an optimization, I think.

boneskull: Sorry, I should have wrote "why _can't_ we use a literal"

legobeat: Given that the expression is static, defaulting to literals rather than constructor-with-string should be preferred from a performance-perspective as it compiles the expression on evaluation rather than on runtime.

If the instantiation of the RegExp is indeed accounting for redundant memory use, it could be that simply switching form without moving the line could be a mitigation in itself.

legobeat: Would be great to see the impact of the two measures indepdendently - maybe the cache isn't needed.

- Broke this part out as #909

packages/core/src/kernelCoreTemplate.js
naugtur: the keys in this map are coming from configuration supplied by humans, so it's unlikely to be of notable size.

naugtur (approved):

boneskull:
Original issue closed: https://github.com/MetaMask/metamask-extension/issues/21006

PR #875, opened 2024-01-17
chore(deps): override @types/node

boneskull:
chore(deps): override @types/node

This makes all `@types/node` versions use the one pinned in the workspace root `package.json`.

chore(yarn-plugin-allow-scripts): remove dev deps

These are installed from the workspace root.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@commitlint/cli": "17.8.1",
         "@commitlint/config-conventional": "17.8.1",
         "@tsconfig/node16": "16.1.1",
-        "@types/node": "20.11.5",
+        "@types/node": "18.19.8",
         "@types/yargs": "17.0.32",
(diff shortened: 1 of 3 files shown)


boneskull:
* `main`
  * **PR #875** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/875?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/875?utm_source=stack-comment).

legobeat (approved):
LGTM!

I think we should additionally drop down the `@types/node` package to 18 as long as we support it in `engines.node` (as it should be the lowest common denominator - we don't want to silently make use of anything present in 20 but not 18):
- #887

legobeat (approved):

PR #887, opened 2024-01-18
chore(deps): downgrade @types/node from 20 to 18

legobeat:
Drop typings package to align with lowest supported major Node.js version.

#### Depends on
- [ ] #875


boneskull (approved):
yeah, probably a good idea.

boneskull:
I've done this in #875, so this can be closed

PR #900, opened 2024-01-18
chore(ci): do not auto-rebase onto main

boneskull:
Renovate was configured to automatically rebase all its PRs onto `main` whenever `main` gets pushed to.  This causes a lot of workflows to get dumped into the CI queue.

This disables that behavior.  I am not sure if it will still automatically attempt to resolve conflicts if things get _too_ out-of-date, but we will see.


legobeat (approved):
This means more manual attention and time is needed (clicking that checkbox) for not having renovate PRs go stale and out of sync. I guess it won't do anything unprompted to existing PRs as long as they are still relevant after this.

Doesn't seem like an unreasonable tradeoff, though.

PR #909, opened 2024-01-19
perf(core): optimize RegExp allocation and compilation in sourceTransforms

legobeat:
Resource-usage optimization targeting regex-transforms in core.

- Switching to literal means static compilation instead of dynamically` compiling expression at runtime
- Moving allocation to outer scope to possibly reduce redundant allocations
- Switching from template string to concatenation in replace function gives measurable runtime difference in local benchmarks

#### Related
- Broken out from: #867 
- Motivation: https://github.com/MetaMask/metamask-extension/issues/21006

---

Co-Authored by: David Murdoch <187813+davidmurdoch@users.noreply.github.com>

### DIFF
--- a/packages/core/src/sourceTransforms.js
+++ b/packages/core/src/sourceTransforms.js
@@ -12,12 +12,12 @@ function applySourceTransforms(source) {
   ])
 }
 
+const DIRECT_EVAL_REPLACE_FN = (_, p1) => '(0,eval)' + p1
 function evadeDirectEvalExpressions(source) {
-  /* eslint-disable-next-line prefer-regex-literals */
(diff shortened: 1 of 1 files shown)


naugtur (approved):

naugtur:
It's a nice bit of cleanup, but I don't see how it would address the memory issue from the motivation link.

legobeat:
> It's a nice bit of cleanup, but I don't see how it would address the memory issue from the motivation link.

from pr description:

> Switching from template string to concatenation in replace function gives measurable runtime difference in local benchmarks

That is, the switch from `new RegExp("expression")` to `/expression/` reduces the number of compilations and thereby temporary allocations in need of GC.

PR #918, opened 2024-01-19
fix(deps): Align and integrity-check cdn resources in perf trials

legobeat:
- chore: add [`integrity`](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) and [`crossorigin`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin) attributes to js and stylesheets loaded from external CDN
- fix(deps): bump d3 from v3 to v5 (latest supported by c3)

c3 is unmaintained and if updating d3 for the perf trials further is actually desired, it would probably have to be replaced with something else, like [`billboard.js`](https://github.com/naver/billboard.js)

- Replaces #917

### DIFF
--- a/packages/perf/trials/bify-simple/chart.html
+++ b/packages/perf/trials/bify-simple/chart.html
@@ -9,7 +9,9 @@
       type="text/css"
     />
     <script
-      src="https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.17/d3.min.js"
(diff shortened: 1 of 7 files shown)


boneskull:
@kumavis pls help

boneskull:
I'd approve this, but I'm still waiting for that CORS talk from @naugtur

PR #959, opened 2024-01-26
feat(webpack): policy generation - experimental

naugtur:
- implemented policy generation using functionality available in core
- implemented policy reading from traditional files
- not a breaking change for the plugin, because the option to pass raw policy as options remains
- using it will be practical as soon as we introduce ESM support to tofu

Remaining issue: 
https://github.com/LavaMoat/LavaMoat/actions/runs/7986482849/job/21806989771?pr=959#step:5:1162
esbuild is missing a mac binary. I'm probably going to ditch esbuild in favor of babel to get it over with. It's not really important which one is building it anymore.


packages/webpack/src/plugin.js
boneskull, line 1: Why not the following for the `Compilation` import? 

```js
const {Compilation} = require('webpack')
```

it fixes the type error

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -0,0 +1,130 @@
+// TODO: scope it properly to compilation
+// const { parse, inspectGlobals } = require('lavamoat-tofu')
+const {
+  createModuleInspector,
+  LavamoatModuleRecord,
+  loadPoliciesSync,
+} = require('lavamoat-core')
+const { getPackageNameForModulePath } = require('@lavamoat/aa')
+const { writeFileSync, mkdirSync, existsSync } = require('fs')
```
boneskull:
```suggestion
const { writeFileSync, mkdirSync, existsSync } = require('node:fs')
```

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -0,0 +1,130 @@
...
+const path = require('path')
```
boneskull:
```suggestion
const path = require('node:path')
```

packages/core/src/loadPolicy.js
boneskull, line 1: I don't like that `async loadPolicy()` now calls sync APIs.  If we want to provide both sync and async APIs, we should provide both.

What's the reason for doing it sync, anyhow?  Did I ask this already?

packages/webpack/examples/webpack.config.js
boneskull, line 1: this example should probably have lots of comments explaining what does what

packages/webpack/package.json
```diff
@@ -38,6 +38,7 @@
   },
   "dependencies": {
     "@lavamoat/aa": "^4.0.1",
+    "json-stable-stringify": "1.1.1",
```
boneskull: add `@types/json-stable-stringify` as a dev dep.  assuming we don't expose its public API, it does not need to be a prod dep

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * @param {Object} opts
+   * @param {import('../types.js').Policy} [opts.policyFromOptions] - The
+   *   hardcoded policy passed in options, takes precedence over reading from
+   *   files
+   * @param {import('@lavamoat/aa').CanonicalNameMap} opts.canonicalNameMap -
+   *   Generated from aa
+   * @param {import('webpack').Compilation} opts.compilation - Webpack
+   *   compilation reference (for emitting assets)
+   * @param {boolean} opts.enabled - Whether to generate a policy
```
boneskull, line 27: this is required?

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -0,0 +1,130 @@
...
+          }
+          if (emit) {
+            compilation.emitAsset(
+              POLICY_SNAPSHOT_FILENAME,
+              new RawSource(stringify(final, { space: 2 }))
+            )
+          }
+          return final
+        },
+      }
```
boneskull: so if policy gen is not enabled, then we return this, which just loads whatever policy was loaded?

if `policyFromOptions` is provided, do we really need to call `loadPoliciesSync` on L42?

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -0,0 +1,130 @@
...
+            // connections are a much better source of information than module.dependencies which contain
+            // all imported references separately along with exports and fluff
+            ...Array.from(connections).reduce((acc, dep) => {
+              // @ts-expect-error - bad types?
+              const depSpecifier = dep.resolvedModule.userRequest
+              acc[depSpecifier] = depSpecifier
+              return acc
+            }, /** @type {Record<string, string>} */ ({})),
+          },
+          //ast: module._ast, - would have to translate to babel anyway
```
boneskull, line 105: remove?

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -0,0 +1,130 @@
...
+        })
+
+        moduleInspector.inspectModule(moduleRecord)
+      },
+      getPolicy: () => {
+        const policy = moduleInspector.generatePolicy({})
+        if (!existsSync(location)) {
+          mkdirSync(location, { recursive: true })
+        }
```
boneskull: no need to check

```suggestion
        mkdirSync(location, { recursive: true })
```

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -0,0 +1,130 @@
...
+        writeFileSync(
```
boneskull, line 113: missing an encoding?

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -0,0 +1,130 @@
...
+          path.join(location, 'policy.json'),
+          stringify(policy, { space: 2 })
+        )
+        const final = applyOverride(policy)
+        if (emit) {
+          compilation.emitAsset(
+            POLICY_SNAPSHOT_FILENAME,
```
boneskull, line 121: maybe this should be configurable

packages/webpack/src/plugin.js
```diff
@@ -22,6 +23,7 @@ const JAVASCRIPT_MODULE_TYPE_AUTO = 'javascript/auto'
 const JAVASCRIPT_MODULE_TYPE_DYNAMIC = 'javascript/dynamic'
 const JAVASCRIPT_MODULE_TYPE_ESM = 'javascript/esm'
 
+// @ts-ignore
```
boneskull, line 25: this is solvable by modifying `includes` to explicitly pull in `.json` files _and_ adding `resolveJsonModule: true` to `compilerOptions`

packages/webpack/src/plugin.js
```diff
@@ -55,10 +57,13 @@ class LavaMoatPlugin {
   /**
    * @param {import('./types.js').LavaMoatPluginOptions} [options]
    */
-  constructor(options = { policy: {} }) {
+  constructor(options = {}) {
     if (!options.lockdown) {
       options.lockdown = lockdownDefaults
     }
+    if (!options.policyLocation) {
+      options.policyLocation = './lavamoat/webpack'
```
boneskull: not portable

packages/webpack/src/plugin.js
```diff
@@ -139,6 +145,16 @@ class LavaMoatPlugin {
           )
         }
 
+        // =================================================================
```
boneskull, line 155: what is the purpose of this drawing

packages/webpack/src/plugin.js
```diff
@@ -161,6 +177,12 @@ class LavaMoatPlugin {
           JAVASCRIPT_MODULE_TYPE_ESM,
         ]
 
+        /**
+         * @param {import('webpack').Module} m
+         * @returns {m is import('webpack').NormalModule}
+         */
+        const isNormalModule = (m) => 'resource' in m
```
boneskull, line 214: 👍

packages/webpack/src/plugin.js
```diff
@@ -354,9 +387,10 @@ class LavaMoatPlugin {
         )
 
         const HtmlWebpackPluginInUse = compiler.options.plugins.find(
-          (plugin) => plugin.constructor.name === 'HtmlWebpackPlugin'
+          (plugin) => plugin && plugin.constructor.name === 'HtmlWebpackPlugin'
```
boneskull, line 419: or

```suggestion
          (plugin) => plugin?.constructor.name === 'HtmlWebpackPlugin'
```

packages/webpack/test/e2e-policy-gen.spec.js
```diff
@@ -0,0 +1,63 @@
+const test = require('ava')
```
boneskull, line 1: if you're interested in adding types for the execution context, an example is in `publish.spec.js`:  https://github.com/LavaMoat/LavaMoat/blob/naugtur/weboack-policy-generation/scripts/publish.spec.js#L53

packages/webpack/test/e2e-policy-gen.spec.js
```diff
@@ -0,0 +1,63 @@
...
+test.before(async (t) => {
+  const webpackConfig = makeConfig({
+    generatePolicy: true,
+    emitPolicySnapshot: true,
+    diagnosticsVerbosity: 1,
+  })
+  await t.notThrowsAsync(async () => {
+    t.context.build = await scaffold(webpackConfig)
+  }, 'Expected the build to succeed')
+  t.context.bundle = t.context.build.snapshot['/dist/app.js']
```
boneskull, line 15: I wonder if this is portable

packages/webpack/test/e2e-policy-gen.spec.js
```diff
@@ -0,0 +1,63 @@
...
+test('webpack/policy-gen - bundle runs without throwing', (t) => {
+  t.notThrows(() => {
+    runScriptWithSES(t.context.bundle)
+  })
+})
+
+test('webpack/policy-gen - policy-overrides get applied', async (t) => {
+  const webpackConfig = makeConfig({
+    emitPolicySnapshot: true,
+    policyLocation: path.resolve(__dirname, 'fixtures/main/policy-broken'),
```
boneskull, line 31: portability

packages/webpack/src/types.js
```diff
@@ -13,7 +18,8 @@
  *   output for lockdown.js if HtmlWebpackPlugin is in use
  * @property {number} [diagnosticsVerbosity] - A number representing diagnostics
  *   output verbosity, the larger the more overwhelming
- * @property {Object} policy - LavaMoat policy
+ * @property {Policy} [policy] - LavaMoat policy object - if programmaticly
```
boneskull, line 23:
```suggestion
 * @property {import('lavamoat-core').LavaMoatPolicy} [policy] - LavaMoat policy object - if programmatically
```

packages/webpack/src/plugin.js
```diff
@@ -206,13 +238,19 @@ class LavaMoatPlugin {
           diag.rawDebug(4, { knownPaths })
           PROGRESS.report('pathsCollected')
 
+          diag.rawDebug(2, 'writing policy')
+          // use the generated policy to save the user one additional pass
+          // getting the policy also writes all files where necessary
+          options.policy = policyGenerator.getPolicy()
```
boneskull: does it matter if we're mutating `options`

packages/webpack/src/plugin.js
```diff
@@ -139,6 +145,16 @@ class LavaMoatPlugin {
           )
         }
 
+        // =================================================================
+        const policyGenerator = createPolicyGenerator({
+          policyFromOptions: options.policy,
+          enabled: !!options.generatePolicy,
```
boneskull, line 158: (style)

```suggestion
          enabled: Boolean(options.generatePolicy),
```

packages/webpack/test/fixtures/main/webpack.config.js
```diff
@@ -3,77 +3,83 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin')
...
+  diagnosticsVerbosity: 0,
+  HtmlWebpackPluginInterop: true,
+}
+
+module.exports = makeConfig()
+
+Object.defineProperty(module.exports, 'makeConfig', {
+  enumerable: false,
+  value: makeConfig,
+})
```
boneskull, line 25: why

packages/webpack/src/buildtime/emitSes.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
     () => {
       // Read the file you want to add to the output
       const sesFile = readFileSync(require.resolve('ses'), 'utf-8')
+      // TODO: instead manually copy to compiler.options.output.path
```
boneskull: why

boneskull (commented):
- This looks like it wants to be rebased onto `main`?
- You can remove all of the `// @ts-check` directives and put `checkJs: true` under `compilationOptions` in its `tsconfig.json`.  There are many type errors, but since we're not generating declaration files for the plugin, it does not affect the build.

packages/webpack/src/plugin.js
naugtur: I had trouble importing it earlier. :shrug: 
fixed

packages/core/src/loadPolicy.js
naugtur: I needed to provide loadPoliciesSync so I turned it into sync implementation. async policy loading would make integrating with webpack harder. Didn't feel like duplicating the code or finding a good async location in webpack. Not sure if it's worth the effort. 
loading policy tends to happen once per process, so I didn't think it matters.

packages/webpack/examples/webpack.config.js
naugtur: yes. eventually

packages/webpack/src/buildtime/policyGenerator.js
naugtur: yes, I defined it to be explicitly specified when calling generator. all levels above are not required. I could change it, but it made sense at the time.

packages/webpack/src/buildtime/policyGenerator.js
naugtur: correct. I'd need to rearrange the module and my brain is falling into an endless loop when I try. Will do.

packages/webpack/package.json
naugtur: done

packages/webpack/src/buildtime/policyGenerator.js
naugtur: ok

packages/webpack/src/buildtime/policyGenerator.js
naugtur: You can decide if you want the emit. It's for humans to read. it does not end up used.
I applied convention over configuration.

packages/webpack/src/plugin.js
naugtur: fixed

packages/webpack/src/plugin.js
naugtur: splits the implementation into bits for faster navigation. saves me time when scrolling. May replace with meaningful comments eventually.

packages/webpack/src/plugin.js
naugtur: learned that from you

packages/webpack/src/buildtime/emitSes.js
naugtur: because I didn't find a way to prevent it from being minified :(

packages/webpack/src/plugin.js
naugtur: I like `!!` better. If prettier didn't fix it, maybe it's fine?

packages/webpack/src/plugin.js
naugtur: not sure what you mean

packages/webpack/test/e2e-policy-gen.spec.js
naugtur: I've got a pending ugly merge in tests here, not gonna make it harder for now.

packages/webpack/test/fixtures/main/webpack.config.js
naugtur: so that the file is consumable from webpack cli and also programatically in tests

packages/webpack/src/buildtime/emitSes.js
naugtur: todo: try telling webpack it's not a source file

packages/webpack/src/plugin.js
naugtur: ok, I noticed I probably don't need to write it back to options anyway

packages/webpack/src/buildtime/emitSes.js
naugtur: it seems webpack is choosing to minify it by extension :disappointed:

naugtur: current attempt: fix it at the cost of user astonishment

packages/webpack/src/plugin.js
weizman: Personally I'd consider changing generatePolicy to shouldGeneratePolicy to indicate it's a bool but mainly to tell it apart from other instances of generatePolicy function in the codebase which are two different things (I was confused by that browsing through the code)

packages/webpack/src/plugin.js
```diff
@@ -55,10 +62,13 @@ class LavaMoatPlugin {
   /**
    * @param {import('./types.js').LavaMoatPluginOptions} [options]
    */
-  constructor(options = { policy: {} }) {
+  constructor(options = {}) {
     if (!options.lockdown) {
       options.lockdown = lockdownDefaults
     }
+    if (!options.policyLocation) {
+      options.policyLocation = path.join('.', 'lavamoat', 'webpack')
```
weizman, line 70: Do we need the default value generation both [here](https://github.com/LavaMoat/LavaMoat/pull/959/files#diff-9647c300f69dfab2af75d7433e10ed31a03f45408883fcea1284b90c35f5a7ddR70) and [here](https://github.com/LavaMoat/LavaMoat/pull/959/files#diff-94c0d6243364db60753823cb2b811194dc7fcba7610f63ec9b5b691353457783R38)? Can we settle for just one?

packages/webpack/README.md
```diff
@@ -11,7 +11,9 @@ LavaMoat Webpack Plugin wraps each module in the bundle in a [Compartment](https
 1. Generate a policy file for your app's entry `npx lavamoat --autopolicy path/to/index.js` and adjust it if needed.
 2. Note that `builtins` in the policy indicate modules that webpack will either polyfill or ignore, so they need to be moved to `packages` and explicitly added as dependencies or just removed.
 3. Create a webpack bundle with the LavaMoat plugin enabled and pass the policy from the policy file to it in options.
-4. Make sure you add a `<script src="./lockdown.js"></script>` before all other scripts or enable the `HtmlWebpackPluginInterop` option if you're using `html-webpack-plugin`.
+4. Make sure you add a `<script src="./lockdown"></script>` before all other scripts or enable the `HtmlWebpackPluginInterop` option if you're using `html-webpack-plugin`.
+
+> The pligin is emitting lockdown without the `.js` extension because that's the only way to prevent it from getting minified, which is likely to break it.
```
weizman:
```suggestion
> The plugin is emitting lockdown without the `.js` extension because that's the only way to prevent it from getting minified, which is likely to break it.
```

packages/webpack/README.md
```diff
@@ -11,7 +11,9 @@ LavaMoat Webpack Plugin wraps each module in the bundle in a [Compartment](https
 1. Generate a policy file for your app's entry `npx lavamoat --autopolicy path/to/index.js` and adjust it if needed.
```
weizman: line 9 (above) can be removed with this PR, right?

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -0,0 +1,129 @@
...
+   * @param {boolean} [opts.emit] - Whether to emit the policy snapshot as an
+   *   asset
+   * @returns
+   */
+  createPolicyGenerator({
+    policyFromOptions,
+    canonicalNameMap,
+    compilation,
+    enabled,
+    location = './lavamoat/webpack',
```
naugtur: TODO: remove

packages/webpack/src/plugin.js
naugtur: we don't. removing one.

packages/webpack/src/runtime/runtime.js
```diff
@@ -45,9 +44,19 @@ const stricterScopeTerminator = freeze(
...
+ * Enforces the policy for resource imports.
+ *
+ * @param {string | undefined} requestedResourceId - The ID of the requested
+ *   resource.
+ * @param {string} referrerResourceId - The ID of the referrer resource.
+ * @throws {Error} Throws an error if the policy does not allow importing the
+ *   requested resource from the referrer resource.
+ */
 const enforcePolicy = (requestedResourceId, referrerResourceId) => {
+  if (typeof requestedResourceId === 'undefined') {
```
weizman, line 57: why not:
```suggestion
  if (requestedResourceId === undefined) {
```
?

weizman (approved):
makes sense

packages/webpack/src/runtime/runtime.js
naugtur: old habits - assumption `undefined` can be overridden

PR #964, opened 2024-01-27
normalize test logging

boneskull:
This was bugging me.  It's tough to visually parse the output of tests, especially when we're running scenarios.  I went through and changed all of the console logging coming out of tests and test harnesses to use Ava's facilities to do so, which logs in a more consistent and readable manner.

Examples. Before:

```
  ✔ waysToRun › use lavamoat-run-command (439ms)
  ✔ index › execute - resolutions (424ms)
  ✔ envConfig › envConfig - intrinsic prototype mutating package running in unfrozen realm does not pollute other package intrinsics (442ms)
  ✔ index › execute - keccak with native modules (424ms)
  ✔ envConfig › envConfig - module.exports from the same Realm as its Compartment (443ms)
Running Node Scenario: security - prevent intrinsic modification
  ✔ globals › globalRef - globalRef - check default containment (454ms)
  ✔ globals › globals - has only the expected global circular refs (459ms)
Running Node Scenario: security - limit platform API
Running Node Scenario: basic - bundle works
Running Node Scenario: basic - npm buffer toLocaleString support
Running Node Scenario: basic - Function constructor for constructing constructor functions
Running Node Scenario: config - dunder proto not allowed in globals path
Running Node Scenario: config - disable access to package
Running Node Scenario: config - config validation fails: invalid "resources" key
Running Node Scenario: config - config validation fails: invalid "packages" key
Running Node Scenario: config - config validation fails: invalid "globals" key
Running Node Scenario: config - config validation fails: invalid "globals" value
Running Node Scenario: config - config validation passes - everything valid
Running Node Scenario: endowments - Date.now works in root and non-root
Running Node Scenario: exportsDefense - readOnly restrictions have override workaround fix - updates the property correctly
```

after (note the order):

```
  ✔ waysToRun › use lavamoat-run-command (416ms)
  ✔ runScenarios › Run scenarios (5.5s)
    ℹ Running Node Scenario: autogen - react-devtools-core hasOwnProperty
    ℹ Running Node Scenario: security - prevent intrinsic modification
    ℹ Running Node Scenario: security - limit platform API
    ℹ Running Node Scenario: basic - bundle works
    ℹ Running Node Scenario: basic - npm buffer toLocaleString support
    ℹ Running Node Scenario: basic - Function constructor for constructing constructor functions
    ℹ Running Node Scenario: config - dunder proto not allowed in globals path
    ℹ Running Node Scenario: config - disable access to package
    ℹ Running Node Scenario: config - config validation fails: invalid "resources" key
    ℹ Running Node Scenario: config - config validation fails: invalid "packages" key
    ℹ Running Node Scenario: config - config validation fails: invalid "globals" key
    ℹ Running Node Scenario: config - config validation fails: invalid "globals" value
    ℹ Running Node Scenario: config - config validation passes - everything valid
    ℹ Running Node Scenario: endowments - Date.now works in root and non-root
    ℹ Running Node Scenario: exportsDefense - readOnly restrictions have override workaround fix - updates the property correctly
```

Importantly, because Ava runs tests in child processes/workers, using `console` directly means the output is printed nondeterministically.  Using Ava's `t.log()`, the output is now nested beneath the test in question, and is always in order.

### DIFF
--- a/packages/allow-scripts/test/index.spec.js
+++ b/packages/allow-scripts/test/index.spec.js
@@ -31,7 +31,7 @@ test('cli - auto command', (t) => {
   )
 
   // forward error output for debugging
-  console.error(result.stderr.toString('utf-8'))
+  t.log(result.stderr.toString('utf-8'))
 
   // get the package.json
   const packageJsonContents = JSON.parse(
@@ -63,7 +63,7 @@ test('cli - auto command with experimental bins', (t) => {
   )
 
   // forward error output for debugging
-  console.error(result.stderr.toString('utf-8'))
+  t.log(result.stderr.toString('utf-8'))
 
   // get the package.json
   const packageJsonContents = JSON.parse(
@@ -101,7 +101,7 @@ test('cli - run command - good dep at the root', (t) => {
   )
 
   // forward error output for debugging
-  console.error(result.stderr.toString('utf-8'))
+  t.log(result.stderr.toString('utf-8'))
 
   // assert the output
   t.deepEqual(result.stdout.toString().split('\n'), [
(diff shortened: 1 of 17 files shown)


legobeat (approved):
LGTM

PR #971, opened 2024-01-30
fix(allow-scripts): add corepack to bannedBins

legobeat:
Part of Node.js bundled package manager management tooling.

https://nodejs.org/api/corepack.html

https://www.npmjs.com/package/corepack

### DIFF
--- a/packages/allow-scripts/src/index.js
+++ b/packages/allow-scripts/src/index.js
@@ -292,7 +292,7 @@ async function runScript({ path, event }) {
   })
 }
 
-const bannedBins = new Set(['node', 'npm', 'yarn', 'pnpm'])
+const bannedBins = new Set(['corepack', 'node', 'npm', 'pnpm', 'yarn'])
 
 /**
  * @param {BinCandidates} binCandidates


boneskull (commented):
Why? 🙂

legobeat:
> Why? 🙂

The idea of `bannedBins` is to prevent package dependencies to override bins that are part of global tooling and use that as a vector to get executed inadvertently. E.g. there is usally no valid reason why you'd want to allow a dep bin `npm` to override the globally installed one. I don't see anything different with `corepack` in this regard.

[origin](https://github.com/LavaMoat/LavaMoat/pull/390/files#diff-9cdd913a8c0d8d146ccb384ffffe54704d29dfa225c2b0db9bfc840c9e796d2cR248-R264)

@naugtur chime in if anything is incorrect in those assumptions

naugtur:
Makes sense to add this. Altho I'm not happy with how complex the overall implementation is and I'm not sure if I want this feature out of experimental.

naugtur (approved):

PR #973, opened 2024-01-30
fix(core): update LavamoatModuleRecord

boneskull:
- add missing `@babel/types` dep

This adds an optional `InitArgs` type argument that different implementations can use to define the arguments to the module initializer function.  For Node.js, that might be the type of `[module, require, exports, __dirname, __filename]` and for browserify/webpack, that's something else.

@kumavis @naugtur I'm not sure if my descriptions are correct; please let me know.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21852,6 +21852,7 @@
       "version": "15.1.1",
       "license": "MIT",
       "dependencies": {
+        "@babel/types": "7.23.6",
         "json-stable-stringify": "1.1.1",
         "lavamoat-tofu": "^7.1.0",
         "merge-deep": "3.0.3",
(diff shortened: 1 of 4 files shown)


naugtur:
I'm reading on the phone app, so apologies if I missed something. Seems like we're defaulting to empty array on init args. I'd rather default to the cjs module ones as they're common (pun!)

boneskull:
@naugtur You mean change the default _of the type argument_ to that, or the default _value in the constructor_ to that?  Because the latter would be a logic change.

naugtur:
Thought it would be a good idea, but let's not make breaking changes.

naugtur (approved):

PR #960, opened 2024-01-26
fix(browserify/test): result output references in error handling

legobeat:
Fixes regression in error-handling in browserify tests introduced in f4292a06c12bbed3e807cc0a2f85f42976255510 (#698) where a non-existent `res` is referred in error-handling.

Ideally cases like this would be caught by lint.

PR #974, opened 2024-01-30
chore: re-enable no-undef rule for tests

boneskull:
~~This needs #960 to pass CI, but I cannot target a branch in a fork.~~

`no-undef` was disabled because we use a lot of undefined variables, but it's better to create lint exceptions on a line-by-line or file-by-file basis, since it can cause the problem fixed in #960.

### DIFF
--- a/packages/browserify/test/globalRef.spec.js
+++ b/packages/browserify/test/globalRef.spec.js
@@ -1,4 +1,6 @@
-/* eslint-disable no-undef */
+// eslint-disable-next-line no-unused-vars
+/* global checkThis: true, checkSelf: true, checkWindow: true, checkGlobal: true */
+
 const test = require('ava')
 const { runScenario } = require('./util')
 const {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -55,9 +55,6 @@ module.exports = {
     // base rules; none of these are in eslint/recommended
     'no-empty': ['error', { allowEmptyCatch: true }],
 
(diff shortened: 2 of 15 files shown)


packages/browserify/test/globalRef.spec.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-undef */
```
legobeat: Rather than entirely disabling the rule here, how about listing the globals actually used?
```suggestion
/* global checkGlobal, checkSelf, checkThis, checkWindow */
```

boneskull: ahh, needs `/* global someVar: write */`: and then another `no-undef` disable above it 😄

boneskull:
* **#974** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/974?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/browserify/test/globalRef.spec.js
boneskull: this is what made it work:

```js
// eslint-disable-next-line no-unused-vars
/* global checkThis: true, checkSelf: true, checkWindow: true, checkGlobal: true */
```

legobeat (approved):

naugtur:
I'm skeptical (as in, all undef warnings will be ignored anyway as we move forward, just with added manual labor) 
But test should not be modified too often, so I don't mind.

PR #984, opened 2024-01-31
Fix viz

legobeat:
The `viz` package seems to not have been working properly since `6.2.0`/`6.0.12` (2023-09). This addresses regressions and should make it run as expected.

- Revert webpack 5 upgrade (#591) and associated plugins
  - Revert changes in `packages/viz` from the following commits:
    - 5ed273d449ddb1efdf93b4f21ba87df6623be29a
    - 24f34cc6879d51c61e9daf6fffad2751cbdd3a51
    - b850dc450f2ebc1e6b30eb7fc5584d3bd3f3a9b7
    - 54bbc4d9f3296207063fe07a33b1b8f6dbb6231f
    - 72577d652a1d87bed122d7a8ce936386b7362d67
  - Would be nice to actually do the webpack 5 upgrade but I got blocked on React not being resolved properly inside `react-force-graph-3d>react-kapsule`.
- Force UMD resolution for `force-graph` packages
  - This has become necessary after recent upgrades to these dependencies. A more correct/proper solution to reconfiguration likely exists.
- Import policies directly through `<script>` tag in html by templating `index.html` at runtime
  - Fixes compatibility with modern Firefox when run from local filesystem. Previous implementation didn't load due to CORS.
  - Rename `index.html` to `index.html.template` for clarity
  - This replaces generated `injectConfigDebugData.js`
- Stop distributing example policies and associated files from distributed
  - These make up ~100MB and I found their apparently mixed purpose confusing. If an example on npm registry is desired it can be distributed as a separate package like `@lavamoat/viz-examples`
- Remove package script `start` and devDependency `webpack-dev-server`
  - The script does not work OotB for some time now any way: it doesn't detect the hoisted `webpack-cli`

This will be followed up with a PR for the usage of viz in metamask-extension (which from what I can tell has been broken independently since 2021)


boneskull:
I'm sure you know this, but the downgrade introduces vulnerable deps.  Mitigating these was the goal of upgrading in the first place

packages/viz/package.json
```diff
@@ -65,13 +65,13 @@
...
+    "style-loader": "^2.0.0",
     "three": "0.160.1",
     "three-forcegraph": "1.41.10",
     "three-spritetext": "1.8.1",
-    "webpack": "5.90.0",
-    "webpack-cli": "5.1.4",
-    "webpack-dev-server": "4.15.1"
+    "webpack": "^4.46.0",
+    "webpack-cli": "^4.9.2",
+    "webpack-dev-server": "4.15.0"
```
boneskull: please pin all these deps (no `^`)

packages/viz/src/webpack.config.js
```diff
@@ -41,14 +41,30 @@ module.exports = {
...
+        '../../../node_modules/three-forcegraph/dist/three-forcegraph.js'
+      ),
+      'force-graph': path.resolve(
+        __dirname,
+        '../../../node_modules/force-graph/dist/force-graph.js'
+      ),
+      'react-force-graph-2d': path.resolve(
+        __dirname,
+        '../../../node_modules/react-force-graph-2d/dist/react-force-graph-2d.js'
+      ),
```
boneskull: this could use some comments as to why it's necessary.  easier to do this, assuming these are resolvable:

```suggestion
      'three-forcegraph': require.resolve('three-forcegraph/dist/three-forcegraph.js'),
      'force-graph': require.resolve('force-graph/dist/force-graph.js'),
      'react-force-graph-2d': require.resolve('react-force-graph-2d/dist/react-force-graph-2d.js'),
```

if that doesn't work, you will need to use `path.resolve` the same way, but omit directory seps for portability.  this is tedious, so maybe:

```suggestion
      'three-forcegraph': path.resolve(
        __dirname,
        ...['../../../node_modules/three-forcegraph/dist/three-forcegraph.js'].split('/')
      ),
      'force-graph': path.resolve(
        __dirname,
        ...['../../../node_modules/force-graph/dist/force-graph.js'].split('/')
      ),
      'react-force-graph-2d': path.resolve(
        __dirname,
        ...['../../../node_modules/react-force-graph-2d/dist/react-force-graph-2d.js'].split('/')
      ),
```

boneskull (commented):
- I can confirm `npm run build` works (with warnings), and the generated `dist/index.html` runs successfully in a browser.
- `npm start` is broken due to an invalid configuration option, seemingly.

What is our plan to mitigate the re-introduced vulnerable dependencies?

If we cannot get Webpack v5 working--which would mitigate the dependency issues--we should archive this project.

cc @naugtur @kumavis

legobeat:
- Corresponding PR for making this work in `metamask-extension`: https://github.com/MetaMask/metamask-extension/pull/22759

packages/viz/package.json
legobeat: Done

packages/viz/src/webpack.config.js
legobeat: If you want Windows support, you'll have to test for it yourself - can merge into this branch if you have something confirmed working. And `require.resolve` requires a little bit more. Added a comment, though!

(Keeping them as one line make them a lot more readable IMO but prettier disagrees)

naugtur:
(nitpick) this being in a fork is getting in the way

naugtur (approved):

naugtur:
Could retarget to https://github.com/boneskull/LavaMoat/tree/viz-only

PR #988, opened 2024-01-31
chore(browserify/test): regenerate fixture

legobeat:
Syncing `npm run test:prep` with changes from #973

### DIFF
--- a/packages/browserify/test/fixtures/secureBundling/lavamoat/node/policy.json
+++ b/packages/browserify/test/fixtures/secureBundling/lavamoat/node/policy.json
@@ -12,9 +12,45 @@
     },
     "@babel/code-frame>@babel/highlight": {
       "packages": {
-        "@babel/code-frame>@babel/highlight>@babel/helper-validator-identifier": true,
+        "@babel/code-frame>@babel/highlight>chalk": true,
         "@babel/code-frame>@babel/highlight>js-tokens": true,
-        "@babel/code-frame>chalk": true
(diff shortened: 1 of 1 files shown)


boneskull:
In this particular case, it doesn't look to me like the file needs to be under VCS at all:

1. `test:prep` gets run via the workspace root `npm test` script. This causes the policy file to be autogenerated
2. Tests run in `browserify`.  The `WRITE_AUTO_POLICY` env flag is not set, so the `run.sh` script expects the `policy.json` to exist.  It should exist, because we just ran `test:prep`.

So why is it under VCS?

boneskull:
I was also going to comment about why is `@babel/types` having _any effect whatsoever_ on the policy, but it looks like it's not actually a type-only module, which is...bizarre.

legobeat:
> I was also going to comment about why is `@babel/types` having _any effect whatsoever_ on the policy, but it looks like it's not actually a type-only module, which is...bizarre.

IMO we should consider moving type-only dependencies to devDeps (optionally also breaking out separate types-only package(s)). `@babel/types` only appears in `.d.ts` files and jsdoc comments and is not an actual runtime dependency.

But that's separate from here.

boneskull:
@legobeat I think @naugtur and I disagree with you on that point

boneskull:
Rather, it's not a _runtime_ dependency, but it's a _production_ dependency.

legobeat:
This was merged as part of #825

PR #990, opened 2024-01-31
chore(dirty-check): fix dirty check to ignore changes in trailing whitespace

legobeat:
- Targeting #825 

#### Changes
- Stage changes before diffing to make `.gitattributes` take effect
- Determine dirtyness by exit code rather than screen-scraping
- Sync fixture (cherry-picked from #988)

### DIFF
--- a/scripts/dirty-check.js
+++ b/scripts/dirty-check.js
@@ -38,31 +38,31 @@ const execFile = promisify(require('node:child_process').execFile)
  */
 async function main() {
   await execFile('git', ['add', '.'])
-  const { stdout } = await execFile(
+  const diffResultPromise = execFile(
     'git',
-    ['diff', '--staged', '-b', '--name-only'],
-    {
-      encoding: 'utf8',
-    }
+    ['diff', '--staged', '-b', '--quiet'],
+    { encoding: 'utf8' },
   )
 
-  const changedFiles = stdout.trim().split(/\r?\n/)
-
-  // if this is true, then something changed, and we should fail the build
-  if (changedFiles.length) {
+  try {
+    await diffResultPromise
+    console.log('::notice::No dirty files detected')
+  } catch (e) {
+    if (e.code !== 1 || e.killed || e.stderr !== '') {
+      throw e
+    }
+    // something changed, and we should fail the build
     process.exitCode = 1
 
-    for (const file of changedFiles) {
-      console.log(
-        '::error file=%s,title=Dirty File::%s changed unexpectedly; see job log for diff',
-        file,
-        file
-      )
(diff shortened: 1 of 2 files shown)


legobeat:
Hmm, this still appears in CI - I'd expected it to not diff:
```
Error: packages/node/test/projects/2/atob.policy.json changed unexpectedly; see job log for diff
diff --git a/packages/node/test/projects/2/atob.policy.json b/packages/node/test/projects/2/atob.policy.json
index 5366045..da96aa2 100644
--- a/packages/node/test/projects/2/atob.policy.json
+++ b/packages/node/test/projects/2/atob.policy.json
@@ -11,4 +11,4 @@
       }
     }
   }
-}
+}
\ No newline at end of file
```

legobeat:
The issue seems to be that `--name-only` and `--name-status` make other configuration of the diffing algorithm be ignored, so first invocation should switch to `--quiet` instead of screen-scraping (and thereby pushing identification of changed files to the second pass)

scripts/dirty-check.js
```diff
@@ -0,0 +1,73 @@
...
+  const diffResultPromise = execFile(
+    'git',
+    ['diff', '--staged', '-b', '--quiet'],
+    {
+      encoding: 'utf8',
+    }
+  )
+
+  try {
+    await diffResultPromise
```
boneskull: why split the promise?

scripts/dirty-check.js
```diff
@@ -0,0 +1,73 @@
...
+    if (e.code !== 1 || e.killed || e.stderr !== '') {
+      throw e
+    }
+    // something changed, and we should fail the build
+    process.exitCode = 1
+
+    console.log(
+      '::error,title=Dirty File::Working area contains unexpected changes; see job log for diff'
+    )
+    const { stdout } = await execFile('git', ['diff', '--staged', '-b'])
```
boneskull: 👍

boneskull: so wait: why stage everything and then diff staged files only?  what's the difference between that and what it was before?

boneskull: also pls use `encoding: 'utf8'` in `execFile` opts if we're expecting a string back

boneskull (approved):

boneskull:
@legobeat looks like you have a conflict to fix

scripts/dirty-check.js
legobeat: > so wait: why stage everything and then diff staged files only? what's the difference between that and what it was before?

To make `.gitattributes` apply (crlf line-ending-only changes won't show as dirty files in `git status`, `git diff --name-only`, etc)

PR #811, opened 2023-12-21
feat(node): add ESM policy generation

boneskull:
**NOTE** This does not actually need to be merged--as #772 does not depend upon it--but can rather serve as an example of how `endomoat` _could_ do things.

This does not add ESM support to `lavamoat`, but rather it enables creation of policies from projects containing ESM packages.

Each `.npmrc` adds the `install-links=true` option, which ensures symlinks do not get created when running `test:prepare`, as Node.js' internal module resolution differs from `resolve.sync`'s resolution in this regard

PR #994, opened 2024-02-05
feat(aa): support symlinks for workspace and devtime linking compatibility

naugtur:
Rebased on #808 _(but now I think it should have been the other way around :see_no_evil: )_

Packages from workspace ended up as `external:../name` because we did not recognize their final location after resolving links and some of the tools that asked for canonical names for locations were using the fully resolved paths. Instead of fighting the ecosystem for consistent link behavior, I propose we throw money (RAM) at it.

What does this fix?
Our policies are bloated with entries like:
`"external:../snaps-sdk/src/ui/components/panel.ts"` 
`"external:../snaps-sdk/src/ui/components/row.ts"` 
when a package from a workspace is being depended on. 
Resolving the symlink in `../../node_modules/@metamask/snaps-sdk` to `../snaps-sdk` will help when lookups end up asking for those locations. (more precisely these would be absolute paths for the most part) 

This should also be useful for workspace support in allow-scripts, but I did not verify. (@legobeat )

Sidenote:
This solution is expensive but general.
I can imagine a solution where we go up to the package.json of the workspace and read the workspace configuration and only add duplicate entries for the workspace items. 
That would be much more performant, but would only solve for workspace and not other cases where links are created (like local folder installations or other trickery)

Both approaches are worth considering.

### DIFF
--- a/packages/aa/src/index.js
+++ b/packages/aa/src/index.js
@@ -12,10 +12,19 @@ module.exports = {
   createPerformantResolve,
 }
 
+/** @type {Resolver|undefined} */
+let performantResolve
+
 /**
+ * performant resolve avoids loading package.jsons if their path is what's being
+ * resolved, offering 2x performance improvement compared to using original
+ * resolve
  * @returns {Resolver}
  */
 function createPerformantResolve() {
+  if (performantResolve) {
+    return performantResolve
+  }
   /**
    * @param {string} self
    * @returns {(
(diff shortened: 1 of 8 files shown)


packages/aa/src/index.js
```diff
@@ -144,6 +152,15 @@ function walkDependencyTreeForBestLogicalPaths({
     }
   } while (currentLevelTodos.length > 0)
 
+  for (const [
+    packageDir,
+    logicalPath,
+  ] of preferredPackageLogicalPathMap.entries()) {
+    if (isSymlink(packageDir)) {
```
boneskull, line 189: is this faster than just _always_ getting the real path?

boneskull (approved):
This may conflict with https://github.com/LavaMoat/LavaMoat/pull/808

and it may be that this would be better implemented on top of that PR such that a consuming module could provide custom resolution with this behavior.  or not.

I just retargeted #808 to `main`, anyway.  #811 needs both #810 and #808 and I'm not sure how to do a diamond-shaped PR yet.

packages/aa/src/index.js
naugtur: I assumed real path is much more work, but maybe only in case the thing is a link or a link to a link...
Gonna need to measure.

naugtur:
I wouldn't mind rebasing on #808 and merging into it. Should avoid all funny shapes like diamonds and circles

boneskull:
If you rebase this on #808, we want to merge #808 before merging this into its target, but OK.

packages/aa/src/index.js
naugtur: Covered with a microbenchmark. isSymlink is almost an order of magnitude faster on both symlinks and not. 
```
 results: {
          'isSymlink(notsymlink)': 16.870908999999983,
          'isSymlink(symlink)': 27.16377,
          'realpathSync(notsymlink)': 142.83877799999993,
          'realpathSync(symlink)': 237.59675600000003,
        },
```

If that's no longer the case or is not the case on one of the node versions, tests will warn us

naugtur:
merged manually into #808

Issue #993, opened 2024-02-01
allow-scripts: config reader does not normalize paths


boneskull:
The configuration for `allow-scripts` within `package.json` does no normalization of the path, e.g.:

```json
"allowBins": {
  "good": "node_modules/good_dep/index.js"
}
```

This is passed verbatim, even if the platform is `win32`.  The above path is invalid on Windows.  It _may_ be processed properly at some point, but it is inconsistent at best.  Currently, for `allow-scripts` to work _equivalently_ on a Windows machine, the consumer would need to change the option to:

```json
"allowBins": {
  "good": "node_modules\\good_dep\\index.js"
}
```

Seems like a non-starter to me 😜

nzbart:
Just adding some more information as a fellow Windows user. I typically use forward slashes on Windows now in many applications as my preferred separator because they work cross-platform. In most scenarios, Windows supports both path separators, with a few exceptions, such as when using extended length path prefixes (`\\?\`) and in some open-file dialogs. In most NodeJS apps, forward slashes just work.

Issue #692, opened 2023-08-30
aa: Support Windows directory separator


legobeat:
Caught in https://github.com/LavaMoat/LavaMoat/actions/runs/6020889921/job/16332885577?pr=691

```
> npm run test:run


> @lavamoat/aa@3.1.2 test:run
> ava


  ✘ [fail]: project 1
  ✘ [fail]: project 2
  ✘ [fail]: project 3
  ─

  project 1

  test\index.spec.js:10

    9:     .map(([packagePath,canonicalName]) => [path.relative(__dirname,packa…
   10:   t.deepEqual(normalizedMapEntries, [                                    
   11:     [                                                                    

  Difference (- actual, + expected):

    [
      [
  -     'projects\\1',
  +     'projects/1',
        '$root$',
      ],
      [
  -     'projects\\1\\node_modules\\aaa',
  +     'projects/1/node_modules/aaa',
        'aaa',
      ],
      [
  -     'projects\\1\\node_modules\\bbb',
  +     'projects/1/node_modules/bbb',
        'bbb',
      ],
      [
  -     'projects\\1\\node_modules\\bbb\\node_modules\\evil_dep',
  +     'projects/1/node_modules/bbb/node_modules/evil_dep',
        'bbb>evil_dep',
      ],
    ]

  › test/index.spec.js:10:5



  project 2

  test\index.spec.js:35

   34:     .map(([packagePath,canonicalName]) => [path.relative(__dirname,packa…
   35:   t.deepEqual(normalizedMapEntries, [                                    
   36:     [                                                                    

  Difference (- actual, + expected):

    [
      [
  -     'projects\\2',
  +     'projects/2',
        '$root$',
      ],
      [
  -     'projects\\2\\node_modules\\aaa',
  +     'projects/2/node_modules/aaa',
        'aaa',
      ],
      [
  -     'projects\\2\\node_modules\\bbb',
  +     'projects/2/node_modules/bbb',
        'bbb',
      ],
      [
  -     'projects\\2\\node_modules\\bbb\\node_modules\\evil_dep',
  +     'projects/2/node_modules/bbb/node_modules/evil_dep',
        'bbb>evil_dep',
      ],
      [
  -     'projects\\2\\node_modules\\good_dep',
  +     'projects/2/node_modules/good_dep',
        'good_dep',
      ],
    ]

  › test/index.spec.js:35:5



  project 3

  test\index.spec.js:64

[...]

PR #1002, opened 2024-02-06
windows-compat: aa, allow-scripts

legobeat:
This is part of introducing cross-OS testing for LavaMoat. It should make tests pass for `allow-scripts` except for the ones associated with the `--experimental-bins` feature (#993)

#### Based on
- [x] #691 
  - [Diff](https://github.com/legobeat/LavaMoat/compare/ci-win-mac...legobeat:LavaMoat:win-compat-allow-scripts)

#### Related
- #1003
- #1004

#### Changes
- chore(ci): Enable Windows-compat testing for `@lavamoat/aa`, `@lavamoat/allow-scripts`
  - chore(allow-scripts): Skip tests for `--experimental-bins` when running on Windows
- chore(allow-scripts): Port package shellscripts to JS
- fix(aa): Fix tests of allow-scripts to run on Windows
- fix(allow-scripts): Fix tests of allow-scripts to run on Windows
- feat(aa): Include unknown directory path in error message

#### Resolves
- #692

### DIFF
--- a/packages/aa/src/index.js
+++ b/packages/aa/src/index.js
@@ -269,7 +269,7 @@ function getPackageNameForModulePath(canonicalNameMap, modulePath) {
   // files should never be associated with a package directory across a package boundary (as tested via the presense of "node_modules" in the path)
   if (relativeToPackageDir.includes('node_modules')) {
     throw new Error(
-      `LavaMoat - Encountered unknown package directory for file "${modulePath}"`
+      `LavaMoat - Encountered unknown package directory "${relativeToPackageDir}" for file "${modulePath}"`
     )
   }
   return packageName
--- a/packages/allow-scripts/package.json
+++ b/packages/allow-scripts/package.json
@@ -35,7 +35,7 @@
   "scripts": {
     "lint:deps": "depcheck",
     "test": "npm run test:run",
-    "test:prep": "for d in ./test/projects/*/ ; do (cd \"$d\" && ../../../src/cli.js auto --experimental-bins); done",
+    "test:prep": "test/prepare.sh",
     "test:run": "ava"
   },
   "dependencies": {
--- a/packages/allow-scripts/package.json
+++ b/packages/allow-scripts/package.json
@@ -35,7 +35,7 @@
   "scripts": {
     "lint:deps": "depcheck",
     "test": "npm run test:run",
-    "test:prep": "test/prepare.sh",
+    "test:prep": "node test/prepare.js",
     "test:run": "ava"
   },
   "dependencies": {
--- a/packages/allow-scripts/test/prepare.js
+++ b/packages/allow-scripts/test/prepare.js
@@ -1,0 +1,20 @@
+const { execSync } = require('child_process')
+const { readdirSync } = require('node:fs')
+const { join, normalize } = require('node:path')
+
+// Execute allow-scripts command inside each test project directory
+readdirSync(join(__dirname, '../test/projects/'), { withFileTypes: true })
+  .filter((p) => p.isDirectory())
+  .forEach((dir) => {
+    try {
+      execSync(`node ${normalize('../../../src/cli.js')} auto --experimental-bins`, {
+        cwd: join(__dirname, '../test/projects/', dir.name),
+        stdio: 'inherit',
+      })
+    } catch (e) {
+      // ignore expected error
+      if (e.status !== 1) {
+        throw e
+      }
+    }
+  })
--- a/packages/allow-scripts/test/prepare.js
+++ b/packages/allow-scripts/test/prepare.js
@@ -1,20 +1,17 @@
-const { execSync } = require('child_process')
-const { readdirSync } = require('node:fs')
+const { execSync } = require('node:child_process')
+const { existsSync, readdirSync } = require('node:fs')
 const { join, normalize } = require('node:path')
 
 // Execute allow-scripts command inside each test project directory
-readdirSync(join(__dirname, '../test/projects/'), { withFileTypes: true })
-  .filter((p) => p.isDirectory())
+const baseDir = join(__dirname, '../test/projects/')
+readdirSync(baseDir, { withFileTypes: true })
+  .filter((p) => p.isDirectory() && existsSync(join(baseDir, p.name, 'package.json')))
   .forEach((dir) => {
-    try {
-      execSync(`node ${normalize('../../../src/cli.js')} auto --experimental-bins`, {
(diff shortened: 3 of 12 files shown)


packages/allow-scripts/package.json
```diff
@@ -35,7 +35,7 @@
   "scripts": {
     "lint:deps": "depcheck",
     "test": "npm run test:run",
-    "test:prep": "for d in ./test/projects/*/ ; do (cd \"$d\" && ../../../src/cli.js auto --experimental-bins); done",
+    "test:prep": "node test/prepare.js",
```
boneskull, line 38: 👍

packages/allow-scripts/test/prepare.js
```diff
@@ -0,0 +1,20 @@
+const { execSync } = require('child_process')
+const { readdirSync } = require('fs')
+const { join } = require('path')
```
boneskull:
```suggestion
const { execSync } = require('node:child_process')
const { readdirSync } = require('node:fs')
const { join } = require('node:path')
```

packages/allow-scripts/test/prepare.js
boneskull, line 1: normalize paths in this file

packages/browserify/package.json
```diff
@@ -57,7 +57,7 @@
     "files": [
       "test/*.spec.js"
     ],
-    "timeout": "2m",
```
boneskull: I saw this in #691 ... what's going on here?

legobeat: browserify tests on macOS in GHA are slow enough that they fail with timeout if this is <=3m.

packages/allow-scripts/test/prepare.js
legobeat: Done! (`path.join` already does os-specific normalization on output)

packages/allow-scripts/test/prepare.js
```diff
@@ -0,0 +1,20 @@
+const { execSync } = require('child_process')
+const { readdirSync } = require('node:fs')
+const { join, normalize } = require('node:path')
+
+// Execute allow-scripts command inside each test project directory
+readdirSync(join(__dirname, '../test/projects/'), { withFileTypes: true })
+  .filter((p) => p.isDirectory())
+  .forEach((dir) => {
+    try {
+      execSync(`node ${normalize('../../../src/cli.js')} auto --experimental-bins`, {
```
boneskull: I was eyeballing this and wondered why it was working, but it's because this uses `exec` and not `execFile`.  If it were `execFile`, you'd want to use `process.execPath` instead of `node`.

packages/allow-scripts/test/prepare.js
```diff
@@ -0,0 +1,20 @@
+const { execSync } = require('child_process')
```
boneskull:
```suggestion
const { execSync } = require('node:child_process')
```

packages/allow-scripts/test/prepare.js
```diff
@@ -0,0 +1,20 @@
...
+const { readdirSync } = require('node:fs')
+const { join, normalize } = require('node:path')
+
+// Execute allow-scripts command inside each test project directory
+readdirSync(join(__dirname, '../test/projects/'), { withFileTypes: true })
```
boneskull:
```suggestion
readdirSync(join(__dirname, '..', 'test', 'projects'), { withFileTypes: true })
```

packages/allow-scripts/test/prepare.js
```diff
@@ -0,0 +1,20 @@
...
+  .filter((p) => p.isDirectory())
+  .forEach((dir) => {
+    try {
+      execSync(`node ${normalize('../../../src/cli.js')} auto --experimental-bins`, {
+        cwd: join(__dirname, '../test/projects/', dir.name),
```
boneskull:
```suggestion
        cwd: join(__dirname, '..', 'test', 'projects', dir.name),
```

packages/allow-scripts/test/prepare.js
```diff
@@ -0,0 +1,20 @@
...
+        stdio: 'inherit',
+      })
+    } catch (e) {
+      // ignore expected error
+      if (e.status !== 1) {
+        throw e
+      }
```
boneskull: We should assert something about _why_ it's failing, because it may be failing for an unexpected reason (e.g., bad paths)

legobeat: this wasn't done previously (it was passing all errors silently) so let's do that separately? (unless you have a commit ready, in which case feel free to push onto this PR)

packages/allow-scripts/test/prepare.js
legobeat: This isn't actually necessary. Node handles this on Windows (tested).

packages/allow-scripts/test/prepare.js
legobeat: This isn't actually necessary. Node handles this on Windows (tested).

packages/aa/src/index.js
```diff
@@ -269,7 +269,7 @@ function getPackageNameForModulePath(canonicalNameMap, modulePath) {
   // files should never be associated with a package directory across a package boundary (as tested via the presense of "node_modules" in the path)
   if (relativeToPackageDir.includes('node_modules')) {
     throw new Error(
-      `LavaMoat - Encountered unknown package directory for file "${modulePath}"`
+      `LavaMoat - Encountered unknown package directory "${relativeToPackageDir}" for file "${modulePath}"`
```
naugtur, line 272: unrelated but useful

packages/allow-scripts/test/prepare.js
naugtur: Does it make sense to leave a comment here with this information?

packages/allow-scripts/test/prepare.js
naugtur: what's the expected error? there was no error capture in the bash for loop this is replacing.

naugtur (approved):
LGTM except I don't get what the expected error is

packages/allow-scripts/test/prepare.js
legobeat: Changed it now, filtering out the test directory without `package.json` (where attempting to run the script will fail) and not swallowing errors here anymore.

packages/allow-scripts/test/prepare.js
legobeat: using `process.execPath` seems more idiomatic, so changing to that in any case.

PR #1003, opened 2024-02-06
test(windows-compat): tofu

legobeat:
This is part of introducing cross-OS testing for LavaMoat. It the tests for `tofu` to pass on Windows.
#### Based on
- [x] #691 
  - [Diff](https://github.com/legobeat/LavaMoat/compare/ci-win-mac...legobeat:LavaMoat:win-compat-tofu)

#### Related
- #1002
- #1004

#### Changes
- chore(ci): Enable Windows-compatability tests for `tofu`
- fix(tofu): Fix tests of tofu to run on Windows

PR #1004, opened 2024-02-06
windows-compat: core

legobeat:
This is part of introducing cross-OS testing for LavaMoat. It the tests for `lavamoat-core` to pass on Windows.

#### Based on
- [x] #691 
  - [Diff](https://github.com/legobeat/LavaMoat/compare/ci-win-mac...legobeat:LavaMoat:win-compat-core)

#### Related
- #1003
- #1002

#### Changes
- chore(ci): Enable Windows-compatibility tests for `lavamoat-core`
- fix(core): Fix tests of core to run on Windows
- feat(core): include attempted path in error message when parsing JSON of policy fails

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -32,7 +32,7 @@
     "lib:ses": "cp ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js",
     "lint:deps": "depcheck",
     "test": "ava && npm run test:ses",
-    "test:ses": "diff -q ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js"
+    "test:ses": "diff -wq ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js"
   },
   "dependencies": {
     "@babel/types": "7.23.9",
--- a/packages/core/src/loadPolicy.js
+++ b/packages/core/src/loadPolicy.js
@@ -17,15 +17,31 @@ function readPolicyFileSync({ debugMode, policyPath }) {
   if (debugMode) {
     console.warn(`Lavamoat looking for policy at '${policyPath}'`)
   }
+  /** @type string */
+  let rawPolicy
   try {
-    const rawPolicy = readFileSync(policyPath, 'utf8')
-    return JSON.parse(rawPolicy)
+    rawPolicy = readFileSync(policyPath, 'utf8')
   } catch (err) {
     if (/** @type {NodeJS.ErrnoException} */ (err).code !== 'ENOENT') {
       throw err
     } else if (debugMode) {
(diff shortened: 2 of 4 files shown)


packages/core/package.json
```diff
@@ -32,7 +32,7 @@
     "lib:ses": "cp ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js",
     "lint:deps": "depcheck",
     "test": "ava && npm run test:ses",
-    "test:ses": "diff -q ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js"
+    "test:ses": "diff -wq ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js"
```
boneskull, line 35: windows has GNU diff??

packages/core/src/loadPolicy.js
```diff
@@ -16,15 +16,31 @@ async function readPolicyFile({ debugMode, policyPath }) {
...
     if (/** @type {NodeJS.ErrnoException} */ (err).code !== 'ENOENT') {
       throw err
     } else if (debugMode) {
       console.warn(`Lavamoat could not find policy at '${policyPath}'`)
     }
+    return
+  }
+  try {
+    return JSON.parse(rawPolicy)
+  } catch (/** @type any */ error) {
```
boneskull, line 34:
```suggestion
  } catch (/** @type {Error} */ (error)) {
```

packages/core/src/loadPolicy.js
```diff
@@ -16,15 +16,31 @@ async function readPolicyFile({ debugMode, policyPath }) {
...
+    if (debugMode) {
+      console.warn({
+        error: error?.message || error,
+        policyPath,
+        rawPolicy,
+      })
+    }
+    throw new Error(
+      `Lavamoat could not parse policy at '${policyPath}': ${/** @type {NodeJS.ErrnoException} */ (error).message}`
```
boneskull, line 43:
```suggestion
      `Lavamoat could not parse policy at '${policyPath}': ${error.message}`
```

it _is_ a `NodeJS.ErrnoException`, but we don't care, since we don't reference its `code` property.

boneskull (approved):
nit or 2

packages/core/src/loadPolicy.js
legobeat: This is not valid syntax..?
Getting `SyntaxError: Unexpected token '('`.

Without the parentheses: `Catch clause variable type annotation must be 'any' or 'unknown' if specified.`

packages/core/src/loadPolicy.js
legobeat: Sure, but it also helps being consistent with the clause just above.

PR #1006, opened 2024-02-07
chore(ci): better lockfile validation

boneskull:
Well, marginally.

This changes the lockfile validation check (via `npm ls`), which previously only lived in CI, to run via `npm run lint`.

Instead of directly executing `npm ls`, this wraps it in a small JS script which suppresses the `npm ls` output if validation succeeded.

If it wasn't for portability's sake, this is easily done in Bash via something like:

```sh
output="$(npm ls)" || {
  echo "$output"
  exit 1
}
```

...but, alas.


boneskull:
* **#1006** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1006?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

naugtur (approved):
BTW. I contributed changes to https://github.com/lirantal/lockfile-lint some time ago so that the basic api package (no CLI wrpapper) barely has any dependencies, and we could create a tool that would lint the lockfile, install without scripts and run allow-scripts after that.

scripts/lint-lockfile.js
```diff
@@ -0,0 +1,22 @@
...
+ * It exists because it should be portable and the only other way I know how to
+ * do this is via a Bash script.
+ *
+ * @packageDocumentation
+ */
+
+const { execFileSync } = require('node:child_process')
+
+try {
+  execFileSync('npm', ['ls', '--color'], { encoding: 'utf-8' })
```
naugtur, line 17: (NOT a change request) there might be a package we could use directly instead of exec+parse. Tho that package might be `arborist`

boneskull:
@naugtur I'm not sure `lockfile-lint` solves the same problem, does it?  Or does it do so for free?  If `npm ls` exits with a non-zero exit code, it means that `package-lock.json` is out-of-sync with one-or-more `package.json` files.  I don't think it actually prevents a `npm ci` from occurring.

boneskull:
Anyhow, I'm open to doing this in a more robust way later, if we find value in it.  For now I wanted to make it a little easier to use.

naugtur:
@boneskull It's a different thing. No overlap. Mentioning it because the context was right. It cannot replace this PR ever.

PR #1007, opened 2024-02-07
chore(browserify): see if running in serial helps tests not timeout

boneskull:
this may fix the browserify tests failing on mac (?).

cc @legobeat


legobeat (approved):
Seems like the sane thing to do in any case!

naugtur (approved):
:ship: :ship: :ship: :ship: :ship:

PR #1014, opened 2024-02-08
fail faster on dirty changes via pre-commit hook

boneskull:
This does three (3) things, and could probably be split up into equivalent PRs, but I'm all out of Graphite stacks:

- chore(browserify): rewrite test:prep bash script as a JS script

This is just for portability's sake.  

- chore: adapt dirty-check script to be run in local dev env

This is trickier, and I especially want @legobeat to look at it re: `--staged`

- chore: update lint-staged to run browserify policy gen

In my limited testing, these changes seem to work, but it probably wants to be tested locally.  To trigger it:

1.  Do something like change a `require('fs')` to `require('node:fs')` (or vice-versa) in `lavamoat-browserify`'s sources somewhere.  
2. Run `npm run setup` to ensure your hooks are up-to-date
3. Stage the change, then commit it
4. It _should_ cause the `test:prep` script to run and the policy file to get updated and will then dump stuff from `dirty-check.js` and fail.

Or, if there's a branch you have handy that _already_ has the problem (which would currently be surfaced in CI), just try to add a commit to that branch.  I was using fc51eea for this, assuming that it still exists.

### DIFF
--- a/.config/lint-staged.config.js
+++ b/.config/lint-staged.config.js
@@ -1,0 +1,12 @@
+// @ts-check
+
+/** @type {import('lint-staged').Config} */
+module.exports = {
+  '*.js': ['eslint --fix', 'prettier --write'],
+  '*.(ts|md|ya?ml|json)': ['prettier --write'],
+  '!((package-lock|policy)*).json': ['prettier --write'],
+  'packages/*/src/**/*.js|package?(-lock).json': () => [
+    'npm run test:prep',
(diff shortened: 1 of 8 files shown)


naugtur:
I can see why this is useful.
I'm worried if our tooling complexity is becoming a challenge to maintain. We're getting close to a line I'm willing to draw.

naugtur (approved):

PR #1018, opened 2024-02-09
stop final newline churn in policy files

boneskull:
Prettier was automatically formatting `node/test/projects/2/atob.policy.json` which keeps causing the file to churn, since when LM writes it, it has no trailing newline. Prettier adds a newline.

This tweaks the `.prettierignore` so that it will not format `*policy.json` instead of just `policy.json`.

It also aligns `.editorconfig` and VS Code to respect this.


legobeat (approved):
TBH I think we should do the inverse change the generation to actually have a newline at EOF. But prettier probably shouldn't be messing around with these files anyway.

PR #1019, opened 2024-02-19
feat(webpack): support webpack replacing  with __webpack_require__.g

naugtur:
~~TODO: retarget to main when #959 is merged~~ (github did it automagically!)

This adds support for an implicit feature. Discovered when building a real project.


packages/webpack/src/plugin.js
```diff
@@ -111,6 +111,20 @@ class LavaMoatPlugin {
     // compiler.options.optimization.moduleIds = "hashed";
     // compiler.options.optimization.chunkIds = "named";
 
+    // sadly regular webpack compilation doesn't allow for synchronous resolver.
```
boneskull, line 114: I'm not a huge fan of leaving commented-out code around.  should this be a draft PR, since it's not targeting `main`?

packages/webpack/test/scaffold.js
```diff
@@ -7,16 +7,25 @@ const { toSnapshotSync } = require('memfs/lib/snapshot')
 const { createContext, Script, runInNewContext } = require('vm')
 const { readFileSync } = require('fs')
 
-exports.scaffold = async function runWebpackBuild(webpackConfig) {
+exports.scaffold = async function runWebpackBuild(
```
boneskull, line 17: types 😄

packages/webpack/test/scaffold.js
```diff
@@ -65,7 +74,13 @@ function convertToMap(input, path = '') {
   return result
 }
 
-function runScript(code, globals = {}) {
+const defaultGlobals = () => ({
```
boneskull, line 96: why does it only contain `console`?

packages/webpack/test/scaffold.js
```diff
@@ -65,7 +74,13 @@ function convertToMap(input, path = '') {
...
+  console: {
+    log: () => {},
+  },
+})
+
+function runScript(code, globals = defaultGlobals()) {
```
boneskull, line 109: remind me if this calls `defaultGlobals()` every time `runScript` is called without a second parameter, or if it's called just once

packages/webpack/test/scaffold.js
```diff
@@ -7,16 +7,25 @@ const { toSnapshotSync } = require('memfs/lib/snapshot')
...
   // Create a compiler instance
   const compiler = webpack(webpackConfig)
+  let fs
 
-  // Use memfs to write the output to memory
-  const memoryFs = memfs.createFsFromVolume(new memfs.Volume())
-  compiler.outputFileSystem = memoryFs
+  if (!writeFS) {
+    // Use memfs to write the output to memory
+    const memoryFs = memfs.createFsFromVolume(new memfs.Volume())
```
boneskull, line 33:
```suggestion
    const {fs: memoryFs} = memfs()
```

weizman (approved):
makes sense

packages/webpack/src/plugin.js
naugtur: I intend to use that code when I get aa to accept async resolve.

packages/webpack/test/scaffold.js
naugtur: every time. `defaultGlobals` doesn't even need to exist at the time of defining runScript

packages/webpack/test/scaffold.js
naugtur: it's useful to have console there by default. The testcases are otherwise not depending on anything else. If you need anything else, you pass your own globals.

packages/webpack/test/scaffold.js
naugtur: I wanted to get a new fs each time

packages/webpack/test/scaffold.js
naugtur: added types to scaffold. boyscout rule principle. will fix types in test specs another time too.

naugtur:
gonna merge before boneskull clarifies he meant all test types :P

PR #1023, opened 2024-02-20
chore(eslint): increase target langugage from es2020 to es2021

legobeat:
This breaks out the es version bump from #753 into a separate commit.

- Change eslint config `parserOptions.ecmaVersion` to `2021`
- Set eslint config `env.es2021`
- Unset eslint config `env.es2020`

### Related
- Lifted from #753 (https://github.com/LavaMoat/LavaMoat/pull/753#discussion_r1491586426) 

#### Blocked by
- [x] #1022

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,13 +18,13 @@
  * The minimum EcmaScript runtime environment to support
  * @see {@link https://eslint.org/docs/latest/use/configure/language-options#specifying-environments}
  */
-const ECMASCRIPT_ENV = 'es2020'
+const ECMASCRIPT_ENV = 'es2021'
 
 /**
  * The minimum EcmaScript language version to support
  * @see {@link https://eslint.org/docs/latest/use/configure/language-options#specifying-parser-options}
  */
-const ECMASCRIPT_VERSION = 2020
+const ECMASCRIPT_VERSION = 2021
 
 /**
  * The minimum supported version of Node.js


leotm (changes requested):
looks like https://github.com/LavaMoat/LavaMoat/pull/753#discussion_r1498160442 needs to happen first ^ before this pr

legobeat:
> looks like [#753 (comment)](https://github.com/LavaMoat/LavaMoat/pull/753#discussion_r1498160442) needs to happen first ^ before this pr

Almost: It's blocked by #1022 (as listed under "Blocked by" in PR description). Hence in draft.

The end result of #1022 + #1023 (this PR) are equivalent to #753.

Wanted to split out the eslint upgrade as a separate change from the rest.

naugtur (approved):

naugtur:
TBH I don't see the point of merging #1022 before this, since this one has all the same commits +1
Could have been one PR

legobeat:
> TBH I don't see the point of merging #1022 before this, since this one has all the same commits +1 Could have been one PR

The point is to have the version increase merged in a commit like [this](https://github.com/LavaMoat/LavaMoat/pull/1023/commits/44c8ce5730dc6e43aec5f80c184bf2c17e949fa4) instead of like [this](https://github.com/LavaMoat/LavaMoat/pull/753/commits/095f9bb6320dd9bf296d86634733f8d749c6b6a4)

leotm (approved):

leotm:
final thought, worth bumping viz eslint config to ec2021 too?

https://github.com/LavaMoat/LavaMoat/blob/e39fec0a9649518cd399053250185610a1c66136/packages/viz/.eslintrc.js#L15

legobeat:
> final thought, worth bumping viz eslint config to ec2021 too?
> 
> https://github.com/LavaMoat/LavaMoat/blob/e39fec0a9649518cd399053250185610a1c66136/packages/viz/.eslintrc.js#L15

It's a good idea and thanks for bringing it up. I'm currently leaning towards getting #984 in released before that though, even if it should be safe.

PR #1027, opened 2024-02-20
chore(browserify): sync exampleApp fixture policy

legobeat:
It seems like the fixture file `test/fixtures/exampleApp/policy.json` has not been updated for quite a while.

This adds regeneration of it to the `test:prep` script (which I intend to refactor to make portable and nicer separately), and commits the result of running that.


boneskull (approved):
needs rebase

PR #1028, opened 2024-02-21
aa benchmark

naugtur:


legobeat:
If these grow any larger (in code-size or resource usage at bench), I think it becomes more desirable to move the benchmarks out of unit tests. So probably a `test:bench` script (which can still be run as part of the standard test suite in CI of course)

packages/aa/test/utils.js
```diff
@@ -2,6 +2,53 @@ const path = require('node:path')
...
+
+const diffEachKey = (a, b) => {
+  const result = {}
+  for (const key in a) {
+    result[key] = b[key] - a[key]
+  }
+  return result
+}
+
+const toHumanReadable = (obj) => {
```
legobeat: Considering the generic names here, maybe break them out to `benchmarkUtils.js` or similar to separate them from the generic test utils?

naugtur:
benchmark1 is actually a unit test guarding against our microoptimization becoming useless in the future. 
but the more serious benchmarking would have to be moved out because CI is not a good place for that as evident from running the above on mac in CI.

packages/aa/test/benchmark2.spec.js
```diff
@@ -0,0 +1,28 @@
+const test = require('ava')
+const path = require('node:path')
+const { benchAsync, benchAsyncAll } = require('./utils')
+const { loadCanonicalNameMap } = require('../src/index.js')
+
+test('[bench] loadCanonicalNameMap on workspace root', async (t) => {
```
boneskull: might be nice to break this up into a test macro

packages/aa/test/benchmark2.spec.js
boneskull, line 1: pls no ✨ magic numbers 😄

packages/aa/test/benchmark2.spec.js
```diff
@@ -0,0 +1,28 @@
...
+  t.assert(
+    result['loadCanonicalNameMap'] < 500, // arbitrary large number so we notice when we degrade by orders of magnitude. Watch the results when developing.
+    'expected loadCanonicalNameMap to be faster than 500ms'
+  )
+  // For development - ensure you're not breaking the result
+  // t.snapshot(
+  //   Array.from(map.entries())
+  //     .map(([k, v]) => k + v)
+  //     .sort()
+  // )
```
boneskull: why isn't this enabled?

packages/aa/test/utils.js
```diff
@@ -2,6 +2,53 @@ const path = require('node:path')
 const { symlink } = require('node:fs/promises')
 const { existsSync } = require('node:fs')
 
+const translateHeapStats = (stats = []) => {
```
boneskull:
```suggestion
/**
 * @param {{space_name: string, space_used_size: number}[]} stats
 */
const translateHeapStats = (stats = []) => {
```

or make a `@typedef` for this somewhere

packages/aa/test/utils.js
```diff
@@ -2,6 +2,53 @@ const path = require('node:path')
...
+  const result = {}
+  for (const { space_name, space_used_size } of stats) {
+    result[space_name] = space_used_size
+  }
+  return result
+}
+
+const updateMaxEachKey = (current, update) => {
```
boneskull:
```suggestion
/**
 * @param {Record<string, number>} current
 * @param {Record<string, number>} update
 */
const updateMaxEachKey = (current, update) => {
```

packages/aa/test/utils.js
```diff
@@ -2,6 +2,53 @@ const path = require('node:path')
...
+  for (const key in current) {
+    current[key] = Math.max(current[key], update[key])
+  }
+}
+
+const diffEachKey = (a, b) => {
```
boneskull:
```suggestion
/**
 * @param {Record<string, number>} a
 * @param {Record<string, number>} b
 */
const diffEachKey = (a, b) => {
```

packages/aa/test/utils.js
```diff
@@ -2,6 +2,53 @@ const path = require('node:path')
...
+  const result = {}
```
boneskull:
```suggestion
  /** @type {Record<string, number>} */
  const result = {}
```

packages/aa/test/utils.js
boneskull:
```suggestion
/**
 * @param {Record<string, number>} obj
 */
const toHumanReadable = (obj) => {
```

packages/aa/test/utils.js
```diff
@@ -2,6 +2,53 @@ const path = require('node:path')
...
+
+const toHumanReadable = (obj) => {
+  const result = {}
+  for (const key in obj) {
+    if (obj[key] > 0) result[key] = `+${(obj[key] / 1024 / 1024).toFixed(4)} MB`
+  }
+  return result
+}
+
+const recordMemorySpike = (frequency = 10) => {
```
boneskull:
```suggestion
/**
 * @param {number} frequency
 */
const recordMemorySpike = (frequency = 10) => {
```

packages/aa/test/utils.js
```diff
@@ -2,6 +2,53 @@ const path = require('node:path')
...
+  const v8 = require('v8')
```
boneskull: why is the require buried in here?

packages/aa/test/utils.js
```diff
@@ -2,6 +2,53 @@ const path = require('node:path')
...
+  const initial = translateHeapStats(v8.getHeapSpaceStatistics())
+  const result = { ...initial }
+  const collect = () =>
+    updateMaxEachKey(result, translateHeapStats(v8.getHeapSpaceStatistics()))
+  const interval = setInterval(collect, frequency)
```
boneskull:
```suggestion
  const interval = setInterval(collect, frequency).unref()
```

unless there's a reason not to do that, it will prevent future bugs encountered by interloping developers of the "why does my test hang???" variety

packages/aa/test/utils.js
```diff
@@ -26,3 +73,37 @@ exports.createProject4Symlink = async () => {
   const target = path.join(__dirname, 'projects', '4', 'packages', 'aaa')
   await exports.osIndependentSymlink(target, src)
 }
+
+exports.bench = (fn, name, iterations = 10000) => {
```
boneskull:
```suggestion
/**
 * @param {(() => void)} fn
 * @param {string} name
 * @param {iterations} number
 * @returns {{MEMORY_SPIKE: string, [x: string]: number}}
 */
exports.bench = (fn, name, iterations = 10000) => {
```

packages/aa/test/utils.js
```diff
@@ -26,3 +73,37 @@ exports.createProject4Symlink = async () => {
...
+  const t0 = performance.now()
+  for (let i = 0; i < iterations; i++) {
+    fn()
+    s.collect()
+  }
+  const t1 = performance.now()
+  return { [name]: (t1 - t0) / iterations, MEMORY_SPIKE: s.getResult() }
+}
+
+exports.benchAsync = async (fn, name, iterations = 1000) => {
```
boneskull:
```suggestion
/**
 * @param {(() => void | Promise<void>)} fn
 * @param {string} name
 * @param {iterations} number
 * @returns {Promise<{MEMORY_SPIKE: string, [x: string]: number}>
 */
exports.benchAsync = async (fn, name, iterations = 1000) => {
```

packages/aa/test/utils.js
```diff
@@ -26,3 +73,37 @@ exports.createProject4Symlink = async () => {
...
+  const t0 = performance.now()
+  for (let i = 0; i < iterations; i++) {
+    await fn()
+    s.collect()
+  }
+  const t1 = performance.now()
+  return { [name]: (t1 - t0) / iterations, MEMORY_SPIKE: s.getResult() }
+}
+
+exports.benchAsyncAll = async (fn, name, iterations = 1000) => {
```
boneskull:
```suggestion
/**
 * @param {(() => void | Promise<void>)} fn
 * @param {string} name
 * @param {iterations} number
 * @returns {Promise<{MEMORY_SPIKE: string, [x: string]: number}>
 */
exports.benchAsyncAll = async (fn, name, iterations = 1000) => {
```

boneskull (approved):

packages/aa/test/benchmark2.spec.js
naugtur: it would fail every time we add/update dependencies. It only makes sense as a local utility while you fiddle with the implementation

packages/aa/test/utils.js
naugtur: I'm working on breaking them out to an unrelated package

packages/aa/test/utils.js
naugtur: so that v8 is not imported unless the function is used. In case it has side effects.

packages/aa/test/utils.js
naugtur: yeah, why not.

packages/aa/test/benchmark2.spec.js
naugtur: Magic numbers are here mostly because I tried to make it a test. it doesn't make a lot of sense to have it as a test. I think I will extract it into something separate.

packages/aa/test/benchmark2.spec.js
naugtur: never really needed to be a test in the first lpace

PR #1032, opened 2024-02-23
fix(core): export types from generatePolicy

boneskull:
The types in `generatePolicy.js` weren't exported, and we need them (e.g., `ModuleInspector`)


leotm (approved):
(generateKernel / mergePolicy / sourceTransforms / makeInitStatsHook exports remaining, after typing those files then .ts or ts-check)

PR #1033, opened 2024-02-24
create @lavamoat/types

### LINKED ISSUE #995: fix lavamoat-tofu/lavamoat-core cyclic dependency of types
`lavamoat-core` depends on `lavamoat-tofu`.  But `lavamoat-tofu` wants types out of `lavamoat-core`.  This is what `lavamoat-tofu` is doing:

- The [`codeSampleFromAstNode()`](https://github.com/LavaMoat/LavaMoat/blob/5cc08cf3cf145d110444b30f4bc375351b516e1a/packages/tofu/src/codeSampleFromAstNode.js#L9) method expects a `LavamoatModuleRecord` which is a class defined in `lavamoat-core`. We could avoid this by moving the function into `lavamoat-core`, as it doesn't appear to have anything else to do with `lavamoat-tofu`.  This would be a breaking change, and both `lavamoat-node` and `survey` would need to be updated as well.
- Several functions in [`util.js`](https://github.com/LavaMoat/LavaMoat/blob/5cc08cf3cf145d110444b30f4bc375351b516e1a/packages/tofu/src/util.js#L194).  These functions aren't exported (not members of `lavamoat-tofu`'s public API) so aren't "production" types--`lavamoat-core` should be a dev dep.  But because TS doesn't understand the difference (nor should it, really), that won't break the cycle.  It's _possible_ to address this _in situ_, but the solution is so onerous and appalling that I'd just rather not.

This leads me to believe what we really need is a **new package** that both `lavamoat-tofu` and `lavamoat-core` can both depend on.  For now, it could be a type-only package which only exports the [policy schema](https://github.com/LavaMoat/LavaMoat/blob/5cc08cf3cf145d110444b30f4bc375351b516e1a/packages/core/src/schema/lavamoat-policy.v0-0-1.schema.ts#L1); potentially providing a [JSON schema](#733) in the future.

So, I'm proposing two tasks. These do not depend on one another, but both are necessary to break the cycle:

1. Move `codeSampleFromAstNode` into `lavamoat-core`.
2. Create a new package (e.g., `@lavamoat/schema` or `@lavamoat/policy-schema`) and move the schema types from `lavamoat-core` into it.

cc @kumavis (who wrote `codeSampleFromAstNode`) and @naugtur who hates TS and @legobeat who hates production deps of types 😄

boneskull:
Closes #995

This creates a new package, `@lavamoat/types`, which provides common types for various packages. It solves a type-only dependency cycle between `lavamoat-core` and `lavamoat-tofu`, at minimum

* * *

**UPDATE May 28 2025**:  The _scope_ of this PR is:

- Resolve the cycle by creating `@lavamoat/types`, populated with types from `lavamoat-core`
- Consume `@lavamoat/types` where `lavamoat-core` was previously consumed (for the appropriate types)
- Use `@import` syntax in JS sources (though I do not claim to have addressed _all_ of these)
- Avoid using deprecated `LavaMoatOverrides` type
- Narrow types if appropriate (this is the only place where types would be meaningfully different, though I expect any such changes to be backwards-compatible)

_Changes to logic are explicitly **not** in scope._


boneskull:
* **#1033** <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1033?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1033?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1782** <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1782?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1593](https://github.com/LavaMoat/LavaMoat/pull/1593) <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1593?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

legobeat (commented):
@boneskull needs conflict-resolution of:

```
package-lock.json
packages/tofu/package.json 
```

legobeat:
@boneskull ping!

packages/core/src/moduleRecord.js
```diff
@@ -1,77 +1,38 @@
 // @ts-check
 
+/**
+ * Do not import this; use `import('@lavamoat/types').LavamoatModuleRecord`
```
legobeat: Considering that this is importing the same (so should be no risk of falling out of sync): Is there any danger in importing this one instead?

After reading into it, it feels kind of like I wouldn't need to know any of this and while not ideal, importing from here should be OK?

(Once the types package has landed properly, I'm thinking we could also look at if there's some linting rule we could have, that could in general prevent direct imports directly from files in other workspace packages?)

legobeat: Also: `@deprecated` might fit here?
https://jsdoc.app/tags-deprecated

packages/types/src/module-record.ts
legobeat, line 1: Could these be `.d.ts` files directly, rather than `.ts`?

boneskull: No, because `.d.ts` files are generally not typechecked.

packages/core/src/moduleRecord.js
boneskull: It's not deprecated.  So the point is that we need to use `LavamoatModuleRecord` from the `types` package as an interface and implement it.  

In JSDoc--and I think this is a bug--`@implements` works _if and only if_ the literal type you're using is a `@typedef` _in the current module_.  That means:

```js
/** @typedef {import('./butts').Butts} Butts */

/**
 * @implements {Butts}
 */
class Pants {
}
``` 

works fine, but:

```js
/**
 * @implements {import('./butts').Butts}
 */
class Pants {
}
``` 

_does not_.


But if we use a `@typedef`, we re-export the type, because that's what typedefs do.  But we only want one source of truth for the `LavamoatModuleRecord` interface.  Because we have to use the `@typedef`, we've now created a "copy" of `LavamoatModuleRecord` named `LMR`.  Nobody should use it.

That said, if we can upgrade TS, we can probably avoid the `@typedef` altogether because of the new `@import` tag.

boneskull:
Reconciling this w/ @naugtur 's changes

packages/types/src/module-record.ts
legobeat: Only because we are inheriting from a base tsconfig which sets `compilerOption.skipLibCheck = true` (which I only realized now).

It will check them if set to `false` (which I really think we should in general): 
https://github.com/LavaMoat/LavaMoat/pull/1264/commits/82e966be469edc56ffbb9e71b97ebb7d958ce91d

(BTW: Setting the same configuration in `packages/types/tsconfig.json` did not take effect during `build:types`. This was surprising to me)

boneskull: No, I don't recommend that.  It will cause things to break if one of our dependencies has bad types.

boneskull: At any rate: this is a package full of types, and IMO TypeScript sources are more than appropriate here.

boneskull:
@naugtur See db4883b838389f2f79726df3c01fffa1b9785cb0

packages/types/src/module-record.ts
legobeat: > It will cause things to break if one of our dependencies has bad types.

.. In which case it will be caught by type-checking before it becomes an actual problem, right?

As opposed to without, in which case it will still affect and break any downstream users who do enable full type-checking, without us noticing.

legobeat: > At any rate: this is a package full of types, and IMO TypeScript sources are more than appropriate here.

`.d.ts` files are guaranteed to only include types sans implementation. Reduces the potential for future fun times because someone somehow puts an `enum` or `class` there.

boneskull: > .. In which case it will be caught by type-checking before it becomes an actual problem, right?

> As opposed to without, in which case it will still affect and break any downstream users who do enable full type-checking, without us noticing.

1. It will break, but we won't be able to fix it, because we do not own those types.
2. Most people don't enable this level of type-checking. We trust that our dependencies ship good types. If they don't ship good types, well, that sucks, but we're not gonna break the build because of it.

> .d.ts files are guaranteed to only include types sans implementation. Reduces the potential for future fun times because someone somehow puts an enum or class there.

I really don't understand the concern here.  This is a package for types.

legobeat: > I really don't understand the concern here. This is a package for types.

`.ts` files can include and export things that are not types (TS `enum` is just an example of something people commonly believe is a type, but is actually also runtime code, a confusion which commonly causes great consternation).

I think the types package should include type interfaces only - no resulting js code - and `.d.ts` files reflect that and makes it more naturally enforcable.

legobeat: > It will break, but we won't be able to fix it, because we do not own those types.

We can, though? Do you have an example of an unfixable issue? If we pull in and rely on types beyond our control, I think that's a bigger issue in itself?

boneskull:
@naugtur This is ready for review again

naugtur (approved):

packages/tofu/src/index.js
```diff
@@ -25,5 +24,4 @@ module.exports = {
   traverse,
   inspectPrimordialAssignments,
   inspectSesCompat,
-  codeSampleFromAstNode,
```
naugtur, line 28: @boneskull this was a breaking change

PR #1034, opened 2024-02-24
chore: revert "adapt dirty-check script to be run in local dev env"

boneskull:
This reverts commit 5c811799fc20ce2c98289b387c3e4e94465e5b21.

It is broken when not staging all changes.


boneskull:
I'm going to abandon the idea for now, as I'd need to do something like take a "snapshot" of unstaged changes and disregard them from whatever the diff output is.

boneskull:
It's unfortunate, because otherwise it works well. 😜

legobeat (approved):

PR #1038, opened 2024-02-27
chore(vscode): organize imports where possible

boneskull:
In TS and ESM projects, this setting causes VSCode to organize the imports in a module by type.  If order is necessary, place a blank line between the imports which should not be changed, e.g.:

```js
import "ses"

import path from "node:path"
import util from "node:util"
```

This will ensure `ses` always comes first.


boneskull:
This does not affect CJS sources.

naugtur (approved):

PR #1054, opened 2024-02-29
fix(tofu): fix logic error in inspectEsmImports

naugtur:


packages/tofu/src/inspectSource.js
```diff
@@ -195,11 +196,7 @@ function inspectEsmImports(ast, packagesToInspect) {
   const handleNodePath = (path) => {
     const importSource = path.node.source?.value
 
-    if (
-      importSource &&
-      ((pkgsToInspect.size && pkgsToInspect.has(importSource)) ||
-        !pkgsToInspect.size)
-    ) {
+    if (importSource && (shouldInspectAll || pkgsToInspect.has(importSource))) {
```
naugtur, line 199: This is the change, everything below is tests

packages/webpack/test/fixtures/main/node_modules/es6-module-package/index.js
```diff
@@ -1,3 +1,6 @@
+import * as um from 'umd-package';
+
 export const hello = () => {
+  um;
```
naugtur, line 4: our newly configured tooling removes unused imports on save

weizman (approved):
lg

packages/tofu/src/inspectSource.js
boneskull: I must have had something that set `packagesToInspect = []` in the function signature, then removed it... weird.

packages/webpack/test/fixtures/main/node_modules/es6-module-package/index.js
boneskull: alternatively, I think this could work:

```
import 'umd-package'
```

packages/tofu/src/inspectSource.js
naugtur: One of the major usecases of this function is no builtins whatsoever in webpack :D 
I'm pondering not calling it at the higher layer but need to doublecheck for sideeffects.

PR #1056, opened 2024-02-29
fix(laverna): understands bad npm API

boneskull:
`npm view <pkg> versions --json` outputs a JSON array of strings _if and only if_ multiple versions of `pkg` have been published. When a _single_ version has been published, it outputs a JSON string.

This change fixes the problem where Lerna always expected a JSON array of strings.


naugtur:
There was a time where it was fashionable to return a single element straight without wrapping it in a list. It dates the API :D

naugtur (approved):

PR #1068, opened 2024-03-01
chore: reduce renovate churn

legobeat:
This is an attempt to reduce the amount of dependency maintenance overhead from Renovate.

- Disable `minor` and `patch` for these frequently updated (>= 4 renovate PRs merged so far) devDependencies (we have other channels for security advisories):
  - `commitlint`
  - `husky`
  - `lint-staged`
  - `linters`
- Disable `major` and `patch` for `@types/node`. `major` because it should be bumped in sync with `engines.node` and `patch` because of the update-frequency, I think we're fine on only getting automation for `minor` here.

Room for improvement: I think Renovate has some functionality for consolidating configured groups of dependencies into larger update PRs. That might be a good way to keep down the merge overhead while still staying on top of new versions. Could also group all GHA actions into one such PR group, if possible.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -51,6 +51,18 @@
       matchUpdateTypes: ['major'],
       enabled: false,
     },
+    {
+      // Disable frequent updates for dev tooling
+      // These can be updated manually but we still want to be notified about new majors
+      matchPackageNames: [
+        'commitlint',
+        'husky',
+        'lint-staged',
+        'linters',
+      ],
+      matchUpdateTypes: ['minor', 'patch'],
+      enabled: false,
+    }
   ],
   transitiveRemediation: true,
 }
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -54,11 +54,11 @@
     {
       // Disable frequent updates for dev tooling
       // These can be updated manually but we still want to be notified about new majors
+      extends: ['packages:linters'],
+      matchPackagePrefixes: ['@commitlint'],
       matchPackageNames: [
-        'commitlint',
         'husky',
-        'lint-staged',
-        'linters',
+        'lint-staged'
       ],
       matchUpdateTypes: ['minor', 'patch'],
       enabled: false,
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -62,6 +62,15 @@
       ],
       matchUpdateTypes: ['minor', 'patch'],
       enabled: false,
+    },
+    {
+      // @types/node should be bumped to major only alongside `engines.node` in package.json
+      // We don't care about every patch release so only notify onnew minors
+      matchPackageNames: [
+        '@types/node',
+      ],
+      matchUpdateTypes: ['major', 'patch'],
+      enabled: false,
     }
   ],
   transitiveRemediation: true,


naugtur:
LGTM but I'll wait for @boneskull

.github/renovate.json5
```diff
@@ -51,6 +51,27 @@
...
     },
+    {
+      // Disable frequent updates for dev tooling
+      // These can be updated manually but we still want to be notified about new majors
+      matchPackageNames: [
+        'commitlint',
+        'husky',
+        'lint-staged',
+        'linters',
+      ],
```
boneskull: [matchPackageNames](https://docs.renovatebot.com/configuration-options/#matchpackagenames) expects a list of _exact_ package names, so `commitlint` (not one of our direct dependencies anyhow) and `linters` should not be here.  

```suggestion
      extends: ['packages:linters'],
      matchPackagePrefixes: ['@commitlint'],
      matchPackageNames: [
        'husky',
        'lint-staged'
      ],
```

see [packages:linters](https://docs.renovatebot.com/presets-packages/#packageslinters) and [matchPackagePrefixes](https://docs.renovatebot.com/configuration-options/#matchpackageprefixes)

.github/renovate.json5
```diff
@@ -51,6 +51,27 @@
...
+    },
+    {
+      // @types/node should be bumped to major only alongside `engines.node` in package.json
+      // We don't care about every patch release so only notify onnew minors
+      matchPackageNames: [
+        '@types/node',
+      ],
+      matchUpdateTypes: ['major', 'patch'],
+      enabled: false,
+    }
```
boneskull, line 74: I agree on holding it back from `major` updates

boneskull (commented):
I'm not sure I like this idea.

What I _don't_ like is all the Renovate PRs for `viz`'s dependencies... which just sit.  These are clogging up the pipeline, as Renovate throttles its total count of open PRs. We should do whatever we're doing with `viz` and close those out.  

There are many levers to pull to make Renovate work better for you/us.  For example, we could change the cadence for updates to these tooling packages to once every _n_ weeks.  We could batch all tooling-related patch/minor updates together into a single PR.  We could decide upon a trusted set of tools (IMO, eslint and typescript-eslint can be trusted) we want to auto-merge.  

I just don't think disabling them entirely and requiring manual updates is the way.

legobeat:
> What I _don't_ like is all the Renovate PRs for `viz`'s dependencies... which just sit. These are clogging up the pipeline, as Renovate throttles its total count of open PRs. We should do whatever we're doing with `viz` and close those out.
> 
> There are many levers to pull to make Renovate work better for you/us. For example, we could change the cadence for updates to these tooling packages to once every _n_ weeks. We could batch all tooling-related patch/minor updates together into a single PR. We could decide upon a trusted set of tools (IMO, eslint and typescript-eslint can be trusted) we want to auto-merge.
> 
> I just don't think disabling them entirely and requiring manual updates is the way.

I don't disagree with any of that but this PR doesn't have anything to do with `viz` so somewhat unrelated to this PR? Left them out here partly to sidestep the "what to do with those packages" question from the scope of this one. Further grouping *also* sounds like a good idea.

boneskull (changes requested):
See discussion elsewhere

PR #1074, opened 2024-03-04
aa async resolver

### LINKED ISSUE #1073: allow async resolver in aa

naugtur:
Fixes #1073 

Introduces a major change in the programmatic API but to a parameter that was seldom used. Will require a major bump but no consequences in adoption or such.

Performance is almost unaffected.

### DIFF
--- a/packages/aa/src/index.js
+++ b/packages/aa/src/index.js
@@ -43,31 +43,52 @@ function createPerformantResolve() {
    */
   const readPackageWithout = (filepath) => {
     /**
-     * @param {(path: string) => StringOrToString} readFileSync - Sync file
-     *   reader
+     * @param {(
+     *   file: string,
+     *   cb: (err: Error | null, file?: StringOrToString) => void
+     * ) => void} readFile
+     *   - Async file reader
+     *
      * @param {string} otherFilepath - Path to another `package.json`
-     * @returns {Record<string, unknown> | undefined}
+     * @param {(err: Error | null, result?: Record<string, unknown>) => void} cb
+     *   - Callback
+     *
+     * @returns {void}
      */
-    return (readFileSync, otherFilepath) => {
+    return (readFile, otherFilepath, cb) => {
       // avoid loading the package.json we're just trying to resolve
       if (otherFilepath.endsWith(filepath)) {
-        return {}
+        return cb(null, {})
       }
       // original readPackageSync implementation from resolve internals:
-      const body = readFileSync(otherFilepath)
-      try {
-        return JSON.parse(`${body}`)
-      } catch (jsonErr) {}
+      readFile(otherFilepath, (err, body) => {
+        if (err) {
+          return cb(err)
+        }
+        try {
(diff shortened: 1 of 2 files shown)


packages/aa/test/benchmark1.spec.js
```diff
@@ -1,7 +1,7 @@
 const { realpathSync, lstatSync } = require('node:fs')
```
naugtur, line 1: This was moved/changed in a different PR. I need to rebase and clean up.

packages/aa/src/index.js
```diff
@@ -371,8 +404,24 @@ function comparePackageLogicalPaths(aPath, bPath) {
...
 /**
- * @typedef Resolver
- * @property {(path: string, opts: { basedir: string }) => string} sync
+ * @template T
+ * @param {number} parallelMax
+ * @param {T[]} items
+ * @param {(item: T) => Promise<any>} mapper
+ * @returns {Promise<void>}
+ */
+async function runParallelMax(parallelMax, items, mapper) {
```
naugtur, line 413: note: this function could be much more sophisticated so that it tracks which async jobs finished and starts new ones as they do. The jobs are similar enough to each other tho, that it might not bring a lot of improvement to the execution time.

naugtur: update - quick test of the concept:
```js
async function runParallelMax(parallelMax, items, mapper) {
  items = items.reverse()
  return new Promise((resolve, reject) => {
    let promises = 0
    const boop = () => {
      if (items.length === 0) {
        if(promises === 0) resolve()
        return
      }
      const p = mapper(items.pop())
      promises++
      p.then(()=>{
        promises--
        boop()
      }, reject)
    }
    for (let i = 0; i < parallelMax; i++) {
      boop()
    }
  })
}
```
with this implementation it's a few percent slower, so there's not much to gain from smarter parallelization

packages/aa/src/index.js
```diff
@@ -209,23 +238,27 @@ function processOnePackageInLogicalTree(
...
     visited = new Set(),
-  } = /** @type {WalkDepTreeOpts} */ (currentLevelTodos.pop())
+  } = currentTodo
+
   const depsToWalk = getDependencies(packageDir, includeDevDeps)
 
   // deps are already sorted by preference for paths
-  for (const depName of depsToWalk) {
-    let depPackageJsonPath = wrappedResolveSync(resolve, depName, packageDir)
+  await runParallelMax(10, depsToWalk, async (depName) => {
```
legobeat, line 246: I'm assuming here that this PR is motivated as a requirement facilitate native ESM and not coming from a  performance perspective.

why is the concurrency introduced here? The synchronous API is simpler and (despite the function name `runParallelMax` hinting otherwise) we can not get actual parallelism here due to Node.js architecture - we still have a single processor thread jumping between up to 100 async IO jobs. Do you notice regressions if dropping concurrency to 1 (refactoring this async wrapping back to a `for const of`?

PR #1078, opened 2024-03-06
chore(core): auto-update vendored lockdown.umd.js

boneskull:
This changes the `lint-staged` configuration so that when `package-lock.json` is updated, the vendored `packages/core/lib/lockdown.umd.js` is copied from the currently-installed `ses` and added to the commit.

I'm not sure if this is something we want to be doing or not, but needing to run the `lib:ses` script in `lavamoat-core` after a `ses` upgrade is non-obvious for maintainers (me).

### DIFF
--- a/.config/lint-staged.config.js
+++ b/.config/lint-staged.config.js
@@ -1,3 +1,5 @@
+const path = require('node:path')
+
 // @ts-check
 
 /** @type {import('lint-staged').Config} */
@@ -5,4 +7,5 @@ module.exports = {
   '*.js': ['eslint --fix', 'prettier --write'],
   '*.(ts|md|ya?ml|json)': ['prettier --write'],
   '!((package-lock|policy)*).json': ['prettier --write'],
+  'package-lock.json': () => ['npm run -w lavamoat-core lib:ses'],
 }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -29,7 +29,7 @@
     "!*.tsbuildinfo"
   ],
   "scripts": {
-    "lib:ses": "cp ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js",
+    "lib:ses": "cp ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js && git add -A ./lib/lockdown.umd.js",
     "lint:deps": "depcheck",
     "test": "ava && npm run test:ses",
     "test:ses": "diff -wq ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js"


packages/core/package.json
```diff
@@ -29,7 +29,7 @@
     "!*.tsbuildinfo"
   ],
   "scripts": {
-    "lib:ses": "cp ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js",
+    "lib:ses": "cp ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js && git add -A ./lib/lockdown.umd.js",
```
legobeat, line 32:
```suggestion
    "lib:ses": "cp ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js",
```
I think the package scripts like this shouldn't be modifying the local git environment, any drawback to moving it into the lint-script?

.config/lint-staged.config.js
```diff
@@ -1,8 +1,11 @@
+const path = require('node:path')
+
 // @ts-check
 
 /** @type {import('lint-staged').Config} */
 module.exports = {
   '*.js': ['eslint --fix', 'prettier --write'],
   '*.(ts|md|ya?ml|json)': ['prettier --write'],
   '!((package-lock|policy)*).json': ['prettier --write'],
+  'package-lock.json': () => ['npm run -w lavamoat-core lib:ses'],
```
legobeat, line 10:
```suggestion
  'package-lock.json': () => ['npm run -w lavamoat-core lib:ses && git add -A ./packages/core/lib/lockdown.umd.js'],
```

legobeat (changes requested):

legobeat:
IMO a more proper solution would be to not have a copy checked into git at all - just vendor a copy from the package `devDependency` at build-time. And maybe add it to `bundledDependencies`? 


Other than or until that, maybe giving a more helpful error message to the failing test [here](https://github.com/LavaMoat/LavaMoat/pull/772#pullrequestreview-1916341177) would also solve most of the need here, without additional magic in commit hooks?

.config/lint-staged.config.js
legobeat: I think the package scripts like this shouldn't be modifying the local git environment, any drawback to moving it into the lint-script?

packages/core/package.json
boneskull: sure (I had it this way originally but changed it for reasons that escape me)

boneskull:
Yeah, I mean, _why do we vendor `lockdown.umd.js`, anyway?_

boneskull:
Closing in lieu of a PR which removes it

PR #1099, opened 2024-03-12
fix(tofu): ClassProperty nodes are not globals

### LINKED ISSUE #1098: tofu: support class properties
~~It looks like `lavamoat-tofu` is now considering some things to be globals which were not previously considered globals, and by all accounts _aren't_ actually globals.  This is likely related to changes merged in #809 or #810~~

Ex: https://github.com/MetaMask/metamask-extension/pull/23423/files#diff-7c92d6ad2a8a25dea901b4d2d1399d086251e96be25bafbbecf7389c95d92be4R2224-R2245

~~Diagnose & fix~~

This is because `lavamoat-tofu` does not know anything about `ClassProperty`-type `Node`s, e.g.:

```js
class Foo {
  bar;
  baz = 1;
}
```

Ensure `ClassProperty`-type `Node`s are not considered globals in `lavamoat-tofu`.

boneskull:
Adds an exception to `findGlobals()` for AST nodes of type `ClassProperty`.

Closes #1098

### DIFF
--- a/packages/browserify/test/fixtures/secureBundling/lavamoat/node/policy.json
+++ b/packages/browserify/test/fixtures/secureBundling/lavamoat/node/policy.json
@@ -1040,17 +1040,9 @@
       },
       "globals": {
         "__dirname": true,
-        "ast": true,
         "console.error": true,
         "console.warn": true,
-        "content": true,
-        "define": true,
-        "file": true,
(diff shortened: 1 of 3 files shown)


naugtur (approved):
:ship: :ship: :ship:

PR #1101, opened 2024-03-13
webpack: lockdown inlining

naugtur:
Some use-cases (including snaps) need to inline SES lockdown into the bundle file. It's impossible to flawlessly guess which file will be the one to load first on a new page in more complex configurations, so I went with a static list you have to provide. separate file lockdown remains the default approach.

Updated the README with explanation of this and got it up to date with implementation,


packages/webpack/README.md
```diff
@@ -14,15 +14,20 @@ LavaMoat Webpack Plugin wraps each module in the bundle in a [Compartment](https
...
-- diagnosticsVerbosity: Optional number property to represent diagnostics output verbosity. A larger number means more overwhelming diagnostics output. Default is 0.  
-  Setting positive verbosity will enable runChecks.
-- readableResourceIds: Decide whether to keep resource IDs human readable (regardless of production/development mode). If false, they are replaced with a sequence of numbers. Keeping them readable may be useful for debugging when a policy violation error is thrown.
-- lockdown: set configuration for [SES lockdown](). Setting the option replaces defaults from LavaMoat.
-- HtmlWebpackPluginInterop: add a script tag to the html output for ./lockdown file if HtmlWebpackPlugin is in use
+The LavaMoat plugin takes an options object with the following properties (all optional):
+
+| property                 | description                                                                                                                                                                                                                                                                                                        | default                  |
```
```suggestion
| Property                 | Description                                                                                                                                                                                                                                                                                                        | Default                  |
```

packages/webpack/README.md
```diff
@@ -14,15 +14,20 @@ LavaMoat Webpack Plugin wraps each module in the bundle in a [Compartment](https
...
+| runChecks                | boolean property to indicate whether to check resulting code with wrapping for correctness.                                                                                                                                                                                                                        | `false`                  |
+| diagnosticsVerbosity     | number property to represent diagnostics output verbosity. A larger number means more overwhelming diagnostics output. Setting positive verbosity will enable runChecks.                                                                                                                                           | `0`                      |
+| policy                   | the LavaMoat policy object. (by default policy is loaded from files, this is an override)                                                                                                                                                                                                                          | `undefined`              |
```
```suggestion
| `policyLocation`           | Folder to store policy files in.                                                                                                                                                                                                                                                                                   | `./lavamoat/webpack`     |
| `generatePolicy`           | Whether to generate the `policy.json` file. Generated policy is used in the build immediately. `policy-override.json` is applied before bundling, if present.                                                                                                                                                        | `false`                  |
| `emitPolicySnapshot`       | If enabled, emits the result of merging policy with overrides into the output folder of Webpack build for inspection. The file is not used by the bundle.                                                                                                                                                          | `false`                  |
| `readableResourceIds`      | Boolean to decide whether to keep resource IDs human readable (possibly regardless of production/development mode). If `false`, they are replaced with a sequence of numbers. Keeping them readable may be useful for debugging when a policy violation error is thrown. By default, follows the Webpack config mode. | `(mode==='development')` |
| `lockdown`                 | Configuration for [SES lockdown][]. Setting the option replaces defaults from LavaMoat.                                                                                                                                                                                                                        | reasonable defaults      |
| `HtmlWebpackPluginInterop` | Boolean to add a script tag to the HTML output for `./lockdown` file if `HtmlWebpackPlugin` is in use.                                                                                                                                                                                                                 | `false`                  |
| `inlineLockdown`           | Array of output filenames in which to inline lockdown (instead of adding it as a file to the output folder).                                                                                                                                                                                                             |
| `runChecks`                | Boolean property to indicate whether to check resulting code with wrapping for correctness.                                                                                                                                                                                                                        | `false`                  |
| `diagnosticsVerbosity`     | Number property to represent diagnostics output verbosity. A larger number means more overwhelming diagnostics output. Setting a positive verbosity will enable `runChecks`.                                                                                                                                           | `0`                      |
| `policy`                   | The LavaMoat policy object (if not loading from file; see `policyLocation`)                                                                                                                                                                                                                          | `undefined`              |
```

packages/webpack/README.md
```diff
@@ -107,12 +102,13 @@ Sadly, even treeshaking doesn't eliminate that module. It's left there and faili
...
+# Security
 
-**This is a _beta_ release and does not provide any guarantees; even those listed below. Use at your own risk!**
+**This is an experimental technology. Use at your own risk!**
 
-- SES must be added to the page without any bundling or transforming for any security guarantees to be sustained.
-  - The plugin is attempting to add it as an asset to the compilation for the sake of Developer Experience. Feedback welcome.
+- [SES lockdown][] must be added to the page without any bundling or transforming for any security guarantees to be sustained.
+  - The plugin is attempting to add it as an asset to the compilation for the sake of Developer Experience. `.js` extension is omitted to prevent minification.
+  - Optionally lockdown can be inlined into the bundle files. It's hard to determine which files are loaded when, so providing filenames for scripts that get to load as the first script on the page to apply lockdown only once is how one uses inlining.
```
boneskull: This sentence is a little hard to understand

packages/webpack/src/plugin.js
```diff
@@ -429,22 +433,38 @@ class LavaMoatPlugin {
...
+            })
+          )
+        } else {
+          const HtmlWebpackPluginInUse = compiler.options.plugins.find(
+            /**
+             * @param {unknown} plugin
+             * @returns {plugin is import('webpack').WebpackPluginInstance}
+             */
+            (plugin) =>
+              !!plugin && plugin.constructor.name === 'HtmlWebpackPlugin'
```
```suggestion
              !!plugin && typeof plugin === 'object' && plugin.constructor.name === 'HtmlWebpackPlugin'
```

Or, would `plugin instanceof HtmlWebpackPlugin` work?  Do we have the class available?

packages/webpack/src/types.js
```diff
@@ -12,17 +12,18 @@
...
- * @property {boolean} [runChecks] - Check resulting code with wrapping for
- *   correctness
  * @property {boolean} [readableResourceIds] - Should resourceIds be readable or
  *   turned into numbers - defaults to (mode==='development')
  * @property {boolean} [HtmlWebpackPluginInterop] - Add a script tag to the html
  *   output for lockdown.js if HtmlWebpackPlugin is in use
+ * @property {string[]} [inlineLockdown] - Prefix the listed files with lockdown
  * @property {number} [diagnosticsVerbosity] - A number representing diagnostics
  *   output verbosity, the larger the more overwhelming
+ * @property {Object} [lockdown] - Options to pass to lockdown
```
boneskull:
```suggestion
 * @property {import('ses').LockdownOptions} [lockdown] - Options to pass to lockdown
```

I'm not sure if that type is exported, but if it isn't, pls use `Record<string, any>` instead of `Object`

packages/webpack/test/e2e-lockdown-inline.spec.js
```diff
@@ -0,0 +1,29 @@
...
+const { scaffold, runScript } = require('./scaffold.js')
+const { makeConfig } = require('./fixtures/main/webpack.config.js')
+
+test.before(async (t) => {
+  const webpackConfig = makeConfig({
+    inlineLockdown: ['app.js'],
+  })
+  await t.notThrowsAsync(async () => {
+    t.context.build = await scaffold(webpackConfig)
+  }, 'Expected the build to succeed')
```
boneskull, line 11: the side-effect here is little malodorous to me, but I'm not sure what a better solution would look like.

boneskull (approved):

packages/webpack/README.md
naugtur: thank you

packages/webpack/README.md
naugtur: It was even harder to write :sweat_smile:

packages/webpack/src/plugin.js
naugtur: I don't trust it to not diverge on webpack versions. but it might be unfounded.

packages/webpack/README.md
```diff
@@ -107,12 +102,13 @@ Sadly, even treeshaking doesn't eliminate that module. It's left there and faili
...
+# Security
 
-**This is a _beta_ release and does not provide any guarantees; even those listed below. Use at your own risk!**
+**This is an experimental technology. Use at your own risk!**
 
-- SES must be added to the page without any bundling or transforming for any security guarantees to be sustained.
-  - The plugin is attempting to add it as an asset to the compilation for the sake of Developer Experience. Feedback welcome.
+- [SES lockdown][] must be added to the page without any bundling or transforming for any security guarantees to be sustained.
+  - The plugin is attempting to add it as an asset to the compilation for the sake of Developer Experience. `.js` extension is omitted to prevent minification.
+  - Optionally lockdown can be inlined into the bundle files. It's hard to determine how entries in the config are deployed and used, so you need to list the scripts that get to load as the first script on theDo we have page to apply lockdown only once when inlined. When you have a single bundle, you just configure a list with one element. It gets more complex with builds for multiple pages.
```
legobeat:
```suggestion
  - Optionally, lockdown can be inlined into the bundle files. The user is expected to provide a list of the scripts allowed to load as the first script on the page.
```

The rest of the text doesn't really seem relevant for the user (and the "single-item-can-be-provided-as-single-element-list" should be obvious)

packages/webpack/README.md
```diff
@@ -107,12 +102,13 @@ Sadly, even treeshaking doesn't eliminate that module. It's left there and faili
 
 This plugin will skip policy enforcement for such ignored modules.
 
-# Security Claims
+# Security
 
-**This is a _beta_ release and does not provide any guarantees; even those listed below. Use at your own risk!**
+**This is an experimental technology. Use at your own risk!**
```
legobeat:
```suggestion
**This is an experimental software. Use at your own risk!**
```

packages/webpack/src/buildtime/emitSes.js
```diff
@@ -1,40 +1,79 @@
...
-            } else {
-              throw Error(
-                'LavaMoat: Could not insert lockdown script tag, no suitable location found in the html template'
-              )
+              if (headTagRegex.test(data.html)) {
+                data.html = data.html.replace(headTagRegex, `$&${scriptTag}`)
+              } else if (scriptTagRegex.test(data.html)) {
+                data.html = data.html.replace(
+                  scriptTagRegex,
+                  `${scriptTag}<script`
```
legobeat:
```suggestion
                  `${scriptTag}$&`
```

packages/webpack/README.md
```diff
@@ -14,15 +14,20 @@ LavaMoat Webpack Plugin wraps each module in the bundle in a [Compartment](https
...
+
+| Property                   | Description                                                                                                                                                                                                                                                                                                           | Default                  |
+| -------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------ |
+| `policyLocation`           | Folder to store policy files in.                                                                                                                                                                                                                                                                                      | `./lavamoat/webpack`     |
```
legobeat:
```suggestion
| `policyLocation`           | Directory to store policy files in.                                                                                                                                                                                                                                                                                      | `./lavamoat/webpack`     |
```

packages/webpack/README.md
```diff
@@ -14,15 +14,20 @@ LavaMoat Webpack Plugin wraps each module in the bundle in a [Compartment](https
...
+| `generatePolicy`           | Whether to generate the `policy.json` file. Generated policy is used in the build immediately. `policy-override.json` is applied before bundling, if present.                                                                                                                                                         | `false`                  |
+| `emitPolicySnapshot`       | If enabled, emits the result of merging policy with overrides into the output folder of Webpack build for inspection. The file is not used by the bundle.                                                                                                                                                             | `false`                  |
```
legobeat:
```suggestion
| `emitPolicySnapshot`       | If enabled, emits the result of merging policy with overrides into the output directory of Webpack build for inspection. The file is not used by the bundle.                                                                                                                                                             | `false`                  |
```

packages/webpack/README.md
```diff
@@ -14,15 +14,20 @@ LavaMoat Webpack Plugin wraps each module in the bundle in a [Compartment](https
...
+| `lockdown`                 | Configuration for [SES lockdown][]. Setting the option replaces defaults from LavaMoat.                                                                                                                                                                                                                               | reasonable defaults      |
+| `HtmlWebpackPluginInterop` | Boolean to add a script tag to the HTML output for `./lockdown` file if `HtmlWebpackPlugin` is in use.                                                                                                                                                                                                                | `false`                  |
+| `inlineLockdown`           | Array of output filenames in which to inline lockdown (instead of adding it as a file to the output folder).                                                                                                                                                                                                          |
```
legobeat:
```suggestion
| `inlineLockdown`           | Array of output filenames in which to inline lockdown (instead of adding it as a file to the output directory).                                                                                                                                                                                                          |
```

packages/webpack/test/e2e-lockdown-inline.spec.js
naugtur: I'm not sure what you mean is a side effect here. scaffold is running the build and returning information it collects. If it fails, we report it as a testsuite failure.

packages/webpack/README.md
naugtur: While the wording could be better (and I can see a copy-paste error now), I insist on explaining that we can't simply infer this from the webpack configuration and need the user to specify files to inline into.

packages/webpack/README.md
naugtur: not sure why, but I don't mind.

packages/webpack/README.md
legobeat: I think it's enough to make it clear that we need the user to specify the files - alluding to our perceived difficulty in automatically deriving it isn't really helpful and comes off at an attempt at an excuse at best and confusing at worst (:

naugtur: It is an excuse for the folks with a basic setup where there's only one resulting bundle file that I decided not to support for consistency. Wanna change that?

legobeat: I think the behavior is fine and I agree we should make it clear what configuration is required. But let's stick to the point and keep any discussion of implementation difficulties in developer docs as opposed to readme?

PR #820, opened 2023-12-21
chore: remove depcheck

boneskull:
This adds [dependency-cruiser](https://npm.im/dependency-cruiser) to replace [depcheck](https://npm.im/depcheck), which does a _very_ thorough check of all referenced packages throughout the codebase.

After configuration, I went through and fixed the problems it highlighted (of which there were many).

Closes #793 

* * *

## Reviewers

You'll want to look at everything here, as there were numerous problems.  Hopefully I split it up in a way that makes sense.

* * *

**UPDATE Jun 7 2024**

I want to try this again, except with [knip](https://knip.dev) instead of `dependency-cruiser`.

Issue #793, opened 2023-12-15
replace depcheck


boneskull:
Recently, @kumavis found that an install of `lavamoat-browserify` was missing a transitive dependency, `espree`.  `espree` is a dependency of `@lavamoat/lavapack`, and [was recently a dev dependency](#792) when it should have been a production dependency.  `depcheck` was not able to catch this.

In addition, `depcheck` is not monorepo-aware nor is it type-aware, which makes it rather awkward to use; we have to add exceptions all over the place (look at the various `.depcheckrc.yml` files across our workspaces).

I recommend we move to a more modern tool to sanity-check the dependencies.  I have had luck with [dependency-cruiser](https://npm.im/dependency-cruiser), which is considerably more powerful (maybe _too_ powerful).  This is not the only tool that solves this.  [midnight-smoker](https://npm.im/midnight-smoker) (my project) could also be helpful in this regard, but I am personally not ready to advocate for its adoption (eventually, I plan to).

My requirements for such a replacement (please add your own):

- should catch prod/dev dependency mismatches
- should be monorepo-aware, which implies compatibility with Node.js' module resolution algorithm / hoisting / depuping
- should be TypeScript-aware such that `import('foo')` in a JSDoc type does not result in package `foo` flagged as extraneous; same for `.ts` sources.

PR #1126, opened 2024-03-21
chore(core): normalize non-existent package references in test files

boneskull:
Essentially, we want to keep use of "phony" module names to a minimum to avoid a tool like `dependency-cruiser` raising errors. They must be explicitly whitelisted, and it's best to keep that list length minimal.

Extracted from #820

### DIFF
--- a/packages/core/test/builtin.spec.js
+++ b/packages/core/test/builtin.spec.js
@@ -79,8 +79,8 @@ test('builtin - paths soft-bindings preserve "this" but allow override', async (
 
     defineTwo: () => {
       const { Buffer } = require('node:buffer')
-      const thisChecker = require('thisChecker')
-      const { SomeClass } = require('someClass')
+      const thisChecker = require('abc')
+      const { SomeClass } = require('xyz')
       // this test ensures "Buffer.prototype.slice" is copied in a way that allows "this" to be overridden
       module.exports.overrideCheck = (buf) =>
         Buffer.prototype.slice.call(buf, 1, 2)[0] === buf[1]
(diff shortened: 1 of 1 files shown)


naugtur:
Where is that whitelist?

boneskull:
@naugtur The whitelist is in the dependency-cruiser configuration contained in #793.

legobeat:
> @naugtur The whitelist is in the dependency-cruiser configuration contained in #793.

@boneskull was that supposed to be a link elsewhere? I don't see a reference there.

boneskull:
* **#1126** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1126?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

weizman:
[Context](https://github.com/LavaMoat/LavaMoat/pull/820/files#diff-0df5ff741b1d1f7d5eb8f64a7beb3055610ddd7553b1f7db7ceb0f2e4de90d03R237) (I suppose @boneskull?)

weizman (approved):
lgtm, although maybe we should come up with names that are both slightly more indicative and easier to find across the repo?

e.g. `fake_prop_1`, `fake_prop_2` etc

boneskull:
@weizman That'd be better, but ... 

I understand the ergonomics of allowing the test fixtures to be written inline like this, but it's also obviously difficult for tooling to understand (TS, ESLint, dependency-cruiser, depcheck, etc., will _all_ take issue with importing non-existent modules).

IMO if we're going to make major changes here (which I'm not necessarily proposing), we should extract these fixtures into a folder & file structure which we can more easily instruct tools to ignore. Note that this is not a hill I'm willing to die on 😄

PR #1156, opened 2024-04-08
fix(lavapack): remove unused sourcemap logic, simplify clearing out sourcemaps from concatenated sources

naugtur:
that sourcemap processing bit was not being used anyway and its regex was unnecessarily complex.

TODO:
- [x] Node 16 tests - ava fails to find the snapshot file. works elsewhere :(


packages/lavapack/src/pack.js
```diff
@@ -299,12 +298,13 @@ function prepareModuleInitializer(
...
 }
 
-function extractSourceMaps(sourceCode) {
-  const converter = convertSourceMap.fromSource(sourceCode)
-  // if (!converter) throw new Error('Unable to find original inlined sourcemap')
-  const maps = converter && converter.toObject()
-  const code = convertSourceMap.removeComments(sourceCode)
-  return { code, maps }
+const sourceMapDropper =
+  /^\s*?(?:\/\/[@#]\s+?sourceMappingURL=.*$|\/\*[@#]\s+?sourceMappingURL=[\s\S]*?\*\/)/gm
```
legobeat, line 302:
```suggestion
  /^\s*(?:\/\/[@#]\s*sourceMappingURL=.*$|\/\*[@#]\s*sourceMappingURL=[^\n]*\*\/)/gm
```
Is this regex sourced from somewhere else? Seems like it could be simplified a bit by dropping redundant optionals and lazies. Should be the same result but simpler and I think slightly cheaper?

Also, is there anything that makes `[\s\S]` different from `.` and why prefer it in one case and not the other..?

naugtur: `[\s\S]` is deliberately used instead of `.` because `.` doesn't match newlines. One of the problems with the previous approach was it would fail to match correctly if a newline ended up inside the block comment.
There's a test for that.

the `\s+?` fragment was a part of a regex sourced from the original implementation we're replacing. The difference between `\s+?` and `\s*` is that the `?` character makes the match non-greedy which doesn't seem necessary in this case, but I didn't see the need to change it.

legobeat: Updated suggestion to use `[^\n]` instead of `[\s\S]` - should be the same effect while making the intention clear?

naugtur: I said we used `[\s\S]` because the dot _doesn't_ match newlines. The intention is to match new lines as part of the string until `*/` is reached. All of the newlines, regardless of discrepancies between what regex considers a newline and what JS parsers consider a newline.

Please checkout the code and run the test.

packages/lavapack/test/sourcemap-regex.spec.js
```diff
@@ -0,0 +1,91 @@
+const test = require('ava')
+
+// Had to extract this part because node16 got confused and believed the sourcemapping comments inside strings. As a result, AVA would malfunction and create a snapshot called foo.js.
+const SRCMAPDEF_STRING = 'sourceMappingURL'
```
naugtur, line 4: This was fun to debug.

boneskull (approved):

PR #1158, opened 2024-04-09
feat: global store with attenuation support

naugtur:
- moved 'write' implementation to endowmentsToolkit
- expanded existing capability to also support attenuating nested fields of values that are writable elsewhere
- changes to core are non-breaking
- PR includes usage in Webpack plugin as it's trivial and provides additional testing to the functionality.

should be immediately usable in endomoat too.


packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -92,7 +92,43 @@ test('getEndowmentsForConfig - siblings', (t) => {
...
+  const getEndowmentsForConfig = prepareTest()
+  const sourceGlobal = {
+    a: 1,
+    b: { c: 2 },
+    d: 3,
+  }
+  const config = {
+    globals: {
+      a: true,
+      'b.c': true,
```
naugtur, line 109: TODO: check what happens if 
```
'b.c':true,
'b.c.x':true
```

- I think the current implementation will fail
- I think copyValueAtPath might also not be handling that correctly.

Can be fixed or forbidden

packages/core/src/endowmentsToolkit.js
```diff
@@ -20,9 +20,14 @@ module.exports = endowmentsToolkit
...
  * @param {object} opts
  * @param {DefaultWrapperFn} [opts.createFunctionWrapper]
+ * @param {boolean} [opts.handleGlobalWrite]
+ * @param {Set<string>} [opts.knownWritableFields] - List of globals that can be
+ *   mutated later
  */
 function endowmentsToolkit({
   createFunctionWrapper = defaultCreateFunctionWrapper,
+  handleGlobalWrite = false,
+  knownWritableFields = new Set(),
```
naugtur, line 39: global write is a decision to make once per implementation, so it belongs here along with the configuration.

packages/core/src/endowmentsToolkit.js
```diff
@@ -80,6 +88,16 @@ function endowmentsToolkit({
...
+          if (!handleGlobalWrite) {
+            return
+          }
+          if (pathParts.length > 1) {
+            throw new Error(
+              `LavaMoat - write access is only allowed at the top level, saw "${path}"`
+            )
+          }
+          allowedWriteFields.add(path)
+          whitelistedReads.push(path)
```
naugtur, line 109: I could let it fall through to the push below, but there's a check for policy field being explicitly `true` not any other accidental value. It's safer to have a clear carve-out for `'write'`

packages/core/src/endowmentsToolkit.js
```diff
@@ -151,6 +183,81 @@ function endowmentsToolkit({
     return !value
   }
 
+  /**
+   * @param {string} key
+   * @param {Record<string, any>} sourceRef
+   * @param {Record<string, any>} targetRef
+   */
+  function makeWritableValueAtPath(key, sourceRef, targetRef) {
```
naugtur: When we make the assumption of global store being the sourceRef, the implementation becomes very simple.

packages/core/src/endowmentsToolkit.js
```diff
@@ -151,6 +183,81 @@ function endowmentsToolkit({
...
+
+  /**
+   * Puts a getter at the end of the path that returns the nested values from a
+   * top-level field that might change at runtime.
+   *
+   * @param {string[]} pathParts
+   * @param {Record<string, any>} sourceRef
+   * @param {Record<string, any>} targetRef
+   */
+  function instrumentDynamicValueAtPath(pathParts, sourceRef, targetRef) {
```
naugtur: This is slightly less simple, because it's doing a whole lot more than the original implementation - it's attenuating access to writable globals via leaf getters which is much higher fidelity than the original implementation that would ignore attenuation and share the top-level key

packages/core/test/endowmentsToolkit.spec.js
naugtur: done

packages/webpack/src/runtime/runtime.js
```diff
@@ -24,8 +26,23 @@ if (LOCKDOWN_ON) {
...
+      }
+    })
+  }
+})
+
 const { getEndowmentsForConfig, copyWrappedGlobals } =
-  LAVAMOAT.endowmentsToolkit()
+  LAVAMOAT.endowmentsToolkit({
+    handleGlobalWrite: true,
+    knownWritableFields,
```
naugtur, line 44: This is all that needed to change in usage to support global store now. Should work the same for endomoat.

packages/core/src/endowmentsToolkit.js
legobeat: Even if this is an internal function, do you think we could add a test or two, to help illustrating intended behavior for future contribution?

naugtur: There's a bunch of tests that are indirectly testing this functionality via `getEndowments...`.
endowmentsToolkit is one of the exports of core and I was hoping to avoid making this method public. 
The closest I could get to calling it directly is `makeMinimalViewOfRef` (which I was hoping to one day make an internal function too.)

Would this satisfy the ask here?
Is it a pre-requisite for approval?

naugtur: Ok, I added a test without making it public API.

packages/core/src/endowmentsToolkit.js
```diff
@@ -15,14 +15,28 @@
  * @packageDocumentation
  */
 
+/**
+ * WARNING: This module is used directly by the runtime in webpack plugin which
+ * uses simple shimming to assemble modules. It doesn't bundle properly. This
+ * file cannot reqire any files or packages.
+ */
```
legobeat, line 22:
```suggestion
 * WARNING: This file may not require any files or packages.
 * This module is used directly by the runtime in webpack plugin which
 * uses simple shimming to assemble modules and does not bundle properly. 
 */

```


 Nit: Putting the actionable message first for clarity (also there was a small typo). 

Aside: If this turns out to be more than just transient, should look at having an extra linting rule for this file/module to catch it IMO

packages/core/src/endowmentsToolkit.js
legobeat: This looks ripe for refactoring to a switch/case but can be done separately as follow-up

packages/core/src/endowmentsToolkit.js
```diff
@@ -640,26 +553,121 @@ function endowmentsToolkit({
...
-      (typeof current === 'object' || typeof current === 'function')
-    ) {
-      protoChain.push(current)
-      current = Reflect.getPrototypeOf(current)
+/**
+ * @param {string} visited
+ * @param {string} next
+ */
+function extendPath(visited, next) {
+  // FIXME: second part of this conditional should be unnecessary
```
legobeat, line 582: Assuming this is indeed still necessary (TBD) and unrelated to what we're doing here.

legobeat (approved):
Left some comment feedback and minor follow-ups but LGTM otherwise!

PR #1166, opened 2024-04-17
Webpack nodejs support

naugtur:
add isBuiltin and its consequences to webpack, so that builds with `target:'node'` can work and be safe

TODO:
- [x] thread isBuiltin for policy generation
- [x] implement considering builtins in policy enforcement
- [x] support attenuating what's returned based on the policy saying `"fs.readFileSync": true`


packages/webpack/test/snapshots/e2e-nodejs.spec.js.md
```diff
@@ -0,0 +1,22 @@
...
+
+## webpack/node.js - policy shape
+
+> Snapshot 1
+
+    `{␊
+      "resources": {␊
+        "nodejs-package": {␊
+          "builtin": {␊
+            "node:fs.readdirSync": true␊
```
naugtur, line 15: this nesting is not yet supported by runtime (see TODOs)

packages/webpack/test/snapshots/e2e-nodejs.spec.js.md
```diff
@@ -0,0 +1,22 @@
...
+          },␊
+          "packages": {␊
+            "external:../../../node:fs": true␊
```
naugtur: this should not be here - we need to use isbuiltin in one more place in policy generation to get rid of it.

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -92,7 +96,7 @@ module.exports = {
           // Knowing the actual specifier is not relevant here, they're used as unique identifiers that match between here and dependencies
           specifier: module.userRequest,
           file: module.userRequest,
-          type: 'js',
+          type: isBuiltin(module.userRequest) ? 'builtin' : 'js',
```
naugtur, line 110: was this not enough to get rid of `external:../../../node:fs` item in policy?

packages/webpack/test/snapshots/e2e-nodejs.spec.js.md
leotm: nice you fixed it adding [filter](https://github.com/LavaMoat/LavaMoat/pull/1166/commits/deb2f73ec51561b2911fa08996588a410cdae1dd#diff-7eaf4b1ffc2d7b78432df1e04b8ff5a068a964fb18ab76285142a6aa9fd8408dR317) to policy generation in core 🎉

packages/webpack/src/buildtime/policyGenerator.js
leotm: yep it wasn't enough, but looks like the `packages/core/src/generatePolicy.js` update was ^

packages/webpack/test/e2e-nodejs.spec.js
```diff
@@ -0,0 +1,35 @@
+const test = require('ava')
+const { scaffold, runScriptWithSES } = require('./scaffold.js')
+const { makeConfig } = require('./fixtures/main/webpack.config.js')
+const { isBuiltin } = require('node:module')
+const path = require('node:path')
+
+test.before(async (t) => {
+  const webpackConfig = makeConfig({
+    // TODO: replace hardcoded policy with generated one when isBuiltin gets implemented
```
weizman: remove todo?

```suggestion
```

packages/core/src/generatePolicy.js
```diff
@@ -314,6 +314,7 @@ function createModuleInspector(opts) {
     // get all requested names that resolve to isBuiltin
     const namesForBuiltins = Object.entries(moduleRecord.importMap)
       .filter(([, resolvedName]) => isBuiltin(resolvedName))
+      .filter((a) => console.log({ a, im: moduleRecord.importMap }) || 1)
```
weizman: what's this for?

packages/core/src/kernelCoreTemplate.js
```diff
@@ -211,16 +211,7 @@
 
       // create minimal selection if its a builtin and the whole path is not selected for
       if (!parentIsEntryModule && moduleData.type === 'builtin' && !parentPackagePolicy.builtin[moduleId]) {
-        const builtinPaths = (
```
weizman, line 214: why was this moved to the toolkit file?

packages/core/src/generatePolicy.js
naugtur: Doh. I missed that in cleanup.

packages/core/src/kernelCoreTemplate.js
naugtur: It's more shareable that way and the logic of what it's doing is better encapsulated there.

packages/webpack/test/fixtures/main/webpack.config.js
```diff
@@ -46,6 +46,9 @@ function makeConfig(lmOptions = {}) {
       }),
       new HtmlWebpackPlugin(),
     ],
+    resolve:{
+      fallback: { "crypto": false }
```
naugtur, line 50: guarantees `crypto` will be an ignored module instead of depending on default behavior that depends on other factors.

packages/core/src/endowmentsToolkit.js
```diff
@@ -130,6 +134,39 @@ function endowmentsToolkit({
     return targetRef
   }
 
+  /**
+   * @param {object} moduleNamespace
+   * @param {string} moduleId
+   * @param {LMPolicy.BuiltinPolicy} policyBuiltin
+   */
```
boneskull: - This appears to transform `moduleNamespace`.  Please note if it mutates `moduleNamespace` or not
- Adding a `@returns` is good practice wherever you can, but for public APIs, you especially want to declare the return type for performance reasons.

packages/core/src/endowmentsToolkit.js
```diff
@@ -130,6 +134,39 @@ function endowmentsToolkit({
...
+      const packagePathParts = packagePath.split('.')
+      if (moduleId === packagePathParts[0]) {
+        const packagePathWithoutPackage = packagePathParts.slice(1).join('.')
+        if (allowed === true) {
+          builtinPaths.push(packagePathWithoutPackage)
+        } else if (allowed === false) {
+          explicitlyBanned.push(packagePathWithoutPackage)
+        }
+      }
+    })
```
boneskull, line 214: please add comment describing the intent

packages/webpack/src/plugin.js
```diff
@@ -69,6 +69,9 @@ class LavaMoatPlugin {
     if (!options.policyLocation) {
       options.policyLocation = path.join('.', 'lavamoat', 'webpack')
```
boneskull: note that `path.join('.', ...anything)` is equivalent to `path.join(...anything)`

packages/webpack/src/plugin.js
```diff
@@ -221,11 +225,38 @@ class LavaMoatPlugin {
...
+          if (
+            m.type === JAVASCRIPT_MODULE_TYPE_DYNAMIC &&
+            // @ts-expect-error BAD TYPES
+            m.identifierStr?.startsWith('ignored')
+          ) {
+            return true
+          }
+          // This function must be very narrow, because it's gatekeeping policy generation too
+          return false
+        }
```
boneskull: - How does the type of an `import('webpack').Module` differ from an "ignored module"?
- Are there any functions accepting either type but whose output differs depending on the type?

packages/webpack/src/plugin.js
```diff
@@ -221,11 +225,38 @@ class LavaMoatPlugin {
...
+         * @param {string} moduleClass
+         * @returns {m is import('webpack').NormalModule} // TODO: this is not
+         *   true anymore, but there's no superclass of all reasonable module
+         *   types
          */
-        const isNormalModule = (m) => 'resource' in m
+        const isInspectableModule = (m, moduleClass) =>
+          'userRequest' in m ||
+          m.type?.startsWith('javascript') ||
+          ['ExternalModule'].includes(moduleClass)
```
boneskull, line 267: Alright, why is this not `moduleClass === 'ExternalModule'`?  this had better be good 😄

packages/webpack/src/plugin.js
```diff
@@ -221,11 +225,38 @@ class LavaMoatPlugin {
...
+          return false
+        }
+
         /**
          * @param {import('webpack').Module} m
-         * @returns {m is import('webpack').NormalModule}
+         * @param {string} moduleClass
+         * @returns {m is import('webpack').NormalModule} // TODO: this is not
+         *   true anymore, but there's no superclass of all reasonable module
+         *   types
```
boneskull, line 262: If this isn't doing anything useful, just return `boolean`

packages/webpack/src/plugin.js
```diff
@@ -248,34 +279,23 @@ class LavaMoatPlugin {
...
                 unenforceableModuleIds.push(moduleId)
               } else {
+                if (isInspectableModule(module, moduleClass)) {
+                  policyGenerator.inspectWebpackModule(
+                    module,
+                    compilation.moduleGraph.getOutgoingConnections(module)
+                  )
+                }
+
+                // typescript is complaining about the use of `resource` here, but it's actually there.
```
boneskull: would be more appropriate to use whatever the actual type is, instead of `any`, but nbd

packages/webpack/src/runtime/runtime.js
```diff
@@ -48,30 +49,56 @@ const stricterScopeTerminator = freeze(
 /**
  * Enforces the policy for resource imports.
  *
- * @param {string | undefined} requestedResourceId - The ID of the requested
- *   resource.
+ * @template T
```
boneskull: Should this be `@template {object} T`?  Is there any case wherein `T` could be a primitive?

packages/webpack/src/runtime/runtime.js
```diff
@@ -48,30 +49,56 @@ const stricterScopeTerminator = freeze(
...
-  requestedResourceId = '' + requestedResourceId
-  referrerResourceId = '' + referrerResourceId
-  // implicitly allow all for root and modules from the same package
+  // skip enforcing what we determined at build time we cannot
   if (
-    referrerResourceId === LAVAMOAT.root ||
-    requestedResourceId === referrerResourceId
+    LAVAMOAT.unenforceable.includes(specifier) ||
+    // implicitly allow all for root
+    referrerResourceId === LAVAMOAT.root
```
boneskull, line 85: is there a reason why `LAVAMOAT.unenforceable` is not a `Set`?  can this array get big enough to worry about?

packages/webpack/src/runtime/runtime.js
```diff
@@ -48,30 +49,56 @@ const stricterScopeTerminator = freeze(
...
   ) {
-    return
+    return wrappedRequire()
+  }
+  const referrerPolicy = LAVAMOAT.policy.resources[referrerResourceId] || {}
+  if (referrerPolicy.builtin) {
+    // @ts-expect-error - missing details in policy type, see TODO in types.js
```
boneskull: can we just use the policy type?

packages/webpack/src/runtime/runtime.js
```diff
@@ -48,30 +49,56 @@ const stricterScopeTerminator = freeze(
...
+    if (referrerPolicy.builtin[specifier]) {
+      return wrappedRequire()
+    }
+    if (keys(referrerPolicy.builtin).some((key) => key.startsWith(specifier))) {
```
boneskull: are we confident that `startsWith(specifier)` is narrow enough?

packages/webpack/src/runtime/runtime.js
```diff
@@ -48,30 +49,56 @@ const stricterScopeTerminator = freeze(
...
+      return getBuiltinForConfig(
+        wrappedRequire(),
+        specifier,
+        referrerPolicy.builtin
+      )
+    }
+  }
+  const requestedResourceId = findResourceId(specifier)
+  if (!requestedResourceId) {
+    throw Error(
```
boneskull, line 110: does this error reach the end-user? is it actionable, and if so, how?

packages/webpack/src/runtime/runtime.js
```diff
@@ -48,30 +49,56 @@ const stricterScopeTerminator = freeze(
...
+  }
+  // allow imports internal to the package
+  if (requestedResourceId === referrerResourceId) {
+    return wrappedRequire()
   }
-  const myPolicy = LAVAMOAT.policy.resources[referrerResourceId] || {}
   // @ts-expect-error - missing details in policy type, see TODO in types.js
-  if (myPolicy.packages && myPolicy.packages[requestedResourceId]) {
-    return
+  if (referrerPolicy.packages && referrerPolicy.packages[requestedResourceId]) {
```
```suggestion
  if (referrerPolicy.packages?.[requestedResourceId]) {
```

packages/webpack/src/runtime/runtime.js
```diff
@@ -150,12 +177,9 @@ const findResourceId = (moduleId) => {
...
   function (specifier) {
-    if (!LAVAMOAT.unenforceable.includes(specifier)) {
-      const requestedResourceId = findResourceId(specifier)
-      enforcePolicy(requestedResourceId, referrerResourceId)
-    }
-    // @ts-ignore - unknown this is the point here
-    return __webpack_require__.apply(this, arguments)
+    // @ts-expect-error - `this` is unknowable
+    const requireThat = __webpack_require__.bind(this, ...arguments)
+    return enforcePolicy(specifier, referrerResourceId, requireThat)
```
boneskull, line 206: You should be able to do something like

```js
  /** @this {object} */
  function (specifier) {
    // ...
  }
```

to avoid the `@ts-expect-error`

packages/webpack/test/e2e-nodejs.spec.js
```diff
@@ -0,0 +1,34 @@
...
+const { isBuiltin } = require('node:module')
+const path = require('node:path')
+
+test.before(async (t) => {
+  const webpackConfig = makeConfig({
+    generatePolicy: true,
+    emitPolicySnapshot: true,
+    diagnosticsVerbosity: 1,
+    isBuiltin,
+    policyLocation: path.resolve(__dirname, 'fixtures/main/policy-node'),
```
boneskull, line 13: windows

packages/webpack/test/fixtures/main/webpack.config.js
boneskull: wat

boneskull (commented):
there are a lot of TODOs in here

packages/webpack/src/plugin.js
naugtur: This stuff would be much nicer if there was one more or one less layer of inheritance there. Now there are 2 or 3 reasonable categories of module types and I would like to select one but I can't do that based on types as they all inherit straight from Module which barely has anything on it.

I'm pretty sure this code will grow over time as we get more experience with real configurations of real projects.

packages/webpack/src/plugin.js
naugtur: There used to be two items there and I'm expecting we may discover more in the future. Went with just one to avoid speculating so we wait till we stumble upon real examples instead of guessing which types could go here too.

packages/webpack/src/plugin.js
naugtur: It is useful as it helps with what fields are available. Module type barely has any fields. Normal module is the most common one, but not the only one we want to let pass through this function. But there's no other type that all modules with actual code in them inherit from and merging the ones we know is not future proof.

This makes types helpful despite being not exactly true.

packages/webpack/src/runtime/runtime.js
naugtur: T extends something didn't want to work for me at the time of writing this originally. Can fix now.

packages/webpack/src/runtime/runtime.js
naugtur: I guess now we can. This part is from before we had the type in main.

packages/webpack/src/runtime/runtime.js
naugtur: I did it for browser support. But didn't check, just went with gut feeling

packages/webpack/test/fixtures/main/webpack.config.js
naugtur: I discovered the ignored modules by testing with an example that had `crypto` in it. Depending on the content of your node modules (and I didn't explore what else) it can get ignored or filled with a package. This setting is here to stabilize that and guarantee the test is still checking our support for ignored modules.

boneskull:
@naugtur I rebased on `main` to fix the conflicts

boneskull:
@naugtur There is still a problem importing types from `ses` here, which is not our fault.  I wonder if this package should be ESM

packages/webpack/src/runtime/runtime.js
naugtur: the content of this array should be limited to the names of node builtins that webpack5 won't polyfill. It could be a Set, but the value is getting serialized to JSON so Set is inconvenient.

packages/webpack/src/runtime/runtime.js
naugtur: Good thinking. 
If the specifier was a package called `fs.exists` it would match.
I don't think this could work as an attack, but we should ensure specifier doesn't contain a dot.

packages/webpack/src/runtime/runtime.js
naugtur: yes. this error means "if you want it to work, you need to edit the policy for `referrerResourceId` to allow `specifier` as a builtin or package, depending on what you know it to be, also it seems the specifier was not found during policy generation"

packages/webpack/src/runtime/runtime.js
naugtur: ok, it was introduced in es2020, so maybe I can use it :D

naugtur:
Since this had main merged in, it's now gonna have to be squashed. disregard the failing commitlint.

packages/webpack/src/plugin.js
legobeat: If there is no appropriate enough type available already, better to make one that captures the behavior of relevance to us, rather than reaching for the closest available one? Just capturing the fields that you think are useful now, I think no need for it to be more complete or exact than `NormalModule`. Could be a union/intersection of adjacent existing webpack types.

> merging the ones we know is not future proof.

Would like to understand why this is

legobeat: (IMO: Overly vague type signatures for return values is generally fine but incomplete ones are not. Inverse for input parameters)

packages/webpack/src/runtime/runtime.js
```diff
@@ -201,7 +226,7 @@ const lavamoatRuntimeWrapper = (resourceId, runtimeKit) => {
     // The following seem harmless and are used by default: ['O', 'n', 'd', 'o', 'r', 's']
     const supportedRuntimeItems = ['O', 'n', 'd', 'o', 'r', 's']
     for (const item of supportedRuntimeItems) {
-      // @ts-expect-error - I'm not gonna do webppack's minified runtime typing
+      // @ts-ignore - I'm not gonna do webppack's minified runtime typing
```
legobeat: Why is this changed from `ts-expect-error` to `ts-ignore`? In general I see no reason to prefer the latter over the former (either we think the code should type, or we know it not to. `ts-ignore` would not let us know if the situation changes)

packages/webpack/test/fixtures/main/node.js
```diff
@@ -0,0 +1,3 @@
+import dir from 'nodejs-package'
+
```
legobeat, line 2: Nit: Missing newline at EOF

tsconfig.json
```diff
@@ -2,23 +2,26 @@
   "extends": "@tsconfig/node16/tsconfig.json",
   "files": [],
   "compilerOptions": {
-    "allowJs": true,
+    "allowJs": true
```
legobeat, line 5: `tsconfig.json` does not need to be valid JSON, we can have trailing commas and comments and stuff. There is still other non-json package tsconfig left after this. If we want to have them actual-JSON, I suggest to do that separate alongside a new linting enforcement?

legobeat (dismissed):

packages/webpack/src/runtime/runtime.js
naugtur: I wanted to indicate that it's not that the types are missing and could show up. It's that I don't intend to have this specific thing typechecked. I don't care either way, was trying to be more specific about the intent. can undo. Let me know if you think I should, knowing the intention behind the change.

tsconfig.json
naugtur: How is this a reason for blocking the entire PR?

naugtur: Also, I didn't do it. It was format on save from prettier. someone's gonna have to commit it or we'll all keep undoing.

packages/webpack/test/fixtures/main/node.js
boneskull: note: we do not automatically do this and we should.  iirc causes problems with the "dirty check", because of test fixture policies being regenerated.  we should either a) remove those from VCS or b) fix policy output such that it emits a newline at EOF.

packages/webpack/src/runtime/runtime.js
boneskull: use of `@ts-ignore` should violate [this rule](https://typescript-eslint.io/rules/ban-ts-comment/).

may be better just to wrap it in a `@type {any}`

tsconfig.json
boneskull: `tsconfig.json` is configured to be formatted as JSON, not JSONC (or whatever it is).  the automatic formatting is correct.

packages/core/src/endowmentsToolkit.js
```diff
@@ -17,35 +17,35 @@
 
 module.exports = endowmentsToolkit
 
-/**
- * @param {object} opts
- * @param {DefaultWrapperFn} [opts.createFunctionWrapper]
- */
+// module.exports.endowmentsToolkit = endowmentsToolkit
```
```suggestion
```

packages/core/src/endowmentsToolkit.js
```diff
@@ -589,3 +613,71 @@ function defaultCreateFunctionWrapper(sourceValue, unwrapTest, unwrapTo) {
  * @param {object} unwrapTo
  * @returns {(...args: any[]) => any}
  */
+
+/**
```
boneskull, line 753: I don't think it really matters, but I am curious why you wanted to extract all these typedefs from the function implementations in the first place.

packages/core/src/endowmentsToolkit.js
```diff
@@ -589,3 +613,71 @@ function defaultCreateFunctionWrapper(sourceValue, unwrapTest, unwrapTo) {
...
+// /**
+//  * @callback EndowmentsToolkitFactory Creates an object with utility functions
+//  *   for generating the endowments object based on a `globalRef` and a
+//  *   `PackagePolicy`.
+//  * @param {object} [opts] - Options for creating the endowments toolkit.
+//  * @param {DefaultWrapperFn} [opts.createFunctionWrapper] - Custom function for
+//  *   wrapping functions. Defaults to `defaultCreateFunctionWrapper`.
+//  * @returns {EndowmentsToolkit} An object containing utility functions for
+//  *   generating endowments.
+//  */
```
```suggestion
```

packages/webpack/src/buildtime/diagnostics.js
```diff
@@ -10,17 +10,18 @@ module.exports = {
...
   },
   /**
    * @param {number} verbosity
    * @param {any} content
    */
-  rawDebug: (lvl, content) => {
-    if (level >= lvl) {
+  rawDebug: (verbosity, content) => {
+    if (level >= verbosity) {
+      // @ts-ignore - trust me, the method exists
```
boneskull: a better fix--though **not at all necessary**--would be to create a `node.d.ts` file and extend the `NodeJS.Process` interface with the definition for `_rawDebug` 😄

packages/webpack/src/plugin.js
boneskull: also, while I trust `process._rawDebug` won't magically disappear, I'm less trusting of webpack's ostensibly-undocumented  API surface (`identifierStr`?).  is there anything else we could use here?  can we maintain our own `WeakMap` instead?

packages/webpack/src/plugin.js
boneskull: okay, so if the type guard is doing actual work--good, just keep it that way.  though I'd probably lean more in the direction of what @legobeat said; create a new typedef (intersecting Module, if appropriate) containing the properties we care about, and use that.  not a big enough deal to block over though

boneskull (approved):

packages/webpack/src/runtime/runtime.js
naugtur: If that's the desired way to indicate "I explicitly don't care what type it is" - I'm fine with that.

packages/core/src/endowmentsToolkit.js
naugtur: because I was struggling to export the whole thing. extraction can now be undone

packages/webpack/src/plugin.js
naugtur: identifierStr is the least obscure way I found to check that a module is ignored. (ignored by webpack itself - specific internal webpack meaning)
This solution is relying on one much more internal thing and if webpack starts breaking things, we'll know (they're not keen on doing that tho)
I think we're good until webpack 6 but I'm planning to look into a less tight integration with webpack over time, opening up the possibility we could support rspack too.

legobeat (commented):
@naugtur FYI: this needs conflict-resolution of `packages/core/src/types.ts` before being mergable.

packages/core/src/endowmentsToolkit.js
naugtur: note: we ended up pairing on this

packages/webpack/src/buildtime/diagnostics.js
naugtur: I prefer not to, it's not something we need in code completion or leaking form our package later ;)

packages/webpack/src/plugin.js
naugtur: There's a bunch of types and non-breaking changes to webpack may add more and there's no good type they all inherit from. the base Module type doesn't have the field. Almost all other SomethingModule types do.

packages/webpack/src/plugin.js
naugtur: We need to talk about this synchronously and open a new PR.
I got the typings to a state that's decent enough to encourage improvements.

packages/core/src/endowmentsToolkit.js
```diff
@@ -172,13 +168,65 @@ function endowmentsToolkit({
...
+    /** @type {string[]} */
+    const explicitlyBanned = []
+
+    // Collect the same paths information as getEndowmentsForConfig to enable
+    // matching behavior of policy between globals and builtins
+    Object.entries(policyBuiltin).forEach(([packagePath, allowed]) => {
+      const packagePathParts = packagePath.split('.')
+      if (moduleId === packagePathParts[0]) {
+        const packagePathWithoutPackage = packagePathParts.slice(1).join('.')
+        if (allowed === true) {
```
weizman, line 208: can `allowed` be non Boolean? If not, this `if` (and the next one) can be simplified

packages/core/src/endowmentsToolkit.js
```diff
@@ -172,13 +168,65 @@ function endowmentsToolkit({
...
+  }
+
+  /**
+   * @template T
+   * @param {T | null} value
+   * @returns {value is null}
+   */
+  function isEmpty(value) {
+    return !value
+  }
```
weizman: I'm pretty sure these lines exist within this file already

weizman (approved):
lgtm (but...)

packages/core/src/endowmentsToolkit.js
naugtur: we intentionally care about them being explicitly boolean so that things like `'write'` value for globals are possible to introduce.

packages/core/src/endowmentsToolkit.js
naugtur: yup, that's a conflict resolution artifact.

packages/webpack/tsconfig.json
```diff
@@ -6,5 +6,9 @@
     "checkJs": true,
     "rootDir": "src",
     "outDir": "types",
+    "paths": {
+      "lavamoat-core": ["../core"]
+    }
   },
+  "references": [{ "path": "../core/tsconfig.json" }]
```
legobeat, line 14: Could this be behind the failing type-check in CI? https://github.com/LavaMoat/LavaMoat/pull/1166/files#diff-94c0d6243364db60753823cb2b811194dc7fcba7610f63ec9b5b691353457783R128

(Error causes should be available from Node 16.9.0 so `MIN_NODE_VERSION` set to `16.20.0` in `.eslintrc.js` does not seem related)

~Maybe~ Also not related:
- #1230

legobeat (changes requested):
Failing CI

packages/webpack/test/e2e-nodejs.spec.js
naugtur: wait, internet says path.resolve normalizes slashes. What windows work do I need here?

naugtur:
> Failing CI

I know, it's after a rebase on main. This branch is a pain to rebase.

packages/webpack/tsconfig.json
naugtur: Error cause seems to require ES2022 and we're targeting earlier. 

I think since we're dropping node 16 we can update https://github.com/LavaMoat/LavaMoat/blob/25405141525e75c214f553e4c082c6b2ed79cfdd/.config/tsconfig.workspace.json to extend node18 not node16 by default and narrow down in packages we want to keep node16 support.

I'll override es version in webpack for this branch and see if it helps.

leotm (approved):

packages/webpack/tsconfig.json
naugtur: Seems to work. 

Now we know dropping node16 is not done until we depend on @tsconfig/node18

And the way our tsconfig is set up doesn't easily allow dropping support selectively.

legobeat: Seems fine to set the esversion selectively for webpack plugin (it supports min 18 already like you note and that should cover es2022)

legobeat (dismissed):

packages/webpack/tsconfig.json
naugtur: I looked at tsconfig node18 and this is what it looks like:

```
{
  "$schema": "https://json.schemastore.org/tsconfig",

  "_version": "18.2.0",

  "compilerOptions": {
    "lib": ["es2023"],
    "module": "node16",
    "target": "es2022",

    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "moduleResolution": "node16"
  }
}
```

I guess we might need to consult with MM team if we want to update the stuff that might work in the browser to es2023 lib target.

packages/core/src/endowmentsToolkit.js
```diff
@@ -26,40 +26,35 @@ module.exports = endowmentsToolkit
 // Exports for testing
 module.exports._test = { instrumentDynamicValueAtPath }
 
-/**
- * @param {object} opts
- * @param {DefaultWrapperFn} [opts.createFunctionWrapper]
- * @param {boolean} [opts.handleGlobalWrite]
- * @param {Set<string>} [opts.knownWritableFields] - List of globals that can be
- *   mutated later
- */
```
boneskull: Why did this get removed?

packages/core/src/endowmentsToolkit.js
```diff
@@ -26,40 +26,35 @@ module.exports = endowmentsToolkit
...
     copyValueAtPath,
     applyGetSetPropDescTransforms,
     applyEndowmentPropDescTransforms,
-    copyWrappedGlobals,
-    createFunctionWrapper,
   }
 
   /**
-   * Creates an object populated with only the deep properties specified in the
-   * packagePolicy
```
boneskull, line 53: Can we keep the description?

packages/core/src/endowmentsToolkit.js
```diff
@@ -172,13 +168,65 @@ function endowmentsToolkit({
...
+          explicitlyBanned.push(packagePathWithoutPackage)
+        }
+      }
+    })
+    const moduleNamespaceView = makeMinimalViewOfRef(
+      moduleNamespace,
+      builtinPaths.sort(),
+      undefined,
+      undefined,
+      explicitlyBanned
```
boneskull, line 220: Why is `builtinPaths` sorted, but `explicitlyBanned` is not?

packages/core/src/endowmentsToolkit.js
```diff
@@ -172,13 +168,65 @@ function endowmentsToolkit({
...
+    return moduleNamespaceView
+  }
+
+  /**
+   * @param {string} visited
+   * @param {string} next
+   */
+  function extendPath(visited, next) {
+    // FIXME: second part of this conditional should be unnecessary
+    if (!visited || visited.length === 0) {
```
boneskull: Just remove it, imo.

```suggestion    
    if (!visited) {
```

boneskull (commented):
I am trying to review this from VSCode, but it is extremely sluggish, so I'm going to just pick up again in the browser.

packages/core/src/endowmentsToolkit.js
boneskull: I realize you only moved this code, but...

The template is not necessary here, since it's unused in the return type:

```suggestion
  /**
   * @param {object|null} value
   * @returns {value is null}
   */
  function isEmpty(value) {
    return !value
  }
```

The above change may cause some type errors, which is completely reasonable.

Also, instead of calling `isEmpty(value)`, it may be more straightforward to just check if `value === null` or `!value`, assuming we know `value` is either an `object` or `null`.

packages/core/src/endowmentsToolkit.js
```diff
@@ -704,3 +735,39 @@ function defaultCreateFunctionWrapper(sourceValue, unwrapTest, unwrapTo) {
  * @param {object} unwrapTo
  * @returns {(...args: any[]) => any}
  */
+
+/**
+ * @callback CopyWrappedGlobals
+ * @param {object} globalRef
+ * @param {Record<PropertyKey, any>} target
```
boneskull: You may want to add a note this type using `any` (instead of `unknown`) is _intentional_ (assuming that is true and `unknown` would not work) because a `CopyWrappedGlobals` will read/write `target`.

packages/core/src/endowmentsToolkit.js
```diff
@@ -704,3 +735,39 @@ function defaultCreateFunctionWrapper(sourceValue, unwrapTest, unwrapTo) {
...
+ * @param {string[]} explicitlyBanned
+ * @param {object} sourceRef
+ * @param {object} targetRef
+ * @param {object} [unwrapTo]
+ * @param {object} [unwrapFrom]
+ * @returns {void}
+ */
+
+/**
+ * @callback ApplyGetSetPropDescTransforms
```
boneskull, line 785: Please add descriptions to all of these callbacks 😄

packages/webpack/src/buildtime/aa.js
```diff
@@ -164,7 +164,7 @@ exports.generateIdentifierLookup = ({
       const translatedPolicy = {
         ...policy,
         resources: Object.fromEntries(
-          Object.entries(policy.resources)
+          Object.entries(policy.resources || {})
```
boneskull: Suggestion: 

```js
const {resources = Object.create(null)} = policy;
```

...and work with that instead.

packages/webpack/src/buildtime/assemble.js
```diff
@@ -41,4 +51,13 @@ const assembleRuntime = (KEY, runtimeModules) => {
   return assembly
 }
 
+/**
+ * @typedef RuntimeModule
+ * @property {string} [file]
+ * @property {any} [data]
```
boneskull:
```suggestion
 * @property {unknown} [data]
```

Unless you are writing to `data` or reading its properties, use `unknown`

packages/webpack/src/buildtime/diagnostics.js
boneskull: OK then, 

```suggestion
      // @ts-expect-error - trust me, the method exists
```

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -30,6 +29,9 @@ module.exports = {
    * @param {string} opts.location - Where to read/write the policy files
    * @param {boolean} [opts.emit] - Whether to emit the policy snapshot as an
    *   asset
+   * @param {(specifier: string) => boolean} opts.isBuiltin - A function that
```
boneskull: (suggestion) Make an `IsBuiltinFn` type

packages/webpack/src/excludeLoader.js
```diff
@@ -1 +1 @@
-module.exports = (source) => source
+module.exports = (/** @type {any} */ source) => source
```
boneskull, line 7: - Is this really `any`?
- What is this module for?  Can you add a module-level description?  e.g.
    ```js
    /**
     * This does that to do the other thing
     *
     * @module
     */
    ```

packages/webpack/src/plugin.js
boneskull: fine by me

packages/webpack/src/plugin.js
```diff
@@ -490,3 +513,45 @@ class LavaMoatPlugin {
...
+ * @property {import('lavamoat-core').LavaMoatPolicy} [policy] - LavaMoat policy
+ *   object - if programmatically created
+ * @property {boolean} [runChecks] - Check resulting code with wrapping for
+ *   correctness
+ * @property {(specifier: string) => boolean} isBuiltin - A function that
+ *   determines if the specifier is a builtin of the runtime platform e.g.
+ *   node:fs
+ */
+
+// Provided inline because import('ses') won't work in jsdoc of a cjs module
```
boneskull, line 541: I think this might work now. It used to be broken, for sure.

packages/webpack/src/runtime/runtime-namespace.ts
```diff
@@ -0,0 +1,14 @@
+import { EndowmentsToolkitFactory, LavaMoatPolicy } from 'lavamoat-core'
+import { LavaMoatPluginOptions } from '../plugin'
+
+type RequiredProperty<T, K extends keyof T> = Omit<T, K> & Required<Pick<T, K>>
+export interface RuntimeNamespace {
+  root: string
+  idmap: [string, string[]][]
+  unenforceable: string[]
+  options: LavaMoatPluginOptions
+  policy: RequiredProperty<LavaMoatPolicy, 'resources'>
```
boneskull, line 10: What happens, exactly, if `resources` is not defined?

packages/webpack/src/runtime/runtime.js
```diff
@@ -65,30 +66,57 @@ const stricterScopeTerminator = freeze(
...
-  if (myPolicy.packages && myPolicy.packages[requestedResourceId]) {
-    return
+  const referrerPolicy = LAVAMOAT.policy.resources[referrerResourceId] || {}
+  if (referrerPolicy.builtin) {
+    if (referrerPolicy.builtin[specifier]) {
+      return wrappedRequire()
+    }
+    if (
+      !specifier.includes('.') &&
+      keys(referrerPolicy.builtin).some((key) => key.startsWith(specifier))
```
```suggestion
      keys(referrerPolicy.builtin).some((key) => key.startsWith(`${specifier}.`))
```

Yeah?

packages/webpack/src/runtime/runtime.js
```diff
@@ -153,10 +181,14 @@ const findResourceId = (moduleId) => {
 }
 
 /**
- * @callback WrappedRequire
+ * @callback WebpackRequire
  * @param {string} specifier
  */
 
+/**
+ * @typedef {WebpackRequire & Record<string, any>} WrappedRequire
```
```suggestion
 * @typedef {WebpackRequire & Record<string, unknown>} WrappedRequire
```

Would this work?  or

```suggestion
 * @typedef {WebpackRequire & Record<PropertyKey, unknown>} WrappedRequire
```

packages/webpack/src/runtime/runtime.js
```diff
@@ -218,12 +246,9 @@ const lavamoatRuntimeWrapper = (resourceId, runtimeKit) => {
     // The following seem harmless and are used by default: ['O', 'n', 'd', 'o', 'r', 's']
     const supportedRuntimeItems = ['O', 'n', 'd', 'o', 'r', 's']
```
boneskull, line 249: image

packages/webpack/test/e2e-nodejs.spec.js
boneskull: you're right.  I'm not sure how I missed that. 😬 

basically the only thing we need to worry about for windows is _bare path strings_.  anything going thru `node:path` looks to be normalized.

packages/webpack/tsconfig.json
```diff
@@ -2,9 +2,14 @@
   "extends": "../../.config/tsconfig.workspace.json",
   "include": ["src"],
   "compilerOptions": {
+    "lib": ["es2022"],
```
boneskull, line 5: hmm, what are we using where this is needed?

I think suggests a newer version of Node.js than we officially support

packages/webpack/src/plugin.js
```diff
@@ -228,11 +233,38 @@ class LavaMoatPlugin {
...
+          if (
+            m.type === JAVASCRIPT_MODULE_TYPE_DYNAMIC &&
+            // @ts-expect-error BAD TYPES
+            m.identifierStr?.startsWith('ignored')
+          ) {
+            return true
+          }
+          // This function must be very narrow, because it's gatekeeping policy generation too
+          return false
+        }
```
```suggestion
        const isIgnoredModule = (m) => {
          return Boolean(m.type === JAVASCRIPT_MODULE_TYPE_DYNAMIC &&
            // @ts-expect-error BAD TYPES
            m.identifierStr?.startsWith('ignored'))
        }
```

packages/webpack/src/plugin.js
```diff
@@ -228,11 +233,38 @@ class LavaMoatPlugin {
           JAVASCRIPT_MODULE_TYPE_ESM,
         ]
 
+        // Webpack has a concept of ignored modules
+        // When a module is ignored a carveout is necessary in policy enforcement for it because the ID that webpack creates for it is not exactly helpful.
+        // example outcome in the bundle: `const nodeCrypto = __webpack_require__(/*! crypto */ "?0b7d");`
+        // Sadly, even treeshaking doesn't eliminate that module. It's left there and failing to work when reached by runtime policy enforcement.
+        // Below is the most reliable way I've found to date to identify ignored modules.
```
boneskull: (suggestion) move this into the docstring for `isIgnoredModule` under a `@remarks` tag, or something

boneskull (approved):
Nothing overly concerning here

boneskull:
> **New dependencies detected.** Learn more about [Socket for GitHub ↗︎](https://socket.dev?utm_medium=gh)
> 
> Package	New capabilities	Transitives	Size	Publisher
> [npm/nodejs-package@1.0.0](https://socket.dev/npm/package/nodejs-package/overview/1.0.0)	None	[`0`](https://socket.dev/npm/package/nodejs-package/dependencies/1.0.0)	0 B	
> [**View full report↗︎**](https://socket.dev/dashboard/org/LavaMoat/diff/986c2ace-9de4-4e8c-aff7-ccc77862eca5/5501b89c-cc5b-4d27-9635-640df9f223eb)

@naugtur Probably want to add some more ignored paths to Socket

packages/webpack/tsconfig.json
legobeat: I think it may just be for the sake of allowing passing `cause` option to the Error constructor. Doing so is safe in older runtimes too (the extra argument gets ignored) so an option could just be to ignore the place where that is done?

packages/core/src/endowmentsToolkit.js
naugtur: good catch. looks like a conflict resolution artifact. we moved the types back and forth in this branch and rebasing on main was rough.

packages/core/src/endowmentsToolkit.js
naugtur: sure, it was git who removed it.

packages/core/src/endowmentsToolkit.js
naugtur: It was there before and we didn't know why. FIXME was added in a conversation where we weren't confident enough to remove it. I don't want to do it as part of this PR.

packages/webpack/src/runtime/runtime.js
naugtur: nice!

packages/webpack/src/runtime/runtime.js
naugtur: they are strings. Im overiding them so not sure if unknown works, but I'll commit and see.

packages/webpack/src/runtime/runtime.js
naugtur: webpack runtime has bits that are minified by hand ;)

packages/webpack/tsconfig.json
naugtur: we dropped node 16 but it's still in the workspace config for some pkgs that still support it. there's a thread about it somewhere here.

packages/webpack/src/runtime/runtime-namespace.ts
naugtur: I need to make it required for webpack and drop the other field. one of the types fixes I want to do in separate PR soon

packages/core/src/endowmentsToolkit.js
naugtur: AFAIR builtinPaths needs to be sorted for the depth to resolve correctly when iterating over it. explicitlyBanned is only for lookup.

packages/webpack/src/excludeLoader.js
naugtur: It's any, because only the first loader gets a string, all other loaders may use whatever object another loader in a sequence can consume and the final one must return a string or Source type.

wrote the description.

packages/webpack/src/plugin.js
naugtur: *grumble* the comment contains a `*/` in example.

packages/webpack/src/plugin.js
naugtur: in the next PR where we also change the Policy types.

packages/webpack/src/runtime/runtime.js
```diff
@@ -166,13 +198,10 @@ const findResourceId = (moduleId) => {
...
 const wrapRequireWithPolicy = (__webpack_require__, referrerResourceId) =>
-  function (specifier) {
-    if (!LAVAMOAT.unenforceable.includes(specifier)) {
-      const requestedResourceId = findResourceId(specifier)
-      enforcePolicy(requestedResourceId, referrerResourceId)
-    }
-    // @ts-ignore - unknown this is the point here
-    return __webpack_require__.apply(this, arguments)
+  /** @this {object} */
+  function (specifier, ...rest) {
```
legobeat, line 204: This is causing [build type error](https://github.com/LavaMoat/LavaMoat/actions/runs/10182291891/job/28164447499?pr=1166): 

```
Error: packages/webpack/src/runtime/runtime.js(202,3): error TS2322: Type '(this: object, specifier: string) => object' is not assignable to type 'WrappedRequire'.
  Type '(this: object, specifier: string) => object' is not assignable to type 'Record<string, unknown>'.
    Index signature for type 'string' is missing in type '(this: object, specifier: string) => object'.
```

If I get it right, seems like it's due to [`object`](https://github.com/LavaMoat/LavaMoat/pull/1166/files#diff-aafaf2ed51103df46104ec90ce67636ce56702818591644fc114c3b512639a16R201) lacking the indexing required by [`Record<string, unknown>`](https://github.com/LavaMoat/LavaMoat/pull/1166/files#diff-aafaf2ed51103df46104ec90ce67636ce56702818591644fc114c3b512639a16R189).

Could it be addressed by asserting `/** @this {Record<string, unknown>} */`, or would we need to make a new typedef here too?

packages/webpack/tsconfig.json
legobeat: Maybe we could just drop the `cause` usage for now to sidestep this entirely?

packages/webpack/src/runtime/runtime.js
naugtur: It's the second time in this PR I fixed a type and typescript didn't warn me in the IDE because some things didn't refresh somewhere :(

packages/webpack/tsconfig.json
naugtur: I don't want to compromise on JavaScript features because of typescript configuration complexity. I'd rather have idiomatic JavaScript and messy tsconfig than the other way around.

packages/webpack/src/runtime/runtime.js
naugtur: it was unrelated to the this value, it was the return type being too strict. Iterating on it required restarting TS server :(

naugtur:
Waiting for a confirmation from @legobeat 

If I'm not available and you want to move with the release, feel free to rebase&merge (I made sure commit history is good enough already)

legobeat (approved):

naugtur:
so teamwork, such reviewed. wow. :dog:

PR #1172, opened 2024-04-29
fix(webpack): fix progress order errors covering compilation errors, …

naugtur:
…improve `__webpack_require__.nmd` compatibility

Minor issues I stumbled upon while working on something else.

### DIFF
--- a/packages/webpack/src/buildtime/progress.js
+++ b/packages/webpack/src/buildtime/progress.js
@@ -6,11 +6,14 @@ const diag = require('./diagnostics')
  *   matches expectedStep
  * @property {(expectedStep: string) => boolean} done - Checks if expectedStep
  *   was already reported.
- * @property {(expectedStep: string) => void} assertDone - Throws if
+ * @property {(expectedStep: string) => void} assertDone - Reports an error if
  *   expectedStep was not already reported.
  * @property {(step: string) => void} report - Moves progress forward if step
  *   passed is the next step. no-op if current step (reporting progress is
  *   idempotent)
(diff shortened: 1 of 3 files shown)


packages/webpack/src/buildtime/progress.js
```diff
@@ -81,9 +94,18 @@ module.exports = function progress({ steps }) {
...
+        `LavaMoatPlugin Plugin: Expected '${query}' to be done, but we're at '${steps[currentStep]}'`
+      )
     )
   }
+  /**
+   * @param {Error[]} errors
+   */
+  API.reportErrorsTo = (errors) => {
+    errors.push(...compilationErrors)
+    compilationErrors = errors
```
legobeat, line 108: Are you explicitly intending to permutate the incoming `errors` here?
If not: 
```suggestion
    compilationErrors = [...errors, ...compilationErrors]
```

naugtur: I am. That's how compilation errors are aggregated in webpack - I have to push to `compilation.errors` to get them reported later instead of throwing (which prevents other compilation errors from getting surfaced to the user)

packages/webpack/src/runtime/runtime.js
```diff
@@ -229,6 +229,7 @@ const lavamoatRuntimeWrapper = (resourceId, runtimeKit) => {
     policyRequire.nmd = (moduleReference) => {
       if (moduleReference === module) {
         module = __webpack_require__.nmd(module)
+        return module
```
legobeat, line 232: For the future would be nice to see separate changes in separate commits, if not PRs.

legobeat (approved):

PR #1201, opened 2024-06-04
 chore(allow-scripts): Indicate peerDependency on @lavamoat/preinstall-always-fail

naugtur:
Reverts https://github.com/LavaMoat/LavaMoat/pull/1200

Bring back the peerDependency after upcoming release

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18351,6 +18351,9 @@
       },
       "engines": {
         "node": "^16.20.0 || ^18.0.0 || ^20.0.0 || ^22.0.0"
+      },
+      "peerDependencies": {
+        "@lavamoat/preinstall-always-fail": "*"
       }
     },
     "packages/browserify": {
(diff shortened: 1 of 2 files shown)


naugtur:
BTW. Why would that be a semver major? 
AFAIR peerDependencies are treated as optional by package maangers. (not 100% sure about NPM, because they went back and forth on that)

legobeat (approved):

legobeat:
> BTW. Why would that be a semver major? AFAIR peerDependencies are treated as optional by package maangers. (not 100% sure about NPM, because they went back and forth on that)

I don't believe it's necessary to treat this as breaking wrt semver (not as much because of how package managers mostly behave by default - and btw I can report NPM does only warn at least on v9+, and yarn only warns as of v3+ - but because it's already been communicated as an effective peerDependency in docs). But I can also see the argument for and if we want to be extra conservative here we could.

PR #1202, opened 2024-06-07
feat: handle excluding modules in policy generation

naugtur:
Our method for ignoring stuff we're not supposed to process was not consistently applied to the policy generation pass.
This fixes it.


packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -87,16 +89,18 @@ module.exports = {
        * @param {Iterable<import('webpack').ModuleGraphConnection>} connections
        */
       inspectWebpackModule: (module, connections) => {
+        if (isExcludedUnsafe(module)) return
```
naugtur, line 94: here we want to ignore all stuff marked as intended to be ignored, no need to protect ourselves from attempts to self-ignore. We're getting more inputs to inspect than to bundle for reasons I didn't research fully.

legobeat (commented):
Change itself looks good!

1. Could we add a test-case for the scenario covered by this PR?
2. For my understanding of context: what makes this `feat` as opposed to `fix`?

naugtur:
re 1 - I didn't have a good idea for a compact fixture that would exhibit the issue at the time. I found this in a real project. I'll try to come up with something. PRs welcome ;) 

re 2 - It's an improvement in ecosystem compatibility of the policy generation functionality, but I agree from the user perspective it's a fix.

naugtur:
@legobeat updated with a test case. Instead of reproducing a case where only policy was failing, I made a stringer but more obvious one. 
Fixed error propagation btw, because the error we're avoiding by exclude here was an uncaught error. The test checks behaviors with and without exclude

packages/webpack/src/buildtime/policyGenerator.js
legobeat: > here we want to ignore all stuff marked as intended to be ignored, no need to protect ourselves from attempts to self-ignore.

Something like this would be a great inline comment!

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -111,7 +115,14 @@ module.exports = {
           //ast: module._ast, - would have to translate to babel anyway
         })
 
-        moduleInspector.inspectModule(moduleRecord)
+        try {
+          moduleInspector.inspectModule(moduleRecord)
+        } catch (e) {
+          throw new Error(
+            `LavaMoatPlugin: Failed to inspect module ${module.userRequest} for policy generation. ${e.message} \n If the file is not intended to be valid JavaScript, consider excluding it using LavaMoat.exclude loader.`,
```
legobeat:
```suggestion
            `LavaMoatPlugin: Failed to inspect module ${module.userRequest} for policy generation:\n ${e.message}\n If the file is not intended to be valid JavaScript, consider excluding it using LavaMoat.exclude loader.`,
```

packages/webpack/src/plugin.js
```diff
@@ -452,7 +463,9 @@ class LavaMoatPlugin {
              * @returns {plugin is import('webpack').WebpackPluginInstance}
              */
             (plugin) =>
-              !!plugin && typeof plugin === 'object' && plugin.constructor.name === 'HtmlWebpackPlugin'
+              !!plugin &&
+              typeof plugin === 'object' &&
+              plugin.constructor.name === 'HtmlWebpackPlugin'
           )
```
legobeat, line 469:
```suggestion
              typeof plugin === 'object' &&
              plugin.constructor.name === 'HtmlWebpackPlugin'
          )
```
the typeof should imply truthiness?

naugtur:
@legobeat all resolved.

legobeat (approved):

packages/webpack/src/buildtime/generator.js
```diff
@@ -112,10 +111,7 @@ exports.wrapGeneratorMaker = ({
       // only accept exclude loader from config, not inline.
       // Inline loader definition uses a `!` character as a loader separator.
       // Defining what to exclude should only be possible in the config.
```
weizman: Comment should move too, right?

packages/webpack/src/plugin.js
```diff
@@ -248,66 +250,75 @@ class LavaMoatPlugin {
...
-              )
+            diag.rawDebug(4, { knownPaths })
+            PROGRESS.report('pathsCollected')
+
+            diag.rawDebug(2, 'writing policy')
+            // use the generated policy to save the user one additional pass
+            // getting the policy also writes all files where necessary
+            const policyToApply = policyGenerator.getPolicy()
+
+            assert(
```
weizman, line 297: why use assert all of the sudden? seems like everywhere else under `pkgs/webpack` we just throw Errors

packages/webpack/src/plugin.js
```diff
@@ -452,7 +463,8 @@ class LavaMoatPlugin {
              * @returns {plugin is import('webpack').WebpackPluginInstance}
              */
             (plugin) =>
-              !!plugin && typeof plugin === 'object' && plugin.constructor.name === 'HtmlWebpackPlugin'
+              typeof plugin === 'object' &&
```
weizman, line 467: former check covered a scenario where `plugin` is `null` thus avoiding a potential unhandled exception, isn't that important?

weizman (commented):
looks good, a few comments though

packages/webpack/src/plugin.js
naugtur: It's a change suggested in review :see_no_evil: 
I accepted the suggestion and checked that typescript did not complain. But that was a lag in processing. it is complaining now.

packages/webpack/src/plugin.js
naugtur: It's more succinct and explains the intention better. Some of our errors should be calls to `assert` and I'm hoping to move in that direction.

packages/webpack/src/buildtime/generator.js
naugtur: yes!

packages/webpack/src/plugin.js
legobeat: I was looking a bit for an appropriate linting rule now but didn't come up with anything good So I guess we just have to keep in mind that `typeof null === 'object'`. Good catch @weizman .

naugtur: I knew it was there for that reason but saw typescript was fine with it and moved on. turns out it was just a lag to display a squiggle. It was complaining now.

PR #1210, opened 2024-06-26
troubleshooting examples

naugtur:
These examples were used in the recording of a tutorial video for the website and can be used to teach how to go about solving common errors under LavaMoat


legobeat:
There are existing examples under `packages/node/examples` (#1248) - feels like they should be merged under either destination?

naugtur:
> There are existing examples under `packages/node/examples` (#1248) - feels like they should be merged under either destination?

they are in need of updating to a point of redoing. consider them legacy.

legobeat:
> > There are existing examples under `packages/node/examples` (#1248) - feels like they should be merged under either destination?
> 
> they are in need of updating to a point of redoing. consider them legacy.

Hmm, does that go for the `browserify` ones too? Is there any antipattern or was just getting them running OK? If not I would still like to pick one from examples either under `/docs/examples` or `packages/*/examples`, not both. The reasoning being that a user ending up on one is likely to miss the existence of the other.

- For consideration: https://github.com/LavaMoat/LavaMoat/pull/1270

boneskull (approved):

boneskull:
@naugtur 

- Renovate should automatically ignore these, but the Socket config might need an update.
- We may want to link to this dir from the docs site and/or `README.md`.

PR #1213, opened 2024-07-06
chore: corepack packageManager support

legobeat:
- set project packageManager to `npm@9.9.3`
    Makes corepack do the right thing when active locally
  - ci: set `COREPACK_ENABLE_PROJECT_SPEC=0` when running cross-package-manager integration tests
- ci: use `corepack enable` to use project npm version
    Prefers corepack over `npm i -g`.
- `yarn-plugin-allow-scripts`: remove `packageManager` field from manifest
  - The package is still managed by npm as part of this monorepo
  - This is required for corepack to not throw errors regarding the mismatch during testing
- `yarn-plugin-allow-scripts`: add `engines.yarn` set to `^3.1.1`
  - v4 support pending #1218


packages/node/test/prepare.js
```diff
@@ -24,6 +24,14 @@ const exec = promisify(execFile)
 const LAVAMOAT_PM = process.env.LAVAMOAT_PM ?? 'npm@next-9'
 const PROJECTS_DIR = path.join(__dirname, 'projects')
 
+/**
+ * Environment variables for corepack support
+ */
+const COREPACK_ENV = {
+  // Corepack refuses to run package managers other than the one declared in packageManager by default
```
legobeat, line 31: Note that this does not affect users of the package - only needed when running inside the project, ie for devs and builds.

naugtur (approved):

PR #1218, opened 2024-07-07
fix(yarnpkg-allow-scripts): yarn berry 

legobeat:
- Fix allow-scripts yarn plugin to properly block preinstall scripts
  - The existing hooks run only after the fact
  - Control execution of lifecycle scripts in dependencies even if they contain a yarn v1 `yarn.lock` 
- Bundle now includes `@lavamoat/allow-scripts` - no extra installation necessary
- Update yarn packages from v3 to v4
- Add support for yarn v4
- BREAKING: Remove support for yarn v3
- BREAKING: Drop support for Node.js v16

#### Blocked by
- [x] #1383 
- [x] #1219
- [x] #1230 
- [x] #1317

#### TODO
- [x] tests
- [x] squash commits and clean up commit messages
- [x] windows/macos test compatibility

---


legobeat:
I can not reproduce the test error locally. Maybe this is one of those cases where yarn behaves differently when it detects running in GitHub Actions...

legobeat:
Still unable to reproduce the failing unit test in CI.

legobeat:
Could use some help getting the tests to run on Windows...

packages/yarn-plugin-allow-scripts/test/projects/protected-unconfigured/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "protected",
+  "private": true,
+  "packageManager": "yarn@4.5.0",
+  "dependencies": {
+    "allowed-dep": "git+https://github.com/LavaMoat/foowriter#commit=d63034fbbbf84978e664187829f599ec8ad4f519",
+    "disallowed-dep": "git+https://github.com/LavaMoat/foowriter#commit=c11948f19858bfb0796cea7c4c80c124133f2732"
```
legobeat: all different commits to [`foowriter`](https://github.com/LavaMoat/foowriter) but the sources are identical, with the only difference being the `version`. If we reuse the same package, yarn may reuse cached result from a prior instance of the same git url.

packages/yarn-plugin-allow-scripts/README.md
```diff
@@ -1,15 +1,29 @@
...
 ```
 
+#### Yarn v3
+
+The current version of the plugin does not support Yarn v3. Yarn v3 users are encouraged to migrate to an actively supported package manager such as Yarn v4.
+
+A legacy version supporting Yarn v3 can still be installed from the `yarn-plugin-allow-scripts-yarn3` tag:
+
+```sh
+yarn plugin import https://raw.githubusercontent.com/LavaMoat/LavaMoat/yarn-plugin-allow-scripts-yarn3/packages/yarn-plugin-allow-scripts/bundles/@yarnpkg/plugin-allow-scripts.js
```
naugtur, line 24: Idea: we could put the built files in GH releases and the URLs would be shorter. Not sure what other consequences it'd bring. Don't remember if releases are mutable.

packages/yarn-plugin-allow-scripts/package.json
```diff
@@ -10,19 +10,49 @@
...
     "rebuild": "npm run build",
-    "test": "exit 0"
+    "test": "npm run build && ava"
   },
-  "dependencies": {
-    "@yarnpkg/builder": "3.3.1",
-    "@yarnpkg/cli": "3.8.5",
-    "@yarnpkg/core": "3.7.0",
-    "clipanion": "3.2.1"
+  "peerDependencies": {
```
naugtur, line 22: If the only way to consume the code is direct copy from github, what is the point of defining peer dependencies?

packages/yarn-plugin-allow-scripts/package.json
```diff
@@ -10,19 +10,49 @@
...
+    "@lavamoat/preinstall-always-fail": "*",
+    "@yarnpkg/cli": "^4.4.1",
+    "@yarnpkg/core": "4.1.2"
+  },
+  "dependencies": {},
+  "devDependencies": {
+    "@lavamoat/allow-scripts": "3.2.0",
+    "@lavamoat/preinstall-always-fail": "2.1.0",
+    "@septh/ts-run": "1.2.6",
+    "@yarnpkg/builder": "4.2.0",
```
naugtur, line 33: Readme for this package says it supports yarn3, but if it works, should be fine.

packages/yarn-plugin-allow-scripts/package.json
```diff
@@ -10,19 +10,49 @@
...
+  "ava": {
+    "files": [
+      "test/*.spec.mts"
+    ],
+    "timeout": "60s",
+    "environmentVariables": {
+      "YARN_ENABLE_HARDENED_MODE": "0"
+    },
+    "extensions": {
+      "ts": "module",
```
naugtur, line 52: Originally the plugin was a ts file because it was contributed externally as such and was too small to warrant switching back.
Now with the typescript specific setup for tests and such, I'm thinking this is becoming inconsistent with the rest of the repo.
May not be a blocker for this PR, but I want the decision to commit to doing TS here or reverting back to jsdoc typings  to be a teamwide decision.

packages/yarn-plugin-allow-scripts/scripts/build.js
```diff
@@ -0,0 +1,33 @@
...
+);
+if (buildResult.status !== 0) {
+  console.error('Build failed');
+  console.log(buildResult.stdout);
+  console.error(buildResult.stderr);
+  process.exit(buildResult.status);
+}
+
+const bundlePath = join(cwd(), 'bundles', '@yarnpkg', 'plugin-allow-scripts.js')
+// patch broken yarn bundling
```
naugtur: (potential blocker)
This looks to me like a very strong argument against using yarn builder. Did you try reporting it?
I looked for open issues on topic and found a different bug left hanging with a strong conclusion https://github.com/yarnpkg/berry/issues/6136

Unless it provides functionality that's unique to yarn plugins I don't see a good reason to use it instead of setting up webpack that we already have in repo dependencies anyway.


[edit]
Looks like the only unique thing yarn builder is doing is this: https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-builder/sources/commands/build/plugin.ts#L80
Which automatically sets dependencies shipped by yarn as external, which we could do with an array in webpack config for our needs.

packages/yarn-plugin-allow-scripts/src/index.ts
```diff
@@ -0,0 +1,143 @@
...
+                    throw new Error(
+                      `Undefined type ${typeof policyValue} for allow-scripts policy value '${policyValue}'`
+                    )
+                }
+              }
+              return missingPolicies
+            })()
+          : new Set<string>()
+
+        if (extraMissingPolicies.size || somePoliciesAreMissing) {
```
naugtur, line 132: What's the difference between missing policy detection in allow-scripts and the additional detection here?
Why is it specific to the plugin and not needed in the main allow-scripts package?

packages/yarn-plugin-allow-scripts/src/index.ts
```diff
@@ -0,0 +1,143 @@
...
+    setupScriptEnvironment: (
+      project: ExtendedProject,
+      { npm_lifecycle_event, npm_package_json, npm_package_name }
+    ) => {
+      isPackageScriptAllowed(project, npm_package_json).then((isAllowed) => {
+        if (!isAllowed) {
+          console.error(
+            isAllowed === null
+              ? `  allow-scripts detected attempted execution of unconfigured package script. ${JSON.stringify([npm_package_name, npm_lifecycle_event, npm_package_json])}`
+              : `  allow-scripts detected attempted execution of disallowed package script. ${JSON.stringify([npm_package_name, npm_lifecycle_event, npm_package_json])}`
```
naugtur: (should) 
I'd avoid introducing the distinction here and print `disallowed` in both cases.
Missing configuration detection is happening elsewhere, and it's harder to miss there.

Also, the same check and message in two different hooks? Is this detecting different cases or will both be displayed for one package?

Why is this hook not attempting to block anything? (It could `process.exit` if there's no other choice)
What are the circumstances when this hook is called?

packages/yarn-plugin-allow-scripts/src/index.ts
```diff
@@ -0,0 +1,143 @@
...
+      project: ExtendedProject,
+      _locator: Locator,
+      npm_lifecycle_event: string,
+      { env: { npm_package_json, npm_package_name } }: { env: any }
+    ): Promise<() => Promise<number>> => {
+      return new Promise((resolve, _reject) =>
+        isPackageScriptAllowed(project, npm_package_json).then((isAllowed) => {
+          if (!isAllowed) {
+            console.error(
+              isAllowed === null
```
naugtur: (should) 
I'd avoid introducing the distinction here and print `disallowed` in both cases.
Missing configuration detection is happening elsewhere, and it's harder to miss there.

packages/yarn-plugin-allow-scripts/src/index.ts
```diff
@@ -0,0 +1,143 @@
...
+                ? `  allow-scripts blocking execution of unconfigured package script. ${JSON.stringify([npm_package_name, npm_lifecycle_event, npm_package_json])}`
+                : `  allow-scripts blocking execution of disallowed package script. ${JSON.stringify([npm_package_name, npm_lifecycle_event, npm_package_json])}`
+            )
+            return resolve(() => Promise.resolve(0))
```
naugtur, line 96: (could) use a named function here to indicate what the function is.
```js
const noopScriptExecutor = async ()=>{}
```

```suggestion
            return resolve(noopScriptExecutor)
```

packages/yarn-plugin-allow-scripts/test/index.spec.mts
```diff
@@ -0,0 +1,238 @@
...
+const run = (
+  t: any,
+  args: string[],
+  cwd: string,
+  env: {} = {},
+): {exitCode: number|null, out: string, err: string, output: (string|Buffer|null)[], signal: string|null, error?: Error|undefined} => {
+  try {
+    const defaultOpts = realisticEnvOptions(cwd);
+    const result = spawnSync(
+      isWindows
```
naugtur: (could) It'd be easier for me to reason about this if we had an `if(isWindows) {} else {}` with the little bit of redundancy but less cognitively demanding structure.

packages/yarn-plugin-allow-scripts/test/index.spec.mts
```diff
@@ -0,0 +1,238 @@
...
+  }
+}
+
+test('install - errors when adding dependency with unspecified preinstall script', async (t: any) => {
+  const projectName = 'uninitialized';
+  const projectRoot = join(dirname(import.meta.url.replace(/^file:/, '')), 'projects', projectName);
+
+  try {
+    // delete any leftover test artifacts
+    rmSync(join(projectRoot, PACKAGE_JSON), { force: true })
```
naugtur: Is this cleanup specific to only this test?

packages/yarn-plugin-allow-scripts/test/index.spec.mts
```diff
@@ -0,0 +1,238 @@
...
+    t.regex(addRes.err, /allow-scripts detected attempted execution of unconfigured package script. \["@lavamoat\/preinstall-always-fail","preinstall"/);
+    t.is(addRes.exitCode, 1);
+
+    // trigger the plugin
+    const installRes = run(t, ['install'], projectRoot)
+    t.regex(installRes.out, /packages missing configuration:\n- @lavamoat\/preinstall-always-fail/);
+    t.regex(installRes.err, /@lavamoat\/allow-scripts has detected dependencies without configuration. explicit configuration required./);
+    t.is(installRes.exitCode, 1);
+  } finally {
+    cleanup(projectName);
```
naugtur: (could) leftovers and cleanup might be more idiomatic to handle with https://github.com/avajs/ava/blob/main/docs/01-writing-tests.md#before--after-hooks

packages/yarn-plugin-allow-scripts/test/index.spec.mts
```diff
@@ -0,0 +1,238 @@
...
+        }
+      },
+    )
+    return {
+      exitCode: result.status,
+      out: result?.stdout?.toString()?.trim(),
+      err: result?.stderr?.toString()?.trim(),
+      output: result?.output,
+      signal: result.signal,
+      error: result.error,
```
naugtur, line 102: kudos for not forgetting the `error` field

You could rethrow it if it's truthy to fall through to error handling below.

packages/yarn-plugin-allow-scripts/test/index.spec.mts
```diff
@@ -0,0 +1,238 @@
...
+    t.regex(installRes.err, /@lavamoat\/allow-scripts has detected dependencies without configuration. explicit configuration required./);
+    t.is(installRes.exitCode, 1);
+  } finally {
+    cleanup(projectName);
+  }
+})
+
+test('install - allows adding and installing package with specified preinstall script', async (t: any) => {
+  const projectName = 'allowed';
+  const projectRoot = join(dirname(import.meta.url.replace(/^file:/, '')), 'projects', projectName)
```
naugtur: (could) extract to a utility function or maybe even hide inside `run`
(could) slight preference for calling the folder fixtures not projects, but up to you.

packages/yarn-plugin-allow-scripts/test/index.spec.mts
```diff
@@ -0,0 +1,238 @@
...
+    const installRes = run(t, ['install'], projectRoot)
+    t.regex(installRes.out, /packages missing configuration:\n- @lavamoat\/preinstall-always-fail/);
+    t.regex(installRes.err, /@lavamoat\/allow-scripts has detected dependencies without configuration. explicit configuration required./);
+    t.is(installRes.exitCode, 1);
+  } finally {
+    cleanup(projectName);
+  }
+})
+
+test('install - allows adding and installing package with specified preinstall script', async (t: any) => {
```
naugtur: Looking at the 'allowed' folder in fixtures, I don't see what this is actually allowing. 

https://github.com/LavaMoat/LavaMoat/pull/1218/files#diff-531f1c2b60c68ac2339c35184659e364dab32ce1431bcae05c136e6d4cfa4a8eR6 says the preinstall is not allowed. 

What's being tested here?

packages/yarn-plugin-allow-scripts/test/index.spec.mts
```diff
@@ -0,0 +1,238 @@
...
+
+    // trigger the plugin
+    const installRes = run(t, ['install'], projectRoot)
+    t.is(installRes.exitCode, 0);
+  } finally {
+    cleanup(projectName);
+  }
+})
+
+test('install - allows execution of allowed yarn classic dependencies', async (t: any) => {
```
naugtur, line 188: Much more explanation is gonna be necessary for this one. A big block comment with links preferably. I know what it's referring to but I'm still having trouble following the test steps :sweat_smile:

packages/yarn-plugin-allow-scripts/test/projects/protected-unconfigured/package.json
naugtur: please put this comment in a README.md of protected-unconfigured

packages/yarn-plugin-allow-scripts/test/index.spec.mts
```diff
@@ -0,0 +1,238 @@
...
+  const projectName = 'protected-configured';
+  const projectRoot = join(dirname(import.meta.url.replace(/^file:/, '')), 'projects', projectName)
+
+  try {
+    run(t, ['plugin', 'import', normalize('../../../bundles/@yarnpkg/plugin-allow-scripts.js')], projectRoot)
+
+    // trigger the plugin
+    const installRes = run(t, ['install', '--inline-builds'], projectRoot, {YARN_IGNORE_SCRIPTS: 'false'})
+    t.is(installRes.exitCode, 0);
+    t.true(existsSync(portablePath(join(projectRoot, 'node_modules', 'allowed-dep', 'foo'))));
```
naugtur: (should) Add a comment indicating this line verifies the lifecycle script worked.

naugtur: (should) I don't see anything specifically removing the file, so I'm assuming git cleanup does it, but it'd make the test more robust if the same assertion with `t.false` was executed before running `install`

packages/yarn-plugin-allow-scripts/scripts/build.js
legobeat: Looks like it may actually be fixed now and patching may no longer be necessary: https://github.com/yarnpkg/berry/pull/6356 (upstream issues linked)

packages/yarn-plugin-allow-scripts/test/index.spec.mts
legobeat: The end-result is we're allowed to run `install` because `preinstall-always-fail` is handled correctly. Maybe there is a better name?

packages/yarn-plugin-allow-scripts/test/index.spec.mts
legobeat: > (could) extract to a utility function or maybe even hide inside run

Seems premature until we see further usage or duplication

> (could) slight preference for calling the folder fixtures not projects, but up to you.

In general agreed but following existing norm from `allow-scripts`. I'd rather rename both in a follow-up.

packages/yarn-plugin-allow-scripts/test/index.spec.mts
legobeat: Yes, it's taken care of by `git clean -fdx`.

And if that would have not been the case, in any case that would already be caught in CI due to the dirty-git-state check? If anything is left over form cleanup, it will fail.

packages/yarn-plugin-allow-scripts/README.md
legobeat: Was thinking this too. I guess it would involve adding it to release-please (and either add another moving branch/tag for it, or auto-update this readme as part of release), while also keeping it from publishing to npm.

packages/yarn-plugin-allow-scripts/src/index.ts
legobeat: > What's the difference between missing policy detection in allow-scripts and the additional detection here?

There shouldn't be any difference - we are now no longer calling the `allow-scripts` cli directly as part of routine operations, so this is why we now import and use the `printMissingPoliciesIfAny` function from `allow-scripts`.

So the actual pre-emption of lifecycle scripts (or failure to do so :p) have already happened by the time we get here.

The previous version of the plugin printed a summary with any outstandings after execution and I thought it a feature worth preserving.

packages/yarn-plugin-allow-scripts/test/index.spec.mts
legobeat: Leftover - it should be removed (or moved to `cleanup`)

packages/yarn-plugin-allow-scripts/scripts/build.js
legobeat: NVM last comment; that was a separate issue that was previously being patched for here as well.

legobeat: I did try some options (directly using esbuild, tsup, webpack). Unfortunately, I come back to this as the least hacky option... This is because:
- Yarn plugins run in ESM context so no native `require` available
- `@lavamoat/allow-scripts` (which we need to bundle, including its deps) causes reference to `require.resolve` via `@npm/run-script`
- Yarn plugins do not have `import` available (regardless of `type:module` or not) so `createRequire(import.meta.url)` does not work as a standin
- Yarn has its own DI of a custom `require` into the plugin, which is reserved specifically for `@yarnpkg/` dependencies (which we use) - though fortunately does provide `require.resolve`

It's specifically the last point that made me give up on webpack for now, since it seems to conflict with how it handles requires itself. If you can come up with a PR for how to make it build with webpack or anything else that doesn't turn out too messy I'm for it, but I don't see a productive path forward here for now..

packages/yarn-plugin-allow-scripts/src/index.ts
legobeat: > I'd avoid introducing the distinction here and print disallowed in both cases.
Missing configuration detection is happening elsewhere, and it's harder to miss there.

I'd like to keep the distinction - it's useful to immediately see why (missing or explicitly blocked).

> Also, the same check and message in two different hooks? Is this detecting different cases or will both be displayed for one package?


> Why is this hook not attempting to block anything? (It could process.exit if there's no other choice)

I ended up removing the `setupScriptEnvironment` hook as it happens prior to `wrapScriptExecution` and doesn't have a way to abort without `process.exit`, so it wasn't able to do anything meaningful. Otherwise I agree it would make sense to hard exit as fallback if the intial prevention was detected ineffective down the line, in case of bugs in any part of the chain.

packages/yarn-plugin-allow-scripts/src/index.ts
legobeat: I'd like to keep the distinction - it's useful to immediately see why (missing or explicitly blocked).

packages/yarn-plugin-allow-scripts/test/index.spec.mts
legobeat: Sure, updated with breaking out the conditionals to a single on outside the `spawnSync`

packages/yarn-plugin-allow-scripts/test/index.spec.mts
legobeat: Updated this test-case and its description. Also added a `blocked` case, to test both scenarios with non-git dependencies.

packages/yarn-plugin-allow-scripts/test/projects/protected-unconfigured/package.json
legobeat: Done

packages/yarn-plugin-allow-scripts/test/index.spec.mts
legobeat: Done

packages/yarn-plugin-allow-scripts/test/index.spec.mts
legobeat: Added a pointer to fixture README which contains explanation and a few links - does that suffice? If not, would you be able to suggest missing parts?

packages/yarn-plugin-allow-scripts/test/index.spec.mts
legobeat: Done

packages/yarn-plugin-allow-scripts/README.md
legobeat: If we're changing plugin path (mildly breaking if we consider installation docs), could also do shorter by moving build output to the distro root and removing the `bundles/`, even if not installing via releases (which does sound preferred but maybe a bit more fiddling to get automated)

naugtur: It'd be a matter of including i in release-please and not including it in laverna configuration - it's release-please that does the github releases thing. 
BUT, all that is pointless, because to publish a JS file not a tgz we need a custom build+publish.
It doesn't make sense to put this up for release management with release-please imho.

packages/yarn-plugin-allow-scripts/scripts/build.js
naugtur: Why does the plugin need to call `@npm/run-script` instead of spawning a call to the `allow-scripts` again? Seems like it's a source of a lot of complexity.

packages/yarn-plugin-allow-scripts/src/index.ts
naugtur: I'm not sure I understand - does this mean the plugin is printing every missing entry twice?

packages/yarn-plugin-allow-scripts/test/index.spec.mts
naugtur: It doesn't yet explain what `yarn classic dependencies` are and what about them we want to work. `{YARN_IGNORE_SCRIPTS: 'false'}` is therefore a mystery too.

Also, if the test has to set this env var, why? And what about the normal usage where it's not set?

packages/yarn-plugin-allow-scripts/test/index.spec.mts
```diff
@@ -0,0 +1,242 @@
...
+      signal: null,
+      out: '',
+      err: '',
+    };
+  }
+}
+
+test('install - errors when adding dependency with unspecified preinstall script', async (t: any) => {
+  t.context.projectName = 'uninitialized';
+  const projectRoot = join(dirname(import.meta.url.replace(/^file:/, '')), 'projects', t.context.projectName);
```
naugtur: This could be extracted to a named function . (sorry if I'm repeating myself. I might have written this comment already but can't see it now)

legobeat:
@naugtur 
> Why does the plugin need to call `@npm/run-script` instead of spawning a call to the allow-scripts again? Seems like it's a source of a lot of complexity.

It doesn't actually call `@npm/run-script` - this is dead code as effect of the bundling and the monolithic `index.js` in `allow-scripts`. I realize if we modularize allow-scripts, we can then easily cut down the bundle-size by a lot, as well as drop this inclusion and thereby the need for patching and `build.js`:
- #1377 
- PoC of the combination of these two PRs: https://github.com/legobeat/LavaMoat/pull/15/commits
  - resulting plugin bundle becomes 88% smaller gzipped

--- 

UPDATE: #1377 has been merged and this PR has been updated as per above. `scripts/build.js` has been removed.

packages/yarn-plugin-allow-scripts/package.json
legobeat: FWIW, a nice side-benefit is that it also acts an an integration-test for `@lavamoat/allow-scripts` type definitions as they're used in the build. If we break the types in some way that affects TypeScript users, it's likely to be caught in the build for the plugin.

packages/yarn-plugin-allow-scripts/scripts/build.js
legobeat: It doesn't actually call `@npm/run-script` - this is dead code as effect of the bundling and the monolithic `index.js` in `allow-scripts`. I realize if we modularize allow-scripts, we can then easily cut down the bundle-size by a lot, as well as drop this inclusion and thereby the need for patching and `build.js`:
- #1377 
- PoC of the combination of these two PRs: https://github.com/legobeat/LavaMoat/pull/15/commits
  - resulting plugin bundle becomes 88% smaller gzipped 

<details>
_For some reason I can reply inline to the same comment here? GH is strange sometimes._
</details>

packages/yarn-plugin-allow-scripts/src/index.ts
legobeat: It logs:
- Individually each time it preempts execution of unrecognized scripts (in the `wrapScriptExecution` hook)
- The output from `printMissingPoliciesIfAny`, for missing scripts

I think it makes sense to have both?

I did remove logging of locked scripts by default, though, as that is more of a "business as usual under daily operations" situation.

packages/yarn-plugin-allow-scripts/test/index.spec.mts
legobeat: Done

packages/yarn-plugin-allow-scripts/package.json
legobeat: - It could still be used as a library, I guess
- More motivating: This communicates requirements to humans. I think it's a decent place to put authoritative version requirements for users even if Yarn currently doesn't have an interface for it.

legobeat:
- Broke out #1383

packages/yarn-plugin-allow-scripts/package.json
naugtur: Nice aside. Doesn't address my last sentence though. Forgot to bring it up in the meeting.

packages/yarn-plugin-allow-scripts/src/index.ts
legobeat: I'd add if it'd been used in more then one place, or been less trivial.

packages/yarn-plugin-allow-scripts/test/index.spec.mts
legobeat: > (could) extract to a utility function

Done

packages/yarn-plugin-allow-scripts/README.md
```diff
@@ -1,15 +1,29 @@
...
 
-A Yarn v3 plugin for running `@lavamoat/allow-scripts` after install.
+A Yarn Berry plugin for integrating `@lavamoat/allow-scripts` lifecycle-script protection.
 
 ### Install
 
-Yarn v3 plugins are installed via public URL.
+#### Yarn v4
+
+First, perform basic [configuration](https://github.com/LavaMoat/LavaMoat/blob/main/packages/allow-scripts/README.md#configure) `@lavamoat/allow-scripts` for the package.
```
boneskull: this sentence seems to be missing something

packages/yarn-plugin-allow-scripts/README.md
```diff
@@ -1,15 +1,29 @@
...
+
+The latest version of the plugin can then be installed by its URL:
 
 ```sh
 yarn plugin import https://raw.githubusercontent.com/LavaMoat/LavaMoat/main/packages/yarn-plugin-allow-scripts/bundles/@yarnpkg/plugin-allow-scripts.js
 ```
 
+#### Yarn v3
+
+The current version of the plugin does not support Yarn v3. Yarn v3 users are encouraged to migrate to an actively supported package manager such as Yarn v4.
```
boneskull, line 19: Isn't Yarn v1 EOL'd?

boneskull (dismissed):

packages/yarn-plugin-allow-scripts/README.md
legobeat: It's a bit ambiguous and subjective if Yarn 1 can be considered EOL:
- It's still downloaded and utilized by Yarn 3/4
  - [Referenced in tests](https://github.com/LavaMoat/LavaMoat/pull/1218/files#diff-28322e804f60a520aed793eef6737b40eaf3f04a896a3f669c0ee4e277422d29R201)
    - [Explained in fixture README](https://github.com/LavaMoat/LavaMoat/pull/1218/files#diff-6f9be4f4e3e42ee652af6f870b5c8800b967642aefdc237f66b26a3484c19ab4)
  - https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-core/sources/scriptUtils.ts#L271
- It's still heavily used in the wild
- Bugfixes are still being considered by maintainers

In any case, how's that relevant here? This is about Yarn 3.

@naugtur required that we consider and accommodate for Yarn 3 users with a maintenance branch even if Yarn 3 is deprecated by Yarn maintainers.

packages/yarn-plugin-allow-scripts/README.md
legobeat: Good catch! Was missing an ` of`.

packages/yarn-plugin-allow-scripts/src/index.ts
naugtur: I don't see the point of having both. We've always printed a summary of what scripts you need added (and stopped the process of running anything too, but that's probably harder with yarn)

The hope is to have people switching between different projects using yarn and npm be able to feel familiaar.

packages/yarn-plugin-allow-scripts/README.md
naugtur: We don't support yarn1 with the plugin but we do support configuring for it in allow-scripts. allow-scripts works fine with yarn1 and there's no reason to stop that.
I requested we keep the v3 version easily reachable and documented because it's value and we already have it. 

Overall, the goal of allow-scripts is to be viable for any repository containing a package.json

packages/yarn-plugin-allow-scripts/package.json
naugtur: with private:true there's no way this has any consequences, so it's fine

packages/yarn-plugin-allow-scripts/src/index.ts
```diff
@@ -0,0 +1,150 @@
...
+  process.platform === 'win32' ? cwd.replace(/^[\/]/, '') : cwd
+
+// Only consider script executions trigered by the below lifecycle events
+const PROTECTED_LIFECYLE_SCRIPTS = new Set([
+  'install',
+  'postinstall', 'postinstallonly', 'preinstall', 'preinstallonly',
+  'preaudit', 'preauditonly', 'postaudit', 'postauditonly',
+  'prepack', 'prepackonly', 'postpack', 'postpackonly',
+  'preprepare', 'preprepareonly', 'postprepare', 'postprepareonly',
+  'prepublish', 'prepublishonly', 'postpublish', 'postpublishonly',
```
naugtur, line 35: note from live conversation: we need to rethink this a bit.

PR #1214, opened 2024-07-06
lavamoat-node: unsupport node 20.15; support 22.1

legobeat:
- **chore(node): unsupport node.js <20.15; support ^22 <22.2**
  lavamoat is affected by a Node.js issue on >=20.15 and >=22.2.
  
  Remove indicated support of affected versions until the situation has been remediated.
  

### Related
- Based on/blocked by: #1212 
- #1211 
- #1184

Issue #1211, opened 2024-07-04
`Cannot read properties of undefined (reading '0')` on newer Node.js versions


legobeat:
There is a known regression affecting Node.js versions >= `20.15.0` and >= `22.2.0`, resulting in unhandled runtime errors:

<details>

https://github.com/LavaMoat/LavaMoat/actions/runs/9816323753/job/27106390352#step:7:339

```npm exec lavamoat -- /home/runner/work/LavaMoat/LavaMoat/packages/browserify/test/fixtures/secureBundling/build.js --policyPath /home/runner/work/LavaMoat/LavaMoat/packages/browserify/test/fixtures/secureBundling/lavamoat/node/policy.json
(TypeError#1)
TypeError#1: Cannot read properties of undefined (reading '0')
  at isInsideNodeModules (node:internal/util:508:17)
  at showFlaggedDeprecation (node:buffer:178:8)
  at new Buffer (node:buffer:266:3)
  at new Buffer (LavaMoat/core/kernel:591:22)
  at new Buffer (LavaMoat/core/kernel:591:22)
  at new Buffer (LavaMoat/core/kernel:591:22)
  at decodeBase64 (/home/runner/work/LavaMoat/LavaMoat/node_modules/combine-source-map/node_modules/convert-source-map/index.js:13:10)
  at new Converter (/home/runner/work/LavaMoat/LavaMoat/node_modules/combine-source-map/node_modules/convert-source-map/index.js:42:28)
  at exports.fromComment (/home/runner/work/LavaMoat/LavaMoat/node_modules/combine-source-map/node_modules/convert-source-map/index.js:109:10)
  at exports.fromSource (/home/runner/work/LavaMoat/LavaMoat/node_modules/combine-source-map/node_modules/convert-source-map/index.js:125:22)
```

</details>

This affects `lavamoat-core` and dependencies. `@lavamoat/allow-scripts` is unaffected.

#### Related

- Upstream issue: https://github.com/nodejs/node/issues/53075
- #1184

legobeat:
- #1251

PR #1184, opened 2024-05-21
support Node.js v22.x

boneskull:
- Updates all `engines` fields in the `package.json` files
- Adds Node.js v22.x to the build matrix
- Updates lockfile

npm is currently dumping mad `EBADENGINE`s upon install

PR #1220, opened 2024-07-08
feat(allow-scripts,tofu):  support node.js v22

legobeat:
Update specifications for supported Node.js engines.

These packages will support Node.js `^22`:

- `aa`
- `allow-scripts`
- `laverna`
- `preinstall-always-fail`
- `tofu`
- `yarn-plugin-allow-scripts`

These packages have been updated to support Node.js `^20 <20.15 || ^22 <22.2` (follow-up to #1214):
- `browserify`
- `perf`

#### Related
- #1211 
- #1214 
- #1184 

--- 

- **feat(allow-scripts): support Node.js v22**
- **ci: run on 22.1**
- **tofu: support node.js 22.x**
- **chore(core and deps): add support for node 22<22.2; unsupport node ^20.15**
  - See #1214


naugtur:
note:  `lavamoat-node` package was updated earlier https://github.com/LavaMoat/LavaMoat/blob/1e263ba7f006c6406c6c03d7fc6650a6799b850a/packages/node/package.json

Question: is it not going to be a problem for installing under node 22 that `lavamoat-core` is not listing it as supported? https://github.com/LavaMoat/LavaMoat/blob/1e263ba7f006c6406c6c03d7fc6650a6799b850a/packages/core/package.json#L17

legobeat:
> note: `lavamoat-node` package was updated earlier https://github.com/LavaMoat/LavaMoat/blob/1e263ba7f006c6406c6c03d7fc6650a6799b850a/packages/node/package.json
> 
> Question: is it not going to be a problem for installing under node 22 that `lavamoat-core` is not listing it as supported?
> 
> https://github.com/LavaMoat/LavaMoat/blob/1e263ba7f006c6406c6c03d7fc6650a6799b850a/packages/core/package.json#L17

Should be fine, but I also think extending it here is even finer. Added.

leotm (approved):

PR #1264, opened 2024-08-02
@lavamoat/types should not expose js

legobeat:
- Targeting #1033 

### Summary
- fix: `@lavamoat/types` should emit and expose only TS type-declarations, not JS implementations
  -  Remove entrypoint `dist/index.js`
- chore: Adjust package author
- fix: Disable `compilerOptions.skipLibCheck` in tsconfig
- fix: Address outstanding type-checking errors

### Before
```
$ tree packages/types/dist
packages/types/dist
 index.d.ts
 index.d.ts.map
 index.js
 module-record.d.ts
 module-record.d.ts.map
 module-record.js
 policy-schema
  ambient.d.ts
  ambient.d.ts.map
  ambient.js
  index.d.ts
  index.d.ts.map
  index.js
  lavamoat-policy.v0-0-1.schema.d.ts
  lavamoat-policy.v0-0-1.schema.d.ts.map
  lavamoat-policy.v0-0-1.schema.js
 tsconfig.tsbuildinfo
```

### After
```
$ tree packages/types/dist
packages/types/dist
 index.d.ts
 index.d.ts.map
 module-record.d.ts
 module-record.d.ts.map
 policy-schema
  ambient.d.ts
  ambient.d.ts.map
  index.d.ts
  index.d.ts.map
  lavamoat-policy.v0-0-1.schema.d.ts
  lavamoat-policy.v0-0-1.schema.d.ts.map
 tsconfig.tsbuildinfo
```


boneskull:
yeah, you're not gonna want to turn `skipLibCheck` off 😄

legobeat:
> yeah, you're not gonna want to turn skipLibCheck off 😄


@boneskull I _do_. The outstanding errors here all seem resolvable?


Also I think f.e. [this](https://github.com/LavaMoat/LavaMoat/actions/runs/10209011198/job/28246352666?pr=1264#step:7:38) is indicative of that provided types are not utilized as expected? Seems like something we want to catch in CI?

```
Error: packages/webpack/src/buildtime/aa.js(63,19): error TS2307: Cannot find module '@lavamoat/types' or its corresponding type declarations.
Error: packages/webpack/src/buildtime/policyGenerator.js(5,13): error TS7016: Could not find a declaration file for module 'lavamoat-core'. '/home/runner/work/LavaMoat/LavaMoat/packages/core/src/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/lavamoat-core` if it exists or add a new declaration (.d.ts) file containing `declare module 'lavamoat-core';`
```

legobeat:
Outstanding type-checking error have been addressed and this is now passing.

LMK is there's anything specific you prefer broken out into separate PR, as some things are not directly a consequence of #1033.

packages/laverna/package.json
```diff
@@ -49,6 +49,7 @@
     "type-fest": "4.23.0"
   },
   "devDependencies": {
+    "@types/node": "^20.14.3",
```
boneskull, line 52: I know you love this: this should actually be a prod dependency, since Laverna exposes types built upon ones from `@types/node` (e.g., `MinimalFs` and its ilk).

That might want to be a different PR.

.config/tsconfig.workspace.json
```diff
@@ -5,6 +5,7 @@
     "allowSyntheticDefaultImports": true,
     "allowJs": true,
     "checkJs": true,
+    "skipLibCheck": false,
```
boneskull, line 8: 👎 

To fix the problem which arose when enabling lib checks, this PR adds `@types/tmp` to our dependencies; this fixed an error caused by the author of `tmp-promise`.  Note that the invalid types which `@types/tmp` fixes are not referenced by our codebase; _we don't  gain any type fidelity_ by fixing this problem. Further, these types _are not exposed to our downstream consumers_--they aren't even exposed to _us_, really--so we cannot argue that we're doing this for the benefit of downstream.  And we _especially_ cannot argue this _because half of our codebase is still completely untyped_.

Depending upon types _for a transitive dependency_ seems dubious, yes?

It's a example of the hazard `skipLibCheck: false`; other people's problems will now be breaking our builds, and it will often be unclear whether the mitigation is a net-positive. A modest proposal: we should also run all of our dependencies' test suites and ensure they pass!!

If a downstream type is broken--and it actually _matters_--we will notice because we will be confronted with an unexpected `any` type.  We can then choose to stub the type out, or take action such as this (but with `skipLibCheck: false`, an upstream fix is the _only_ option; we have little agency).  Here, nobody (not us nor downstream) would have noticed a problem.  This poor tradeoff is the main reason why `skipLibCheck: true` is the default behavior.

Admittedly, the problems the lib check surfaced were few, but it's a mistake to think this will always be so. More problems of this nature will be surfaced the closer we get to complete type coverage. Some problems will not be so easily fixable short of using  `patch-package`.

That said, to ensure _we're_ shipping correct types--for the benefit of downstream--purpose-built tools such as [tsd](https://npm.im/tsd) or [expect-type](https://npm.im/expect-type) will prove a good investment. Even so, top priority should be type coverage. 

Also, `skipLibCheck: false` makes compilation way slower. 😜 

At the very least, I hope we can agree to defer a decision to enable lib checks until we're shipping complete type declarations for all of our packages.  I'll still argue vociferously against doing so, but I'm open to being swayed the other way once it doesn't seem so premature.

legobeat: > To fix the problem which arose when enabling lib checks, this PR adds @types/tmp to our dependencies

No, it adds it to `devDependencies`, which does not have the drawback of adding overhead in the shipped packages that adding it to `dependencies` would. It simply serves to resolve the type-checking, and does not have impact beyond that.

packages/laverna/package.json
legobeat: Why does it need to be added to `dependencies` as a consequence of `packages/laverna/test/types.d.ts` being removed?!

`packages/laverna/src/types.d.ts` is still there (as you know I'd like to see that one being culled as well, but that is not part of this PR as it is).

boneskull: to be clear, it's broken in `master` and doesn't have anything to do with this change.  this change makes `tsc` shut up, but downstream consumers will be missing the types unless they are shipped as prod deps.

PR #1266, opened 2024-08-02
feat(core): add allowDynamicRequire option to generatePolicy

legobeat:
Defaults to previous `false`.

Broken out from #772 as independent PR.

### DIFF
--- a/packages/core/src/schema/lavamoat-policy.v0-0-1.schema.ts
+++ b/packages/core/src/schema/lavamoat-policy.v0-0-1.schema.ts
@@ -84,6 +84,10 @@ export interface ResourcePolicy {
    * Allow native modules
    */
   native?: boolean
+  /**
+   * Allow dynamic requires
+   */
+  dynamic?: boolean
 }
 
 /**
@@ -109,6 +113,7 @@ export interface BuiltinPolicy {
 export interface PackagePolicy {
   [k: string]: boolean
 }
+
 /**
  * Custom run-time module resolutions by direct dependency
  */
--- a/packages/core/src/generatePolicy.js
+++ b/packages/core/src/generatePolicy.js
@@ -44,7 +44,12 @@ function createModuleInspector(opts) {
   const packageToGlobals = new Map()
   /** @type {Map<string, string[]>} */
   const packageToBuiltinImports = new Map()
+  /** @type {Map<string, import('./moduleRecord').LavamoatModuleRecord[]>} */
   const packageToNativeModules = new Map()
+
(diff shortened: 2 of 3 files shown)


packages/core/src/schema/lavamoat-policy.v0-0-1.schema.ts
```diff
@@ -84,6 +84,10 @@ export interface ResourcePolicy {
    * Allow native modules
    */
   native?: boolean
+  /**
+   * Allow dynamic requires
+   */
+  dynamic?: boolean
```
legobeat, line 90: The schema version is still at v0.0.1. Wouldn't this mean a bump to v0.0.2?

naugtur:
What's the relationship between this change and #772 ?
Do we want to merge it before or after? why?

Also, how final is the decision to explicitly call out dynamic require in policy? Last time I was involved, we were hesitating

boneskull:
Wait, why is this assigned to me?

boneskull:
@naugtur A package can import any other package it's allowed to import; the specifics are  generally irrelevant to Endo. This means we cannot prevent _any_ or _all_ dynamic requires  unless we choose to add a new flag to Endo (which, IMO, we should not). 

So without adding a new field to Endo's policy, the only place we could approve or deny use of a dynamic require would be if the case that the dynamic require invoked the dynamic-import-now exit module hook.  

Currently, this hook would be invoked if the module specifier was an arbitrary path which _did not match_ any known compartment _or_ was otherwise decided to be an exit module (e.g., a builtin).  We can enforce policy in that hook using the `options` prop of Endo's policy.  

It _seems_ like it would be security-positive to have a flag for that behavior vs. no flag at all.  I am not sure how to distill that condition into a single word, though.  `dynamic` may be OK if we're careful about documenting what it means.

naugtur:
The last good reason we had left for `dynamic` was a case where someone dynamically chooses what builtin to require and that has a nicer solution now, so it looks like we'll be removing that flag altogether.

I'd avoid making changes related to its existence for now.

boneskull:
I don't understand what this PR is doing?

#772 was created from a set of stacked PRs which became too onerous to manage due to pacing and cross-cutting changes.  So instead of splitting things up nicely, we purposely did the opposite--lowered developer (mainly me) overhead for a difficult-to-review PR.  

Once we feel that #772 is in a good place, then I can take a "final" pass at splitting it up again. But it's still premature.

boneskull:
...and regarding the changes here, we are not going to be changing the schema.  We will need to retain a flag in `core` to squelch warnings about dynamic requires, though.

PR #1248, opened 2024-07-26
chore(node): fix and update express example

legobeat:
- Related: #1210

---

- **chore(node): update patch-package in test fixture**
  - **chore(node): set packageManager on test fixtures**
- **chore(node): bump dependencies for express example**
- **chore(node): fix package.json for express example**
- **chore(node): fix out of date lavamoat example**
- **chore(node): sync express example policy file**

PR #1270, opened 2024-08-03
chore: remove legacy examples

legobeat:
https://github.com/LavaMoat/LavaMoat/pull/1210#issuecomment-2265129845

> > There are existing examples under packages/node/examples (https://github.com/LavaMoat/LavaMoat/pull/1248) - feels like they should be merged under either destination?
> they are in need of updating to a point of redoing. consider them legacy.

### Related
- #1248 (alternative to this which attempts to fix the node examples)
- #1210 (adding new examples)


boneskull (approved):

boneskull:
I've just rebased this. @naugtur anything to say about it?

PR #1282, opened 2024-08-12
webpack fixes for metamask config

naugtur:
- rootDir setting to let aa find the right package.json while webpack `compiler.context` is set to a subpath
- prevent error in policy generation for virtual modules webpack uses for juggling things that don't actually have sources
- added some debugging helpers

### DIFF
--- a/packages/webpack/src/buildtime/policyGenerator.js
+++ b/packages/webpack/src/buildtime/policyGenerator.js
@@ -11,7 +11,8 @@ const {
   sources: { RawSource },
 } = require('webpack')
 
-const { isExcludedUnsafe } = require('./exclude.js')
+const { isExcludedUnsafe } = require('./exclude')
+const diag = require('./diagnostics')
 
 const POLICY_SNAPSHOT_FILENAME = 'policy-snapshot.json'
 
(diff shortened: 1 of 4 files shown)


packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -99,6 +100,14 @@ module.exports = {
         // Skip modules the user intentionally excludes.
         // This is policy generation so we don't need to protect ourselves from an attack where the module has a loader defined in the specifier.
         if (isExcludedUnsafe(module)) return
+        if (module.userRequest === undefined) {
+          diag.rawDebug(
+            1,
+            `LavaMoatPlugin: Module ${module} has no userRequest`
+          )
+          diag.rawDebug(2, { skippingInspectingModule: module })
```
boneskull, line 108: what is this option?

packages/webpack/src/plugin.js
```diff
@@ -102,6 +102,18 @@ class LavaMoatPlugin {
     })
 
     const options = this.options
+
+    diag.run(2, () => {
```
boneskull, line 106: what is this magic number?

packages/webpack/src/plugin.js
```diff
@@ -102,6 +102,18 @@ class LavaMoatPlugin {
...
+      // Log stack traces for all errors on higher verbosity because webpack won't
+      compiler.hooks.done.tap(PLUGIN_NAME, (stats) => {
+        if (stats.hasErrors()) {
+          stats.compilation.errors.forEach((error) => {
+            console.error(error.stack || error)
```
```suggestion
            console.error(error)
```

Maybe better to let Node.js decide how to display it?

boneskull (dismissed):
I don't understand how any of this stuff works, but the changes seem fine!

packages/webpack/src/buildtime/policyGenerator.js
naugtur: It's just a console log based on the diagnostics level. The object is what gets logged

packages/webpack/src/plugin.js
naugtur: The log level

boneskull:
```suggestion
    const LOG_LEVEL = 2
    diag.run(LOG_LEVEL, () => {
```

😄

weizman (approved):
lgtm

PR #1291, opened 2024-08-20
test(core): add a test to document the support scope of unwrapping this with bind

naugtur:
This test documents the gap in unwrapping `this` support in lavamoat core. 
(this was first identified in Snaps: https://github.com/MetaMask/metamask-extension/pull/26431/files#diff-63ab7391d870a62d8bcd3cc5d5371432068538deb98e5effcb899434ed8649bbL2640)

@kumavis anticipates security issues if we handle unwrapping these. 

We could try to add support (regardless?) but it's a very niche thing and we might be better off eliminating endowments limited to `functionName.bind` from policy generation instead. the return value of bind is the function anyway, so why restrict it?

Therefore in policy generation we could trim `.bind` (and probably also `.call` and `.apply`) to solve this in the default case.  

Update:
Added a commit changing the behavior in Tofu to skip certain path endings. 
How likely is it that a nested global actually has a `.bind` method that's custom, not a Function method?
Is it too much of a POLA scope widening by default in the case of these not being function methods?
I'm thinking it's fine to have that default if we don't support wrapping.

### DIFF
--- a/packages/core/test/endowmentsToolkit.spec.js
+++ b/packages/core/test/endowmentsToolkit.spec.js
@@ -45,6 +45,57 @@ test('getEndowmentsForConfig - function on proto', (t) => {
   t.is(resultGlobal.lookAtMyProto.appendChild(), assertMe)
 })
 
+test('getEndowmentsForConfig - function call/bind unwrapping', (t) => {
+  const getEndowmentsForConfig = prepareTest()
+
+  const sourceGlobal = {
+    checkThis: function () {
+      return this === sourceGlobal
+    },
+  }
+  const config = {
+    globals: {
+      checkThis: true,
+    },
+  }
+  const resultGlobal = getEndowmentsForConfig(sourceGlobal, config)
+  t.is(typeof resultGlobal.checkThis, 'function')
+  t.is(resultGlobal.checkThis(), true)
+  t.is(resultGlobal.checkThis.bind({})(), false)
+  t.is(resultGlobal.checkThis.bind(resultGlobal)(), true)
+  t.is(resultGlobal.checkThis.call(resultGlobal), true)
+})
+
+test.failing(
+  'getEndowmentsForConfig - function call/bind unwrapping when only methods endowed',
+  (t) => {
+    const getEndowmentsForConfig = prepareTest()
+
+    const sourceGlobal = {
+      checkThis: function () {
+        return this === sourceGlobal
+      },
+    }
+
+    const config = {
+      globals: {
+        'checkThis.bind': true,
+        'checkThis.call': true,
+      },
+    }
+    const resultGlobal = getEndowmentsForConfig(sourceGlobal, config)
+    t.is(typeof resultGlobal.checkThis, 'object') // the function is never wrapped
+
+    // This case is not supported and might be too risky to support. We should rather avoid producing it.
+    // In policy generation we could trim .bind (and probably also .call and .apply) to avoid hitting this problem.
+
+    // When a function is endowed, call and bind should work as expected because of the wrapping/unwrapping
+    // When only the call/bind methods are endowed, they are not taken from a wrapped function but from the source directly and `this` doesn't get correctly unwrapped.
+    t.is(resultGlobal.checkThis.call(resultGlobal), true)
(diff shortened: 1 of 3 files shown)


packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -45,6 +45,57 @@ test('getEndowmentsForConfig - function on proto', (t) => {
...
+      globals: {
+        'checkThis.bind': true,
+        'checkThis.call': true,
+      },
+    }
+    const resultGlobal = getEndowmentsForConfig(sourceGlobal, config)
+    t.is(typeof resultGlobal.checkThis, 'object') // the function is never wrapped
+
+    // This case is not supported and might be too risky to support. We should rather avoid producing it.
+    // In policy generation we could trim .bind (and probably also .call and .apply) to avoid hitting this problem.
```
boneskull, line 90: I'd like to better understand what's risky here.

(please fix) I'd recommend splitting this up into two tests, one for each assertion; when the first assertion fails (as it presumably should), the test should exit and the second assertion will never be called. This assumes AVA's behavior is not completely wacko.

packages/tofu/src/inspectSource.js
```diff
@@ -102,6 +104,11 @@ function inspectGlobals(
         if (path.length === 0) {
           return
         }
+        // Sometimes nesting further doesn't make sense, like in case of Function methods
+        // Eliminating cases of nesting unwanted path endings like func.bind.call was considered not worth the complexity
+        if (avoidEndingPathsIn.includes(path[path.length - 1])) {
+          path.pop()
+        }
```
boneskull, line 111: I don't understand what this does.

boneskull (commented):
I don't understand if you mean to make more changes here

packages/core/test/endowmentsToolkit.spec.js
naugtur: It'd be really hard to avoid creating a way to leak a reference that the script should not have according to policy while trying to get the binding to work correctly.

packages/tofu/src/inspectSource.js
naugtur: If a path in generated policy ends with one of the items on the list, we pop that item off (mutate `path` in place.

so `document.getElementById.bind` becomes `document.getElementById`

legobeat:
> Update:
Added a commit changing the behavior in Tofu to skip certain path endings.
How likely is it that a nested global actually has a .bind method that's custom, not a Function method?
Is it too much of a POLA scope widening by default in the case of these not being function methods?
I'm thinking it's fine to have that default if we don't support wrapping.

Is there a way to avoid the new changes in `tofu` alltogether? Comment mentions `When only the call/bind methods are endowed, they are not taken from a wrapped function but from the source directly and 'this' doesn't get correctly unwrapped.`. Could the practical issue be addressed on a higher level?

To me it's not POLA that functions (which may or may not match the `call`/`bind`/`apply` being covered for here) with certain names get treated differently.

If not: The scope of this PR is `test(core)` while these changes make it `feat!(tofu)`. Could it either be broken out separately, or the PR title updated?

packages/core/test/endowmentsToolkit.spec.js
boneskull: can you change the test though? 😄

boneskull:
This is a breaking change in `tofu`?

legobeat:
> This is a breaking change in `tofu`?

I believe so, due to https://github.com/LavaMoat/LavaMoat/pull/1291/files#r1781284256

boneskull:
OK.  I don't take issue with that.

packages/tofu/src/inspectSource.js
boneskull: can you change the comment to be more clear about it?

boneskull:
@naugtur What is the status of this

naugtur:
Waiting for opinions

[edit]

Wait, I think we've introduced a similar change to tofu later on and this is now obsolete. Would have to check.
The bind and call test could be useful to adopt to the updated tests (I expanded then in the endowAll PR)

PR #1296, opened 2024-08-21
feat(webpack): change how inlining lockdown is configured 

naugtur:
- from array of specific strings to a regexp - this is how it's most often done in webpack plugins and it makes a lot of sense when I tried to use the feature. asset names can have hashes prepended to them for cache breaking and version tracking etc.
- a breaking change, but we're still in beta


packages/webpack/src/buildtime/emitSes.js
```diff
@@ -5,30 +5,28 @@ const {
 } = require('webpack')
 
 const lockdownSource = readFileSync(
-  path.join(require.resolve('ses'), '../lockdown.umd.min.js'),
+  path.join(require.resolve('ses'), '../lockdown.umd.js'),
```
boneskull:
```suggestion
  require.resolve('ses'),
```

I'm pretty sure that's all that's necessary, because all of the distfiles are the same.

packages/webpack/src/buildtime/emitSes.js
```diff
@@ -5,30 +5,28 @@ const {
...
-    ({ compilation, inlineLockdown: files }) =>
+    ({ compilation, inlineLockdown }) =>
     () => {
-      files.forEach((file) => {
-        const asset = compilation.assets[file]
-        if (!asset) {
-          throw new Error(
-            `LavaMoatPlugin: file specified in inlineLockdown not found in compilation: ${file}`
-          )
+      for (const file in compilation.assets) {
```
boneskull, line 19: imo `Object.entries()` would be helpful here.  the `in` throws me off a little

packages/webpack/README.md
```diff
@@ -108,7 +108,7 @@ This plugin will skip policy enforcement for such ignored modules.
 
 - [SES lockdown][] must be added to the page without any bundling or transforming for any security guarantees to be sustained.
   - The plugin is attempting to add it as an asset to the compilation for the sake of Developer Experience. `.js` extension is omitted to prevent minification.
-  - Optionally lockdown can be inlined into the bundle files. You need to list the scripts that get to load as the first script on the page to apply lockdown only once when inlined. When you have a single bundle, you just configure a list with one element. It gets more complex with builds for multiple pages. The plugin doesn't attempt to guess where to inline lockdown.
+  - Optionally lockdown can be inlined into the bundle files. You need to match the scripts that get to load as the first script on the page to apply lockdown only once when inlined. When you have a single bundle, you just configure a regex with one unique file name or a `.*`. It gets more complex with builds for multiple pages. The plugin doesn't attempt to guess where to inline lockdown.
```
boneskull: > You need to match the scripts that get to load as the first script on the page to apply lockdown only once when inlined.

Can you rephrase this?

Also, maybe example usage would be helpful, since it's a little complicated.  e.g., given `<script src="bundle-0.js">` and `<script src="bundle-1.js">`, what would be the regex

packages/webpack/src/buildtime/emitSes.js
```diff
@@ -5,30 +5,28 @@ const {
 } = require('webpack')
 
 const lockdownSource = readFileSync(
-  path.join(require.resolve('ses'), '../lockdown.umd.min.js'),
+  path.join(require.resolve('ses'), '../lockdown.umd.js'),
   'utf-8'
 )
-const lockdownSourcePrefix = `;(function(){\n${lockdownSource}\n})();\n`
+const lockdownSourcePrefix = `/* SES sources included by LavaMoat. Do not optimize or minify. */\n;\n${lockdownSource}\n;/* end SES */\n`
```
boneskull:
```suggestion
const lockdownSourcePrefix = `/*! SES sources included by LavaMoat. Do not optimize or minify. */\n;\n${lockdownSource}\n;/*! end SES */\n`
```

boneskull: Also: does adding inlining this befoul source maps?

packages/webpack/test/e2e-lockdown-inline.spec.js
```diff
@@ -4,7 +4,7 @@ const { makeConfig } = require('./fixtures/main/webpack.config.js')
 
 test.before(async (t) => {
   const webpackConfig = makeConfig({
-    inlineLockdown: ['app.js'],
+    inlineLockdown: /app/,
```
boneskull, line 7: idk if this is a snapshot test, but it seems like there should be such a thing

boneskull (dismissed):
just a couple non-blocking suggestions

packages/webpack/src/buildtime/emitSes.js
naugtur: except for the min one I used initially :D

packages/webpack/README.md
naugtur: In case only one bundle file results this is trivial. With a setup that has many entries, the regex needs to match all of them. If you're chunking, you're better off configuring webpack to emit the runtime chunk and match that. I was trying to avoid writing all this, but I could.

packages/webpack/src/buildtime/emitSes.js
naugtur: I'm afraid it does. But I'm also pretty sure wrapping each module does too. Might have to give it a separate push one day.

packages/webpack/src/buildtime/emitSes.js
naugtur: It's the most basic form of iterating fields of an object. How is it surprising? I switched to a loop for the sake of using continue

packages/webpack/test/e2e-lockdown-inline.spec.js
naugtur: this test runs the bundle and checks if lockdown was applied. 
a snapshot test would have to snapshot the bundle source and it would be brittle or even flaky.

packages/webpack/src/buildtime/emitSes.js
```diff
@@ -5,30 +5,28 @@ const {
 } = require('webpack')
 
 const lockdownSource = readFileSync(
-  path.join(require.resolve('ses'), '../lockdown.umd.min.js'),
```
legobeat, line 8: Need to remove now unused `path` import to satisfy linting

https://github.com/LavaMoat/LavaMoat/actions/runs/10849844996/job/30109919317?pr=1296

legobeat (dismissed):

packages/webpack/src/buildtime/emitSes.js
weizman: Why did we give up on the anon func? Wasn't it important to avoid polluting upper scope by accident?

weizman (dismissed):

packages/webpack/src/buildtime/emitSes.js
naugtur: SES is only polluting global scope in ways that are desired. There's plenty of wrapping involved.

weizman (approved):

PR #1298, opened 2024-08-21
feat(webpack): support scuttling

weizman:
Introduce support in the concept of scuttling to the webpack package


packages/core/src/endowmentsToolkit.js
```diff
@@ -26,6 +26,11 @@ module.exports = endowmentsToolkit
 // Exports for testing
 module.exports._test = { instrumentDynamicValueAtPath }
 
+const {
+  Reflect,
+  Object,
+} = globalThis
+
```
naugtur: I don't see how that's necessary - endowmentsToolkit runs under lockdown.

packages/webpack/src/plugin.js
```diff
@@ -74,6 +74,10 @@ class LavaMoatPlugin {
    */
   constructor(options = {}) {
     /** @type {LavaMoatPluginOptions} */
+    if (options.scuttleGlobalThis === true) {
```
naugtur, line 77: We don't need backwards compatibility when it's first introduced

packages/webpack/src/runtime/runtime.js
```diff
@@ -27,6 +34,11 @@ if (LOCKDOWN_ON) {
   )
 }
 
+// harden appears on globalThis only after lockdown is called
+const {
+  harden,
+} = globalThis
```
naugtur, line 40: no need to cache it in local scope tho, now that lockdown was called.

packages/webpack/src/runtime/runtime-namespace.ts
```diff
@@ -3,6 +3,7 @@ import { LavaMoatPluginOptions } from '../plugin'
 
 type RequiredProperty<T, K extends keyof T> = Omit<T, K> & Required<Pick<T, K>>
 export interface RuntimeNamespace {
+  scuttle: {scuttle: <T, U>(arg1: T, arg2: U) => {}}
```
naugtur: I'd go with `LAVAMOAT.<noun>.scuttle` - `scuttle.scuttle` is a bit confusing

packages/webpack/src/runtime/runtime.js
```diff
@@ -304,3 +316,5 @@ const lavamoatRuntimeWrapper = (resourceId, runtimeKit) => {
 
 // defaultExport is getting assigned to __webpack_require__._LM_
 LAVAMOAT.defaultExport = freeze(lavamoatRuntimeWrapper)
+
+LAVAMOAT.scuttle.scuttle(theRealGlobalThis, LAVAMOAT.options.scuttleGlobalThis)
```
naugtur: (big deal)
I'm not sure this is the right moment for scuttling. I think at this point the entry compartment was not created yet so we've not had the change to capture everything.

We might need this to happen right after root compartment's globalThis gets fed fields from real globalThis.

packages/webpack/test/e2e-scuttle.spec.js
```diff
@@ -0,0 +1,64 @@
...
+  'scuttling mode. To learn more visit https://github.com/LavaMoat/LavaMoat/pull/360.'
+
+test.before(async (t) => {
+  const webpackConfigDefault = makeConfig({
+    scuttleGlobalThis: {
+      enabled: true,
+      exceptions: ['JSON'],
+      scuttlerName: 'SCUTTLER',
+    },
+    generatePolicy: true,
```
naugtur, line 29: we don't want to overwrite another test's policy, so let's set this to false or provide a different policyLocation

packages/webpack/test/e2e-scuttle.spec.js
```diff
@@ -0,0 +1,64 @@
...
+    ...webpackConfigDefault,
+    entry: {
+      app: './simple.js',
+    },
+  }
+  await t.notThrowsAsync(async () => {
+    t.context.build = await scaffold(webpackConfig)
+  }, 'Expected the build to succeed')
+  t.context.bundle = t.context.build.snapshot['/dist/app.js']
+  t.context.globalThis = runScriptWithSES(
```
naugtur: this belongs in a testcase IMHO

packages/webpack/test/scaffold.js
```diff
@@ -113,7 +113,10 @@ function runScript(code, globals = defaultGlobals()) {
     throw new Error('runScript requires a bundle string as the first argument')
   }
   const context = createContext(globals)
-  return runInNewContext(code, context)
+  return {
+    context,
+    result: runInNewContext(code, context),
```
naugtur, line 125: refactoring this requires an update to all tests now.

packages/webpack/test/snapshots/e2e-scuttle.spec.js.md
```diff
@@ -0,0 +1,16 @@
+# Snapshot report for `test/e2e-scuttle.spec.js`
```
naugtur: The only snapshot we might need for this test is the snapshot of scuttled context. I'm not sure we want that.

packages/webpack/test/fixtures/main/simple.js
```diff
@@ -0,0 +1,8 @@
+function start(arr) {
+  arr.push(111)
+  return arr
+}
+
+if (start(new Array())[0] !== 111) {
```
naugtur: Is this enough to check if scuttling did not disrupt the compartment globals? From the current implementation I thought it should :sweat_smile:

packages/webpack/src/runtime/runtime.js
weizman: great catch c87ab1e (leaving unresolved so we make sure this is still the right way to do it)

packages/webpack/test/e2e-scuttle.spec.js
weizman: 9f1815a

packages/webpack/test/scaffold.js
weizman: no one's using it 🎉

packages/webpack/src/plugin.js
weizman: This isn't backwards compatibility, we decided `true` is acceptable for quick and easy enablement

packages/core/src/endowmentsToolkit.js
weizman: c95526bb0

packages/webpack/src/runtime/runtime-namespace.ts
weizman: I agree, but this is kind of how the assembler currently works, its current convention for assembling files is `<file-path/file-name>.js : <file-name>`, and since the word `scuttle` stands for both the file name and the exported function, we end up with that...

```
{
  name: 'scuttle',
  shimRequire: 'lavamoat-core/src/scuttle.js',
},
```

We could do:

```
{
  name: 'scuttling',
  shimRequire: 'lavamoat-core/src/scuttle.js',
},
```

But it introduces `scuttling` vs `scuttle` which isn't consistent with the rest of the assembled directives which I find unfortunate.

Thoughts?

packages/webpack/src/runtime/runtime.js
weizman: `harden` isn't skipped over by the scuttling process, so without these lines, it is being scuttled (see  https://github.com/LavaMoat/LavaMoat/blob/main/packages/core/src/scuttle.js#L132).

We can add `harden` to that list, or cache it here, but without both a later use of `harden` encounters a scuttled property instead of the expected value of the `harden` property of the globalThis (here https://github.com/LavaMoat/LavaMoat/blob/main/packages/webpack/src/runtime/runtime.js#L247).

weizman:
I addressed everything you asked for @naugtur , either by comment or in general in context of tests refactor 0cfc73f27 (maybe except for the snapshot comment which I could use some clarification on)

packages/webpack/src/runtime/runtime.js
```diff
@@ -1,7 +1,6 @@
 /// <reference path="./lavamoat.d.ts" />
 /* global LAVAMOAT */
-/* global lockdown, harden, Compartment */
-
+/* global Compartment */
```
boneskull, line 3: Should this be `/// <reference types="ses" />`?

packages/webpack/test/e2e-scuttle.spec.js
```diff
@@ -0,0 +1,88 @@
+const test = require('ava')
```
boneskull: if you want strict types for your test context, you could do something like

```js
// eslint-disable-next-line ava/use-test
const { default: ava } = require('ava')

/**
 * Test context object for our tests
 *
 * @typedef ScuttlingTestContext
 * @property {type} x
 * @property {type} y
 * @property {type} z
 */

const test = /** @type {import('ava').TestFn<ScuttlingTestContext>} */(ava)
```

now `t.context` will be of type `ScuttlingTestContext`

boneskull (commented):
I don't remember what I just wrote

packages/webpack/src/runtime/runtime.js
weizman: I have no idea, I just took out the parts that were no longer relevant

packages/webpack/test/e2e-scuttle.spec.js
weizman: Pretty cool. Do I want this though?

naugtur: It could suggest correct assertions fromt the `t` reference inside the test case (As opposed to copilot, which often makes up assertions that don't exist)

packages/webpack/src/plugin.js
```diff
@@ -431,6 +436,10 @@ class LavaMoatPlugin {
                     json: true,
                   },
                   { name: 'options', data: runtimeOptions, json: true },
+                  {
+                    name: 'scuttle',
```
naugtur: (should)
We should make this conditional on`options.scuttleGlobalThis` to avoid adding bulk if scuttling is off.

packages/webpack/src/runtime/runtime.js
naugtur: makes sense

naugtur:
make scuttling runtime chunk conditional on settings (see https://github.com/LavaMoat/LavaMoat/pull/1298#discussion_r1781234352) and this is good to go for me.

packages/webpack/src/plugin.js
weizman: How do i make this look good without getting the type checker to fail me? f2a6f1aa93d3103fc685553e080451f83191a903

weizman:
added 318f3c5 to support a change in https://github.com/LavaMoat/LavaMoat/pull/1329/files

packages/webpack/src/runtime/runtime-namespace.ts
naugtur: The latter looks better, because it's easier to match which name comes from where.

packages/webpack/src/runtime/runtime.js
naugtur: I remember I had an idea for a refactor that would make this nicer, but I'm not sure if it's feasible. This is for now the good way (tm)

packages/webpack/src/runtime/runtime.js
naugtur: `/// <reference types="ses" />` used to be in the local d.ts but it was impossible to untangle the esm/cjs problems with loading types from ses from dependencies - I remember we both tried that over zoom and settled on what's there now. Happy to reopen one day.

packages/webpack/src/plugin.js
naugtur: Keep the assembler unchanged, add the item to the array conditionally. I'd allow `.filter(x=>!!x)` use in the assembler to make it more convenient

packages/webpack/src/buildtime/assemble.js
```diff
@@ -43,7 +43,9 @@ const assembleRuntime = (KEY, runtimeModules) => {
         LAVAMOAT['${name}'] = module.exports;
       })();`
     }
-    assembly += `\n;/*${name}*/;\n${sourceString}`
+    if (sourceString) {
```
naugtur, line 46: This is clever, but I don't like it for one reason - passing a broken item to asseble is now a silent error.

If you want the ability to skip items, I'd rather go with 
```
 runtimeModules.filter(x=>!!x).map(({...
 ```

File: packages/webpack/src/runtime/runtime-namespace.ts (hunk 4 above)
Comment (weizman) [reply packages/webpack/src/runtime/runtime-namespace.ts]: f4a3bee33

File: packages/webpack/src/plugin.js (hunk 13 above)
Comment (weizman) [reply packages/webpack/src/plugin.js]: The type check error I'm seeing isn't about the empty object, it surfaces if I don't strictly check the type of the config before accessing it:

```javascript
 (typeof runtimeOptions?.scuttleGlobalThis === 'boolean' && runtimeOptions?.scuttleGlobalThis === true) ||
                  (typeof runtimeOptions?.scuttleGlobalThis === 'object' && runtimeOptions?.scuttleGlobalThis?.enabled === true)
```

What I'm asking is whether there's a way to tell whether scuttling is enabled or not in a type-check safe way that isn't this long and ugly

packages/webpack/src/buildtime/assemble.js
weizman: For sourceString to be updated, it takes only one of `{ file, data, name, json, shimRequire }` to be truthy, which means that such filter won't cut it, and instead we'll have to test all of them in order to filter stuff out.

Essentially, the filter suggestion is no different than the current one in context of "silent error" - filtering out broken items without letting anyone know about it is essentially the same in both scenarios, I just chose the one scenario that requires the smaller amount of checks, which is the final sink of whether the item caused a string change or not.

naugtur: Candidate for a refactor later to make the items that go into assemble more strict.

packages/webpack/test/snapshots/e2e-scuttle.spec.js.md
naugtur: I mean skip the snapshot in scuttling test because this is not useful for anything.

naugtur:
Looks good. Squash and push a proper commit name and maybe clean up the test a bit and it looks good to go.

packages/webpack/test/snapshots/e2e-scuttle.spec.js.md
weizman: like this [74de4c6](https://github.com/LavaMoat/LavaMoat/pull/1298/commits/74de4c674808e5f58835b198429c7f3f308a255d)?

boneskull:
looks like a snapshot problem

packages/webpack/src/plugin.js
```diff
@@ -74,6 +74,10 @@ class LavaMoatPlugin {
    */
   constructor(options = {}) {
     /** @type {LavaMoatPluginOptions} */
+    if (options.scuttleGlobalThis === true) {
+      options.scuttleGlobalThis = { enabled: true }
+    }
+    options.scuttleGlobalThis = Object.assign({}, options.scuttleGlobalThis)
```
boneskull: could this not be expressed as

```suggestion
    options.scuttleGlobalThis = {...options.scuttleGlobalThis}
```

?

packages/webpack/src/plugin.js
```diff
@@ -540,6 +545,12 @@ class LavaMoatPlugin {
                     json: true,
                   },
                   { name: 'options', data: runtimeOptions, json: true },
+                  (typeof runtimeOptions?.scuttleGlobalThis === 'boolean' && runtimeOptions?.scuttleGlobalThis === true) ||
+                  (typeof runtimeOptions?.scuttleGlobalThis === 'object' && runtimeOptions?.scuttleGlobalThis?.enabled === true) ?
```
```suggestion
                  (typeof runtimeOptions?.scuttleGlobalThis === 'boolean' && runtimeOptions.scuttleGlobalThis === true) ||
                  (typeof runtimeOptions?.scuttleGlobalThis === 'object' && runtimeOptions.scuttleGlobalThis.enabled === true) ?
```

packages/webpack/src/runtime/runtime-namespace.ts
```diff
@@ -7,6 +7,7 @@ type DebugTools = {
 
 type RequiredProperty<T, K extends keyof T> = Omit<T, K> & Required<Pick<T, K>>
 export interface RuntimeNamespace {
+  scuttling: {scuttle: <T, U>(arg1: T, arg2: U) => {}}
```
boneskull: 1. Why do we have generics here?
2. Do you mean to return type `{}`, which means "an object which is not required to have properties but could."

packages/webpack/src/runtime/runtime.js
boneskull: IIRC this was a bug in `ses` that I fixed.  We can look at it later though

packages/webpack/test/e2e-scuttle.spec.js
boneskull: It will provide type inference and checks for `t.context`, which you're using, so I'm not sure why you wouldn't want this.

packages/webpack/test/e2e-scuttle.spec.js
```diff
@@ -0,0 +1,87 @@
...
+const path = require('node:path')
+// eslint-disable-next-line ava/no-import-test-files
+const { scaffold, runScriptWithSES } = require('./scaffold.js')
+const {makeConfig} = require('./fixtures/main/webpack.config.js')
+
+const err = (intrinsic) =>
+  'LavaMoat - property "' +
+  intrinsic +
+  '" of globalThis is inaccessible under ' +
+  'scuttling mode. To learn more visit https://github.com/LavaMoat/LavaMoat/pull/360.'
```
boneskull, line 24: 1. (opinion) I prefer template literals
2. Linking a PR seems maybe bad and we should either a) reference an issue instead or b)  if this is a permanent situation, link a document

packages/webpack/test/e2e-scuttle.spec.js
```diff
@@ -0,0 +1,87 @@
...
+
+async function scuttle(t, scuttleGlobalThis, globals) {
+  const webpackConfigDefault = makeConfig({
+    scuttleGlobalThis,
+    generatePolicy: true,
+    diagnosticsVerbosity: 1,
+    policyLocation: path.resolve(__dirname, 'fixtures/main/policy-scuttling'),
```
boneskull, line 31: (suggestion) `__dirname` is absolute so `path.resolve` can be `path.join`.

packages/webpack/test/e2e-scuttle.spec.js
```diff
@@ -0,0 +1,87 @@
...
+  })
+  const webpackConfig = {
+    ...webpackConfigDefault,
+    entry: {
+      app: './simple.js',
+    },
+  }
+  await t.notThrowsAsync(async () => {
+    t.context.build = await scaffold(webpackConfig)
+    t.context.bundle = t.context.build.snapshot['/dist/app.js']
```
boneskull, line 41: are these paths guaranteed to be POSIX?

packages/webpack/test/fixtures/main/simple.js
```diff
@@ -0,0 +1,3 @@
+// this throws if we scuttled props of compartment globalThis instead of startCompartment globalThis by mistake
```
boneskull, line 1: (suggestion) I like Endo's convention of adding a `README` for each fixture

packages/webpack/test/fixtures/main/webpack.config.js
```diff
@@ -4,6 +4,11 @@ const HtmlWebpackPlugin = require('html-webpack-plugin')
 const path = require('path')
 
 const defaultLmOptions = {
+  scuttleGlobalThis: {
+    enabled: false,
+    exceptions: [],
+    scuttlerName: '',
```
boneskull: since the last two options are optional, can they just be omitted?

packages/webpack/test/e2e-scuttle.spec.js
```diff
@@ -0,0 +1,87 @@
...
+})
+
+test(`webpack/scuttled - hosting globalThis's environment is not scuttled`, async (t) => {
+  await scuttle(t)
+  try {
+    const global = t.context.globalThis
+    Object.getOwnPropertyNames(global).forEach(name => global[name])
+  } catch (e) {
+    t.fail(`Unexpected error in scenario: ${e.message}`)
+  }
```
boneskull: (suggestion) `try` / `catch` / `t.fail()` this is better expressed `t.notThrows(fn)`.  I'd normally just omit the assertion altogether, but I think AVA treats a test case w/o an assertion as an error.

boneskull (approved):
nothing blocking, but I do have a concern about the `RuntimeNamespace` type. if that type is externally-visible, especially

boneskull (changes requested):
actually, wait, I _do_ have something blocking: please add a description to this PR. 😄

packages/webpack/test/e2e-scuttle.spec.js
weizman: umm something like this? cb4027641

packages/webpack/test/e2e-scuttle.spec.js
weizman: 1. me too, but i further prefer keeping lines short and well indentated (am open to suggestions)
2. why an issue though? the occurrence of such an error isn't an issue to be fixed but rather something that happens intentionally which was introduced and explained in the PR. A document might make more sense, but I'm not sure it's worth focusing on atm

packages/webpack/test/e2e-scuttle.spec.js
weizman: this is a copy-paste from the rest of the code, should we maybe address all occurrences in a separate or?

packages/webpack/test/e2e-scuttle.spec.js
weizman: this is a copy-paste from the rest of the code, should we maybe address all occurrences in a separate or?
also, i am not sure

packages/webpack/test/fixtures/main/simple.js
weizman: can i see an example?

packages/webpack/test/fixtures/main/webpack.config.js
weizman: `enabled: false` is default, so basically we can omit this diff all together, I just think being explicit here is worthwhile, do you disagree?

packages/webpack/test/e2e-scuttle.spec.js
weizman: like this? 4e41f842e

packages/webpack/src/runtime/runtime-namespace.ts
weizman: 1. is this better? 4f621d3
2. i want it to return nothing, what should i use then?

packages/webpack/test/fixtures/main/webpack.config.js
boneskull: asserting the defaults work as expected has more value imo

packages/webpack/test/e2e-scuttle.spec.js
boneskull: nbd

packages/webpack/test/e2e-scuttle.spec.js
boneskull: it's very minor

packages/webpack/test/fixtures/main/simple.js
boneskull: see `endojs/endo/packages/compartment-mapper/test/fixture/*/node_modules/*/README.md` :D

packages/webpack/src/runtime/runtime-namespace.ts
```diff
@@ -7,6 +7,7 @@ type DebugTools = {
 
 type RequiredProperty<T, K extends keyof T> = Omit<T, K> & Required<Pick<T, K>>
 export interface RuntimeNamespace {
+  scuttling: {scuttle: (globalThis: any, scuttleGlobalThis: LavaMoatPluginOptions["scuttleGlobalThis"]) => {}}
```
boneskull: you said it's supposed to return nothing?

```suggestion
  scuttling: {scuttle: (globalThis: any, scuttleGlobalThis: LavaMoatPluginOptions["scuttleGlobalThis"]) => void}
```

I don't like `any` here -- it should be `Record<string, unknown>` if it doesn't cause type errors elsewhere; otherwise `Record<string, any>` is more appropriate. (`string` because IIRC we don't care about `number | symbol` which are the _other_ types of properties that could live in any given object)

I'd _strongly suggest_ to do the following, so we aren't tempted to tightly couple the type of the `scuttleGlobalThis` arg to the `LavaMoatPluginOptions` type:

1. Rename `ScuttlerConfig` to `ScuttlerObjectConfig` or `ScuttlerConfigObject` or the like
2. Create new `type ScuttlerConfig = ScuttlerObjectConfig | boolean`
3. Instead of `LavaMoatPluginOptions["scuttleGlobalThis"]` above, use the new `ScuttlerConfig`

boneskull (commented):
- please see requested typing change 
- please see my comment on the test code

everything else is nbd

packages/webpack/test/e2e-scuttle.spec.js
boneskull: yes exactly like that 😄

packages/webpack/test/fixtures/main/simple.js
weizman: i think it'll be strange if `simple` ends up being the only file there with a README

packages/webpack/test/fixtures/main/webpack.config.js
weizman: it seems that the parent object is called `defaultLmOptions` and it indicates more (probably default) stuff, i prefer not to deviate from the intention of the original author (@naugtur thoughts?)

packages/webpack/src/runtime/runtime-namespace.ts
weizman: like this? 5c7226de5

weizman: like this? a40a28044

packages/webpack/test/e2e-scuttle.spec.js
naugtur: I have a stronger opinion on 2

This should be a link to our docs page or markdown file in the repo (whichever we decide is more likely to survive longer under the same URL) that explains everything

can be a separate PR

packages/webpack/test/e2e-scuttle.spec.js
naugtur: it's a memFs, so they should be all posix.

packages/webpack/test/fixtures/main/webpack.config.js
naugtur: these are defaults for tests. As opposed to test-specific setup. 
I agree there's value in asserting default settings work in the main test and then you provide that custom setting in specific tests as you do.

packages/webpack/test/fixtures/main/simple.js
naugtur: @weizman l - start a new PR draft on a branch and do this one, then either do the rest or leave that for me.

packages/webpack/src/runtime/runtime.js
naugtur: let's fix that separately

naugtur (approved):
approve, but see latest comments, there's 2 left to address.

packages/webpack/test/fixtures/main/simple.js
weizman: not sure what to do exactly, because it's not fully comparable with Endo - do we want fixtures (main, hack1, hack2) to become folders? as in:

present:
```
main.js
```

future:
```
/main/
  index.js
  README.md
```
?

packages/webpack/test/fixtures/main/webpack.config.js
weizman: is this better? 3d64180f9

packages/webpack/test/fixtures/main/simple.js
naugtur: yes, or one README with a section for each. whichever feels nicer to maintain up to date.

weizman: https://github.com/LavaMoat/LavaMoat/pull/1517

packages/webpack/src/runtime/runtime-namespace.ts
naugtur: wait, but isn't that `globalThis` THE `globalThis`? In that case, it's far from being a record and we're about to mess it up so bad no typescript server is gonna figure it out.

packages/webpack/src/plugin.js
```diff
@@ -634,6 +645,17 @@ LavaMoatPlugin.exclude = EXCLUDE_LOADER
 
 module.exports = LavaMoatPlugin
 
+/**
+ * @typedef {Object} ScuttlerObjectConfig
```
boneskull, line 649: not even a suggestion, but "just saying": `Object` is the default extension type and can always be omitted from `@typedef`s having `@property` tags.

if the `@typedef` does _not_ extend `Object` (like [here](https://github.com/LavaMoat/LavaMoat/pull/1298/files/3d64180f957317a85231b811049a4b5fa3412962#diff-9647c300f69dfab2af75d7433e10ed31a03f45408883fcea1284b90c35f5a7ddR656)), any `@property` tags _will be ignored_.

image

packages/webpack/src/plugin.js
```diff
@@ -634,6 +645,17 @@ LavaMoatPlugin.exclude = EXCLUDE_LOADER
...
+ * @property {boolean} [enabled] - Indicates whether the feature is enabled
+ * @property {string[]} [exceptions] - A list of exceptions as strings
+ * @property {string} [scuttlerName] - The name of the scuttler
+ */
+
+/**
+ * @typedef {ScuttlerObjectConfig | boolean | undefined} ScuttlerConfig
```
boneskull, line 656: 👍

packages/webpack/test/e2e-scuttle.spec.js
```diff
@@ -0,0 +1,89 @@
...
+})
+
+test(`webpack/scuttled - provided scuttlerName successfully invoked defined scuttlerFunc`, async (t) => {
+  const scuttlerName = 'SCUTTLER';
+  await scuttle(t, {
+    enabled: true,
+    scuttlerName,
+  }, {
+    [scuttlerName]: (globalRef, scuttle) => {
+      t.context.scuttler_func_called = true
```
boneskull, line 84: it confounds me why AVA packs in assertions but not mocks

packages/webpack/src/buildtime/assemble.js
```diff
@@ -43,7 +43,9 @@ const assembleRuntime = (KEY, runtimeModules) => {
...
+      assembly += `\n;/*${name}*/;\n${sourceString}`
```
boneskull, line 47: (curiosity) is it possible for `name` to contain `*/`? if so, what happens?  is this what the semicolons are for?

boneskull (approved):

boneskull:
- [ ] rebase
- [ ] squash w/ reasonable commit message 😄

packages/webpack/src/buildtime/assemble.js
weizman: `name` is a property of a static object which is passed in here as part of an array of such objects that we hardcoded define and control ourselves (see `const runtimeChunks = [` @ `plugin.js`)

packages/webpack/src/runtime/runtime-namespace.ts
weizman: what is it then? sorry for not fully getting ts yet

naugtur: if this is the atual reference to the realm global aka globalThis aka window, it's not a string-keyed record and trying to type it properly is a futile endeavor.

PR #1301, opened 2024-08-22
feat(webpack): support thisAsExports runtime requirement

naugtur:
In case you didn't know, in cjs modules under node.js `exports === this` and webpack is capable of emulating that, so some old modules still use it.

TODO:
 - [x] add a fixture for this to the main tests

### DIFF
--- a/packages/webpack/src/buildtime/generator.js
+++ b/packages/webpack/src/buildtime/generator.js
@@ -14,7 +14,7 @@ const { RUNTIME_KEY } = require('../ENUM.json')
 
 const { isExcluded } = require('./exclude.js')
 
-// TODO: processing requirements needs to be a tiny bit more clever yet.
+// TODO: There's potential for a few more flags in runtimeFlags if we want to support all things webpack supports. Proceed with common sense.
 // Look in JavascriptModulesPlugin for how it decides if module and exports are unused.
 /**
  * @param {Set<string>} requirements
@@ -22,28 +22,26 @@ const { isExcluded } = require('./exclude.js')
  * @returns
  */
 function processRequirements(requirements, module) {
-  // TODO: the approach of passing a runtimneKit is a minimal solution.
-  // It may be possible to do more at compile time to simplify the runtime.
-  // Handling "thisAsExports" may require that.
   const runtimeKit = new Set()
   const runtimeFlags = {}
 
   for (const requirement of requirements) {
-    const chunks = requirement.split('.')
-    if (chunks[0] === RUNTIME_GLOBALS.thisAsExports) {
+    // requirements can be more precise than just `module` - webpck will list nested fields in requirements, meanwhile we're only interested in passing the top level references.
+    const requirementReferenceName = requirement.split('.')[0]
+    if (requirementReferenceName === RUNTIME_GLOBALS.thisAsExports) {
       runtimeFlags.thisAsExports = true
       continue
     }
-    if (chunks[0] === RUNTIME_GLOBALS.returnExportsFromRuntime) {
+    if (requirementReferenceName === RUNTIME_GLOBALS.returnExportsFromRuntime) {
       // TODO: should be doable to introduce support in wrapper.js by conditionally adding a return statement. feels too niche to support
       continue
     }
-    if (chunks[0] === '__webpack_exports__') {
+    if (requirementReferenceName === '__webpack_exports__') {
       runtimeKit.add(module.exportsArgument)
-    } else if (chunks[0] === 'module') {
+    } else if (requirementReferenceName === 'module') {
       runtimeKit.add(module.moduleArgument)
     } else {
-      runtimeKit.add(chunks[0])
+      runtimeKit.add(requirementReferenceName)
     }
   }
   diag.run(2, () => {
@@ -57,7 +55,6 @@ function processRequirements(requirements, module) {
 // this is to avoid wrapping the same instance twice
 const wrappedGeneratorInstances = new WeakSet()
 
-// TODO: this should probably be extracted to a separate file for easier navigation
 /**
  * @param {object} options
  * @param {string[]} options.excludes
@@ -79,6 +76,7 @@ exports.wrapGeneratorMaker = ({
     // Monkey-patching JavascriptGenerator. Yes, this could be nicer.
     // Using features of the generator itself we might be able to achieve the same
     // but it would be more suseptible to changes in webpack.
+    // And there aren't any official or private hooks that would give us access to runtime requirements that I could find.
 
     if (wrappedGeneratorInstances.has(generatorInstance)) {
       return generatorInstance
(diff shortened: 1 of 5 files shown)


packages/webpack/src/buildtime/generator.js
```diff
@@ -26,16 +26,16 @@ function processRequirements(requirements, module) {
   // It may be possible to do more at compile time to simplify the runtime.
   // Handling "thisAsExports" may require that.
   const runtimeKit = new Set()
+  const runtimeFlags = {}
+
   for (const requirement of requirements) {
     const chunks = requirement.split('.')
     if (chunks[0] === RUNTIME_GLOBALS.thisAsExports) {
```
naugtur: @boneskull would be less scared if this was not called `chunk`

packages/webpack/src/buildtime/generator.js
```diff
@@ -14,50 +14,47 @@ const { RUNTIME_KEY } = require('../ENUM.json')
...
   const runtimeKit = new Set()
+  const runtimeFlags = {}
+
   for (const requirement of requirements) {
-    const chunks = requirement.split('.')
-    if (chunks[0] === RUNTIME_GLOBALS.thisAsExports) {
-      // TODO: not sure what to do with it.
-      //github.com/webpack/webpack/blob/07ac43333654280c5bc6014a3a69eda4c3b80273/lib/javascript/JavascriptModulesPlugin.js#L560
+    // requirements can be more precise than just `module` - webpck will list nested fields in requirements, meanwhile we're only interested in passing the top level references.
+    const requirementReferenceName = requirement.split('.')[0]
```
legobeat, line 30: Can this result in duplicates for requirements coming from the same top-level ref? If so, should/can we skip/filter already-processed entries?

packages/webpack/src/buildtime/generator.js
```diff
@@ -14,50 +14,47 @@ const { RUNTIME_KEY } = require('../ENUM.json')
...
-  // Handling "thisAsExports" may require that.
   const runtimeKit = new Set()
+  const runtimeFlags = {}
+
   for (const requirement of requirements) {
-    const chunks = requirement.split('.')
-    if (chunks[0] === RUNTIME_GLOBALS.thisAsExports) {
-      // TODO: not sure what to do with it.
-      //github.com/webpack/webpack/blob/07ac43333654280c5bc6014a3a69eda4c3b80273/lib/javascript/JavascriptModulesPlugin.js#L560
+    // requirements can be more precise than just `module` - webpck will list nested fields in requirements, meanwhile we're only interested in passing the top level references.
```
legobeat, line 29:
```suggestion
    // requirements can be more precise than just `module`: webpack will list nested modules as .-separated keys. Here we filter out everything except the top-level references.
```

packages/webpack/src/buildtime/generator.js
naugtur: runtimeKit is a set. Flags are an object. There's no room for duplication in data structures returned.

packages/webpack/src/buildtime/wrapper.js
```diff
@@ -3,13 +3,18 @@ const diag = require('./diagnostics')
 const fs = require('node:fs')
 const q = JSON.stringify
 
+/**
+ * @typedef {object} RuntimeFlags
+ * @property {boolean} [thisAsExports]
+ */
```
boneskull:
```js
/**
 * Describe me
 * 
 * @typedef {object} RuntimeFlags
 * @property {boolean} [thisAsExports] Describe me
 */
```

packages/webpack/src/buildtime/wrapper.js
```diff
@@ -33,13 +38,26 @@ exports.wrapper = function wrapper({
...
 
+  // This adds support for mapping `this` to `exports` or `module.exports` if webpack detected it's necessary
+  let optionalBinding = ''
+
+  if (runtimeFlags.thisAsExports) {
+    if (runtimeKitArray.includes('exports')) {
+      optionalBinding = '.bind(exports)'
+    } else if (runtimeKitArray.includes('module')) {
+      optionalBinding = '.bind(module.exports)'
+    }
```
boneskull, line 61: in case it has any bearing on this logic: you can use both `exports` and `module.exports` in the same damn script if you want

boneskull (dismissed):
no major complaints

packages/webpack/src/buildtime/generator.js
naugtur: that's untrue. 

`module` here is the functor's argument `(module, exports, __dirname...`  not the word module

and here's an example of what's in requirements:
`__webpack_require__,__webpack_require__.n,__webpack_require__.r,__webpack_exports__`

packages/webpack/src/buildtime/wrapper.js
naugtur: yes, this is here because if 'exports' are not used, webpack doesn't list them as a requirement and they're actually undefined in the scope of the module, you have to reach for `module.exports` to grab that reference.

packages/webpack/test/fixtures/main/node_modules/commonjs-quirks/this.js
```diff
@@ -0,0 +1,8 @@
+const thisThis = this
+
+// a line to make sure that the file is considered cjs
+exports.zz=123;
+
+if(thisThis !== exports){
```
weizman, line 6: why not just `this`?

weizman: why not:

```javascript
exports.zz=123;
if (exports !== this || exports.zz !== this.zz || this.zz !== 123) throw 
```

?

packages/webpack/src/buildtime/wrapper.js
```diff
@@ -33,13 +41,26 @@ exports.wrapper = function wrapper({
   runtimeKit,
   evalKitFunctionName,
   runChecks = true,
+  runtimeFlags = {},
 }) {
+  const runtimeKitArray = Array.from(runtimeKit)
   // validateSource(source);
 
   // No AST used in these transforms, so string cmparison should indicate if anything was changed.
```
weizman, line 49:
```suggestion
  // No AST used in these transforms, so string comparison should indicate if anything was changed.
```

packages/webpack/src/buildtime/wrapper.js
```diff
@@ -33,13 +41,26 @@ exports.wrapper = function wrapper({
...
   const sesCompatibleSource = applySourceTransforms(source)
   const sourceChanged = source !== sesCompatibleSource
 
+  // This adds support for mapping `this` to `exports` or `module.exports` if webpack detected it's necessary
+  let optionalBinding = ''
+
+  if (runtimeFlags.thisAsExports) {
```
weizman, line 56: I think this is great, but it took me a while to realize that `exports` and `module` are hardcoded statements coming from webpack, I feel like if there's a way to grab them and use them instead, it would make this code more clear (if possible):

```javascript
const {exportsArgument, moduleArgument} = Webpack.consts.arguments;
```

```javascript
if runtimeKitArray.includes(exportsArgument) 
if runtimeKitArray.includes(moduleArgument) 
```

weizman (approved):
lgtm (some comments)

PR #1302, opened 2024-08-22
fix(webpack): overwrite global aliases even if endowed

naugtur:
I think that's the desired behavior. We need to consider both the root compartment and package compartments.

cc @kumavis


boneskull (dismissed):

packages/webpack/test/fixtures/main/node_modules/globals-package/index.js
```diff
@@ -4,6 +4,13 @@ if(typeof window === 'undefined' || !window.console) {
 if(typeof globalThis === 'undefined' || !globalThis.console) {
   throw Error('globalThis is missing')
 }
+// Utter globalThis.self to trigger it being explicitly added to policy.
+// When one of the global aliases is listed in policy as an endwment, 
+// it might clash with the pre-defined global alias we're adding to each compartment.
+// There used to be an error thrown from endowments at runtime:
+// `Cannot redefine property: self` that this line would cause.
+globalThis.self;
```
naugtur, line 12: this IS the test case

legobeat (approved):

PR #1308, opened 2024-08-26
fix(webpack): an implementation of builtins that actually works

naugtur:
- previous implementation relied on information that was not available in webpack production mode
- current implementation is more resilient as it only looks up known external modules as builtins, not everything in a module that has builtins allowed.

### DIFF
--- a/packages/webpack/src/buildtime/aa.js
+++ b/packages/webpack/src/buildtime/aa.js
@@ -51,6 +51,7 @@ const crossReference = (neededIds, policyIds) => {
  * @typedef {Object} IdentifierLookup
  * @property {string} root
  * @property {(string | number)[]} unenforceableModuleIds
+ * @property {Record<string | number, string>} externals
  * @property {[string, (string | number)[]][]} identifiersForModuleIds
  * @property {(path: string) => string | undefined} pathToResourceId
  * @property {(id: string) => string} policyIdentifierToResourceId
@@ -63,6 +64,7 @@ const crossReference = (neededIds, policyIds) => {
  * @param {import('lavamoat-core').LavaMoatPolicy} options.policy
  * @param {import('@lavamoat/aa').CanonicalNameMap} options.canonicalNameMap
  * @param {(string | number)[]} options.unenforceableModuleIds
+ * @param {Record<string | number, string>} options.externals
  * @param {boolean | undefined} options.readableResourceIds
  * @returns {IdentifierLookup}
  */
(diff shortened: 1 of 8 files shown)


packages/webpack/src/buildtime/aa.js
```diff
@@ -63,6 +64,7 @@ const crossReference = (neededIds, policyIds) => {
  * @param {import('lavamoat-core').LavaMoatPolicy} options.policy
  * @param {import('@lavamoat/aa').CanonicalNameMap} options.canonicalNameMap
  * @param {(string | number)[]} options.unenforceableModuleIds
+ * @param {Record<string | number, string>} options.externals
```
legobeat, line 67: Can the (at least input) type be constrained to `Record<string, string>`? If not: under what circumstances is the `number` index required?

naugtur:
fixing the types :face_in_clouds:

packages/webpack/src/buildtime/aa.js
naugtur: webpack uses strings and numbers interchangeably depending on mode.

legobeat:
> fixing the types 😶‍🌫️

@naugtur Still have anything to address here or do you consider this good to go?

naugtur:
> > fixing the types 😶‍🌫️
> 
> @naugtur Still have anything to address here or do you consider this good to go?

I pushed a commit titled "fix types" soon after posting the comment. The types are fixed and it's been ready for review. The comment was here to acknowledge that I see the CI failing on type check.

legobeat (approved):

packages/webpack/test/e2e-nodejs.spec.js
```diff
@@ -13,7 +13,7 @@ test.before(async (t) => {
     policyLocation: path.resolve(__dirname, 'fixtures/main/policy-node'),
   })
   webpackConfig.target = 'node'
-  webpackConfig.mode = 'development'
+  webpackConfig.optimization.minimize = false
```
leotm, line 16: why disable minification on node app bundles btw?

i note removing `webpackConfig.mode = 'development'` means `production` by default (so minimizing, tree shakes, etc)

naugtur: this particular test needs to be in production mode to avoid webpack-generated module names coinciding with builtins that they're wrapping. I disable minification so that it's easier to look into the output. In production mode the module names are replaced with numbers and the old builtins implementation was naively relying on webpack using the right names which is not a great assumption.

PR #1317, opened 2024-08-29
chore: corepack is devDependency of lavamoat-node

legobeat:
This changes `corepack` from being in `optionalDependencies` of root project to `devDependencies` of project where it is actually used in tests:
- chore(lavamoat-node): add `corepack` to `devDependencies` 
- chore: remove `corepack` from `optionalDependencies`

---

#### Blocking
- #1218


naugtur (approved):
I don't see why not

PR #1319, opened 2024-08-29
feat(webpack): support ContextModule - skip enforcement of policy

naugtur:


legobeat (dismissed):

legobeat:
Needs conflict-resolution

legobeat (approved):

legobeat:
@naugtur  Needs conflict-resolution again, if you re-request review when rebased I can re-approve/merge

packages/webpack/src/plugin.js
```diff
@@ -254,6 +254,17 @@ class LavaMoatPlugin {
...
+        /**
+         * Checks if a module is a context module.
+         *
+         * @param {any} m - The module to check.
+         * @param {string} moduleClass - The class of the module.
+         * @returns {boolean} - Returns true if the module is a context module,
+         *   otherwise false.
+         */
+        const isContextModule = (m, moduleClass) =>
+          moduleClass === 'ContextModule'
```
boneskull, line 266:
```suggestion
        /**
         * Checks if a module is a context module.
         *
         * @param {string} moduleClass - The class of the module.
         * @returns {moduleClass is 'ContextModule'} - Returns true if the module is a context module,
         *   otherwise false.
         */
        const isContextModule = (moduleClass) =>
          moduleClass === 'ContextModule'
```

`m` is doing nothing here?

boneskull: I'm not sure we need a function for this, if so

PR #1324, opened 2024-08-29
chore(ci): no windows tests for allow-scripts, yarn-plugin-allow-scripts

legobeat:
- chore(ci): Disable Windows tests for `allow-scripts`, `yarn-plugin-allow-scripts`


naugtur (changes requested):
the reason for this no longer exists

PR #1326, opened 2024-08-29
fix(deps): elliptic@6.5.5->6.5.7

legobeat:
- CVE-2024-42461 / CVE-2024-42460 / CVE-2024-42459

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8450,9 +8450,9 @@
       "dev": true
     },
     "node_modules/elliptic": {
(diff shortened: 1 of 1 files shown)


naugtur (approved):
any idea why didn't we get this from renovate?

legobeat:
> any idea why didn't we get this from renovate?

renovate only looks at our direct dependencies as declared in `package.json`. In this case it's a transitive dependency only visible in the lockfile. dependabot would catch it, though.

PR #1329, opened 2024-09-02
feat(webpack): policy debugging capabilities and tighter tests

naugtur:
The way this works is when policy generation fails to notice something is necessary, this runtime addition will display warnings in the console and collect what it finds in a format that's matching policy.json

It's likely to detect many missing policy entries before the first error is thrown. Before this tool you'd have to run the bundle, pause on an error, find out what the field name is, rebuild with an override and proceed to the next error. More difficult if no error is thrown, but `undefined` is silently but incorrectly accepted.

The tool is prototyped here, but a more general version of it could be created in core. The drawback of it being in core for the webpack plugin would be that now the debug implementation is included in the bundle conditionally.


legobeat:
Needs conflict-resolution with `main`

weizman:
I'm a bit lost.. where might this be useful? Mind demonstrating?

naugtur:
> I'm a bit lost.. where might this be useful? Mind demonstrating?

Edited the description to explain. I used it to find out what I need to put in policy-override

packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -2,15 +2,15 @@ const test = require('ava')
 const endowmentsToolkit = require('../src/endowmentsToolkit.js')
 
 function prepareTest({ knownWritable } = {}) {
```
boneskull, line 4: `prepareTest` may make sense as an AVA macro

packages/webpack/src/runtime/debug.js
```diff
@@ -0,0 +1,96 @@
+/**
+ * @type {Record<string, { globals: Record<string, boolean> }>}
+ */
+const incrementalPolicy = {}
+
+const printPolicyDebug = () => {
+  const policy = JSON.stringify(incrementalPolicy, null, 2)
```
boneskull, line 7: does it matter that we're not using `json-stable-stringify` here?

packages/webpack/src/runtime/debug.js
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Adds a key to the incremental policy.
+ *
+ * @param {string} hint - The hint for the incremental policy.
+ * @param {string} key - The key to be added to the incremental policy.
+ * @returns {void}
+ */
+function addToPolicy(hint, key) {
+  if (!incrementalPolicy[hint]) {
```
boneskull:
```suggestion
  if (Object.hasOwn(incrementalPolicy, hint)) {
```

**UPDATE**: 🤣

packages/webpack/src/runtime/debug.js
```diff
@@ -0,0 +1,96 @@
...
+    incrementalPolicy[hint] = { globals: Object.create(null) }
+  }
+  if (!incrementalPolicy[hint].globals[key]) {
+    incrementalPolicy[hint].globals[key] = true
```
boneskull, line 31: Would `globals` make more sense as a `Set`?

packages/webpack/src/runtime/debug.js
```diff
@@ -0,0 +1,96 @@
...
+    const informativeStack =
+      '\n' + (Error().stack || '').split('\n').slice(2).join('\n')
+    console.log(`-- missing ${key} from ${hint}`, informativeStack)
+    clearTimeout(debounceTimer)
+    debounceTimer = setTimeout(printPolicyDebug, 3000)
```
boneskull: magic number

packages/webpack/src/runtime/debug.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+}
+/**
+ * Creates a recursive proxy object.
```
boneskull: that does what?

packages/webpack/src/runtime/debug.js
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/**
+ * @param {object | null} obj
+ */
+function getAllKeys(obj) {
+  const keys = []
+  for (; obj != null; obj = Object.getPrototypeOf(obj)) {
+    keys.push(...Object.getOwnPropertyNames(obj))
+  }
```
boneskull, line 76: isn't there a way to do this by calling `getOwnPropertyDescriptors` or something

packages/webpack/src/runtime/debug.js
```diff
@@ -0,0 +1,96 @@
...
+        path = [...path, key]
+        addToPolicy(hint, path.join('.'))
+        return recursiveProxy(hint, path)
+      },
+    }
+  )
+}
+
+/**
+ * @param {object | null} obj
```
boneskull, line 70: should add a description about the intent: we want all keys from the object and its prototype chain?  also we cannot get `Symbol` props this way, if it matters

packages/webpack/src/runtime/debug.js
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * @param {any} target - The target object to create a debug proxy for.
+ * @param {object} source - The keys to check for in the target object.
+ * @param {string} hint - The hint to identify the source of the missing key.
+ */
+const debugProxy = (target, source, hint) => {
+  const inheritedFromObj = Object.getOwnPropertyNames(Object.prototype)
+
+  const keys = getAllKeys(source)
+  keys.forEach((key) => {
```
boneskull: use `for` loop?  seems inconsistent

packages/webpack/src/runtime/runtime-namespace.ts
```diff
@@ -1,6 +1,10 @@
 import { EndowmentsToolkitFactory, LavaMoatPolicy } from 'lavamoat-core'
 import { LavaMoatPluginOptions } from '../plugin'
 
+type DebugTools = {
```
boneskull, line 4: seems like this should be exported if `runtime.js` is using it?

boneskull (approved):
this seems to instrument objects for dumping.  what I don't understand is the relationship to policy.  I feel like you might want to walk me through it.

could it be useful outside of this package's context?  

I don't see anything obviously wrong or bad.

packages/webpack/src/runtime/debug.js
naugtur: It's only here to make transferring the information collected into the policy-override.json - so it's a copy+paste+adjust and not copy-paste each global individually from console.

packages/webpack/src/runtime/debug.js
naugtur: they're in a shape that resembles policy more - makes it easier to grab them from `printPolicyDebug`

packages/webpack/src/runtime/debug.js
naugtur: not if you want to go down to the prototypes. The idea is to instrument all keys that could have been available.

packages/webpack/src/runtime/runtime-namespace.ts
naugtur: It seemed to propagate without that. I'll doublecheck.

boneskull (commented):
_ping_

packages/core/test/endowmentsToolkit.spec.js
naugtur: I rewrote it to macros and gotta say it was much harder to understand what's going on and what's coming from where.

packages/webpack/src/runtime/runtime-namespace.ts
naugtur: It's only reachable through LAVAMOAT.debug below and therefore seems suffice to have it be local to here.

naugtur:
@boneskull I addressed the review comments. Should be good to go.

packages/webpack/src/runtime/debug.js
```diff
@@ -14,6 +14,8 @@ globalThis.LM_printPolicyDebug = printPolicyDebug
  */
 let debounceTimer
 
+const PRINT_AFTER_NO_NEW_POLICY_DISCOVERED_MS = 3000
```
boneskull, line 17: 👍

packages/webpack/src/runtime/debug.js
```diff
@@ -58,6 +64,9 @@ function recursiveProxy(hint, path) {
 }
 
 /**
+ * Finds all non-symbol keys for an object and its prototype chain. Symbol keys
+ * are not included to simplify the implementation of debugProxy.
```
boneskull, line 68: 👍

boneskull (approved):

PR #1332, opened 2024-09-04
test(allow-scripts): fix allow-scripts tests on windows

naugtur:
The problem was a security fix that blocks dangerous ways of invoking things for Windows. Had to add a param to spawn.

https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2
https://github.com/nodejs/node/issues/52554


packages/allow-scripts/test/index.spec.js
```diff
@@ -77,10 +77,9 @@ test('cli - auto command', (t) => {
     realisticEnvOptions(projectRoot)
   )
 
-  // forward error output for debugging
-  if (typeof initResult.stderr !== 'undefined') {
-    t.log(initResult.stderr.toString('utf-8'))
```
naugtur, line 82: when spawn results in an error before spawning anything, `stderr` is null.

legobeat (approved):

PR #1335, opened 2024-09-05
chore: align internal dependency ranges

legobeat:
Aligning workspace dependency versions to not pull old releases from npmjs.org.

- Follow-up to #1228


naugtur (approved):
don't see why not

PR #1350, opened 2024-09-11
feat(core): sort policy in a diff-friendly way

naugtur:
## Provides more stable order for the policy file entries.

If dependency tree gets rearranged and the canonical name of a package changes, the order will not be affected in most cases, as the resources are now sorted primarily by the package name and not the entire canonical name. This is especially useful when a package (and its dependencies) end up with a shorter canonical name as one of the indirect dependencies becomes a direct dependency somewhere in the tree.

Introducing this update will rearrange the policy once. While it's not a breaking change, it's recommended to do it independently of updating the dependency tree for clarity. 

---
- [x] Is this a breaking change?
  - I now strongly believe it's not a breaking change. All it's changing is the order of fields but no actual content is changed.
- [x] add a test with snapshots to verify it works (no current tests were broken after sorting)
  - added a snapshot test


boneskull:
Why _wouldn't_ it be a breaking change?  Is it possible to still support the legacy format, but soft-deprecate it?  Then sunset support for the old format in the next major.

naugtur:
It's just changing the order of the JSON file, not a single character is changed in the output.  
It's not the initial proposal where resource names were rearranged.

boneskull:
> It's just changing the order of the JSON file, not a single character is changed in the output. It's not the initial proposal where resource names were rearranged.

wee-bey

legobeat (approved):
LGTM and also think this is non-breaking. Might still want to highlight and expand a little bit on this for release notes anyway, though.

PR #1358, opened 2024-09-17
chore(allow-scripts): handle spaces in execPath in Windows tests

legobeat:
Fixes breaking test on Windows when nodejs installed under `C:\Program Files`

### DIFF
--- a/packages/allow-scripts/test/utils.js
+++ b/packages/allow-scripts/test/utils.js
@@ -1,0 +1,12 @@
+const { platform } = require('node:os')
+
+const isWindows = platform() === 'win32'
+
+// https://github.com/nodejs/node/issues/38490
+const fixWindowsExecPath = (inPath) =>
+  '"' + inPath.replace(/"/g, '\\"') + '"'
+
+module.exports = {
+  fixWindowsExecPath,
+  isWindows,
+};
--- a/packages/allow-scripts/test/utils.js
+++ b/packages/allow-scripts/test/utils.js
@@ -6,7 +6,19 @@ const isWindows = platform() === 'win32'
 const fixWindowsExecPath = (inPath) =>
   '"' + inPath.replace(/"/g, '\\"') + '"'
 
+const portableExecPath = (inPath, debugLogging) => {
+  if (!isWindows) {
+    return inPath;
+  }
+  const portablePath = fixWindowsExecPath(inPath);
+  if (debugLogging && inPath !== portablePath) {
+    console.log(`@lavamoat/allow-scripts: transforming absolute path from ${inPath} to ${portablePath}`);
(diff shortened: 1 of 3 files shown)


packages/allow-scripts/test/utils.js
```diff
@@ -0,0 +1,15 @@
+const { platform } = require('node:os')
+
+const isWindows = platform() === 'win32'
+
+// https://github.com/nodejs/node/issues/38490
+const fixWindowsExecPath = (p) =>
+  '"' + p.replace(/"/g, '\\"') + '"'
```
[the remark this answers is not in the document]
legobeat: backslash characters don't need escaping here

packages/allow-scripts/test/utils.js
```diff
@@ -0,0 +1,12 @@
+const { platform } = require('node:os')
+
+const isWindows = platform() === 'win32'
+
+// https://github.com/nodejs/node/issues/38490
+const fixWindowsExecPath = (p) =>
```
naugtur: (should) It'd be harder to forget to use it if it was a general function for compatibility fixes that is always called on paths in allow-scripts test and applies the fixes conditionally inside instead of being applied conditionally in the setup.

legobeat: Added wrapping `portableExecPath` in e4529a329548cb7bc6ef20b7cc1ae19b6bb2e418

packages/allow-scripts/test/utils.js
```diff
@@ -0,0 +1,24 @@
...
+  '"' + p.replace(/"/g, '\\"') + '"'
```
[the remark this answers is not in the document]
legobeat: On Windows, `"` and `\` are forbidden in directory names.

packages/allow-scripts/test/utils.js
weizman: p --> path

packages/allow-scripts/test/utils.js
```diff
@@ -0,0 +1,24 @@
...
+
+const portableExecPath = (p, debugLogging) => {
```
weizman: p --> path

weizman (dismissed):
great work

packages/allow-scripts/test/utils.js
legobeat: I don't like using `path` specifically as a local var due to common usage of it as reference to `path` module. I went with the (uglier, but hopefully addressing the motivation behind not wanting `p`) `inPath` as a compromise here.

legobeat:
> great work

Thanks! Just replaced `p` with `inPath`, otherwise identical to before.

PR #1375, opened 2024-09-30
feat(webpack): unlockedChunksUnsafe option to refrain from protecting…

naugtur:
… selected chunks

The purpose of this feature is to allow skipping policy enforcement on a chunk of the bundle that is meant to result from the same build process but is not supposed to be protected (eg. inpage.js in metamask-extension)

### DIFF
--- a/packages/webpack/src/plugin.js
+++ b/packages/webpack/src/plugin.js
@@ -565,76 +565,95 @@ class LavaMoatPlugin {
                   )
                   return
                 }
-                diag.rawDebug(
-                  1,
-                  '> adding runtime (additionalChunkRuntimeRequirements)'
-                )
-                // narrow down the policy and map to module identifiers
-                const policyData = identifierLookup.getTranslatedPolicy()
-
-                const runtimeChunks = [
-                  {
-                    name: 'root',
-                    data: identifierLookup.root || null,
-                    json: true,
-                  },
-                  {
-                    name: 'idmap',
-                    data: identifierLookup.identifiersForModuleIds || null,
-                    json: true,
-                  },
-                  {
-                    name: 'unenforceable',
-                    data: identifierLookup.unenforceableModuleIds || null,
-                    json: true,
-                  },
-                  {
-                    name: 'ctxm',
-                    data: identifierLookup.contextModuleIds || null,
-                    json: true,
-                  },
-                  {
-                    // known chunk ids
-                    name: 'kch',
-                    data: chunkIds,
-                    json: true,
-                  },
-                  {
-                    name: 'externals',
-                    data: identifierLookup.externals || null,
-                    json: true,
-                  },
-                  { name: 'options', data: runtimeOptions, json: true },
-                  (typeof runtimeOptions?.scuttleGlobalThis === 'boolean' &&
-                    runtimeOptions.scuttleGlobalThis === true) ||
-                  (typeof runtimeOptions?.scuttleGlobalThis === 'object' &&
-                    runtimeOptions.scuttleGlobalThis.enabled === true)
-                    ? {
-                        name: 'scuttling',
-                        shimRequire: 'lavamoat-core/src/scuttle.js',
-                      }
-                    : {},
-                  { name: 'policy', data: policyData, json: true },
-                  {
-                    name: 'ENUM',
-                    file: path.join(__dirname, './ENUM.json'),
-                    json: true,
-                  },
-                  {
-                    name: 'endowmentsToolkit',
-                    shimRequire: 'lavamoat-core/src/endowmentsToolkit.js',
-                  },
-                  {
-                    name: 'runtime',
-                    file: path.join(__dirname, './runtime/runtime.js'),
-                  },
-                ]
-
-                if (options.debugRuntime) {
-                  runtimeChunks.push({
-                    name: 'debug',
-                    shimRequire: path.join(__dirname, './runtime/debug.js'),
-                  })
+                let runtimeChunks = []
+                if (
+                  chunk.name &&
+                  options.unlockedChunksUnsafe?.test(chunk.name)
+                ) {
+                  diag.rawDebug(
+                    1,
+                    `> adding UNLOCKED runtime for chunk ${chunk.name}`
+                  )
+                  runtimeChunks = [
+                    {
+                      name: 'ENUM',
+                      file: path.join(__dirname, './ENUM.json'),
+                      json: true,
+                    },
+                    {
+                      name: 'runtime',
+                      file: require.resolve('./runtime/runtimeUnlocked.js'),
+                    },
+                  ]
+                } else {
+                  diag.rawDebug(1, `> adding runtime for chunk ${chunk.name}`)
+                  // narrow down the policy and map to module identifiers
+                  const policyData = identifierLookup.getTranslatedPolicy()
+  
+                  runtimeChunks = [
+                    {
+                      name: 'root',
+                      data: identifierLookup.root || null,
+                      json: true,
+                    },
+                    {
+                      name: 'idmap',
+                      data: identifierLookup.identifiersForModuleIds || null,
+                      json: true,
+                    },
+                    {
+                      name: 'unenforceable',
+                      data: identifierLookup.unenforceableModuleIds || null,
+                      json: true,
+                    },
+                    {
+                      name: 'ctxm',
+                      data: identifierLookup.contextModuleIds || null,
+                      json: true,
+                    },
+                    {
+                      // known chunk ids
+                      name: 'kch',
+                      data: chunkIds,
+                      json: true,
+                    },
+                    {
+                      name: 'externals',
+                      data: identifierLookup.externals || null,
+                      json: true,
+                    },
+                    { name: 'options', data: runtimeOptions, json: true },
+                    (typeof runtimeOptions?.scuttleGlobalThis === 'boolean' &&
+                      runtimeOptions.scuttleGlobalThis === true) ||
+                    (typeof runtimeOptions?.scuttleGlobalThis === 'object' &&
+                      runtimeOptions.scuttleGlobalThis.enabled === true)
+                      ? {
+                          name: 'scuttling',
+                          shimRequire: 'lavamoat-core/src/scuttle.js',
+                        }
+                      : {},
+                    { name: 'policy', data: policyData, json: true },
+                    {
+                      name: 'ENUM',
+                      file: path.join(__dirname, './ENUM.json'),
+                      json: true,
+                    },
+                    {
+                      name: 'endowmentsToolkit',
+                      shimRequire: 'lavamoat-core/src/endowmentsToolkit.js',
+                    },
+                    {
+                      name: 'runtime',
+                      file: path.join(__dirname, './runtime/runtime.js'),
+                    },
+                  ]
+  
+                  if (options.debugRuntime) {
+                    runtimeChunks.push({
+                      name: 'debug',
+                      shimRequire: path.join(__dirname, './runtime/debug.js'),
+                    })
+                  }
                 }
                 const lavaMoatRuntime = assembleRuntime(
                   RUNTIME_KEY,
--- a/packages/webpack/src/plugin.js
+++ b/packages/webpack/src/plugin.js
@@ -589,7 +589,7 @@ class LavaMoatPlugin {
                   diag.rawDebug(1, `> adding runtime for chunk ${chunk.name}`)
                   // narrow down the policy and map to module identifiers
                   const policyData = identifierLookup.getTranslatedPolicy()
-  
+
                   runtimeChunks = [
                     {
                       name: 'root',
@@ -647,7 +647,7 @@ class LavaMoatPlugin {
                       file: require.resolve('./runtime/runtime.js'),
                     },
                   ]
-  
+
                   if (options.debugRuntime) {
                     runtimeChunks.push({
                       name: 'debug',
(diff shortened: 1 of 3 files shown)


naugtur:
Started with the exception to protections
Planning to add some defensive tricks to make this at least safer than nothing

naugtur:
Policies updated.  
👀 Please review the diff for suspicious new powers.  

🧠 Learn how: https://lavamoat.github.io/guides/policy-diff/#what-to-look-for-when-reviewing-a-policy-diff

packages/webpack/src/plugin.js
```diff
@@ -464,50 +464,72 @@ class LavaMoatPlugin {
...
+                  options.unlockedChunksUnsafe?.test(chunk.name)
+                ) {
+                  diag.rawDebug(
+                    1,
+                    `> adding UNLOCKED runtime for chunk ${chunk.name}`
+                  )
+                  runtimeChunks = [
+                    {
+                      name: 'ENUM',
+                      file: path.join(__dirname, './ENUM.json'),
```
```suggestion
                      file: path.join(__dirname, 'ENUM.json'),
```

packages/webpack/src/plugin.js
```diff
@@ -464,50 +464,72 @@ class LavaMoatPlugin {
...
+                      json: true,
+                    },
+                    {
+                      name: 'runtime',
+                      file: path.join(
+                        __dirname,
+                        './runtime/runtimeUnlocked.js'
```
```suggestion
                      file: path.join(
                        __dirname,
                        'runtime',
                        'runtimeUnlocked.js'
```

packages/webpack/src/plugin.js
boneskull: suggestion:

- make string constants for filenames e.g., `const ENUM_FILENAME = 'ENUM.json'`
- other constants for other strings having significant names (e.g., `endowmentsToolkit`)

packages/webpack/src/plugin.js
```diff
@@ -464,50 +464,72 @@ class LavaMoatPlugin {
...
+                  ]
+                } else {
+                  diag.rawDebug(1, `> adding runtime for chunk ${chunk.name}`)
+                  // narrow down the policy and map to module identifiers
+                  const policyData = identifierLookup.getTranslatedPolicy()
+
+                  runtimeChunks = [
+                    {
+                      name: 'root',
+                      data: identifierLookup.root || null,
```
boneskull: do we mean `||` or `??`?

packages/webpack/src/runtime/runtimeUnlocked.js
```diff
@@ -0,0 +1,47 @@
+/// <reference path="./lavamoat.d.ts" />
+/* global LAVAMOAT */
```
boneskull, line 2: may be better defined in `.eslintrc.js`

packages/webpack/src/runtime/runtimeUnlocked.js
```diff
@@ -0,0 +1,47 @@
...
+      exportsReference = value
+    },
+  })
+  freeze(runtimeHandler)
+
+  return {
+    [NAME_scopeTerminator]: create(null),
+    [NAME_runtimeHandler]: runtimeHandler,
+    [NAME_globalThis]: globalThis, // TODO: apply some defensive coding to the global here to give at least some benefits to the user
+  }
```
boneskull: if we are careful to `create(null)`, why not define the return value the same way?

packages/webpack/src/runtime/runtimeUnlocked.js
```diff
@@ -0,0 +1,47 @@
...
+  LAVAMOAT.ENUM
+
+/**
+ * Wraps the webpack runtime with Lavamoat security features.
+ *
+ * @param {string} resourceId - The identifier of the resource.
+ * @param {any} runtimeKit - The runtime kit containing bits from the webpack
+ *   runtime.
+ * @returns {Object} - An object containing the wrapped runtime and other
+ *   related properties.
```
boneskull: these two types leave something to be desired 😉

boneskull (approved):
nothing blocking.

what's the definition of an _unlocked_ runtime?

boneskull:
needs a rebase

packages/webpack/src/plugin.js
weizman: I read this out load "do we mean or or or?"

naugtur: I chose the or with wider engine support ;)

packages/webpack/src/runtime/runtimeUnlocked.js
naugtur: I think this is also relevant for TS

packages/webpack/src/runtime/runtimeUnlocked.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+const lavamoatRuntimeWrapper = (resourceId, runtimeKit) => {
+  let { module } = runtimeKit
+  const runtimeHandler = runtimeKit
```
weizman: why the assignment?

packages/webpack/src/plugin.js
```diff
@@ -464,50 +464,72 @@ class LavaMoatPlugin {
                   )
                   return
                 }
-                diag.rawDebug(
-                  1,
-                  '> adding runtime (additionalChunkRuntimeRequirements)'
+                let runtimeChunks = []
+                if (
+                  chunk.name &&
+                  options.unlockedChunksUnsafe?.test(chunk.name)
```
weizman, line 571: we trust this part can't be manipulated due to RegExp hardening? Any other attack scenarios we might be forgetting here?

weizman (commented):
Looks good, not approving because I think this could use some missing tests

packages/webpack/src/plugin.js
boneskull: it _does_ say `unlockedChunksUnsafe` ... 😉

packages/webpack/src/plugin.js
boneskull: @naugtur which are you concerned about?

packages/webpack/src/plugin.js
```diff
@@ -464,50 +464,72 @@ class LavaMoatPlugin {
...
+                    {
+                      name: 'unenforceable',
+                      data: identifierLookup.unenforceableModuleIds || null,
+                      json: true,
+                    },
+                    { name: 'options', data: runtimeOptions, json: true },
+                    { name: 'policy', data: policyData, json: true },
+                    {
+                      name: 'ENUM',
+                      file: path.join(__dirname, './ENUM.json'),
```
```suggestion
                      file: path.join(__dirname, 'ENUM.json'),
```

packages/webpack/src/plugin.js
```diff
@@ -464,50 +464,72 @@ class LavaMoatPlugin {
...
+                      json: true,
+                    },
+                    {
+                      name: 'endowmentsToolkit',
+                      shimRequire: 'lavamoat-core/src/endowmentsToolkit.js',
+                    },
+                    {
+                      name: 'runtime',
+                      file: path.join(__dirname, './runtime/runtime.js'),
```
```suggestion
                      file: path.join(__dirname, 'runtime', 'runtime.js'),
```

packages/webpack/src/plugin.js
```diff
@@ -464,50 +464,72 @@ class LavaMoatPlugin {
...
+                    { name: 'options', data: runtimeOptions, json: true },
+                    { name: 'policy', data: policyData, json: true },
+                    {
+                      name: 'ENUM',
+                      file: path.join(__dirname, './ENUM.json'),
+                      json: true,
+                    },
+                    {
+                      name: 'endowmentsToolkit',
+                      shimRequire: 'lavamoat-core/src/endowmentsToolkit.js',
```
boneskull, line 643: I'm guessing this is found via Node.js module resolution?  if so, it may be better to "fail fast" and `require.resolve` it.

packages/webpack/src/runtime/runtimeUnlocked.js
boneskull: the triple-slash directive is—assuming `lavamoat.d.ts` declares a global `LAVAMOAT`. otherwise you'd need another directive pulling in whatever declares it, as well...

packages/webpack/src/runtime/runtimeUnlocked.js
```diff
@@ -0,0 +1,47 @@
...
+
+  // allow setting, but ignore value for /* module decorator */ module = __webpack_require__.nmd(module)
+  defineProperty(runtimeHandler, 'module', {
+    get: () => module,
+    set: () => {},
```
boneskull, line 38: is the intent to provide a no-op, or is it to provide a read-only prop?

packages/webpack/src/runtime/runtimeUnlocked.js
```diff
@@ -0,0 +1,47 @@
+/// <reference path="./lavamoat.d.ts" />
+/* global LAVAMOAT */
+
+const { create, freeze, defineProperty } = Object
+const warn = typeof console === 'object' ? console.warn : () => {}
+warn('LavaMoat: using unlocked runtime')
+
+const { NAME_globalThis, NAME_scopeTerminator, NAME_runtimeHandler } =
```
boneskull, line 8: fwiw, this naming convention is an abomination 😈

packages/webpack/src/runtime/runtimeUnlocked.js
```diff
@@ -0,0 +1,47 @@
+/// <reference path="./lavamoat.d.ts" />
+/* global LAVAMOAT */
+
+const { create, freeze, defineProperty } = Object
+const warn = typeof console === 'object' ? console.warn : () => {}
```
boneskull, line 5: what are we targeting that does not define `console`?

packages/webpack/src/plugin.js
```diff
@@ -590,6 +612,8 @@ module.exports = LavaMoatPlugin
  * @property {boolean} [HtmlWebpackPluginInterop] - Add a script tag to the html
  *   output for lockdown.js if HtmlWebpackPlugin is in use
  * @property {RegExp} [inlineLockdown] - Prefix the matching files with lockdown
+ * @property {RegExp} [unlockedChunksUnsafe] - Give matching chunks an unsafe
```
boneskull: re: usability. is use of `RegExp` typical for matching filepaths in webpack plugins? if so, then this is fine, but otherwise globs make the easy cases easier.

packages/webpack/src/plugin.js
```diff
@@ -464,50 +464,72 @@ class LavaMoatPlugin {
...
+                      name: 'ENUM',
+                      file: path.join(__dirname, './ENUM.json'),
+                      json: true,
+                    },
+                    {
+                      name: 'runtime',
+                      file: path.join(
+                        __dirname,
+                        './runtime/runtimeUnlocked.js'
+                      ),
```
boneskull: alternatively:

```suggestion
                      file: require.resolve('./runtime/runtimeUnlocked.js'),
```

boneskull (approved):
questions and non-blocking suggestions only

packages/webpack/src/plugin.js
naugtur: yup, but it should only be unsafe to the selected chunk not the rest of the app. 

the attack would need to happen at build time, should not be possible to craft a package that messes with it without being part of the dependencies for webpack and its plugins/loaders etc.

And that's a concern of `@lavamoat/node` eventually. 

Am I missing anything? @weizman

packages/webpack/src/plugin.js
naugtur: When I refactor plugin.js into a few files, I intend to make this slightly more declarative maybe, but I don't see much value in more indirection in this as is.

packages/webpack/src/plugin.js
naugtur: Didn't mean anything specific, Just saying I'm defaulting to `||` when coding. Should I not?

packages/webpack/src/plugin.js
naugtur: 1. It's quite common
2. AFAIR, we don't have other options that are flexible enough other than forcing the user to resolve a path to a file that's yet to be created

We could revisit this decision, I don't have the mental context in which I made this decision.

packages/webpack/src/runtime/runtimeUnlocked.js
naugtur: I liked the co-location of these two comments.

packages/webpack/src/runtime/runtimeUnlocked.js
naugtur: hostile environments of someone else's websites

packages/webpack/src/runtime/runtimeUnlocked.js
naugtur: PRs welcome :P

packages/webpack/src/runtime/runtimeUnlocked.js
naugtur: yes, I admit this is where I called "diminishing returns"

packages/webpack/src/runtime/runtimeUnlocked.js
naugtur: the intent is to avoid throwing errors on attempts of breaking module.

packages/webpack/src/runtime/runtimeUnlocked.js
naugtur: um. Not sure, might be an indication that I initially intended to make a copy but backed away.

packages/webpack/src/plugin.js
boneskull: (opinion) prefer no coercion over coercion.  use `||` if you want coercion; otherwise use `??`

naugtur: I do prefer coercion, but
I looked more carefully at the types and that null fallback doesn't make sense here.

packages/webpack/src/plugin.js
naugtur: shimRequire is passed straight into require.resolve in `assembleRuntime`

packages/webpack/src/runtime/runtimeUnlocked.js
naugtur: expanded the type a bit

packages/webpack/src/runtime/runtimeUnlocked.js
naugtur: Actually, I think you spotted a bug.

naugtur: fixed by making a shallow copy

packages/webpack/src/plugin.js
naugtur: I remember - a popular way to set up webpack is so that it adds a checksum or a cachebuster to the end of the script filename in dist. that's why we wanted regex

PR #1377, opened 2024-10-01
refactor(allow-scripts): break apart index.js into submodules

legobeat:
This breaks apart the monolothic `index.js` of allow-scripts into a couple of internal submodules. Common types that would need to be shared across them are declared and exported from package global types. Existing exports in `index.js` are retained. No functional changes.

This allows more selective import and bundling of allow-scripts functionality without needing to parse the whole thing.

Also adds a return type annotation for `getOptionsForBin`.

Files:
- `types/common.ts`: Common and exposed types broken out into TS definition
- `config.js`: functions related to reading and updating package configuration
  - contains code broken out from `setup.js` 
- `report.js`: functions printing results to stdout
- `runAllowedPackages.js`: The runAllowedPackages function


packages/allow-scripts/src/setup.js
legobeat, line 1: Note: I kind of wanted to break this out to a separate `config.js` but seemed more appropriate to reuse existing `setup.js`, contents of which has overlapping scope. As current contents are exported as `setup`, not changing the name is more consistent and without having to change the public API.

packages/allow-scripts/src/setup.js
```diff
@@ -186,13 +164,395 @@ function editPackageJson() {
...
+    if (binsFromDirectDependencies.length === 1 && !bannedBins.has(bin)) {
+      // there's no conflicts, seems fairly obvious choice to put it up
+      policy[bin] = binsFromDirectDependencies[0].fullLinkPath
+    }
+  }
+  return policy
+}
+
+/**
+ * @typedef SavePackageConfigurationsOpts
```
legobeat, line 443: Note: Previously the inlined typedefs were a bit all over the place. Aside from the `import` of allow-scripts library types in the top, they are now declared just prior to their first usage.

packages/allow-scripts/src/setup.js
```diff
@@ -186,13 +164,395 @@ function editPackageJson() {
...
 /**
  * @typedef WriteRcFileContentOpts
  * @property {string} file
  * @property {string} entry
  */
 
 /**
- * @typedef AreBinsBlockedOpts
- * @property {boolean} [noMemoization] Turn off memoization, make a fresh lookup
+ * @param {WriteRcFileContentOpts} param0
```
legobeat, line 410: Note: Previously the inlined typedefs were a bit all over the place. Aside from the `import` of allow-scripts library types in the top, they are now declared just prior to their first usage.

packages/allow-scripts/src/setup.js
```diff
@@ -186,13 +164,395 @@ function editPackageJson() {
...
+}
+
 /**
  * @typedef WriteRcFileContentOpts
  * @property {string} file
  * @property {string} entry
  */
 
 /**
- * @typedef AreBinsBlockedOpts
```
legobeat, line 196: Note: Previously the inlined typedefs were a bit all over the place. Aside from the `import` of allow-scripts library types in the top, they are now declared just prior to their first usage.

packages/allow-scripts/src/index.js
```diff
@@ -1,671 +1,30 @@
...
-/**
- * @param {SavePackageConfigurationsOpts} param0
- * @returns {Promise<void>}
- */
-async function savePackageConfigurations({
-  rootDir,
-  conf: {
-    packageJson,
-    configs: { lifecycle, bin },
+  setup: {
```
legobeat, line 19: Note: This keeps the `setup` export identical to before, instead of exporting some functions twice. If we want to also change the external API, we can do so in a follow-up.

packages/allow-scripts/src/setup.js
legobeat: Went ahead and broke out `config.js` from `setup.js`, while maintaining API and current `setup` exports.

packages/allow-scripts/src/config.js
```diff
@@ -0,0 +1,358 @@
...
+ * @property {Record<string, any>} [allowScripts]
+ * @property {Record<string, any>} [allowBins]
+ * @property {Record<string, any>} [allowConfig]
+ * @property {Record<string, any>} [allowedPatterns]
+ * @property {Record<string, any>} [disallowedPatterns]
+ * @property {Record<string, any>} [missingPolicies]
+ * @property {Record<string, any>} [excessPolicies]
+ */
+/**
+ * @typedef {import('type-fest').PackageJson & { lavamoat: PkgLavamoatConfig }} LavamoatPackageJson
```
boneskull, line 25: please use the new [@import tag](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#the-jsdoc-@import-tag)--it's way nicer

packages/allow-scripts/src/runAllowedPackages.js
boneskull, line 1: aside:

macOS and Windows default to case-insensitive filesystems. if you ever happen to rename this to `runallowedpackages.js`, version control systems (historically; it seems to be a rare occurrence and I'm unsure if it's been somehow addressed in Git) will face corruption on those filesystems; the reflog claims there was a rename operation, but the filesystem is reporting that no rename took place.

for that reason I _always_ use lowercase kebab-casing for filenames.  IMO, we should adopt the convention, but it is unpopular in certain sections of the ecosystem.

packages/allow-scripts/src/config.js
```diff
@@ -0,0 +1,358 @@
...
+}
+
+/**
+ * @typedef GetOptionsForBinOpts
+ * @property {string} rootDir
+ * @property {string} name
+ */
+
+/**
+ * @param {GetOptionsForBinOpts} param0
```
boneskull, line 166:
```suggestion
 * @param {GetOptionsForBinOpts} params
```

packages/allow-scripts/src/config.js
```diff
@@ -0,0 +1,358 @@
...
+  return {
+    packageJson,
+    configs,
+    somePoliciesAreMissing,
+    canonicalNamesByPath,
+  }
+}
+
+/**
+ * @typedef GetOptionsForBinOpts
```
boneskull, line 160: I consider options to be _optional_, but these are not. Can we call this `GetOptionsForBinParams` instead? I'd like to adopt a naming convention around stuff like this ...

... or avoid it entirely.  FWIW, if I was writing this function, the signature of `getOptionsForBin` would be `(rootDir: string, name: string) => Promise<BinInfo[] | undefined>`, but we're not refactoring here. 😄

packages/allow-scripts/src/config.js
```diff
@@ -0,0 +1,358 @@
...
+  return binCandidates.get(name)
+}
+
+/**
+ * @typedef SetDefaultConfigurationOpts
+ * @property {string} rootDir
+ */
+
+/**
+ * @param {SetDefaultConfigurationOpts} param0
```
boneskull, line 185:
```suggestion
 * @param {SetDefaultConfigurationOpts} params
```

packages/allow-scripts/src/config.js
```diff
@@ -0,0 +1,358 @@
...
+    },
+  } = await loadAllPackageConfigurations({ rootDir })
+
+  return binCandidates.get(name)
+}
+
+/**
+ * @typedef SetDefaultConfigurationOpts
+ * @property {string} rootDir
+ */
```
boneskull, line 182: same idea as re: `GetOptionsForBinOpts`

packages/allow-scripts/src/report.js
```diff
@@ -0,0 +1,120 @@
...
+      '\n# packages with bin scripts that no longer need configuration (package or script removed or script path outdated)'
+    )
+    excessPolicies.forEach((bin) => {
+      console.log(`- ${bin}`)
+    })
+  }
+}
+
+/**
+ * @param {ScriptsConfig} param0
```
boneskull, line 74:
```suggestion
 * @param {ScriptsConfig} config
```

packages/allow-scripts/src/runAllowedPackages.js
```diff
@@ -0,0 +1,152 @@
...
+const { printMissingPoliciesIfAny } = require('./report.js')
+const { FEATURE } = require('./toggles.js')
+
+/**
+ * @typedef RunAllowedPackagesOpts
+ * @property {string} rootDir
+ */
+
+/**
+ * @param {RunAllowedPackagesOpts} param0
```
boneskull, line 19: I hate `param0` 😄

boneskull (dismissed):
Just some _optional_ (wink) suggestions.

And a general suggestion: I like to add a docstring at the top of each file describing its purpose, e.g.: 

```js
/**
 * This module provides x, y and z
 * 
 * @module
 */
```

The `@module` (or `@packageDocumentation`) tag flags the docstring as referring to the file itself; if this was not present, the docstring would refer to the _next statement_ (usually an `import` or `require`).

This is a good time to add these docstrings, as we're creating new modules.

packages/allow-scripts/src/config.js
legobeat: Fair point and I'd like to do some further cleaning up of some of these typedefs.

This is copied as verbatim, from the existing code being moved, though. Improving on and changing the types outside of facilitating the refactor is something I'd like to do in a follow-up in a separate PR to keep the diff clean.

legobeat:
@boneskull I just added minimal `@module` docstrings to the new modules but thinking we can address further improvements of the existing typedefs in follow-up to keep the diff clean. This way it can also be addressed separately from the unrelated changes here.

packages/allow-scripts/src/config.js
legobeat: I'm just copying over your code from f10ac687 > 82f9d3aa - can we address that separately from the structural refactoring here?

packages/allow-scripts/src/report.js
legobeat: Lets do actual cleanup separately from shuffling things around?

packages/allow-scripts/src/config.js
legobeat: Lets do actual cleanup separately from shuffling things around?

packages/allow-scripts/src/config.js
legobeat: Lets do actual cleanup and renaming separately from shuffling things around?

packages/allow-scripts/src/config.js
legobeat: Lets do actual cleanup separately from shuffling things around?

legobeat:
- Suggestions addressed in #1381

PR #1379, opened 2024-10-02
fix(aa): only require used sync submodule from resolve package

legobeat:
- Only importing `sync` by path allows bundled usage (like allow-scripts yarn plugin) to shake ~20KiB due to dropped branches and deps
- Fill in missing type from `@types.resolve`

### DIFF
--- a/packages/aa/src/index.js
+++ b/packages/aa/src/index.js
@@ -2,7 +2,7 @@
 
 const { readFileSync, realpathSync, lstatSync } = require('node:fs')
 const path = require('node:path')
-const nodeResolve = require('resolve')
+const nodeResolve = require('resolve/sync')
 
 module.exports = {
   loadCanonicalNameMap,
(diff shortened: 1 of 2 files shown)


packages/aa/src/index.js
```diff
@@ -2,7 +2,7 @@
 
 const { readFileSync, realpathSync, lstatSync } = require('node:fs')
 const path = require('node:path')
-const nodeResolve = require('resolve')
+const nodeResolve = require('resolve/sync')
```
boneskull, line 5: it bears repeating that `resolve` does not understand the `exports` field in `package.json` and will likely cause problems soon.  I understand we use it to _avoid_ reading `package.json`, but there's really no way to avoid it now. 

¯\\_(ツ)_/¯

boneskull (approved):

packages/aa/src/index.js
legobeat: Note that this doesn't rely on exports in `package.json` (in fact, [`resolve`](https://www.npmjs.com/package/resolve) has none) but requires the module by subpath, which should not be problematic in the same way? Or foreseeing problems here too?

PR #1381, opened 2024-10-02
chore(allow-scripts): improve type definitions

legobeat:
- chore(allow-scripts): unify `PkgInfo` and `PkgLavamoatConfig` types
  - Use `type-fest`.`JsonObject` to type objects serialized to/from json 
- chore(allow-scripts): break out @typedefs to package-global types
- chore(allow-scripts): Rename in interface names `s/Opts/Params/` as none of these are actually optional
  - For aesthetics, `s/param0/params/`  

I personally find the improved readability and familiar syntax of TypeScript definitions vs JsDoc to be worth losing colocating definitions and usage. Trivial non-named one-off definitions are still kept in-line, though.

### Related
- Follow-up to #1377

### DIFF
--- a/packages/allow-scripts/src/types/common.ts
+++ b/packages/allow-scripts/src/types/common.ts
@@ -58,29 +58,29 @@ declare global {
   }
 
   // config.js
-  interface SetDefaultConfigurationParams {
+  export interface SetDefaultConfigurationParams {
     rootDir: string;
     lifecycleEvents?: string[];
   }
 
-  interface SavePackageConfigurationsParams {
+  export interface SavePackageConfigurationsParams {
     rootDir: string;
     conf: PkgConfs;
   }
 
-  interface GetOptionsForBinParams {
+  export interface GetOptionsForBinParams {
     rootDir: string;
     name: string;
     lifecycleEvents?: string[];
   }
 
   // setup.js
-  interface AreBinsBlockedParams {
+  export interface AreBinsBlockedParams {
     /** Turn off memoization, make a fresh lookup */
     noMemoization?: boolean;
   }
 
-  interface PkgConfs {
+  export interface PkgConfs {
     packageJson: LavamoatPackageJson;
     configs: {
       lifecycle: ScriptsConfig;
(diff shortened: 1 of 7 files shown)


packages/allow-scripts/src/types/common.ts
```diff
@@ -1,15 +1,29 @@
+import type { JsonObject, PackageJson } from 'type-fest';
+
 declare global {
+  export interface PkgLavamoatConfig {
+    ['allowBins']?: JsonObject;
```
legobeat: `PackageJson` is already using `JsonObject` so using the same pattern here makes the unified type `LavamoatPackageJson` more consistently typed.

packages/allow-scripts/src/types/common.ts
```diff
@@ -1,15 +1,29 @@
...
+    ['allowConfig']?: JsonObject;
+    ['allowScripts']?: JsonObject;
+    ['allowedPatterns']?: JsonObject;
+    ['disallowedPatterns']?: JsonObject;
+    ['excessPolicies']?: JsonObject;
+    ['missingPolicies']?: JsonObject;
```
boneskull:
```suggestion
    allowBins?: JsonObject;
    allowConfig?: JsonObject;
    allowScripts?: JsonObject;
    allowedPatterns?: JsonObject;
    disallowedPatterns?: JsonObject;
    excessPolicies?: JsonObject;
    missingPolicies?: JsonObject;
```

unsure if there was a reason for writing it that way?

packages/allow-scripts/src/types/common.ts
```diff
@@ -1,15 +1,29 @@
+import type { JsonObject, PackageJson } from 'type-fest';
```
boneskull, line 1: This wants `type-fest` to now be a prod dependency.

packages/allow-scripts/src/types/common.ts
```diff
@@ -42,4 +56,60 @@ declare global {
      excessPolicies: string[];
      somePoliciesAreMissing: boolean;
   }
+
```
boneskull, line 59: if any of the below are user-accessible via a function signature (referenced directly or otherwise), they should be exported

packages/allow-scripts/src/types/common.ts
boneskull: also `JSONObject` means "any JSON object", which could be something like  `Record<string, Record<string, number[]|null>>`. Aren't these all just `Record<string, boolean>`?  That's how I see `allowScripts` used, anyhow.

legobeat: > unsure if there was a reason for writing it that way?

Keeping at least the key types consistent with the previous interface but yeah, I also prefer this form. Updated.

> also JSONObject means "any JSON object", which could be something like Record<string, Record<string, number[]|null>>. Aren't these all just Record<string, boolean>? That's how I see allowScripts used, anyhow.

Yes, but unless I am missing something there is no input validation at runtime beyond that it parses as JSON? Therefore it's not safe to assume values are `boolean` and `JSONObject` is more correct type here, at least until runtime validation is done prior to casting.

packages/allow-scripts/src/types/common.ts
legobeat: What is the assumption behind that importing it here vs in the jsdoc changes is status?

boneskull: for posterity:

Exporting a type containing `JsonObject` means that unless `type-fest` is a production type, downstream users won't know what `JsonObject` is.

boneskull (approved):

boneskull:
I've added `type-fest` to the prod deps.  This LGTM

PR #1382, opened 2024-10-03
fix(allow-scripts): consider wider range of lifecycle scripts

legobeat:
There may be other lifecycle scripts relevant than the currently recognized (`preinstall`, `install`, `postinstall`).

New ones have historically been added somewhat inconsistently and arbitrarily to package managers, so this updates allow-scripts to recognize with a wider range of lifecycle events out of the box.

I think this may be considered a breaking change?

### Related
- #1218 

#### Blocked by
- [x] #1383

### DIFF
--- a/packages/allow-scripts/src/config.js
+++ b/packages/allow-scripts/src/config.js
@@ -13,7 +13,14 @@ const bannedBins = new Set(['corepack', 'node', 'npm', 'pnpm', 'yarn'])
 /**
  * Scripts with names other than these are ignored and unenforced by default.
  */
-const DEFAULT_LIFECYCLE_EVENTS = ['preinstall', 'install', 'postinstall']
+const DEFAULT_LIFECYCLE_EVENTS = [
+  'install',
+  'postinstall', 'postinstallonly', 'preinstall', 'preinstallonly',
+  'preaudit', 'preauditonly', 'postaudit', 'postauditonly',
+  'prepack', 'prepackonly', 'postpack', 'postpackonly',
+  'preprepare', 'preprepareonly', 'postprepare', 'postprepareonly',
+  'prepublish', 'prepublishonly', 'postpublish', 'postpublishonly',
+];
 
 /**
  * @typedef PkgLavamoatConfig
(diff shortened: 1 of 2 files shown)


packages/allow-scripts/src/config.js
```diff
@@ -13,7 +13,14 @@ const bannedBins = new Set(['corepack', 'node', 'npm', 'pnpm', 'yarn'])
...
  */
-const DEFAULT_LIFECYCLE_EVENTS = ['preinstall', 'install', 'postinstall']
+const DEFAULT_LIFECYCLE_EVENTS = [
+  'install',
+  'postinstall', 'postinstallonly', 'preinstall', 'preinstallonly',
+  'preaudit', 'preauditonly', 'postaudit', 'postauditonly',
+  'prepack', 'prepackonly', 'postpack', 'postpackonly',
+  'preprepare', 'preprepareonly', 'postprepare', 'postprepareonly',
+  'prepublish', 'prepublishonly', 'postpublish', 'postpublishonly',
+];
```
boneskull, line 23: Does this account for case-sensitivity? Do _package managers_ care about case-sensitivity?  I've only ever seen `prepublishOnly`, for instance.

boneskull (dismissed):
How is this a breaking change?  Is it a breaking change for _all_ supported package managers?

legobeat:
> How is this a breaking change? Is it a breaking change for all supported package managers?

Considering this scenario:
- Existing setup of `@lavamoat/allow-scripts`
- There is an existing devDependency with a required lifecycle script which was not convered previously but will be after this
- After the upgrade, the lifecycle script will be blocked until the allow-scripts configuration has been updated to allowlist it

I think this can be considered breaking but could be convinced otherwise.

packages/allow-scripts/src/config.js
legobeat: Hm, I thought I addressed that part but looks like I forgot it. Good catch!

I don't really see a valid use-case to disambiguate and would rather not assume things regarding if package managers make the distinction or not - so making this case-insensitive as well.

naugtur (changes requested):
As far as I understand lifecycle script this is now incorrect.

allow-scripts supports installation time scripts and turns off all of them, so it was never running any scripts other than `.*install` because that's what runs on a proper installation for every package. Any other scripts that run during regular install are the ones that run when you install a script from a folder instead of a repository.
I don't think we want to support that.

Correct me if I'm wrong, but this change looks to me like executing `allow-scripts` would now run all of the listed scripts if present. Which is not the correct behavior for packages installed from npm.

PR #1396, opened 2024-10-08
chore(vscode): format on save & disable insertion of final newline

boneskull:
Insertion of final newline causes certain generated files to churn if they are opened by VS Code.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,5 +6,5 @@
     "source.formatDocument": "explicit"
   },
   "prettier.useEditorConfig": true,
-  "files.insertFinalNewline": false
+  "editor.formatOnSave": true
 }


boneskull:
@legobeat would appreciate the reasoning behind this

.vscode/settings.json
```diff
@@ -6,5 +6,5 @@
     "source.formatDocument": "explicit"
   },
   "prettier.useEditorConfig": true,
-  "files.insertFinalNewline": false
```
legobeat, line 9: Title says `disable insertion of final newline` but this seems to be the opposite...?

Either way's fine, though. Since I think I may be the only current maintainer not actively using vscode on this repo:

I'll leave it to someone who does to sign off on if `formatOnSave` should be introduced as a repo-level config or if it's better to just keep it at the removal and leave that to user config.

(That's my reasoning, not "the" reasoning - don't believe I ever edited this file before)

boneskull: ahh, you know what, I think I just screwed it up.  there are directory-based configs, workspace-based configs and user-based configs, and this ended up in one when it should have been in the other.  I'll close this

PR #1395, opened 2024-10-08
export endowmentsToolkit from lavamoat-core

boneskull:
This exports `endowmentsToolkit` from `lavamoat-core`.  It's painful to need to reach into `lavamoat-core/src/endowmentsToolkikt` to get at it; likewise it can break type resolution.

It's likely `@lavamoat/webpack` will want to use this.  cc @naugtur

- Bonus: simplification tweak of `core/test/tsconfig.json`

PR #1349, opened 2024-09-11
feat: default attenuator initialization for write policy support

naugtur:
Core's endowmentsToolkit has all the capability to support writable fields, but it needs to receive a result of scanning the policy for them as a prerequisite. 
That in turn forced an initialization step for the default attenuator to provide the context.

This is a draft implementation with that step. 

Alternatively, we could let attenuators export an `initialize` method and pass policy into that method, collect the return value and give that return value to all attenuator calls as argument.

TODO:
- [ ] fix modules attenuator
- [ ] tests

PR #969, opened 2024-01-30
feat(endomoat): policy generation implementation

boneskull:
This branch is for @naugtur and I to collaborate on policy-gen changes.

## Setup

1. Checkout PR https://github.com/endojs/endo/pull/2190  in a working copy of `endojs/endo`
2. Run `yarn` to update deps, etc.
3. Run this: 
    ```sh
    npx lerna build:types && \ 
    npm link -w @endo/compartment-mapper -w ses -w @endo/env-options -w @endo/cjs-module-analyzer \
      -w @endo/static-module-record -w @endo/zip
    ```
4. Navigate to your `LavaMoat/LavaMoat` working copy.
5. Run this:
    ```sh
    npm ci && \
    npm link @endo/compartment-mapper ses @endo/env-options @endo/cjs-module-analyzer \
      @endo/static-module-record @endo/zip -w packages/endomoat && \
    npm run rebuild:types # verify linking
    ```

**The above will only work properly if all versions of `@endo/compartment-mapper` are the same**; run `npm ls @endo/compartment-mapper` to verify this.  If this isn't the case, deduping won't occur upon linking, and the type generation will fail.

> [!NOTE]
>
> If you find linking fails because `ses`' `prepare` script doesn't work, delete the `prepare` script from its `package.json`.

PR #1397, opened 2024-10-08
core: features & fixes for @lavamoat/node

boneskull:
This contains two _enhancements_ and one _fix_:

- fix(core): move ses to production dependencies

    This is due to use of types which are exported

- feat(core): add "allowDynamicRequires" flag to policy generation

    This flag suppresses warnings about dynamic requires if the flag is set.

- feat(core): second param to mergePolicy is now explicitly optional

    Simplifies calling code. Also added types.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -39,12 +39,12 @@
     "json-stable-stringify": "1.1.1",
     "lavamoat-tofu": "^8.0.2",
     "merge-deep": "3.0.3",
+    "ses": "1.9.0",
     "type-fest": "4.30.0"
   },
   "devDependencies": {
     "@types/json-stable-stringify": "1.1.0",
     "@types/merge-deep": "3.0.3",
-    "ses": "1.9.0",
     "tmp-promise": "3.0.3"
   },
   "ava": {
--- a/packages/core/src/mergePolicy.js
+++ b/packages/core/src/mergePolicy.js
@@ -1,9 +1,17 @@
+// @ts-check
+
 const {
   reduceToTopmostApiCalls,
   objToMap,
   mapToObj,
 } = require('lavamoat-tofu/src/util')
 const mergeDeep = require('merge-deep')
+
+const { values, prototype: objectPrototype } = Object
+
+/**
+ * @import {LavaMoatPolicy, LavaMoatPolicyOverrides, GlobalPolicy, BuiltinPolicy, GlobalPolicyValue} from './schema'
+ */
 
 /**
  * Merges two policies together.
@@ -38,10 +46,17 @@ function mergePolicy(policyA, policyB) {
   return policyA
 }
 
+/**
+ * @template {BuiltinPolicy | GlobalPolicy} T
+ * @param {T} packagePolicy
+ * @returns {T}
+ */
 function dedupePolicyPaths(packagePolicy) {
-  const itemMap = objToMap(packagePolicy)
+  const itemMap = /** @type {Map<string, GlobalPolicyValue>} */ (
+    objToMap(packagePolicy)
+  )
   reduceToTopmostApiCalls(itemMap)
-  return mapToObj(itemMap)
+  return /** @type {T} */ (mapToObj(itemMap))
 }
 
 module.exports = { mergePolicy }
--- a/packages/core/src/mergePolicy.js
+++ b/packages/core/src/mergePolicy.js
@@ -5,19 +5,37 @@ const {
 } = require('lavamoat-tofu/src/util')
 const mergeDeep = require('merge-deep')
 
-module.exports = { mergePolicy }
-
+/**
+ * Merges two policies together.
+ *
+ * `policyB` overwrites `policyA` where concatenation is not possible
+ *
+ * @param {LavaMoatPolicy} policyA First policy
+ * @param {LavaMoatPolicy | LavaMoatPolicyOverrides} [policyB] Second policy or
+ *   policy override
+ * @returns {LavaMoatPolicy} Merged policy or `policyA` if `policyB` not
+ *   provided
+ */
 function mergePolicy(policyA, policyB) {
-  const mergedPolicy = mergeDeep(policyA, policyB)
-  Object.values(mergedPolicy.resources).forEach((packagePolicy) => {
-    if ('globals' in packagePolicy) {
-      packagePolicy.globals = dedupePolicyPaths(packagePolicy.globals)
-    }
-    if ('builtin' in packagePolicy) {
-      packagePolicy.builtin = dedupePolicyPaths(packagePolicy.builtin)
-    }
-  })
-  return mergedPolicy
+  if (policyB) {
+    const mergedPolicy = mergeDeep(policyA, policyB)
+    values(mergedPolicy.resources ?? {}).forEach((packagePolicy) => {
+      if (
+        objectPrototype.hasOwnProperty.call(packagePolicy, 'globals') &&
+        packagePolicy.globals
+      ) {
+        packagePolicy.globals = dedupePolicyPaths(packagePolicy.globals)
+      }
+      if (
+        objectPrototype.hasOwnProperty.call(packagePolicy, 'builtin') &&
+        packagePolicy.builtin
+      ) {
+        packagePolicy.builtin = dedupePolicyPaths(packagePolicy.builtin)
+      }
+    })
+    return /** @type {LavaMoatPolicy} */ (mergedPolicy)
+  }
+  return policyA
 }
 
 function dedupePolicyPaths(packagePolicy) {
@@ -25,3 +43,5 @@ function dedupePolicyPaths(packagePolicy) {
   reduceToTopmostApiCalls(itemMap)
   return mapToObj(itemMap)
 }
+
+module.exports = { mergePolicy }
--- a/packages/core/src/mergePolicy.js
+++ b/packages/core/src/mergePolicy.js
@@ -7,7 +7,7 @@ const {
 } = require('lavamoat-tofu/src/util')
 const mergeDeep = require('merge-deep')
 
-const { values, prototype: objectPrototype } = Object
+const { values, hasOwn } = Object
 
 /**
  * @import {LavaMoatPolicy, LavaMoatPolicyOverrides, GlobalPolicy, BuiltinPolicy, GlobalPolicyValue} from './schema'
@@ -28,16 +28,10 @@ function mergePolicy(policyA, policyB) {
   if (policyB) {
     const mergedPolicy = mergeDeep(policyA, policyB)
     values(mergedPolicy.resources ?? {}).forEach((packagePolicy) => {
-      if (
-        objectPrototype.hasOwnProperty.call(packagePolicy, 'globals') &&
-        packagePolicy.globals
-      ) {
+      if (hasOwn(packagePolicy, 'globals') && packagePolicy.globals) {
         packagePolicy.globals = dedupePolicyPaths(packagePolicy.globals)
       }
-      if (
-        objectPrototype.hasOwnProperty.call(packagePolicy, 'builtin') &&
-        packagePolicy.builtin
-      ) {
+      if (hasOwn(packagePolicy, 'builtin') && packagePolicy.builtin) {
         packagePolicy.builtin = dedupePolicyPaths(packagePolicy.builtin)
       }
     })
--- a/.config/tsconfig.workspace.json
+++ b/.config/tsconfig.workspace.json
@@ -1,5 +1,5 @@
 {
-  "extends": "@tsconfig/node16/tsconfig.json",
+  "extends": "@tsconfig/node18/tsconfig.json",
   "compilerOptions": {
     "composite": true,
     "allowSyntheticDefaultImports": true,
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -34,11 +34,6 @@ jobs:
           image: 'dev:latest'
           run: |
             npm run setup && npm run rebuild
-      - name: lint
-        uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185  # v3
-        with:
-          image: 'dev:latest'
-          run: 'npm run lint'
       - name: test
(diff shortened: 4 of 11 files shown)


boneskull:
* **#772** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/772?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#969](https://github.com/LavaMoat/LavaMoat/pull/969) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/969?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1349](https://github.com/LavaMoat/LavaMoat/pull/1349) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1349?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1397** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1397?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1397?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1442** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1442?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

boneskull:
I could split this up into 3 PRs, but getting one PR reviewed is a chore, let alone three 😄

boneskull:
Also note that this targets #1395

packages/core/src/mergePolicy.js
```diff
@@ -5,23 +5,38 @@ const {
...
+  return policyA
 }
 
 function dedupePolicyPaths(packagePolicy) {
   const itemMap = objToMap(packagePolicy)
   reduceToTopmostApiCalls(itemMap)
   return mapToObj(itemMap)
 }
+
+module.exports = { mergePolicy }
```
naugtur, line 56: (nit) forgotten is the tradition of leveraging function hoisting to put exports up top

packages/core/src/generatePolicy.js
```diff
@@ -414,7 +433,7 @@ function createModuleInspector(opts) {
       // get native modules
       native = packageToNativeModules.has(packageName)
       // skip package policy if there are no settings needed
-      if (!packages && !globals && !builtin) {
+      if (!packages && !globals && !builtin && !native) {
```
naugtur: This should have caused trouble, but I guess we were lucky that native packages always use builtins to load or rely on another package

naugtur:
I'm still not sure if we want to suppress the warnings. There's a fair chance some of the dynamic requires won't work and even if they do, you often need to add entries to policy-overrides manually to get them to work at all and be reachable next time policy generation is happening.

BTW - that's why policies are merged and taken into account in policy generation - if policy-override indicates a dependency for a package, that dependency is then also taken into account in scanning for policy so non-dynamic dependencies of it will be discovered and the developer doesn't have to manually create the policy for the entire tree starting at dynamic require. 

Have we lost any of that along the way?

packages/core/src/generatePolicy.js
```diff
@@ -203,17 +214,25 @@ function createModuleInspector(opts) {
...
+  ) {
     const { packageName } = moduleRecord
     // @ts-expect-error `ParseResult` / `AST` mismatch
     const compatWarnings = inspectSesCompat(ast)
 
     const { primordialMutations, strictModeViolations, dynamicRequires } =
       // @ts-expect-error `SesCompat` / `InspectSesCompatResult` mismatch
       /** @type {import('./schema').SesCompat} */ (compatWarnings)
+
+    if (allowDynamicRequires && dynamicRequires.length) {
```
weizman:
```suggestion
    if (dynamicRequires?.length) {
```

packages/core/src/mergePolicy.js
```diff
@@ -5,23 +5,38 @@ const {
...
+    const mergedPolicy = mergeDeep(policyA, policyB)
+    Object.values(mergedPolicy.resources).forEach((packagePolicy) => {
+      if ('globals' in packagePolicy) {
+        packagePolicy.globals = dedupePolicyPaths(packagePolicy.globals)
+      }
+      if ('builtin' in packagePolicy) {
+        packagePolicy.builtin = dedupePolicyPaths(packagePolicy.builtin)
+      }
+    })
+    return /** @type {LavaMoatPolicy} */ (mergedPolicy)
```
weizman, line 38:
```suggestion
    return /** @type {LavaMoatPolicy} */ mergedPolicy
```

packages/core/src/mergePolicy.js
```diff
@@ -5,23 +5,38 @@ const {
...
-    }
-    if ('builtin' in packagePolicy) {
-      packagePolicy.builtin = dedupePolicyPaths(packagePolicy.builtin)
-    }
-  })
-  return mergedPolicy
+  if (policyB) {
+    const mergedPolicy = mergeDeep(policyA, policyB)
+    Object.values(mergedPolicy.resources).forEach((packagePolicy) => {
+      if ('globals' in packagePolicy) {
```
weizman:
```suggestion
      if (packagePolicy.hasOwnProperty('globals')) {
```

packages/core/src/mergePolicy.js
```diff
@@ -5,23 +5,38 @@ const {
...
+        packagePolicy.globals = dedupePolicyPaths(packagePolicy.globals)
+      }
+      if ('builtin' in packagePolicy) {
```
weizman:
```suggestion
      if (packagePolicy.hasOwnProperty('builtin')) {
```

weizman (commented):
looks good, regarding `mergePolicy` function:
* (nit) I highly recommend going with `if (!policyB) return policyA` to avoid an extra indentation level and to trigger far less diff within the function so that we can later on be able to focus on actual diffs when using git-blame 
* I can tell this predates this PR, but the `in` operator isn't a good idea here, as it searches the property along the entire chain of the object, while we know it's necessarily a made up own property of this object. Theoretically, someone could prototype pollute a `packagePolicy` object's prototype that happens to not have own `globals` prop with its own malicious version of `globals` prop, and because of the `in` we'll adopt it. `hasOwnProperty` will assure such a scenario won't affect us, as well as is more explicit regarding the operation we attempt on executing.
  * (Practically, this can't happen because objects can't be polluted this way since they are hardened, but that's still not a good enough reason - it's best if our code expresses our intentions more accurately)

boneskull:
@weizman ESLint says not to use that because `hasOwnProperty` is not present on objects created via `Object.create(null)`.

Instead, I can use `Object.prototype.hasOwnProperty.call(obj, prop)`.

I use `in` because it's the TS-blessed way to perform this sort of type narrowing.  I probably use it everywhere...

packages/core/src/mergePolicy.js
boneskull: `/** @type {T} */ (x)` is how you format a type assertion in JS; your suggestion doesn't work as intended.

packages/core/src/generatePolicy.js
boneskull: That seems to change the intention.  Why do this?

boneskull:
@naugtur 

> I'm still not sure if we want to suppress the warnings. There's a fair chance some of the dynamic requires won't work and even if they do, you often need to add entries to policy-overrides manually to get them to work at all and be reachable next time policy generation is happening.

I'll rip out the suppression for now.

> BTW - that's why policies are merged and taken into account in policy generation - if policy-override indicates a dependency for a package, that dependency is then also taken into account in scanning for policy so non-dynamic dependencies of it will be discovered and the developer doesn't have to manually create the policy for the entire tree starting at dynamic require.
> Have we lost any of that along the way?

Yes, we did, but we fixed it. 😄

packages/core/src/generatePolicy.js
weizman: Misread, sorry

weizman:
While ts is generally great, it fails us here, as the in operator represents a question we don't intend to ask.
We're not asking "is this prop part of this object's prototype?", we're asking "does this prop belong to this object?".
Imo, I think this matters in a project like this, the questions we ask via code mustn't be interpreted in more than one way.

boneskull:
@naugtur Due to changes in https://github.com/LavaMoat/LavaMoat/commit/428ff061750a016ef329d16e87f9c886d0e35dbe which writes to stderr, the `globalRef - webpack code in the wild` scenario now fails in lavamoat-node under current installed version of SES. it seems.

boneskull:
Those CI-Docker tests shouldn't be linting.

packages/core/package.json
```diff
@@ -39,12 +39,12 @@
     "json-stable-stringify": "1.1.1",
     "lavamoat-tofu": "^8.0.2",
     "merge-deep": "3.0.3",
+    "ses": "1.8.0",
```
boneskull: - [x] fix this version mismatch

packages/core/src/generatePolicy.js
```diff
@@ -116,7 +120,10 @@ function createModuleInspector(opts) {
     if (!packageToNativeModules.has(packageName)) {
       packageToNativeModules.set(packageName, [])
     }
-    const packageNativeModules = packageToNativeModules.get(packageName)
+    const packageNativeModules =
+      /** @type {import('./moduleRecord').LavamoatModuleRecord[]} */ (
```
boneskull: - [x] replace these with `@import`s

packages/lavamoat-node/test/runScenarios.spec.js
```diff
@@ -13,9 +13,18 @@ test('Run scenarios', async (t) => {
...
+      )
+      continue
+    }
     t.log(`Running Node Scenario: ${scenario.name}`)
-    await runAndTestScenario(t, scenario, ({ scenario }) =>
-      runScenario({ scenario })
-    )
+    await runAndTestScenario(t, scenario, ({ scenario }) => {
+      return runScenario({ scenario })
+    })
```
boneskull: - [x] revert

packages/lavamoat-node/test/runScenarios.spec.js
```diff
@@ -13,9 +13,18 @@ test('Run scenarios', async (t) => {
...
     }
+    // SKIP FOR NOW
+    if (
+      scenario.name.startsWith('globalRef - Webpack code in the wild works')
+    ) {
+      t.log(
+        `Skipping scenario "${scenario.name}" due to expected failure; re-enable when fixed`
+      )
+      continue
+    }
```
boneskull: **UPDATE**

@naugtur 

This scenario fails in #772 for everything that _isn't_ `@lavamoat/node`.

We added a policy to it, as seen here:

https://github.com/LavaMoat/LavaMoat/blob/e056d88246e6f7218bc2b00cd6a8b9eb41b5bc4e/packages/core/test/scenarios/globalRef.js#L32-L41

Why does this cause the everything else to fail?  I'm guessing it's because the presence of `config` overrides some boilerplate policy that was expected to be set, and the defaults are not applied _deeply_.

- [x] I'll see if that's the case

packages/core/src/generatePolicy.js
boneskull: I just ripped out all changes to `packages/core/src/generatePolicy.js` since I reverted my original changes and the others had nothing to do with what I need here

packages/lavamoat-node/test/runScenarios.spec.js
boneskull: ...no, it seems to be getting policy OK.  But I'm getting this warning:

```
LavaMoat: ignoring write attempt to read-access global "Function"
```

I tried a `console.trace()` to find where it was happening, but the stack isn't so helpful.

Reverting our changes fixes it.  Changing `Function: true` to `Function: write` causes some _other_ error. I think there must be some implicit write-access that we're blowing away. also I don't know where the hell we're writing to the global `Function`?!

naugtur:
> Instead, I can use `Object.prototype.hasOwnProperty.call(obj, prop)`.

`Object.hasOwn` is the nice way to express it and it can be destructured out of `Object` without complaining because it's already uncurried by design. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

---

Let's figure out the tests in our working session.

boneskull:
I figured out the problem.

The kernel does not correctly handle a resource policy containing `globals: {Function: true|'write'}`. If we have `Function: true`, `Function` will be replaced with an accessor which makes `Function` read-only:

https://github.com/LavaMoat/LavaMoat/blob/95ac308a213f2ad94ccac531305c6b492c677679/packages/core/src/makePrepareRealmGlobalFromConfig.js#L61-L77

Then, the following operation a few lines later will fail (and print the warning from L73):

https://github.com/LavaMoat/LavaMoat/blob/95ac308a213f2ad94ccac531305c6b492c677679/packages/core/src/makePrepareRealmGlobalFromConfig.js#L113-L123C7

If we do not provide such a policy, the code which replaces the global `Function` with an accessor is never executed, because `Function` does not appear in the arrays returned by `getTopLevelWriteAccessFromPackageConfig()` nor `getTopLevelReadAccessFromPackageConfig()`.

So in the first code block from above, if we filter `Function` out or otherwise skip the `Object.defineProperty()` call if `key === 'Function'`, it gets us out of the woods.

I don't know what the "correct" behavior is, but I think there's a bug here in the kernel.

This is not a problem in `@lavamoat/node` because we do not use   `makePrepareRealmGlobalFromConfig()`.

naugtur:
> So in the first code block from above, if we filter `Function` out or otherwise skip the `Object.defineProperty()` call if `key === 'Function'`, it gets us out of the woods.
> 
> I don't know what the "correct" behavior is, but I think there's a bug here in the kernel.

Since Function is already tamed, I think it's safe to give it a special treatment. Not sure if skipping is _the_ special treatment it needs yet. Gonna need a testcase.

boneskull:
I've reverted where I disabled the "problem" scenario and fixed the `Function` bug in core.

naugtur (approved):
not sure about this being the best place to bump to node18 in types, but it's been overdue so why not.

everything else seems fine to finally merge.

PR #1421, opened 2024-10-22
feat(browserify): configurable resolve implementation

legobeat:
- feat(browserify): allow overriding `resolve` implementation via plugin options

For tackling https://github.com/MetaMask/metamask-extension/pull/27923#issuecomment-2430491614

### DIFF
--- a/packages/browserify/src/index.js
+++ b/packages/browserify/src/index.js
@@ -10,7 +10,7 @@ const { loadCanonicalNameMap } = require('@lavamoat/aa')
 const browserResolve = require('browser-resolve')
 
 // these are the reccomended arguments for lavaMoat to work well with browserify
-const reccomendedArgs = {
+const recommendedArgs = {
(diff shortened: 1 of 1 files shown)


packages/browserify/src/index.js
```diff
@@ -177,6 +177,10 @@ function getConfigurationFromPluginOpts(pluginOpts) {
       pluginOpts.scuttleGlobalThisExceptions
   }
 
+  if (!pluginOpts.resolve) {
+    pluginOpts.resolve = require('browser-resolve')
```
naugtur, line 181: not sure why moving the require was necessary, but it shouldn't change anything.

naugtur (approved):

packages/browserify/src/index.js
legobeat: Just a minor optimization so it's not pulled in if unused.

PR #1443, opened 2024-11-19
chore(renovate): enable @types/node versioning

boneskull:
This updates the Renovate config to include [workarounds:typesNodeVersioning](https://docs.renovatebot.com/presets-workarounds/\#workaroundstypesnodeversioning) which should keep `@types/node` in sync with our supported Node.js versions.

Unfortunately I do not know if it syncs with the minimum or maximum version.


naugtur (approved):
I guess this is a good thing ;)

PR #1445, opened 2024-11-19
chore: update CODEOWNERS

boneskull:
Removed `@Lavamoat/devs` from each item since the wildcard should always work


naugtur (approved):
I won't stop you, but this looks a bit like reinforcing a bus factor of 1

PR #1448, opened 2024-11-19
fix(browserify): use @lavamoat/sourcemap-validator

boneskull:
I started investigating an `npm audit` issue and wound up creating a patch for `lodash.template`, but realized that Lego had created a fork of `sourcemap-validator` which removes the `lodash.template` dep. He never got around to actually using it--I am sure that this is what he created it for.


naugtur (approved):
sounds good

PR #1451, opened 2024-11-20
feat(webpack): avoid emiting resource assets from packages by default

naugtur:
A fix to an audit finding.

Prevent ambient assets from dependencies.
Could be used to add a new HTML file to dist and put an xss or open redirect in there.

This is not a vulnerability in the plugin itself, but it does need to be prevented by default.


packages/webpack/src/plugin.js
```diff
@@ -284,6 +284,20 @@ class LavaMoatPlugin {
...
           moduleClass === 'ContextModule'
 
+        /**
+         * Identifies an asset that webpack includes in dist by default without
+         * setting any loaders explicitly.
+         *
+         * @param {import('webpack').Module} m
+         * @returns {m is import('webpack').NormalModule}
+         */
+        const isAmbientAsset = (m) =>
```
weizman, line 294: same as i told @boneskull, i don't think `in` is the question we intend to ask here, and i think it's important we'd be explicit about the question we do intend to ask

packages/webpack/src/plugin.js
```diff
@@ -339,6 +353,37 @@ class LavaMoatPlugin {
...
+                  )
+
+                  // chunkGraph.disconnectChunkAndModule(chunk, module) // can't use this, the require statement remains and errors out
+
+                  if (module.generatorOptions) {
+                    // genertorOptions was not present in testnig, but types indicate it might be there
+                    module.generatorOptions.emit = false
+                  }
+                  if (module.generator) {
+                    module.generator = Object.create(module.generator, {
```
weizman, line 377: doesn't this have a potential for unexpected behaviour? this object comes from webpack internals, could it be that webpack relies/will rely on the original object and not on its reference via the `generator` property, thus creating discrepancy?

packages/webpack/src/runtime/runtime.js
```diff
@@ -247,8 +247,8 @@ const lavamoatRuntimeWrapper = (resourceId, runtimeKit) => {
     // TODO: print a warning for other functions on the __webpack_require__ namespace that we're not supporting.
     //   It's probably best served at build time though - with runtimeRequirements or looking at the items in webpack runtime when adding lavamoat runtime.
 
-    // The following seem harmless and are used by default: ['O', 'n', 'd', 'o', 'r', 's', 't']
-    const supportedRuntimeItems = ['O', 'n', 'd', 'o', 'r', 's', 't']
+    // The following seem harmless and are used by default: ['O', 'n', 'd', 'o', 'r', 's', 't', 'b']
+    const supportedRuntimeItems = ['O', 'n', 'd', 'o', 'r', 's', 't', 'b']
```
weizman, line 253: what's b?

packages/webpack/test/fixtures/main/main.js
```diff
@@ -2,12 +2,17 @@ import { hello as helloCommonJS } from 'commonjs-package'
...
+import 'resource-hack'
 import 'side-effects-package/styles.css'
 import { hello as helloTypeScript } from 'typescript-package'
 import { hello as helloUMD } from 'umd-package'
 import { what } from './src/hello.ts'
 import { thing } from './src/other.mjs'
 
+// This is a webpack feature where the file referenced here is included in the
+// dist as an asset. LavaMoat plugin allows this in app, but not in packages.
+const x = new URL('./src/ambientasset.html', import.meta.url)
```
weizman, line 14: How does this feature work under the hood? Is URL overwritten by webpack? How does webpack know to take these when they're invoked and do its voodoo?

packages/webpack/test/fixtures/main/main.js
```diff
@@ -24,4 +29,4 @@ function run() {
   )
 }
 
-run()
+run(x)
```
weizman, line 32: why is this diff importnat?

weizman (commented):
Kinda makes sense, could use some answers to my questions in order to make a better call

packages/webpack/src/plugin.js
boneskull: it's possible the object is frozen, but yeah, I would try to mutate it first

packages/webpack/src/plugin.js
```diff
@@ -339,6 +353,37 @@ class LavaMoatPlugin {
...
+                // Fixes the issue with assets being emitted to dist without a loader
+                // TODO: refactor to move random hardening of the build somewhere it's easier to track.
+                if (
+                  isAmbientAsset(module) &&
+                  module.resource.includes('node_modules') // FIXME: would be better to use canonicalName lookup and match with root
+                ) {
+                  // add a warning about removing the asset
+                  mainCompilationWarnings.push(
+                    new WebpackError(
+                      `LavaMoatPlugin: the following resource was being silently emitted to the dist and LavaMoat has prevented it: '${module.resource}'. If you want to add this resource, explicitly define a file-loader for it in your webpack configuration.`
```
```suggestion
                      `LavaMoatPlugin: the following resource was being silently included in the bundle LavaMoat has prevented it: '${module.resource}'. If you want to add this resource, explicitly define a file-loader for it in your webpack configuration.`
```

packages/webpack/src/plugin.js
```diff
@@ -339,6 +353,37 @@ class LavaMoatPlugin {
...
+                    )
+                  )
+
+                  // chunkGraph.disconnectChunkAndModule(chunk, module) // can't use this, the require statement remains and errors out
+
+                  if (module.generatorOptions) {
+                    // genertorOptions was not present in testnig, but types indicate it might be there
```
```suggestion
                    // generatorOptions was not present in testing, but types indicate it might be there
```

packages/webpack/src/plugin.js
```diff
@@ -339,6 +353,37 @@ class LavaMoatPlugin {
...
+                  module.resource.includes('node_modules') // FIXME: would be better to use canonicalName lookup and match with root
+                ) {
+                  // add a warning about removing the asset
+                  mainCompilationWarnings.push(
+                    new WebpackError(
+                      `LavaMoatPlugin: the following resource was being silently emitted to the dist and LavaMoat has prevented it: '${module.resource}'. If you want to add this resource, explicitly define a file-loader for it in your webpack configuration.`
+                    )
+                  )
+
+                  // chunkGraph.disconnectChunkAndModule(chunk, module) // can't use this, the require statement remains and errors out
```
```suggestion
                  // We can't use `chunkGraph.disconnectChunkAndModule` here
                  // because the require statement remains and errors out
```

packages/webpack/src/runtime/runtime.js
boneskull: @weizman you know the rest of them? 🤣

packages/webpack/test/e2e-main.spec.js
```diff
@@ -12,6 +12,7 @@ test.before(async (t) => {
 })
 
 test('webpack/main - dist shape', (t) => {
+  // Note: The fixture contains a few cases of resources that are emitted by default. This test is to ensure that the default behavior is not changed for the main app (hence the oddly named html file and the svg file from a dependency are emitted). resources from dependencies are prevented from getting emitted, so the other html resource is not showing up in dist. For the svg to be there a loader had to be added to the webpack config.
```
boneskull, line 15: I'd probably retitle the test to something based on this description.  Also the last sentence could be transposed. "A loader needs to be added to the webpack config.."

packages/webpack/test/fixtures/main/main.js
boneskull: I'd be curious to know how prevalent this sort of thing is in 3rd-party packages.

packages/webpack/test/fixtures/main/webpack.config.js
```diff
@@ -33,6 +33,7 @@ function makeConfig(lmOptions = {}) {
     output: {
       filename: '[name].js',
       path: '/dist',
+      publicPath: '', // running bundles under node errors out when webpack attempts to figure out relative paths at runtime if this is not set
```
boneskull, line 36: wtf

there used to be a `node` flag for webpack (haven't needed it in years); unsure if it's applicable in this case.

packages/webpack/test/fixtures/main/webpack.config.js
```diff
@@ -73,6 +74,18 @@ function makeConfig(lmOptions = {}) {
...
           sideEffects: true,
         },
+        {
+          // this explicitly allows svg assets to be emited from dependencies
+          test: /\.svg$/,
+          use: [
+            {
+              loader: 'file-loader',
+              options: {
+                name: '[name].[ext]',
```
boneskull, line 84: I know this is just a fixture, but what happens when the filename is `bar.baz.svg`?

boneskull (approved):
nothing blocking, but some questions/suggestions

packages/webpack/src/plugin.js
naugtur: note that the `in` questions are there for typescript to shut up about me later using loaders. What I'm asserting that's relevant to security is m.type and m.loaders being an empty array. 
The rest is to stop typescript from complaining I use loaders and resource on a Module instance because there's no better type I can give it without lying and I know for a fact these fields are there.

packages/webpack/src/plugin.js
naugtur: that's not true. 
It doesn't get included in the bundle, it's a separate file alongside the bundle, in the dist folder, that's created (in webpack terms emitted) during compilation.

packages/webpack/src/plugin.js
naugtur: The idea is to put a field on generator that guarantees emit is false even if generator itself is immutable. Extending it this way preserves any functionality (think methods) it could carry. There's some inconsistency between `generator` and `generatorOptions` being there, but actual generator options being in `generatpr` and `genertorOptions` being empty in practice. Looks like a leftover from previous version maybe. 
I arrived at this with trial and error, seems like the most compatible way to override `emit`

packages/webpack/src/runtime/runtime.js
naugtur: It's where webpack keeps its basepath to figure out how dynamic requires fit into the bundle.
Fragment from runtime:
```
__webpack_require__.b = document.baseURI || self.location.href;.
```

packages/webpack/test/fixtures/main/main.js
naugtur: Webpack extracts this at compile time (I think without running the code, just from AST) and does its best to ensure the file is included in dist for when you put it in a link or use as img src.

I've never used it, but it seems to be in use unfortunately.

packages/webpack/test/fixtures/main/main.js
naugtur: I'm preventing the URL from being unused and therefore skipped.

packages/webpack/test/fixtures/main/webpack.config.js
naugtur: node flag is orthogonal to this. when you use features that rely on figuring out paths, webpack needs this field to exist, otherwise it fails. 
Also, we're running bundles targeting the browser in node when we test.

packages/webpack/test/fixtures/main/webpack.config.js
naugtur: this part is giving a permission for the file to be emitted. nothing bad can come out of it not covering all cases.
But if you mean the pattern [name].[ext] - it tells webpack how to shape teh filename in dist, which ends up as something like `d5234bcf8c67c6.svg`

packages/webpack/src/runtime/runtime.js
boneskull: imo this warrants a comment or something

naugtur: Following logical conclusions, we'd need to document the entire set of things we consider harmless. TBH I'd still prefer reading the source code of these in a bundle that webpack produces...

Would https://github.com/webpack/webpack/blob/main/lib/RuntimeGlobals.js be helpful enough as a reference?
The implementations are classes generating strings and harder to read than the bundle output, but they live here: https://github.com/webpack/webpack/tree/main/lib/runtime/

The vast majority of these we don't support are related to HMR and federation, which we can't really reliably protect on our own.

boneskull: more concerned about wtf theses magic letters mean. 

- [ ] a comment permalinking to whatever appropriate source(s) for reference

packages/webpack/src/plugin.js
boneskull:
```suggestion
                      `LavaMoatPlugin: the following resource was being silently emitted to the dist directory and LavaMoat has prevented it: '${module.resource}'. If you want to add this resource, explicitly define a file-loader for it in your webpack configuration.`
```

packages/webpack/src/plugin.js
weizman: that is a frowny face sticker to ts @boneskull

packages/webpack/src/plugin.js
weizman: Not sure i follow, does this answer my question?
I'm worried the way webpack internals make use of the object referred by `module.generator` won't be compatible with this modification.
What if they store another reference to it and sometimes use it instead? Not saying this is necessarily a problem, but rather trying to learn to what extent it might be, since I can see a scenario where sometime in the future such an internal change would break our software and it would be hard telling why.

weizman (approved):
lg

PR #1454, opened 2024-11-25
fix(webpack): prevent toString manipulation on a specifier

naugtur:


leotm (commented):
looks like CI [failing](https://github.com/LavaMoat/LavaMoat/actions/runs/12010738331/job/33478228259?pr=1454) on node 20 `npm run test` in `aa`
`project 4 - workspace symlink [Rejected promise returned by test]`

```console
ERROR in ./node_modules/commonjs-quirks/invalid.js 1:5
Module parse failed: Unexpected token (1:5)
File was processed with these loaders:
 * ../../../src/excludeLoader.js
You may need an additional loader to handle the result of these loaders.
> this is !not! valid js
 @ ./excludes.js 1:0-35
```

giving it a re-run

packages/webpack/test/fixtures/main/node_modules/prototype-poisoning-package/tostring.js
```diff
@@ -0,0 +1,13 @@
+// poisoning the specifier with a toString
+
+let count = 0;
+require('path') // this is here for policy generator to find
+const stolenFs = __webpack_require__({toString: () => {
+  count++;
+  // console.trace("toString called",count)
```
leotm: is this worth keeping for debugging, or forgot to delete

```suggestion
```

packages/webpack/test/fixtures/main/policy-dependencies/policy.json
```diff
@@ -0,0 +1,12 @@
+{
+  "resources": {
+    "prototype-poisoning-package": {
+      "builtin": {
+        "path": true
+      },
+      "globals": {
+        "console.trace": true
```
leotm, line 8:
```suggestion
        "__webpack_require__": true
```

packages/webpack/test/e2e-policy-dependencies.spec.js
```diff
@@ -0,0 +1,40 @@
+const test = require('ava')
+const path = require('node:path')
+// eslint-disable-next-line ava/no-import-test-files
+const { scaffold, runScriptWithSES } = require('./scaffold.js')
+const { makeConfig } = require('./fixtures/main/webpack.config.js')
+const { isBuiltin } = require('node:module')
+
+test.before(async (t) => {
+  const webpackConfigDefault = makeConfig({
+    generatePolicy: false,
```
leotm, line 10: (nb: switch to `true` locally to check `policy.json` change)

leotm (dismissed):
CI passing now looks like was flakey test

added couple minor comments

packages/webpack/test/fixtures/main/policy-dependencies/policy.json
naugtur: nope, that's the point - it's explicitly not included in the policy and still works. That's inevitable (and we don't put `require` in policy as a global either)

packages/webpack/test/fixtures/main/node_modules/prototype-poisoning-package/tostring.js
leotm: reworded the GH suggestion commit to fix CI (commitlint)

leotm (approved):

packages/webpack/test/fixtures/main/node_modules/prototype-poisoning-package/tostring.js
```diff
@@ -0,0 +1,12 @@
+// poisoning the specifier with a toString
+
+let count = 0;
+require('path') // this is here for policy generator to find
+const stolenFs = __webpack_require__({toString: () => {
+  count++;
+  if (count < 3) {
```
leotm, line 7: why `3` btw?

naugtur: first 2 times that field is being inspected we want to pretend we're importing `path` so that it ends up validating against the policy and then we switch to something more powerful.

weizman (approved):
lgtm

PR #1467, opened 2024-12-04
feat(core): policy sorting cli for cleaner sort order migration

naugtur:
To help reorder the policy independently of any reviewable changes, a cli tool can be used when introducing the new lavamoat version to have reordering and newly updated policy in two separate commits.


weizman (approved):
lgtm

PR #1469, opened 2024-12-05
fix: make policy ordering consistntly manifest itself in json files produced

naugtur:
json-stable-stringify was undoing our sorting in real usecases but not in tests, at least not in a way that would be noticable.
Sorting and tests fixed.


packages/aa/src/index.js
```diff
@@ -268,6 +268,7 @@ function getPackageNameForModulePath(canonicalNameMap, modulePath) {
   const relativeToPackageDir = path.relative(packageDir, modulePath)
   // files should never be associated with a package directory across a package boundary (as tested via the presense of "node_modules" in the path)
   if (relativeToPackageDir.includes('node_modules')) {
+    console.log({ relativeToPackageDir, packageDir, modulePath })
```
weizman: is this here on purpose?

packages/core/src/stringifyPolicy.js
```diff
@@ -0,0 +1,21 @@
...
+
+function diffFriendlyPolicyKeysComparator({ key: a }, { key: b }) {
+  if (a === b) return 0
+  if (a.includes('>')) {
+    a = a.split('>').reverse().join('<')
+  }
+  if (b.includes('>')) {
+    b = b.split('>').reverse().join('<')
+  }
+  return a < b ? -1 : 1
```
naugtur, line 20: inlining the reversing and using `<` instead of localeCompare for performance. No noticable performance gain from avoiding the split.reverse.join (I tested)
Also, localeCompare could work differently on different machines, which we also don't want.

packages/aa/src/index.js
naugtur: :hand_over_mouth:

packages/core/src/stringifyPolicy.js
weizman: maybe lose the `if`s? whether `>` is included or not gives the same end result

naugtur: on 1000 iterations of sorting metamask policy I get from 7s to 5s with the optimizations and the ifs are the main source of speedup a lot of sorting happens on globals also and they don't have arrows

packages/core/test/stringifyPolicy.spec.js
```diff
@@ -0,0 +1,74 @@
+/* eslint-disable no-undef, @typescript-eslint/no-unused-vars, no-unused-expressions, no-extend-native */
```
weizman, line 1: is there a fundamental change to this spec file from its former version at all, or was it just renamed basically?

packages/lavamoat-node/test/index.spec.js
```diff
@@ -14,14 +15,26 @@ test('parseForPolicy - resolutions', async (t) => {
...
   const policy1 = await parseForPolicy({ entryId, projectRoot })
+
+  // snapshot to prevent regression in policy sorting or output
+  t.snapshot(
+    readFileSync(path.join(projectRoot, 'lavamoat/node/policy.json'), 'utf8')
+  )
+
   t.deepEqual(policy1, {
     resources: {
+      'a>0zero': {
```
weizman, line 26: any way to do `a,b,c` instead of `0zero,a,b`? would make far more sense at first glance

weizman (approved):
a couple of nits, lgtm

packages/core/test/stringifyPolicy.spec.js
naugtur: only the way we assert at the end has changed

packages/lavamoat-node/test/index.spec.js
naugtur: I was going for a minimal diff. could be done, but IMHO not worth it.

package-lock.json
```diff
@@ -19190,7 +19190,6 @@
         "concat-stream": "2.0.0",
         "convert-source-map": "2.0.0",
         "duplexify": "4.1.3",
-        "json-stable-stringify": "1.1.1",
```
leotm, line 19193: at first wondered why `packages/browserify/examples/02-scuttler-js/package-lock.json` still has this

(turns out the examples on browserify16, when bumped to browserify17 it'll be gone

packages/browserify/src/index.js
```diff
@@ -232,7 +231,7 @@ function writeAutoPolicy(policy, configuration) {
     })
     fs.writeFileSync(
       configuration.policyPaths.debug,
-      jsonStringify(policy, { space: 2 })
+      jsonStringifySortedPolicy(policy)
```
leotm, line 234: why we no longer specifying space 2? or is it the `lavamoat-core` default

leotm (approved):

packages/browserify/src/index.js
weizman: I think it now happens within this new sink function

naugtur: It's the json-stable-stringify settings, which I'm providing just once, where it's now used

PR #1471, opened 2024-12-05
fix: 2 bugs manifested in walking content from policy-overrides: atte…

naugtur:
…mpt to go down `$root$` and failure to load content from the main of the package

### DIFF
--- a/packages/core/src/generatePolicy.js
+++ b/packages/core/src/generatePolicy.js
@@ -74,6 +74,9 @@ function createModuleInspector(opts) {
     moduleRecord,
     { isBuiltin, includeDebugInfo = false }
   ) {
+    if (moduleRecord === undefined) {
+      return
+    }
     const { packageName, specifier, type } = moduleRecord
     // record the module
     moduleIdToModuleRecord.set(specifier, moduleRecord)
--- a/packages/lavamoat-node/src/parseForPolicy.js
+++ b/packages/lavamoat-node/src/parseForPolicy.js
@@ -186,14 +186,22 @@ function makeImportHook({
     const extension = path.extname(filename)
     const packageName = getPackageNameForModulePath(canonicalNameMap, filename)
 
-    if (commonjsExtensions.includes(extension)) {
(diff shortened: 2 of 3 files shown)


packages/core/src/generatePolicy.js
```diff
@@ -74,6 +74,9 @@ function createModuleInspector(opts) {
     moduleRecord,
     { isBuiltin, includeDebugInfo = false }
   ) {
+    if (moduleRecord === undefined) {
```
naugtur, line 77: This is redundant for the bugfix in question but it's the correct thing to do here anyway.

packages/lavamoat-node/src/parseForPolicy.js
```diff
@@ -186,14 +186,22 @@ function makeImportHook({
...
     const packageName = getPackageNameForModulePath(canonicalNameMap, filename)
 
-    if (commonjsExtensions.includes(extension)) {
-      return makeJsModuleRecord(specifier, filename, packageName)
-    }
     if (extension === '.node') {
       return makeNativeModuleRecord(specifier, filename, packageName)
     }
+    try {
+      var content = await fs.readFile(filename, 'utf8')
```
naugtur, line 193: intentional use of var for nicer code.

packages/lavamoat-node/src/parseForPolicy.js
```diff
@@ -264,8 +270,18 @@ function makeImportHook({
...
         }
+        // do not pursue root as a dependency to scan
+        if (packageName === '$root$') {
+          return
+        }
         // resolve and add package main in override
         const packageRoot = getMapKeyForValue(canonicalNameMap, packageName)
+        if (!packageRoot) {
+          console.warn(
+            `could not find package root for ${packageName} as found in policy-override`
```
naugtur:
```suggestion
            `lavamoat could not find package's entry script for ${packageName} as found in policy-override`
```

weizman (approved):
Lgtm

PR #1484, opened 2025-01-10
docs: rewrite readme for @lavamoat/node

naugtur:
I went through many iterations with my main trouble being avoiding repeating myself in paragraphs.

### DIFF
--- a/packages/node/README.md
+++ b/packages/node/README.md
@@ -1,34 +1,38 @@
 # @lavamoat/node
 
-> Nether-age secure runtime for Node.js
+> Warning: You are looking at an early version of this packages. Breaking changes and bugs are to be expected.
 
-**`@lavamoat/node` is a [Hardened JavaScript][] runtime for Node.js v18.0.0+.**
+**`@lavamoat/node` is a [Hardened JavaScript][] runtime for Node.js v18.0.0+ that provides per-package security policy enfocement**
 
 `@lavamoat/node`:
 
 - Uses [`lockdown`][lockdown] to prevent tampering with the execution environment:
-  - User code cannot meddle with global objects or shared intrinsics (i.e. "the
+  - User code cannot meddle with JavaScript's global objects or shared intrinsics (i.e. "the
     stuff in the prototype chain")
-  - Neutralizes prototype pollution attacks
+  - Neutralizes most prototype pollution attacks
 - Isolates dependencies _within the same process_:
   - By default, packages do not share references to any global objects
   - Access to resources (global objects, other packages, Node.js builtins,
     native modules) is controlled by user-defined policy
-- Secures your supply chain by generating and maintaining (with your help) policies
+- Provides tools for generating and maintaining the policy without the need to manually create all of it.
 
-LavaMoat is built on top of [SES][].
+LavaMoat can protect your Node.js application, but it can _also_ protect your development environment (e.g., build scripts) — you can run your tools with [the `lavamoat` CLI](#usage).
 
 ## Supply Chain Security
 
-Over time, your application's dependencies will change. When they do, LavaMoat detects any changes to resource access. For example, if an upgraded transitive dependency now uses the `left-pad` package, this will be evident from an auto-generated policy file. You can then choose to allow or deny this access (or stop using `left-pad`).
+Over time, your application's dependencies will get updated. It's easy to update to a version of a package that introduces malicious code or a new dependency on a malicious package.
 
-In this way, LavaMoat is a little like [Socket][], which alerts you to dubious changes in a package's behavior (among other things). LavaMoat, however, surfaces _all_ such changes; it's _up to you_ to decide whether or not they're reasonable.
+`@lavamoat/node` provides a runtime that should work just as before, but with protection against most supply chain attacks.
+
+Powerful APIs and Node builtins are available globally or through imports without restriction despite only very few dependencies of your application actually need to use them. LavaMoat can stop packages trying to abuse these resources without blocking their intended use.
+
+Each package can only access resources that are explicitly allowed by the policy. If a resource is not allowed for a package, the environment will behave as if that resource doesn't even exist. This is a powerful way to prevent malicious packages from doing harm.
 
 ## Runtime Security
 
-LavaMoat also guards against deceptive packages. If a package obfuscates its intentions to the degree that LavaMoat's own policy generation didn't detect it, the existing policy will prevent naughty behavior.
+If a package obfuscates its intentions to the degree that LavaMoat's own policy generation didn't detect what it's using, the existing policy will prevent naughty behavior because LavaMoat doesn't rely on reading the code for the protections it provides. It makes sure resources not allowed by the policy for the package simply don't exist in the package's execution scope. (thanks to [Hardened JavaScript][] Compartments)
 
-LavaMoat can protect your Node.js application, but it can _also_ protect your development environment (e.g., build tooling)—you can run your tools with [the `lavamoat` CLI](#usage).
+The only time LavaMoat reads the code is when generating a policy. Policy generation is there to provide a starting point for customization and it allows every resource it detects. You can use policy-overrides to restrict resources that are allowed for a package. And you need to review the policy you generate. (More on policy review in the [LavaMoat docs](https://lavamoat.github.io/guides/policy-diff/))
 
 ## Differences from [lavamoat][]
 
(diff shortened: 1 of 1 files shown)


packages/node/README.md
```diff
@@ -1,34 +1,38 @@
 # @lavamoat/node
 
-> Nether-age secure runtime for Node.js
+> Warning: You are looking at an early version of this packages. Breaking changes and bugs are to be expected.
```
boneskull:
```suggestion
> [!WARNING]
>
> This software is _alpha_ quality. Expect bugs and breaking changes!
```

boneskull: Omit L3 and L4 if you care about how it looks on npmjs.com

packages/node/README.md
```diff
@@ -1,34 +1,38 @@
...
 
-**`@lavamoat/node` is a [Hardened JavaScript][] runtime for Node.js v18.0.0+.**
+**`@lavamoat/node` is a [Hardened JavaScript][] runtime for Node.js v18.0.0+ that provides per-package security policy enfocement**
```
boneskull:
```suggestion
**`@lavamoat/node` is a [Hardened JavaScript][] runtime** for Node.js v18.0.0+ that provides _per-package_ security policy enforcement.
```

packages/node/README.md
```diff
@@ -1,34 +1,38 @@
...
+  - User code cannot meddle with JavaScript's global objects or shared intrinsics (i.e. "the
     stuff in the prototype chain")
-  - Neutralizes prototype pollution attacks
+  - Neutralizes most prototype pollution attacks
 - Isolates dependencies _within the same process_:
   - By default, packages do not share references to any global objects
   - Access to resources (global objects, other packages, Node.js builtins,
     native modules) is controlled by user-defined policy
-- Secures your supply chain by generating and maintaining (with your help) policies
+- Provides tools for generating and maintaining the policy without the need to manually create all of it.
```
boneskull:
```suggestion
- Provides tooling to _generate_ and _maintain_ the policy
```

- I don't think we need the last bit because:
  - at this point, the new reader does not know what a policy really is
  - nor do they understand what's involved in maintaining a policy 
  - and why a policy would or would not need to be created manually
- in a list, we can omit periods
- consistent tense

packages/node/README.md
```diff
@@ -1,34 +1,38 @@
...
 
-LavaMoat is built on top of [SES][].
+LavaMoat can protect your Node.js application, but it can _also_ protect your development environment (e.g., build scripts) — you can run your tools with [the `lavamoat` CLI](#usage).
 
 ## Supply Chain Security
 
-Over time, your application's dependencies will change. When they do, LavaMoat detects any changes to resource access. For example, if an upgraded transitive dependency now uses the `left-pad` package, this will be evident from an auto-generated policy file. You can then choose to allow or deny this access (or stop using `left-pad`).
+Over time, your application's dependencies will get updated. It's easy to update to a version of a package that introduces malicious code or a new dependency on a malicious package.
```
boneskull:
```suggestion
Over time, your application's dependencies will need updating. How can you ensure a new version of a package—or newly-added dependencies—aren't doing something malicious?
```

Asking a question like this is a pretty good hook into the section, imo

packages/node/README.md
```diff
@@ -1,34 +1,38 @@
...
 
-In this way, LavaMoat is a little like [Socket][], which alerts you to dubious changes in a package's behavior (among other things). LavaMoat, however, surfaces _all_ such changes; it's _up to you_ to decide whether or not they're reasonable.
+`@lavamoat/node` provides a runtime that should work just as before, but with protection against most supply chain attacks.
+
+Powerful APIs and Node builtins are available globally or through imports without restriction despite only very few dependencies of your application actually need to use them. LavaMoat can stop packages trying to abuse these resources without blocking their intended use.
```
boneskull:
```suggestion
Nodes.js provides powerful global APIs and builtins without restriction—but most of your app's dependencies _don't need them_. LavaMoat can stop packages from abusing these resources _without_ blocking legitimate use.
```

packages/node/README.md
```diff
@@ -1,34 +1,38 @@
...
+
+Each package can only access resources that are explicitly allowed by the policy. If a resource is not allowed for a package, the environment will behave as if that resource doesn't even exist. This is a powerful way to prevent malicious packages from doing harm.
```
boneskull:
```suggestion
Each package can only access resources that are explicitly allowed by the policy. _If a package is disallowed access to a resource, the environment will behave as if that resource does not exist._ This is a powerful way to prevent malicious packages from doing harm.
```

thought this was worth emphasizing

packages/node/README.md
```diff
@@ -1,34 +1,38 @@
...
 
 ## Runtime Security
 
-LavaMoat also guards against deceptive packages. If a package obfuscates its intentions to the degree that LavaMoat's own policy generation didn't detect it, the existing policy will prevent naughty behavior.
+If a package obfuscates its intentions to the degree that LavaMoat's own policy generation didn't detect what it's using, the existing policy will prevent naughty behavior because LavaMoat doesn't rely on reading the code for the protections it provides. It makes sure resources not allowed by the policy for the package simply don't exist in the package's execution scope. (thanks to [Hardened JavaScript][] Compartments)
```
boneskull:
```suggestion
If a package obfuscates its intentions to the degree that LavaMoat's own policy generation cannot detect what resources it needs, the generated policy will _still_ prevent naughty behavior. This is because LavaMoat operates using the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege). Resources **not explicitly allowed by the policy do not exist** in the package's execution scope—thanks to the `Compartment` provided by [Hardened JavaScript][].
```

I think it may not have been obvious that a policy is of the deny-by-default nature.  feel free to add URL to list at bottom of doc instead of inline

packages/node/README.md
```diff
@@ -1,34 +1,38 @@
...
 
-LavaMoat can protect your Node.js application, but it can _also_ protect your development environment (e.g., build tooling)—you can run your tools with [the `lavamoat` CLI](#usage).
+The only time LavaMoat reads the code is when generating a policy. Policy generation is there to provide a starting point for customization and it allows every resource it detects. You can use policy-overrides to restrict resources that are allowed for a package. And you need to review the policy you generate. (More on policy review in the [LavaMoat docs](https://lavamoat.github.io/guides/policy-diff/))
```
boneskull:
```suggestion
LavaMoat parses code _only_ during policy generation—never at runtime. Policy generation provides a _starting point_ for customization by creating a policy which allows _every resource_ it detects. While reviewing the policy, you can choose to create **policy overrides** to restrict or grant access to resources.  

> [!TIP]
>
> Read more about reviewing policy files in the [LavaMoat docs](https://lavamoat.github.io/guides/policy-diff/).
```

using italics for emphasis and bold for terminology here. maybe it should be applied elsewhere

boneskull (changes requested):
the only stuff that I'd block on is of the typo/spelling/grammar variety (which, I'm sorry, is mixed in with everything else). if you take issue with any of the suggestions otherwise, just ignore it for now

packages/node/README.md
boneskull: also pls add url at bottom of doc

packages/node/README.md
naugtur: yup, would rather keep npm looking nicer than github

naugtur:
```suggestion
> Warning: This software is _alpha_ quality. Expect bugs and breaking changes!
```

packages/node/README.md
naugtur: thanks for the corrections. 
I still want to sell this as an upgrade from having to craft the whole policy manually, because I'm on a mission to leave hints of "you're responsible for the content of the policy file" everywhere.

packages/node/README.md
naugtur: "resource is not allowed for a package" sounded more natural to me than "package is disallowed access to a resource" but I''m not the native speaker here :sweat_smile:

packages/node/README.md
naugtur: very well put

packages/node/README.md
boneskull: hmm, "if policy disallows access to a resource" is what I meant to write, I think.  active voice is less painful to read

boneskull (approved):
looks good with one last comment

packages/node/README.md
naugtur: Then pls change that and merge if you don't mind. This stuff is ready to go.

PR #1489, opened 2025-01-10
chore: attempt to fix initial release of @lavamoat/node

boneskull:
RP is confused by a new package taking the same path as an old package (`packages/node`). This change sets the `bootstrapSha` to the version immediately after the SHA of the commit which moved the directory, so that RP has no knowledge of another package at this path. Maybe it will even work!

Ref: #1477


naugtur (approved):
Worth trying. might be enough to fix

naugtur:
AFAICT it worked

PR #1505, opened 2025-01-21
fix(core): block circular global object endowments

weizman:
Currently, when a package asks for "self"/"parent"/"top"/"frames"/"window" it gets access to the rootCompartment globalThis instead of the Compartment globalThis.

It turns out this behaviour was dictated intentionally (see `packages/core/src/makePrepareRealmGlobalFromConfig.js#prepareCompartmentGlobalFromConfig`), stating "if globalRef is actually an endowment, ignore".

In reality, there shouldn't be a scenario in which a package can ask for the rootCompartment globalThis as an endowment - that's practically like asking LavaMoat to execute outside of LavaMoat.

> this seems to successfully not break MM (at least locally)


packages/core/src/makePrepareRealmGlobalFromConfig.js
```diff
@@ -104,17 +104,7 @@ function makePrepareRealmGlobalFromConfig({ createFunctionWrapper }) {
...
-      if (topLevelReadAccessKeys.includes(key)) {
-        return
-      }
-      if (topLevelWriteAccessKeys.includes(key)) {
-        return
-      }
-      // set circular ref to global
-      packageCompartmentGlobal[key] = packageCompartmentGlobal
-    })
+    globalThisRefs.forEach((key) => packageCompartmentGlobal[key] = packageCompartmentGlobal)
```
naugtur: I was trying to defend the legacy approach but I'm out of arguments. This is the right thing to do and I don't see a single reason why not (other than some cases might require more work from the user to put stuff in overrides when we don't detect nested fields.
This is a breaking change (even if we could talk around it being one, I want it to be indicated as a breaking change)

packages/core/src/generatePolicy.js
```diff
@@ -283,7 +283,7 @@ function createModuleInspector(opts) {
       // browserify commonjs scope
       ignoredRefs: [...moduleRefs, ...globalObjPrototypeRefs],
       // browser global refs + browserify global
-      globalRefs: ['globalThis', 'self', 'window', 'global'],
+      globalRefs: ['window', 'self', 'global', 'globalThis', 'top', 'frames', 'parent'],
```
naugtur: `window.frames === window` always throws me off.

packages/core/test/globals.spec.js
```diff
@@ -317,6 +317,60 @@ test('globals - nested property true.false.true', async (t) => {
...
+        self: globalThis === globalThis.self,
+        warn: typeof globalThis.top.window.frames.parent.self.console.warn,
+        URL: typeof globalThis.top.window.frames.parent.self.URL
+      }
+    },
+    ...shared,
+  })
+
+  const testResult = await runScenario({ scenario: handlesAccess })
+  t.is(testResult.top, true)
```
naugtur: This could be asserted all at once with a deepEqual and would produce more useful output when failing.

packages/lavamoat-node/src/kernel.js
```diff
@@ -55,7 +55,7 @@ function createKernel({
     getRelativeModuleId,
     prepareModuleInitializerArgs,
     getExternalCompartment,
-    globalThisRefs: ['global', 'globalThis'],
+    globalThisRefs: ['window', 'self', 'global', 'globalThis', 'top', 'frames', 'parent'],
```
naugtur: (must) Do not introduce new globals in node. None of these exist or are supposed to.

packages/core/src/kernelCoreTemplate.js
```diff
@@ -59,7 +59,7 @@
     getRelativeModuleId,
     prepareModuleInitializerArgs,
     getExternalCompartment,
-    globalThisRefs = ['globalThis'],
+    globalThisRefs = ['window', 'self', 'global', 'globalThis', 'top', 'frames', 'parent'],
```
naugtur: (must) doublecheck - I think this is getting overriden whereever kernel is being used. I think this change should not be here.

packages/core/src/endowmentsToolkit.js
```diff
@@ -525,7 +525,7 @@ function endowmentsToolkit({
   function copyWrappedGlobals(
     globalRef,
     target,
-    globalThisRefs = ['globalThis']
+    globalThisRefs = ['window', 'self', 'global', 'globalThis', 'top', 'frames', 'parent']
```
naugtur: (must) do not change this default. we are supposed to pass these in if we want more than just `globalThis`

naugtur (changes requested):
only expand the browser specific globals lists

packages/core/src/makePrepareRealmGlobalFromConfig.js
weizman: "that's practically like asking LavaMoat to execute outside of LavaMoat"

packages/core/src/generatePolicy.js
weizman: :shrug:

packages/core/test/globals.spec.js
weizman: 59f5c3f3e

packages/core/src/endowmentsToolkit.js
weizman: 11f8e45

packages/core/src/kernelCoreTemplate.js
weizman: Doesn't seem like it b37a74db7

packages/lavamoat-node/src/kernel.js
weizman: 9258717

packages/core/src/kernelCoreTemplate.js
naugtur: in that case we might be missing an override. this should not be included in lavamoat-node and should be specified by either lavapack or lm-browserify

naugtur: > Doesn't seem like it [b37a74d](https://github.com/LavaMoat/LavaMoat/commit/b37a74db7cfa074ed71e249ae0fd45be5995451d)

oh, maybe that's because this test belongs in lavapack or browserify or needs to itself be parametrized?

naugtur: I know what's the best way out - make this field mandatory and throw if not set - we'll quickly learn where it requires updating.

weizman:
Introduced an important change via b95462b88eaf6a0aaaa31003432635c452e24823, which uncovered how my original implementation wasn't really working.

Now, browserify, node and core are supported both implementation-wise and on the tests level (each has its own test based on its own aliases for globalThis)

packages/core/src/kernelCoreTemplate.js
weizman: addressed https://github.com/LavaMoat/LavaMoat/pull/1505#issuecomment-2619300114

packages/core/src/generatePolicy.js
boneskull: does it matter that these properties are not always present in `globalThis`?

packages/core/src/makePrepareRealmGlobalFromConfig.js
```diff
@@ -104,17 +106,7 @@ function makePrepareRealmGlobalFromConfig({ createFunctionWrapper }) {
...
```
boneskull:
```suggestion
    globalThisRefs.forEach((key) => {
      packageCompartmentGlobal[key] = packageCompartmentGlobal
    })
```

packages/core/src/makePrepareRealmGlobalFromConfig.js
boneskull: this is a breaking change in multiple packages, yes?

packages/lavapack/src/runtime-template.js
```diff
@@ -31,7 +31,7 @@
     loadModuleData,
     getRelativeModuleId,
     prepareModuleInitializerArgs,
-    globalThisRefs: ['window', 'self', 'global', 'globalThis'],
+    globalThisRefs: ['window', 'self', 'global', 'globalThis', 'top', 'frames', 'parent'],
```
boneskull, line 34: is it possible to stuff this list somewhere where we don't have to copy/paste it to a bunch of places?

packages/lavamoat-node/test/globals.spec.js
```diff
@@ -19,6 +19,47 @@ test('globals - has only the expected global circular refs', async (t) => {
   scenario.checkResult(t, testResult.sort(), scenario)
 })
 
+test('globals - circular refs taming', async (t) => {
```
boneskull, line 22: (suggestion) add a similar test for `@lavamoat/node`

though I am not entirely sure why we can't add a single scenario in `core`, since `lavamoat-node`, `lavamoat-browserify` and `@lavamoat/node` run all of the core scenarios?

weizman: that was the case originally, then i realized each env has its own set of globalThis aliases, and we want to test each separately, partly to tell passing config from each of them actually works

packages/core/src/generatePolicy.js
weizman: this happens within a function that statically looks up globalThis aliases in the ast, so it is important it covers all options. we _could_ redo this to accept config, but it doesn't seem very straightforward and arguably worth it

packages/lavapack/src/runtime-template.js
weizman: not really, there are 5 occurences of this:
* 2 are within template files that are meant to be static and therefore cannot import anything really
* 1 in a spec file so, maybe, yea, but not supoer important
* 2 in places where importing would make sense

is it worth it for effectively just 2 places?

packages/core/src/generatePolicy.js
boneskull: (theoretical) this really _should_ change at some point such that it can be defined by each consumer of `lavamoat-core`.  there may be a situation in which one of these globals exists in the environment but it is _not_ a reference to `globalThis`.

packages/lavamoat-node/test/globals.spec.js
boneskull: ah, I see. there wouldn't be any elegant way to do this, then.

hopefully I can address that one of these days

packages/lavapack/src/runtime-template.js
boneskull: well, if it exists in the source but not the tests, when the source changes, the tests won't reflect those changes.   so yeah, IMO it's worth it to share the data structure as much as is reasonable

packages/core/src/generatePolicy.js
weizman: agreed

packages/lavapack/src/runtime-template.js
weizman: ok sure.
how would you got about @boneskull?
we don't have some shared space for such stuff, do we? should i just place this in `core`? and then import it from `browserify`? (these 2 cover all non-static occurrences)

boneskull: it seems reasonable to define some frozen objects in `core` and export them from e.g., `constants`. in this case, I'd probably want a "const enum"-type object:

```js
const DEFAULT_GLOBALTHIS_REFS = Object.freeze(/** @type {const} */({
  WINDOW: 'window',
  SELF: 'self',
  GLOBAL: 'global',
  GLOBALTHIS: 'globalThis',
  TOP: 'top',
  FRAMES: 'frames',
  PARENT: 'parent',
}));
```

This would probably be the most flexible data structure (easy conversion to array) and enables expressions like `DEFAULT_GLOBALTHIS_REFS.WINDOW !== globalThis.window` to mitigate bugs of the "typo" variety 😄 

The `const` type causes TS to interpret the keys and values as literals, otherwise it would infer to `Readonly<Record<string, string>>`.

weizman: how's this? https://github.com/LavaMoat/LavaMoat/pull/1539/files

weizman: merged

packages/core/src/makePrepareRealmGlobalFromConfig.js
naugtur: Yes, but only formally. It's the desired behavior and previous was less secure.

packages/core/src/constants.js
```diff
@@ -0,0 +1,19 @@
...
+    TOP: 'top',
+    FRAMES: 'frames',
+    PARENT: 'parent',
+  })
+)
+
+const DEFAULT_GLOBAL_THIS_REFS = Object.freeze(Object.values(GLOBAL_THIS_REFS))
+
+/**
+ * @typedef {(typeof DEFAULT_GLOBAL_THIS_REFS)[number]} GlobalThisRef
```
naugtur, line 16: Where is this type used?

packages/webpack/src/runtime/runtime.js
```diff
@@ -141,7 +141,7 @@ const enforcePolicy = (specifier, referrerResourceId, wrappedRequire) => {
 const theRealGlobalThis = globalThis
 /** @type {any} */
 let rootCompartmentGlobalThis
-const globalAliases = ['globalThis', 'window', 'self']
+const globalAliases = ['window', 'self', 'global', 'globalThis', 'top', 'frames', 'parent']
```
naugtur, line 144: If we want to reuse these from core, we can include them in the runtime assembler thingy

naugtur (approved):
- doublechecked that fictitiousglobals won't appear in lavamoat-node
-  logic in core makes sense

It is a breaking change for every package that depends on core's kernel or endowmentsToolkit

packages/core/src/constants.js
naugtur: @boneskull does this make sense to you?

PR #1526, opened 2025-02-03
feat(webpack): introduce enhancements to support webpack integration

weizman:
This PR focuses on preparing webpack plugin to work well with MetaMask effort to integrate it (see https://github.com/MetaMask/metamask-extension/pull/30134).

In order to do that, I had to:

* cac9e9aa70d809797aaa5acf79d24908a80a3c4e - fix emitSes part regarding how to load SES (I'm not sure about this part, I'm a bit confused regarding where this part came from, this might be redudent)
* 33e836d0f87ea90f82025a29b618a6123751a008 - cache natives in the debug file so that scuttling doesn't collide with it
* feb2bd6015dbdfbea1de9f686208ee001335b129 - add support to the webpack `e` function which support context modules which required support since MetaMask added support in dynamic imports
* 6fee72a97b5cee69d99ab4ed78c4d85299622316 - add a chunk that caches natives needed for webpack core logic so that it doesn't clash with scuttling

These make it so that when integrated with the MetaMask PR, compilation and execution of the app works well

### DIFF
--- a/packages/webpack/README.md
+++ b/packages/webpack/README.md
@@ -105,8 +105,7 @@ This plugin will skip policy enforcement for such ignored modules.
 
 #### HMR
 
-LavaMoat is not compatible with Hot Module Replacement and any form of it would not make much sense with the security guarantees, so you're better off using lavamoat in your production build.
-It's not easy to spot when a tool or setting is enabling HMR though, so LavaMoat plugin will stub out some of the HMR functionality to do nothing.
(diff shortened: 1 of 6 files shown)


weizman:
This bug was introduced at rebase I believe, this fixes it e0044c8

naugtur:
More elaborate versions of these were introduced separately. This can be closed.

Some fixes to debug.js might be portable still. I'll look into it at some point.

PR #1529, opened 2025-02-04
fix(webpack): support regex in scuttling exceptions

weizman:
`core/src/scuttle.js` knows that if a string in the array starts with '/' it should first be turned into a RE, and other packages know to transform REs to strings before passing them on (because if you don't, when embedding conf as a hardcoded string to the final bundle, REs will become `"{}"` instead of the actual RE statement).

This was missed at #1298 - #1529 addresses that

### DIFF
--- a/packages/webpack/src/plugin.js
+++ b/packages/webpack/src/plugin.js
@@ -483,6 +483,15 @@ class LavaMoatPlugin {
         // I stole the idea from Zach of module federation fame
 
         const onceForChunkSet = new WeakSet()
+
+        if (typeof options.scuttleGlobalThis === 'object') {
+          if (typeof options.scuttleGlobalThis.exceptions === 'object') {
+            options.scuttleGlobalThis.exceptions.forEach((exception, i, exceptions) => {
+              exceptions[i] = exception.toString()
+            })
+          }
+        }
+
         const runtimeOptions = {
           scuttleGlobalThis: options.scuttleGlobalThis,
           lockdown: options.lockdown,


packages/webpack/src/plugin.js
```diff
@@ -483,6 +483,8 @@ class LavaMoatPlugin {
         // I stole the idea from Zach of module federation fame
 
         const onceForChunkSet = new WeakSet()
+
+        options?.scuttleGlobalThis?.exceptions?.forEach((n, i, a) => a[i] = n.toString())
```
leotm: curious what the `n` stands for? was expecting `e` for exception 😄 (or element)

leotm (approved):
lgtm (could be worth adding test)

leotm (commented):
ah CI failing looks like 👀

leotm (changes requested):
just need to fix the type error the happy to approve

```console
> lavamoat-monorepo@0.0.0 build:types
> tsc -b

Error: packages/webpack/src/plugin.js(487,37): error TS2339: Property 'exceptions' does not exist on type 'boolean | ScuttlerObjectConfig'.
  Property 'exceptions' does not exist on type 'false'.
Error: packages/webpack/src/plugin.js(487,58): error TS7006: Parameter 'n' implicitly has an 'any' type.
Error: packages/webpack/src/plugin.js(487,61): error TS7006: Parameter 'i' implicitly has an 'any' type.
Error: packages/webpack/src/plugin.js(487,64): error TS7006: Parameter 'a' implicitly has an 'any' type.
Error: Process completed with exit code 2.
```

packages/webpack/src/plugin.js
boneskull: - The callback to `forEach` returns `void`, so we should not return anything.
- Would prefer better variable names

```suggestion
        options?.scuttleGlobalThis?.exceptions?.forEach((exception, i, exceptions) => {
          exceptions[i] = exception.toString()
        })
```

or 

```suggestion
        options?.scuttleGlobalThis?.exceptions?.forEach((exception, i, exceptions) => {
          exceptions[i] = `${exception}`
        })
```

or even

```suggestion
        if (options?.scuttleGlobalThis?.exceptions) {
          options.scuttleGlobalThis.exceptions = optionsScuttleGlobalThis.exceptions.map(String)
        }
```

`ScuttlerObjectConfig` needs an `exceptions` prop added to its definition.  This should fix the other type errors.

boneskull: Providing empty default values for these options may help avoid null-checking everywhere, too; but TS would want you to differentiate between the user-provided `Partial<ScuttlerObjectConfig>` and `ScuttlerObjectConfig` (which would now want all  properties having default empty values to be "required")

weizman: i don't get ts, but this works :shrug: 74acc1ca62e0a5792faf137c7623b90ff8c41e8d

packages/webpack/src/plugin.js
```diff
@@ -483,6 +483,15 @@ class LavaMoatPlugin {
...
 
         const onceForChunkSet = new WeakSet()
+
+        if (typeof options.scuttleGlobalThis === 'object') {
+          if (typeof options.scuttleGlobalThis.exceptions === 'object') {
+            options.scuttleGlobalThis.exceptions.forEach((exception, i, exceptions) => {
+              exceptions[i] = exception.toString()
+            })
+          }
+        }
```
leotm, line 493: https://github.com/LavaMoat/LavaMoat/blob/74acc1ca62e0a5792faf137c7623b90ff8c41e8d/packages/webpack/src/plugin.js#L660

typeof `exceptions` is always `string[] | undefined` right?
(so currently not reaching inside the condition)

leotm: could maybe tidy the nested `if` too, i'll add suggestion

leotm:
```suggestion
        if (
          typeof options.scuttleGlobalThis === 'object' &&
          typeof options.scuttleGlobalThis.exceptions
        ) {
          options.scuttleGlobalThis.exceptions =
            options.scuttleGlobalThis.exceptions.map((exception) =>
              exception.toString()
            )
        }
```

edit: discussed, ignore this suggestion
`typeof options.scuttleGlobalThis.exceptions === 'object'` is what we're after

leotm: (the new var names lgtm)

leotm (approved):

PR #1539, opened 2025-02-16
refactor(core/browserify): export global this aliases to one place

weizman:
export globalThis aliases to a single file to be consumed by the rest of the codebase

### DIFF
--- a/packages/core/src/constants.js
+++ b/packages/core/src/constants.js
@@ -1,20 +1,19 @@
-const GLOBAL_THIS_REFS = Object.freeze(/** @type {const} */({
-  WINDOW: 'window',
-  SELF: 'self',
-  GLOBAL: 'global',
-  GLOBAL_THIS: 'globalThis',
-  TOP: 'top',
-  FRAMES: 'frames',
-  PARENT: 'parent',
-}))
+const GLOBAL_THIS_REFS = Object.freeze(
+  /** @type {const} */ ({
+    WINDOW: 'window',
+    SELF: 'self',
+    GLOBAL: 'global',
+    GLOBAL_THIS: 'globalThis',
+    TOP: 'top',
+    FRAMES: 'frames',
+    PARENT: 'parent',
+  })
+)
 
 const DEFAULT_GLOBAL_THIS_REFS = Object.freeze(Object.values(GLOBAL_THIS_REFS))
 
 /**
- * @typedef {typeof DEFAULT_GLOBAL_THIS_REFS[number]} GlobalThisRef
+ * @typedef {(typeof DEFAULT_GLOBAL_THIS_REFS)[number]} GlobalThisRef
  */
 
-
-module.exports = {GLOBAL_THIS_REFS, DEFAULT_GLOBAL_THIS_REFS}
-
-module.exports = {DEFAULT_GLOBAL_THIS_REFS}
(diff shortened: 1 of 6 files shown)


packages/core/src/constants.js
```diff
@@ -0,0 +1,11 @@
+const DEFAULT_GLOBAL_THIS_REFS = Object.freeze(/** @type {readonly string[]} */(Object.values({
+  WINDOW: 'window',
+  SELF: 'self',
+  GLOBAL: 'global',
+  GLOBAL_THIS: 'globalThis',
+  TOP: 'top',
+  FRAMES: 'frames',
+  PARENT: 'parent',
+})));
```
boneskull: - Not sure of the point of defining an object if all we're going to do is export an array.  
- The type of this variable evaluates to `Readonly<string[]>` which is lossy. What we really want is to be able--at the type level--to know that e.g., `top` is a member of this data structure; as written, we cannot.

```suggestion
const GLOBAL_THIS_REFS = Object.freeze(/** @type {const} */({
  WINDOW: 'window',
  SELF: 'self',
  GLOBAL: 'global',
  GLOBAL_THIS: 'globalThis',
  TOP: 'top',
  FRAMES: 'frames',
  PARENT: 'parent',
}))

const DEFAULT_GLOBAL_THIS_REFS = Object.freeze(Object.values(GLOBAL_THIS_REFS))

/**
 * @typedef {typeof DEFAULT_GLOBAL_THIS_REFS[number]} GlobalThisRef
 */


module.exports = {GLOBAL_THIS_REFS, DEFAULT_GLOBAL_THIS_REFS}
```

With the above, the type of `GLOBAL_THIS_REFS` is `Readonly<{WINDOW: 'window', SELF: 'self', GLOBAL: 'global' ...}>`. The type of `DEFAULT_GLOBAL_THIS_REFS` is then `Readonly<('window' |  'self' | 'global' | ...)[]>`. And the convenience type `GlobalThisRef` is `'window'|'self'|'global'|...` (`typeof arr[number]` is how you get a union of all items in an array; if `arr` is instead `string[]`, then `typeof arr[number]` is just `string`).

So we can use what I've written this way:

```js
/**
 * @param {GlobalThisRef} ref
 */
function foo(ref) {
  // ref must be one of the values of `GLOBAL_THIS_REFS` 
  // or items in `DEFAULT_GLOBAL_THIS_REFS`
  assert(DEFAULT_GLOBAL_THIS_REFS.includes(ref))

  if (ref === GLOBAL_THIS_REFS.TOP) {
    // ...
  }
  
  if (ref === 'butts') { // type error!
    // ...
  } 
}
```

Also, I'm not sure how semicolons are creeping in here. Do you have prettier/eslint enabled in your IDE? Do you have precommit hooks enabled (`npm run setup`)? The hooks should automatically fix this.

boneskull: Otherwise yes, I think this is the right way to go 😄

weizman: Thanks @boneskull that makes sense! Also just ran `setup` so I'll be wary of that from now on

weizman: https://github.com/LavaMoat/LavaMoat/pull/1539#discussion_r1959677575

boneskull (approved):

boneskull:
- [ ] fix commit msg 😄 
- [ ] add description to PR

PR #1547, opened 2025-02-27
feat(node): add custom errors

boneskull:
This adds a type definition and factory for custom error classes, including a set of custom errors.  Custom errors may or may not have specific implementations, but they all have a `code` property.

My ultimate aim here is to have all errors explained on the docs site; when one of our custom errors is thrown, it will direct the user to a URL with more information.

This sets the groundwork for adding more custom implementations and any other fancy stuff we want to do later.

### DIFF
--- a/packages/node/src/error-code.js
+++ b/packages/node/src/error-code.js
@@ -1,0 +1,28 @@
+/**
+ * Provides {@link ErrorCodes} enum
+ *
+ * @packageDocumentation
+ */
+
+/**
+ * Enum of error codes thrown by this package
+ *
+ * Each one of these corresponds to a specific implementation of
+ * `LavaMoatError`.
+ */
+export const ErrorCodes = Object.freeze(
+  /** @type {const} */ ({
+    AttenuationFailure: 'LMN_ATTENUATION_FAILURE',
+    GenerationFailure: 'LMN_GENERATION_FAILURE',
+    ExecutionFailure: 'LMN_EXECUTION_FAILURE',
+    ConversionFailure: 'LMN_CONVERSION_FAILURE',
+    InvalidArguments: 'LMN_INVALID_ARGUMENTS',
+    InvalidCompartment: 'LMN_INVALID_COMPARTMENT',
+    InvalidPolicy: 'LMN_INVALID_POLICY',
+    NoExecutable: 'LMN_NO_EXECUTABLE',
+    NoPolicy: 'LMN_NO_POLICY',
+    NoWorkspace: 'LMN_NO_WORKSPACE',
+    PermissionDenied: 'LMN_PERMISSION_DENIED',
+    TrustMismatch: 'LMN_TRUST_MISMATCH',
+  })
+)
--- a/packages/node/src/error.js
+++ b/packages/node/src/error.js
@@ -1,0 +1,140 @@
+/**
+ * @import {ErrorCode, LavaMoatErrorClass, LavaMoatErrorClassParams} from './errors.js'
+ * @import {Class} from 'type-fest'
+ */
+import { ErrorCodes } from './error-code.js'
+
+/**
+ * Given a name, code, and implementation, create a new
+ * {@link LavaMoatErrorClass}.
+ *
+ * This handles some boilerplate when creating new custom error classes.
+ *
+ * @template {ErrorCode} Code
+ * @template {ErrorOptions} [Options=ErrorOptions] Default is `ErrorOptions`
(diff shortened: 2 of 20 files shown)


boneskull:
* **#1547** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1547?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1547?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1620** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1620?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

boneskull:
I kind of just did this for my own satisfaction, but I would _also_ die on the hill of "errors need error codes", so, take that as you will.

packages/node/src/error-code.js
boneskull, line 1: Each unique type of error has its own error code, and they are defined here

packages/node/src/error.js
boneskull, line 1: This file defines all the custom errors and contains a factory helper function which massages the base implementations into things which implement the `LavaMoatErrorClass` interface.

packages/node/src/errors.ts
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @template Cause Reason for the error; usually another error. Specify if known
+ * @template Options Additional options for the error; if this object has any
+ *   required fields, the second parameter will be required.
+ */
+export type LavaMoatErrorClassParams<
+  Options extends ErrorOptions = ErrorOptions,
+> =
+  // tuples seem to struggle with conditional types, so it doesn't look like I can embed it in the tuple
+  Partial<Options> extends Options
```
boneskull, line 40: this means, essentially, "are all the fields in `Options` optional"?  and we can change the constructor parameters as needed.

packages/node/src/errors.ts
```diff
@@ -0,0 +1,58 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import type { Class, ValueOf } from 'type-fest'
+import type { ErrorCodes } from './error-code.js'
+
+export type ErrorCode = ValueOf<typeof ErrorCodes>
```
boneskull, line 5: this is a union of all possible error codes (the ones in SCREAMING_SNAKE_CASE)

packages/node/src/errors.ts
```diff
@@ -0,0 +1,58 @@
...
+> =
+  // tuples seem to struggle with conditional types, so it doesn't look like I can embed it in the tuple
+  Partial<Options> extends Options
+    ? [message: string, errOptions?: Options]
+    : [message: string, errOptions: Options]
+
+/**
+ * The _class_ of a {@link LavaMoatError}.
+ */
+export type LavaMoatErrorClass<
```
boneskull, line 47: in TS, if we have class `Foo`, type `Foo` is an instance of that class.

the _class_ `Foo` would be `typeof Foo` or could be expressed as a constructor e.g., `new (...args: any[]) => Foo` (a constructor function in TS is just a function type prepended with the `new` keyword).

`type-fest` provides a shortcut `Class` to define "a constructor function which returns a Foo", but it doesn't let us define static properties.  Below, static props are defined via an intersection type.

boneskull:
Note to reviewers:

- I've made some annotations in the code that may or may not be helpful
- If you see a plain `Error`, `TypeError`, `ReferenceError`, please call it out (it may have snuck in via merge)

packages/node/src/error.js
```diff
@@ -0,0 +1,135 @@
...
+ * Given a name, code, and implementation, create a new
+ * {@link LavaMoatErrorClass}.
+ *
+ * This handles some boilerplate when creating new custom error classes.
+ *
+ * @template {ErrorCode} Code
+ * @template {ErrorOptions} [Options=ErrorOptions] Default is `ErrorOptions`
+ * @param {string} name
+ * @param {Code} code
+ * @param {Class<Error, LavaMoatErrorClassParams<Options>>} ctor
```
boneskull, line 17: change this such that `Error extends typeof ctor` or some such

naugtur (approved):

packages/node/src/error.js
boneskull: I don't remember what this means. I'll do it later if I remember.

PR #1553, opened 2025-03-04
webpack context modules

naugtur:
Support for:
- ContextModule type module, that encapsulates loading fragments of the bundle dynamically, lazy, over the time of the app working.
- loading chunks with `__webpack_require__.e` is limited to chunks known at build time.

`context` field in ContextModule points to a path that can be matched to a package.


packages/webpack/src/plugin.js
```diff
@@ -430,12 +467,15 @@ class LavaMoatPlugin {
...
               readableResourceIds: options.readableResourceIds,
               unenforceableModuleIds,
+              contextModules,
               externals,
               paths: knownPaths,
               policy: policyToApply,
               canonicalNameMap,
             })
 
+            // =================================================================
```
boneskull, line 477: what is this?

packages/webpack/src/plugin.js
```diff
@@ -548,6 +588,17 @@ class LavaMoatPlugin {
                     data: identifierLookup.unenforceableModuleIds || null,
                     json: true,
                   },
+                  {
+                    name: 'ctxm',
```
boneskull, line 592: is `ctxm` more esoteric webpack apocrypha?

packages/webpack/src/runtime/runtime.js
```diff
@@ -121,8 +122,13 @@ const enforcePolicy = (specifier, referrerResourceId, wrappedRequire) => {
...
   const requestedResourceId = findResourceId(specifier)
   if (!requestedResourceId) {
+    if (LAVAMOAT.ctxm.includes(specifier)) {
+      throw Error(
+        `'${referrerResourceId}' is trying to load a module dynamically from a dependency it has not been allowed using the following Webpack Context Module: '${specifier}'`
+      )
+    }
     throw Error(
-      `Requested specifier ${specifier} is not allowed as a builtin and not a known dependency of ${referrerResourceId}. Regenerate policy or add it to policy-override.json.`
+      `Requested specifier '${specifier}' is not allowed as a builtin and not a known dependency of '${referrerResourceId}'. Regenerate policy or add it to policy-override.json.`
```
```suggestion
      `Requested specifier '${specifier}' is not allowed as a builtin and is not a known dependency of '${referrerResourceId}'; regenerate policy or add it to policy-override.json.`
```

packages/webpack/src/runtime/runtime.js
```diff
@@ -261,6 +270,20 @@ const lavamoatRuntimeWrapper = (resourceId, runtimeKit) => {
     // wrap webpack runtime for policy check and hardening
     const policyRequire = wrapRequireWithPolicy(__webpack_require__, resourceId)
 
+    /**
+     * Wrapped chunk loader
+     *
+     * @param {string | number} chunkId
+     */
+    policyRequire.e = (chunkId) => {
```
boneskull, line 278: does the type of `policyRequire` have an `e` prop as returned by `wrapRequireWithPolicy`?

packages/webpack/src/runtime/runtime.js
```diff
@@ -261,6 +270,20 @@ const lavamoatRuntimeWrapper = (resourceId, runtimeKit) => {
...
+      // Chunk resolution at runtime may be a lookup table or a string concatenation. In the latter case it can be fooled into loading any file from under publicPath. That's why we only want to accept known chunks.
+      if (!LAVAMOAT.kch.includes(chunkId)) {
```
boneskull, line 280: could this array get large enough to want to convert it to a `Set`?

packages/webpack/src/runtime/runtime.js
```diff
@@ -261,6 +270,20 @@ const lavamoatRuntimeWrapper = (resourceId, runtimeKit) => {
...
+        throw Error(
+          `Attempt to load a chunk that was not known at compile time: ${chunkId} from ${resourceId}`
```
boneskull, line 282: is this actionable? how?

packages/webpack/test/aa.spec.js
```diff
@@ -59,23 +59,26 @@ test('aa - generateIdentifierLookup', (t) => {
...
+        '/home/user/app': '$root$',
+        '/home/user/app/node_modules/leftpad/': 'leftpad',
+        '/home/user/app/node_modules/fast-json-patch': 'fast-json-patch',
+        '/home/user/app/node_modules/fast-json-patch/module': 'fast-json-patch',
+        '/home/user/app/node_modules/@ethereumjs/util/dist/':
+          '@ethereumjs/util',
+        '/home/user/app/node_modules/@ethereumjs/util/node_modules/ethereum-cryptography/dist/':
+          '@ethereumjs/util>ethereum-cryptography',
+      })
+    ),
```
boneskull: if possible, I'd omit `/home/user/app` and just use `/` as `$root$`

packages/webpack/test/e2e-dynamic-import.spec.js
```diff
@@ -0,0 +1,316 @@
+const test = require('ava')
+const { scaffold, runScriptWithSES } = require('./scaffold.js')
+const { makeConfig } = require('./fixtures/main/webpack.config.js')
+
+test.before(async (t) => {
```
boneskull, line 5: (suggestion) define the context type ([example](https://github.com/LavaMoat/LavaMoat/blob/68fa29918d76deec44c955c9ac1f085c6a3a906e/packages/node/test/e2e/cli-generate.spec.js#L18-L25))

if you choose _not_ to do this, it's helpful to get actual types for `test`:

```js
/**
 * @import {TestFn} from 'ava'
 */
const test = /** @type {TestFn} */(require('ava')
```

...since the version of `ava` we're using has broken types for CJS; the type assertion is not necessary in ESM.  🙄

packages/webpack/test/e2e-nodejs.spec.js
```diff
@@ -28,7 +27,21 @@ test('webpack/node.js - policy shape', (t) => {
 })
 
 test('webpack/node.js - bundle runs without throwing', (t) => {
+  let capture
```
boneskull, line 30:
```suggestion
  let capture
  t.plan(3)
```

packages/webpack/test/scaffold.js
```diff
@@ -120,6 +128,7 @@ function runScript(code, globals = defaultGlobals()) {
     throw new Error('runScript requires a bundle string as the first argument')
   }
   const context = createContext(globals)
+  context.self = context // minimal browser emulation
```
boneskull, line 131: what does this mean?

boneskull (approved):
mostly some questions.

packages/webpack/src/buildtime/aa.js
```diff
@@ -90,6 +92,20 @@ exports.generateIdentifierLookup = ({
         }
       }
     }
+    for (const c of contextModules) {
+      const resourceId = getPackageNameForModulePath(
```
weizman, line 96: nit: why `resourceId`? I mean, `getPackageNameForModulePath` sounds like it returns a name, right? I get that the name is effectively an identifier, but at the same time a bit confusing because id reminds me of something different than a name? I'm stating this just in case there's an explanation of this worth hearing, but either way feel free to ignore

packages/webpack/src/plugin.js
```diff
@@ -343,6 +357,11 @@ class LavaMoatPlugin {
             const chunkGraph = compilation.chunkGraph
 
             Array.from(chunks).forEach((chunk) => {
+              // Collect chunk IDs and info while we're here
+              if (chunk.id !== null) {
```
weizman, line 361: I sense a weird, hard-to-uncover, bug around `chunk.id` being "of-false" without being specifically `null` sometime in the future... perhaps I'm wrong

packages/webpack/src/runtime/runtime.js
naugtur: It gives you a chance to figure out what's happening if you have plugins that add chunks to your build later than we anticipated that to be possible. chunkIds are stable

packages/webpack/src/runtime/runtime.js
naugtur: All other stuff that happens with chunks is orders of magnitude more runtime complexity. Each chunk must be added to the page and executed and so on. This is most likely to contain between 0 and 20 items for all reasonable apps

packages/webpack/test/e2e-dynamic-import.spec.js
naugtur: I'm looking forward to switching to node's built-in test runner now that it has snapshots.

packages/webpack/src/buildtime/aa.js
naugtur: we're calling it multiple things interchangeably. canonical name == resource. aa uses canonical name, policy file has a key called resources.
canonical name is a decent internal name, but I'm slowly switching to calling them resources to be consistent with policy.

packages/webpack/src/plugin.js
naugtur: it should not be possible for a chunk we're interested in supporting to not have an ID but types allowed it :shrug:

packages/webpack/src/plugin.js
naugtur: helps navigate the plugin file visually until I split it into more files. hopefully this spring!

packages/webpack/src/plugin.js
naugtur: that's short for context modules. it doesn't have to be that ugly. I was in a mood to optimize bundle size prematurely

packages/webpack/src/runtime/runtime.js
naugtur: the type is  `@typedef {WebpackRequire & { [key: string]: any }} WrappedRequire` because it's a function that's specific and it has a bunch of stuff on it that's very internal to webpack.

packages/webpack/test/aa.spec.js
naugtur: okay

packages/webpack/test/e2e-dynamic-import.spec.js
naugtur: I hate the `// eslint-disable-next-line ava/use-test` bit, but the type override is ok.
I optimized it to one line
```
const test = /** @type {import('ava').TestFn} */(require('ava'))
```

packages/webpack/test/scaffold.js
naugtur: browsers have a `self` global that some things rely on.

PR #1556, opened 2025-03-10
fix(git-safe): improve git specifier detection and tests

naugtur:
- makes errors in resolving non git URLs as git URLs easier to debug
- fixes incorrectly identified yarn specifiers 
- adds tests around the issue


boneskull:
ref: #1548

packages/git-safe-dependencies/package.json
```diff
@@ -37,7 +37,8 @@
     "supply-chain"
   ],
   "scripts": {
-    "test": "ava"
+    "test": "ava",
+    "test-manual": "cd test/manual && bash run.sh > snapshot.txt 2>&1"
```
boneskull: normally I'd insist on "no bash scripts in the repo" but this seems like a thing that is not meant to be run as part of a typical workflow, so I'll let it slide (at least until I rewrite it myself).

...but it looks like essentially a bunch of E2E snapshot tests.  why not actually write those tests?

note that I did not previously complain about this because the original PR adding the workspace was merged while I was OOO. 😄

packages/git-safe-dependencies/src/isgit.js
```diff
@@ -11,7 +11,8 @@ const isUnsafeUrl = (url) =>
 
 const isGitUrl = (url) => gitUrlRegex.test(url)
 const isGHRepoSpecifier = (specifier) =>
-  specifier.startsWith('github:') || specifier.match(/^[^/]+\/[^/]+/)
+  !!(specifier.startsWith('github:') || specifier.match(/^[^/@]+\/[^/]+$/)) &&
+  !specifier.startsWith('npm:')
```
boneskull:
```suggestion
  !specifier.startsWith('npm:') && (specifier.startsWith('github:') || /^[^/@]+\/[^/]+$/.test(specifier))  
```

this defers regex execution until the last step and avoids the need for casting to boolean

- [ ] would prefer a `const ghRepoSpecifierRegex = /.../` akin to `gitUrlRegex`
- [ ] please show example of what the regex is trying to match or otherwise document the `isGHRepoSpecifier` function so we know what it's trying to do

packages/git-safe-dependencies/src/isgit.js
```diff
@@ -32,6 +33,12 @@ exports.gitInfo = (url) => {
       .replace('/tar.gz/', '.git#')
   }
   const info = gitinfo.fromUrl(url)
+  if (!info) {
+    throw new Error(`Invalid git url: ${url}`)
+  }
+  if (!Object.hasOwn(info, 'committish')) {
+    info.committish = null
+  }
```
boneskull, line 40: why wouldn't this suffice?

```suggestion
  info.committish ??= null
```

packages/git-safe-dependencies/test/isgit.spec.js
```diff
@@ -1,5 +1,35 @@
 const test = require('ava')
-const { gitInfo } = require('../src/isgit')
+const { gitInfo, isGitUrl, isGitSpecifier } = require('../src/isgit')
+
+test('isGitUrl should identify git URLs correctly', (t) => {
```
boneskull, line 4: please use `t.plan` or split into multiple tests (either one per assertion or one per `true` / `false` assertions, respectively)

packages/git-safe-dependencies/test/manual/snapshot.txt
```diff
@@ -0,0 +1,411 @@
...
+      // problem id: fork-short-head:e1be28b4
+
+ fork-short-merge:
+  [ValidateGitUrl] The specified commit hash for fork-short-merge does not exist in the repository. It might exist on a fork, but you should not trust that.
+  specifier: naugtur/not-really-a-package#1/merge
+  url: https://codeload.github.com/naugtur/not-really-a-package/tar.gz/2a0d442b8a68a802b72f6a4a8240d2ae8ef950aa
+      // problem id: fork-short-merge:cf613cb8
+
+ git-http@git+http://github.com/naugtur/not-really-a-package:
+  [ValidateProtocol] detected invalid scheme(s) for package: git-http@git+http://github.com/naugtur/not-really-a-package
```
boneskull: why does this tool bother validating the protocol?  if `npm` won't understand it, it seems relatively harmless.

packages/git-safe-dependencies/test/manual/snapshot.txt
```diff
@@ -0,0 +1,411 @@
...
+        // problem id: fork-longer-hash:ea34e0b9
+
+ fork-longer-hash:
+  [packageJson] GIT URLs in package.json dependencies must use a full-length commit hash.
+    expected: naugtur/not-really-a-package#COMMITHASH_40_CHARACTERS_LONG
+    actual: naugtur/not-really-a-package#05ba7b1263a4ba5b98
+        // problem id: fork-longer-hash:2561615b
+
+ fork-merge-hash:
+  [ValidateGitUrl] The specified commit hash for fork-merge-hash does not exist in the repository. It might exist on a fork, but you should not trust that.
```
boneskull: some ideas

- [ ] name the repository instead of writing "the repository"
- [ ] validation errors should be easily visually distinguishable from missing commit hashes and appear more "severe"
- [ ] if possible, determine _which_ fork the commit lives in. if that means brute-force, forget it
- [ ] I don't know what "problem id" refers to or why it is formatted the way it is
- [ ] the identifier in the brackets (e.g., `[packageJson]`) does not provide enough context into its meaning.  is it supposed to be an error code?
- [ ] `expected` vs `actual` should appear as green/red
- [ ] if the type of `expected` is an array, it should be formatted nicely 😄 
- [ ] consider omitting validation problems entirely (maybe beyond scope of the tool?)

> checkboxes do not imply anything; they are for your convenience if you choose to use them

packages/git-safe-dependencies/src/index.js
```diff
@@ -21,7 +21,7 @@ const runAllValidations = async ({ cwd, type }) => {
   } catch (e) {
     errors.push({
```
boneskull, line 22: instead of an array of validation errors, what about using an `AggregateError`?  this may make more ergonomic to consume as a library (if that is or will become in scope).  each type of child error could have its own class and custom formatter, if desired.

boneskull (approved):
~~some~~ many suggestions, but nothing that I'd block about. much of it could be addressed in later PRs, if desired

boneskull:
Oh, also:

- [ ] if the stuff Socket found is _actually_ dangerous, we should probably not make references to those packages because of the (obvious) risk of automated security scanners running amok
- [ ] We should dogfood this somehow.

packages/git-safe-dependencies/package.json
naugtur: these started out as a pile of examples I was manually testing on and I'm adapting them to be proper tests slowly. Might just ask AI to convert all of it. Not sure if it's gonna be less frustrating tho.

packages/git-safe-dependencies/src/index.js
naugtur: The array gets its items enriched with other info later on and it'd feel odd to put extra fields on errors within AggregateError for me. The structure is a bit flexible and may still change. I'm not ready to make it more fancy at this point.

packages/git-safe-dependencies/src/isgit.js
naugtur: This exists because error output is better if destructuring committish out of the object doesn't fail. 
If `undefined` was explicitly returned, I wanted to keep it to make it easier to figure out what's going on.

packages/git-safe-dependencies/test/isgit.spec.js
naugtur: I intuitively disagree with splitting. Seems excessive. What do I gain for the boilerplate?

I understand the argument behind adding a `t.plan` altho in this particular case don't see how it prevents any issues. Will add it for the sake of posing as a good citizen, but curious if there's an objective reason I'm missing.

naugtur: fun fact: AI suggestions for the number to put in t.plan were consistently off by 1

packages/git-safe-dependencies/package.json
naugtur: That was actually pretty decent to work with

packages/git-safe-dependencies/test/manual/snapshot.txt
naugtur: > some ideas
> 
>     * [ ]  name the repository instead of writing "the repository"
can do
> 
>     * [ ]  validation errors should be easily visually distinguishable from missing commit hashes and appear more "severe"
Not sure how. Will try something
> 
>     * [ ]  if possible, determine _which_ fork the commit lives in. if that means brute-force, forget it
nope, that info is not available without querying each fork :D 
> 
>     * [ ]  I don't know what "problem id" refers to or why it is formatted the way it is
you can use it in the ignore file
> 
>     * [ ]  the identifier in the brackets (e.g., `[packageJson]`) does not provide enough context into its meaning.  is it supposed to be an error code?
it's the name of the validator that reported the problem. I'm thinking it could be helpful for powerusers.
> 
>     * [ ]  `expected` vs `actual` should appear as green/red
how strongly do you advocate for that?
> 
>     * [ ]  if the type of `expected` is an array, it should be formatted nicely 😄
> 
>     * [ ]  consider omitting validation problems entirely (maybe beyond scope of the tool?)
What do you mean?
> 
> 
> > checkboxes do not imply anything; they are for your convenience if you choose to use them

naugtur:
> Oh, also:
> 
>     * [ ]  if the stuff Socket found is _actually_ dangerous, we should probably not make references to those packages because of the (obvious) risk of automated security scanners running amok
the example file had some useful cases, but 
> 
>     * [ ]  We should dogfood this somehow.
this and some of the rest of it too

naugtur:
> Oh, also:
> 
>     * [ ]  if the stuff Socket found is _actually_ dangerous, we should probably not make references to those packages because of the (obvious) risk of automated security scanners running amok

> 
>     * [ ]  We should dogfood this somehow.

naugtur:
Folks are waiting on a fix, so I'll merge this and look into releasing. Most improvements listed are worth doing.

PR #1576, opened 2025-03-24
E2E dogfOOding test - run a webpack project with @lavamoat/node

naugtur:
Current issues:
- [ ] `@lavamoat/node` complains graceful-fs is missing, had to install it
- [x]  `@lavamoat/node` on main is incapable of handling the `#!`
- [ ] merged with `boneskull/untrusted-entrypoints` throws `Globals attenuation errors: Error while attenuating globals for "webpack" with "<default attenuator>": "Root compartment globalThis already initialized; this is a bug"`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -77,7 +77,6 @@
       "version": "2.3.0",
       "dev": true,
       "license": "Apache-2.0",
-      "peer": true,
       "dependencies": {
         "@jridgewell/gen-mapping": "^0.3.5",
         "@jridgewell/trace-mapping": "^0.3.24"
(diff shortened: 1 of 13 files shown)


boneskull:
> Globals attenuation errors: Error while attenuating globals for "webpack" with "<default attenuator>": "Root compartment globalThis already initialized; this is a bug"

afaict this could happen is if the first attenuated compartment was not the root compartment (which we were assuming).  if that's not true, then we need to either fix it (whatever "it" is in this case) such that it _becomes_ true, or we need to change the code to be agnostic about the order in which it receives compartments.

boneskull:
- [ ] Try to find a case in which a non-entry compartment has a dependency on the entry compartment. This theoretically happens in `webpack`, where it depends on `@webpack/cli` which depends on `webpack`. This is due `@webpack/cli` being a dev dep of `webpack` and `webpack` being a peer dep of `@webpack/cli`. If so, there's a `ReferenceError` tossed out of `policy-gen-compartment-class` that should be hit.

PR #1590, opened 2025-04-04
chore(eslint): validate config files against JSON schemas

boneskull:
This change adds [eslint-plugin-json-schema-validator](https://npm.im/eslint-plugin-json-schema-validator) to provide validation of config files against JSON schemas. It supports JSON, JSON5 and YAML config files.

The plugin will either a) download and cache schemas from the URL provided by the top-level `$schema` property (if present in file being linted); or b) it will load a catalog from [SchemaStore](https://www.schemastore.org) and try to match the file. If it matches, it will download and cache the schema. Cached schemas are stored in `node_modules/eslint-plugin-json-schema-validator/.cached_schemastore`. This is essentially the same thing that VSCode does by default.

I understand we may be concerned about this behavior, but I could not find a quick way around it without loss of functionality; there's no way to manually configure an association between any given file and a vendored schema _without_ using the `$schema` prop. This is needed because `$schema` is just a convention and some strict schemas do not allow the property (specifically the GHA workflow schema)!

If we are truly concerned and feel it's worth taking further steps to improve our posture, I can think of a couple options:

- We _may_ be able to vendor schemas and pre-populate `eslint-plugin-json-schema-validator`'s cache with symlinks of vendored schemas (and the catalog, I guess?) as part of the `setup` script.
- Upstream a PR to allow manually mapping config files to schemas.

Meanwhile, if we decide to merge this, we should carefully inspect any proposed upgrades to `eslint-plugin-json-schema-validator`.

---

The subsequent change lints the `nodejs` workflow.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -11,6 +11,7 @@ import { readFileSync } from 'node:fs'
 import eslintJs from '@eslint/js'
 import prettierConfig from 'eslint-config-prettier/flat'
 import avaPlugin from 'eslint-plugin-ava'
+import jsonSchemaValidatorPlugin from 'eslint-plugin-json-schema-validator'
 import jsoncPlugin from 'eslint-plugin-jsonc'
 import nodePlugin from 'eslint-plugin-n'
 import nodeImportPlugin from 'eslint-plugin-node-import'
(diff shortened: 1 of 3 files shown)


boneskull:
* **#1590** <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1590?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1590?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1600** <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1600?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

Issue #1583, opened 2025-03-26
lavamoat-core: add "redefine" global policy to policy schema


boneskull:
In addition to `read` and `true`, a global policy can now have a value of `redefine`, which allows a compartment to call `Object.defineProperty` or `Object.defineProperties` against `globalThis` and the property name.

`refine` is only valid for _nested_ global properties, which we should be able to  express.

PR #1592, opened 2025-04-09
feat(core): add write+define policy option to allow defineProperty on globals

### LINKED ISSUE #1584: @lavamoat/node: support "redefine" global policy in attenuator
Depends on both #1583 and #1582 to be implemented first.

May not require much other than tests, because `getEndowmentsForConfig()` accepts the entire global policy object.

naugtur:
fixes #1584 
Handles the case where a package needs to be allowed write to a global with define semantics

decided to go with `write+define` instead of `redefine` so that `write` is a common prefix for all write permissions we may wanna introduce.

- [ ] typescript is panicking about the Object reference
- [ ] The feature requires getEndowmentsForConfig to be called with all 4 arguments  
  The thing is, it could be a separate function that gets applied where getEndowmentsForConfig has been used but it would require access to the `allowRedefine` list so it can't be a separate function.
- [x] verify that the right references (unwrapTo unwrapFrom) are used and that the test emulates real use sufficiently
- [ ] Update the logic that produces knownWritableFields to use `.startsWith('write')`

### DIFF
--- a/packages/core/src/endowmentsToolkit.js
+++ b/packages/core/src/endowmentsToolkit.js
@@ -23,6 +23,8 @@
 
 module.exports = endowmentsToolkit
 
+const { assign, create, defineProperty } = Object
+
 // Exports for testing
 module.exports._test = { instrumentDynamicValueAtPath }
 
@@ -84,6 +86,8 @@ function endowmentsToolkit({
     const allowedWriteFields = new Set()
     /** @type {string[]} */
     const explicitlyBanned = []
+    /** @type {string[]} */
+    const allowRedefine = []
 
     Object.entries(packagePolicy.globals).forEach(
       ([path, packagePolicyValue]) => {
@@ -103,7 +107,10 @@ function endowmentsToolkit({
           return
         }
         // write access handled elsewhere
-        if (packagePolicyValue === 'write') {
+        if (
+          typeof packagePolicyValue === 'string' &&
+          packagePolicyValue.startsWith('write')
+        ) {
           if (!handleGlobalWrite) {
             return
           }
@@ -114,6 +121,14 @@ function endowmentsToolkit({
           }
           allowedWriteFields.add(path)
           whitelistedReads.push(path)
+          if (packagePolicyValue === 'write+define') {
+            if (!unwrapFrom) {
+              throw Error(
+                'LavaMoat - write+define value in policy requires unwrapFrom to be set when generating endowments'
+              )
+            }
+            allowRedefine.push(path)
+          }
           return
         }
         if (packagePolicyValue !== true) {
@@ -126,7 +141,7 @@ function endowmentsToolkit({
     )
     // sort by length to optimize further steps
     whitelistedReads.sort((a, b) => a.length - b.length)
-    return makeMinimalViewOfRef(
+    const endowments = makeMinimalViewOfRef(
       sourceRef,
       whitelistedReads,
       unwrapTo,
@@ -134,6 +149,41 @@ function endowmentsToolkit({
       explicitlyBanned,
       allowedWriteFields
     )
+
+    if (allowRedefine.length > 0 && unwrapFrom && unwrapTo) {
+      const expectedTarget = unwrapTo || endowments
+      const originalDefineProperty = unwrapFrom.Object.defineProperty
+      /**
+       * @param {any} target
+       * @param {PropertyKey} key
+       * @param {PropertyDescriptor} descriptor
+       * @returns {boolean}
+       */
+      const redefine = (target, key, descriptor) => {
+        if (
+          target === expectedTarget &&
+          typeof key !== 'symbol' &&
+          allowRedefine.includes(`${key}`)
+        ) {
+          target[key] = descriptor.value
+          return true
+        }
+        originalDefineProperty(target, key, descriptor)
+        return true
+      }
+      endowments.Object = assign(
+        create(unwrapFrom.Object.prototype),
+        unwrapFrom.Object
+      )
+      defineProperty(endowments.Object, 'defineProperty', {
+        value: redefine,
+        writable: false,
+        configurable: false,
+        enumerable: true,
+      })
+    }
+
+    return endowments
   }
 
   /**
--- a/packages/core/src/endowmentsToolkit.js
+++ b/packages/core/src/endowmentsToolkit.js
@@ -150,8 +150,8 @@ function endowmentsToolkit({
       allowedWriteFields
     )
 
-    if (allowRedefine.length > 0 && unwrapFrom && unwrapTo) {
-      const expectedTarget = unwrapTo || endowments
+    if (allowRedefine.length > 0 && unwrapFrom) {
+      const expectedTarget = unwrapFrom || endowments
       const originalDefineProperty = unwrapFrom.Object.defineProperty
       /**
        * @param {any} target
--- a/packages/core/src/endowmentsToolkit.js
+++ b/packages/core/src/endowmentsToolkit.js
@@ -23,7 +23,7 @@
 
 module.exports = endowmentsToolkit
 
-const { assign, create, defineProperty } = Object
+const { assign, create, entries, defineProperty } = Object
 
 // Exports for testing
 module.exports._test = { instrumentDynamicValueAtPath }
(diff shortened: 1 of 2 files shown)


packages/core/src/endowmentsToolkit.js
```diff
@@ -126,14 +141,49 @@ function endowmentsToolkit({
...
       sourceRef,
       whitelistedReads,
       unwrapTo,
       unwrapFrom,
       explicitlyBanned,
       allowedWriteFields
     )
+
+    if (allowRedefine.length > 0 && unwrapFrom && unwrapTo) {
+      const expectedTarget = unwrapTo || endowments
```
naugtur: I think this is wrong. But it works in tests, so I probably confused the order in tests...

packages/core/src/endowmentsToolkit.js
```diff
@@ -114,6 +121,14 @@ function endowmentsToolkit({
           }
           allowedWriteFields.add(path)
           whitelistedReads.push(path)
+          if (packagePolicyValue === 'write+define') {
```
boneskull: I guess I don't love `write+define`. at first glance, this seems like _multiple_ permissions, and multiple permissions want to be expressed as an object or array.

what of `overwrite`?  or is this _specifically_ for `Object.defineProperty()`?

packages/core/src/endowmentsToolkit.js
```diff
@@ -114,6 +121,14 @@ function endowmentsToolkit({
...
+            if (!unwrapFrom) {
+              throw Error(
```
boneskull: (doesn't really matter)

```suggestion
              throw ReferenceError(
```

packages/core/src/endowmentsToolkit.js
```diff
@@ -126,14 +141,49 @@ function endowmentsToolkit({
...
       whitelistedReads,
       unwrapTo,
       unwrapFrom,
       explicitlyBanned,
       allowedWriteFields
     )
+
+    if (allowRedefine.length > 0 && unwrapFrom) {
+      const expectedTarget = unwrapFrom || endowments
+      const originalDefineProperty = unwrapFrom.Object.defineProperty
```
boneskull, line 153: the test failures here indicate something is wrong

packages/core/src/endowmentsToolkit.js
```diff
@@ -126,14 +141,49 @@ function endowmentsToolkit({
...
+      /**
+       * @param {any} target
```
```suggestion
       * @param {Record<PropertyKey, unknown>} target
```

or, if TS complains about L168

```suggestion
       * @param {Record<PropertyKey, any>} target
```

packages/core/src/endowmentsToolkit.js
```diff
@@ -126,14 +141,49 @@ function endowmentsToolkit({
...
+          typeof key !== 'symbol' &&
+          allowRedefine.includes(`${key}`)
+        ) {
+          target[key] = descriptor.value
+          return true
+        }
+        originalDefineProperty(target, key, descriptor)
+        return true
+      }
+      endowments.Object = assign(
```
boneskull, line 172: the return type of `makeMinimalViewOfRef` should include `Object?: U` (where `U` is the type of whatever you're assigning to it here). which probably means that the type argument `T` of `makeMinimalViewOfRef` extends `Object?: U`

if you don't want to deal with this (I don't blame you), I can just push a commit addressing the typing issues myself after you're satisfied with everything else.

packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -16,12 +16,12 @@ test('getEndowmentsForConfig', (t) => {
...
     },
   }
-  const config = {
+  const policy = {
     globals: {
       'namespace.stringValue.includes': true,
     },
   }
-  const resultGlobal = getEndowmentsForConfig(sourceGlobal, config)
+  const resultGlobal = getEndowmentsForConfig(sourceGlobal, policy)
```
boneskull, line 24: (suggestion) seems like we should rename `getEndowmentsForConfig` to `getEndowmentsForPolicy`, then alias the former to the latter, and `@deprecate` the former

packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -457,3 +457,83 @@ test('copyWrappedGlobals - copy from prototype too', (t) => {
...
+  const { getEndowmentsForConfig } = prepareTest({ knownWritable })
+  const sourceGlobal = {
+    ...globalThis,
+    x: 1,
+    y: 2,
+    z: 3,
+  }
+  const policy = {
+    globals: {
+      x: 'write+define',
```
boneskull, line 466: I'd recommend replacing global policy value strings with constants.  e.g., add `GLOBAL_POLICY_VALUE_WRITE='write'` and `GLOBAL_POLICY_VALUE_REDEFINE='write+define'` to `core/src/constants.js` then consume them everywhere

packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -457,3 +457,83 @@ test('copyWrappedGlobals - copy from prototype too', (t) => {
 
   t.is(Object.keys(target).sort().join(), 'onTheObj,onTheProto,window')
 })
+
+test('getEndowmentsForConfig - allow redefine on properties', (t) => {
+  'use strict'
```
boneskull: this looks meaningful. is it?

packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -457,3 +457,83 @@ test('copyWrappedGlobals - copy from prototype too', (t) => {
 
   t.is(Object.keys(target).sort().join(), 'onTheObj,onTheProto,window')
 })
+
+test('getEndowmentsForConfig - allow redefine on properties', (t) => {
```
boneskull, line 455: would prefer that we reduce the count of assertions in this test.  e.g. instead of 

```js
t.is(derivedGlobal.x, 1)
t.is(derivedGlobal.y, 2)
t.is(derivedGlobal.z, 3)
```

use

```js
t.like(derivedGlobal, {x: 1, y: 2, z: 3})
```

but splitting into multiple tests would be even better.

also use `t.plan()`

packages/core/src/endowmentsToolkit.js
```diff
@@ -84,6 +86,8 @@ function endowmentsToolkit({
     const allowedWriteFields = new Set()
     /** @type {string[]} */
     const explicitlyBanned = []
+    /** @type {string[]} */
+    const allowRedefine = []
 
     Object.entries(packagePolicy.globals).forEach(
```
boneskull: might as well add `entries` on L26

boneskull (commented):
nothing blocking, but would like to reconsider the value of `write+define` because (IMO) it implies `define` is a valid policy value since `write` is

packages/core/src/endowmentsToolkit.js
naugtur: I was trying to come up with a convention that would avoid `[].includes()` being the way we handle writeable stuff. 
It's literally the same as `write` but with the added allowance to use a monkey-patch on defineProperty

packages/core/src/endowmentsToolkit.js
naugtur: TS is surprised we're getting Object here. The only way to avoid that is to have the whole thing be in a separate function called on the compartment globalThis with the list of fields allowed to be redefined.
This is the thing I didn't figure out here yet. It works and is guaranteed to work whenever unwrapFrom is a global reference of a compartment (which is always) but types are gonna complain.

packages/core/src/endowmentsToolkit.js
naugtur: I'm not yet sure this is the way to go. But I don't see nicer alternatives

packages/core/test/endowmentsToolkit.spec.js
naugtur: Didn't wanna go that far in this PR. Generally agree.

packages/core/test/endowmentsToolkit.spec.js
naugtur: oops, no. this is AI slop from boilerplate it generated :see_no_evil:

packages/core/test/endowmentsToolkit.spec.js
naugtur: lots of assertions is kinda my style, but I have no doubt you're right.

packages/core/src/endowmentsToolkit.js
naugtur: yup. missed that

packages/core/src/endowmentsToolkit.js
boneskull:
```ts
interface SomeGlobalObj {
  Object: typeof Object;
  [x: string]: unknown;
}
```

naugtur:
> nothing blocking, but would like to reconsider the value of `write+define` because (IMO) it implies `define` is a valid policy value since `write` is

Ok, so we have the following concerns:

- it'd be nice to have it named `write.*` because it's literally an extension of write and we can use startsWith instead of [].includes
- there's already been a name prepared for it here https://github.com/LavaMoat/LavaMoat/issues/1583 but the name has some drawbacks (it's not clear what the difference between `write` and `redefine` is. Would a global have to exist prior to being set to be allowed by `redefine`? That's not true, but it sounds like it.)

naugtur:
Looking at it now, I would rather implement it as a custom capability instead of permanently adding it here.

Issue #1134, opened 2024-03-27
aa: flaky symlink test


boneskull:
Here's an excerpt from a failing build:

```
> @lavamoat/aa@4.2.0 test:run
> ava


  ✘ [fail]: symlink-perf › [bench] isSymlink is significantly faster than realpathSync in a naive microbenchmark Rejected promise returned by test
  ✔ index › project 1
  ✔ index › project 2
  ✔ index › project 3
  ✔ index › project 1 - with custom resolver
  ✔ index › project 1 - resolution missing silently
  ✔ index › project 1 - resolution failure
  ✔ index › project 4 - workspace symlink
  ─

  symlink-perf › [bench] isSymlink is significantly faster than realpathSync in a naive microbenchmark

  Rejected promise returned by test. Reason:

  Error {
    code: 'EEXIST',
    dest: '/home/runner/work/LavaMoat/LavaMoat/packages/aa/test/projects/4/node_modules/aaa',
    errno: -17,
    path: '/home/runner/work/LavaMoat/LavaMoat/packages/aa/test/projects/4/packages/aaa',
    syscall: 'symlink',
    message: 'EEXIST: file already exists, symlink \'/home/runner/work/LavaMoat/LavaMoat/packages/aa/test/projects/4/packages/aaa\' -> \'/home/runner/work/LavaMoat/LavaMoat/packages/aa/test/projects/4/node_modules/aaa\'',
  }

  ─

  1 test failed
npm ERR! Lifecycle script `test:run` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: @lavamoat/aa@4.2.0 
npm ERR!   at location: /home/runner/work/LavaMoat/LavaMoat/packages/aa 
npm ERR! Lifecycle script `test` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: @lavamoat/aa@4.2.0 
npm ERR!   at location: /home/runner/work/LavaMoat/LavaMoat/packages/aa 
```

This is likely some sort of race condition; if so, it should be addressable by using `test.serial()` somewhere.

> [!NOTE]
>
> This flake is _not_ the same flake as noted in #1118.

boneskull:
This is still an issue and appears to be cross-platform; here's a recent fail on Windows:

```
> @lavamoat/aa@4.3.1 test:run
> ava


[...]

PR #1593, opened 2025-04-09
chore(node): add support for webpack

boneskull:
This is here because I didn't want to pile more stuff on #1543.

At some point, @naugtur had requested we rename `--root` to `--project-root` so that's happening here.  

**UPDATE Oct 31 2025**: There are many other changes which integrate with endojs/endo#2988.  Thus, CI has no hope of passing until that PR is merged, which is why this is DRAFT.


boneskull:
* **#1593** <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1593?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3  dependent PRs ([#1670](https://github.com/LavaMoat/LavaMoat/pull/1670) <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1670?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1683](https://github.com/LavaMoat/LavaMoat/pull/1683) <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1683?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1713](https://github.com/LavaMoat/LavaMoat/pull/1713) <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1713?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1593?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1782** <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1782?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1033](https://github.com/LavaMoat/LavaMoat/pull/1033) <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1033?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

boneskull:
ref: #1134

boneskull:
~~hm, I didn't mean to stack this on #1670~~

ah, n/m #1670 is stacked on this.

boneskull:
CI will fail on this until endojs/endo#2988 is released and we can update

packages/node/src/cli.js
```diff
@@ -420,48 +412,43 @@ const main = async (args = hideBin(process.argv)) => {
...
             projectRoot,
           })
-        } else {
-          policy = await loadPolicies(policyPath, {
-            policyOverridePath,
-            projectRoot,
-          })
         }
 
         stripProcessArgv(
```
naugtur: stripProcessArgv is incorrect. It takes argv['--'] from yargs and splices it in front of arguments that follow the first `--` , so it should be duplicating it if I'm reading it correctly.

seems like we want to replace process.argv with the items array, not splice it in.

Added a failing test for that.

naugtur:
Instead of reviewing this as diff, I'll make the first pass reviewing current state as a whole. 
Commenting options might be insufficient, but we'll see.

packages/node/src/policy-util.js
naugtur, line 1: Some of the stuff here would be good to have in core so that it benefits all implementations

packages/node/src/exec/exec-compartment-class.js
naugtur, line 60: (must) reuse the one from core

packages/node/src/exec/default-attenuator.js
naugtur: (must) disallow the deprecated type of scuttling definition. If you need a boolean arg in cli, convert it to the new shape before it reaches any further. 
Use the type that doesn't allow it to be boolean. And finally, remove this block.


```suggestion
```

packages/node/src/compartment/node-compartment-map.js
```diff
@@ -0,0 +1,180 @@
...
+          location,
+          /** @type {PackageJson} */ (packageDescriptor)
+        )
+      }
+    },
+    /**
+     * Adds any missing dependencies from `policyOverride` to the list of
+     * dependencies for a package
+     */
+    packageDependenciesHook: ({ canonicalName, dependencies }) => {
```
naugtur: (could) Consider making this hook's existence conditional on policyOverride as an optimization

packages/node/src/exec/load.js
```diff
@@ -7,15 +7,18 @@
...
+import { ExecutionError } from '../error.js'
 import { log as defaultLog } from '../log.js'
-import { toEndoURL } from '../util.js'
+import { reportInvalidCanonicalNames } from '../report.js'
 
 /**
- * @import {ReadNowPowers, LoadLocationOptions} from '@endo/compartment-mapper'
+ * @import {ImportLocationOptions, SyncImportLocationOptions, CanonicalName, PackageCompartmentMapDescriptor} from '@endo/compartment-mapper'
  * @import {ApplicationLoader, ExecuteOptions} from '../types.js'
+ * @import {PackageJson} from 'type-fest'
```
naugtur: unused import?

packages/node/src/report.js
```diff
@@ -0,0 +1,224 @@
...
+ *
+ * @param {Set<CanonicalName>} unknownCanonicalNames - Set of canonical names
+ *   that were referenced in policy but not found
+ * @param {Set<CanonicalName>} knownCanonicalNames - Set of all canonical names
+ *   found in the compartment map
+ * @param {ReportInvalidOverridesOptions} options
+ * @returns {void}
+ * @internal
+ */
+export const reportInvalidCanonicalNames = (
```
naugtur, line 83: (not in this PR) Having this for webpack as well would be great. I'd like to migrate this to core or somewhere shared.

packages/node/src/report.js
```diff
@@ -0,0 +1,224 @@
...
+    })
+    .filter(({ source }) => source !== undefined)
+
+  // if we have any invalid overrides, we will search through the canonical names from the compartment map and make suggestions for the user to fix them
+  if (invalidCanonicalNames.length) {
+    /** @type {Map<string, string[]>} */
+    const suggestions = new Map()
+    for (const { name: invalidName } of invalidCanonicalNames) {
+      const invalidNameParts = invalidName.split('>')
+      const invalidPackageName = invalidNameParts[invalidNameParts.length - 1]
```
naugtur: (nit) the package name is not actually invalid. `misplacedPackageName` maybe?

packages/node/src/report.js
```diff
@@ -0,0 +1,224 @@
...
+    .map((canonicalName) => {
+      const keypath = findCanonicalNameKeypath(policy, canonicalName)
+      return {
+        name: canonicalName,
+        source: keypath || `unknown location for "${canonicalName}"`,
+      }
+    })
+    .filter(({ source }) => source !== undefined)
+
+  // if we have any invalid overrides, we will search through the canonical names from the compartment map and make suggestions for the user to fix them
```
naugtur: clever!

packages/node/src/exec/load.js
```diff
@@ -29,36 +32,73 @@ import { toEndoURL } from '../util.js'
...
     ...DEFAULT_ENDO_OPTIONS,
-    ...options,
+    ...otherOptions,
+    dev,
+    policy: endoPolicy,
     log: log.debug.bind(log),
   }
 
-  const { import: importApp, sha512 } = await loadLocation(
+  const { import: importApp, sha512 } = await loadFromMap(
```
naugtur, line 91: (not for this PR)
We never look at the contents of `packageCompartmentMap`

mapNoceModules might as well return an object with loadFrom map method. 

In a cleaner fashion:

```js
// full chain
makeApp({ subset of loadLocation options })
  .mapNodeModules({existing options})
  .load({same as loadFromMap})
  .import({the existing one})

// all-default mapping
makeApp({ subset of loadLocation options })
  .load({same as loadFromMap})
  .import({the existing one})

// simplest
makeApp({ subset of loadLocation options })
  .import({the existing one})
```

There's room to return more, so mapNodeModules could still return a reference to compartmentMap, but the user is not responsible for passing it around to load.

packages/node/src/types.ts
```diff
@@ -275,27 +282,52 @@ export type CompartmentMapToDebugPolicyOptions = SetFieldType<
...
+  [
+    WithReadPowers,
+    WithIsBuiltin,
+    WithLoadForMapOptions,
+    WithWritePolicyOptions,
+    WithPolicyOverrideOrPath,
+    WithTrustRoot,
+    WithScuttleGlobalThis,
+    WithReadFile,
+    WithProjectRoot,
```
naugtur: (uncertain) 
This contains something like 40 options. There's no good way to read about them all in one place. 
I think generatePolicy options got out of hand. 

Maybe splitting them into groups would help. policy paths, root and scuttling separately from the bag of tools like readPowers and isBuiltin separately from load options etc.

packages/node/src/types.ts
```diff
@@ -275,27 +282,52 @@ export type CompartmentMapToDebugPolicyOptions = SetFieldType<
...
+ *
+ * @remarks
+ * If testing with a virtual filesystem, `projectRoot` _must_ be provided, since
+ * it defaults to `process.cwd()`, which references not-your-filesystem.
+ */
+export type GeneratePolicyOptions = ComposeOptions<
+  [
+    WithReadPowers,
+    WithIsBuiltin,
+    WithLoadForMapOptions,
```
naugtur, line 312: Is his a misspelling of `LoadFromMapOptions`?

packages/node/src/types.ts
```diff
@@ -167,16 +195,25 @@ export interface WithReadPowers {
...
 }
 
+/**
+ * Options for `scuttleGlobalThis`
+ */
 export interface ScuttleGlobalThisOptions {
   enabled?: boolean
   exceptions?: string[]
   scuttlerName?: string
 }
```
naugtur: (must) this is wrong. Import the type from core. exception are `Array<string|RegExp>`

packages/node/src/policy-gen/generate.js
```diff
@@ -5,306 +5,120 @@
...
-    await writePolicy(policyDebugPath, debugPolicy, { fs: writableFs })
-    const nicePolicyDebugPath = hrPath(policyDebugPath)
-    log.info(`Wrote debug policy to ${nicePolicyDebugPath}`)
-    // do not attempt to use the `delete` keyword with typescript. you have been
-    // warned!
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    const { debugInfo: _, ...corePolicy } = /** @type {LavaMoatPolicyDebug} */ (
-      debugPolicy
+  } else if (rawPolicyOverridePath) {
+    log.warning(
```
naugtur: (should) I'd make this an error if input has both

packages/node/src/policy-gen/generate.js
```diff
@@ -5,306 +5,120 @@
...
+    readPowers,
+    policyOverride,
+    projectRoot,
+  })
 
   if (shouldWrite) {
     await writePolicy(policyPath, policy, { fs: writableFs })
-    log.info(`Wrote policy to ${hrPath(policyPath)}`)
+    log.info(
+      `${chalk.greenBright('✓')} ${chalk.bold('Wrote policy')} ${chalk.white('to')} ${hrPath(policyPath)}`
```
naugtur, line 120: (should) Eliminate the dependency on chalk. We should limit our prod dependencies to necessities. Colorful output is not worth it. It'll run in CI half the time anyway.

packages/node/src/policy-gen/load-for-policy.js
```diff
@@ -0,0 +1,463 @@
...
+          if (builtinPolicy) {
+            const currentBuiltinPolicy = builtinsForPackage.get(canonicalName)
+            builtinsForPackage.set(canonicalName, {
+              ...currentBuiltinPolicy,
+              ...builtinPolicy,
+            })
+          }
+          if (violations) {
+            const currentViolations = violationsForPackage.get(canonicalName)
+            if (currentViolations) {
```
naugtur: (should) this function mixes come very distinct responsibilities. It understands messages from workers, displaying progress but also structure of policy and types of violations. 
If we add a new violation, I don't want to go and update this function. I'd rather generalize the aggregation and/or make it happen elsewhere.

packages/node/src/policy-gen/load-for-policy.js
```diff
@@ -0,0 +1,463 @@
...
+  const builtinsForPackage = new Map()
+
+  /**
+   * Stores package policies for each package.
+   *
+   * Updated as we receive messages from the worker pool.
+   *
+   * @type {Map<CanonicalName, PackagePolicy>}
+   */
+  const packagesForPackage = new Map()
```
naugtur: (could) rename to `connectionsForPackage` ?

packages/node/src/report.js
```diff
@@ -0,0 +1,224 @@
...
+  // Create list of invalid canonical names with their keypaths and source representations
+  const invalidCanonicalNames = [...unknownCanonicalNames]
+    .map((canonicalName) => {
+      const keypath = findCanonicalNameKeypath(policy, canonicalName)
+      return {
+        name: canonicalName,
+        source: keypath || `unknown location for "${canonicalName}"`,
+      }
+    })
+    .filter(({ source }) => source !== undefined)
```
naugtur: source can't be undefined if it has a fallback to a string.

packages/node/src/policy-gen/load-for-policy.js
```diff
@@ -0,0 +1,463 @@
...
+       * compatibility violations.
+       */
+      moduleSourceHook: ({ moduleSource, canonicalName }) => {
+        if (canonicalName === ROOT_COMPARTMENT) {
+          return
+        }
+        if ('location' in moduleSource) {
+          return inspectModuleSource(moduleSource, canonicalName)
+        } else if ('error' in moduleSource) {
+          throw new GenerationError(
```
naugtur: Do you know what circumstances this could happen under?

packages/node/src/policy-gen/inspector.js
```diff
@@ -0,0 +1,380 @@
...
+
+  return result
+}
+
+/**
+ * Performs global inspection on source code.
+ *
+ * This function:
+ *
+ * 1. Decodes the bytes into a string using `TextDecoder` (utf-8 encoding).
```
naugtur: that's not true. looks like the comment is stale

packages/node/src/policy-gen/inspector.js
```diff
@@ -0,0 +1,380 @@
...
+  }
+}
+
+/**
+ * @param {ParseResult} ast
+ * @param {string} id - The file path (file:// URL)
+ * @returns {StructuredViolationsResult | null}
+ */
+const inspectViolations = (ast, id) => {
+  const compatWarnings = inspectSesCompat(/** @type {any} */ (ast))
```
naugtur: the type casting here should not be needed. This being necessary suggests we might be depending on a different version of babel

packages/node/src/policy-gen/inspector.js
```diff
@@ -0,0 +1,380 @@
...
+ * 1. Converts `globalMap` to a `Record<string, GlobalPolicyValue>`
+ * 2. Replaces any `GlobalPolicyValue` which is string literal `read` with `true`
+ * 3. Handles the current inability for LavaMoat to work with writable nested
+ *    properties. Finds any entry in the map in which the key contains a `.`
+ *    (which is _not_ the last character) _and_ the value is `write`, then:
+ *
+ *    - Splits the key on the first `.` (this array must have length >= 2)
+ *    - Creates a new entry choosing the head element of the array as the new key and
+ *         the value as `true`
+ *    - Deletes the original entry
```
naugtur, line 118: (must) 
Why would htere be a nested property marked as writable? 

`"someproperty":true` means you get a reference to someproperty and `someproperty.nested` is writeable without asking (this gets complicated with global getters and functions)

`"someproperty": "write"` means you can assign to it and it will propagate to other compartments.

AFAIR tofu should not return nested properties for writable. If it does, collapsing them to the top property doesn't seem to make sense. 

We should look at an example of this and fugure out why this is here.

packages/node/src/policy-gen/load-for-policy.js
```diff
@@ -0,0 +1,463 @@
...
+  violationsForPackage,
+  pendingInspections
+) => {
+  let messageCount = 0
+
+  /**
+   * @param {Omit<
+   *   LocalModuleSource,
+   *   'location' | 'sourceLocation' | 'sourceDirname' | 'parser' | 'record'
+   * > & { language: Language; location: FileUrlString }} moduleSource
```
naugtur: Is this an artifact of waiting for latest Endo, or does the type really differ frmo the value we're getting here? I f so, why is it rewired here and not defined somewhere as a new type for our internal use? I'd like to get the origin story of what's going on here.

packages/node/src/policy-gen/generate.js
```diff
@@ -5,306 +5,120 @@
...
-    msg += invalidOverrides
-      .map((invalidOverride) => `  - ${hrLabel(invalidOverride)}`)
-      .join('\n')
-    log.warning(msg)
-  }
-}
 
 /**
  * Generates a LavaMoat policy or debug policy from a given entry point using
  * `@endo/compartment-mapper`
```
naugtur, line 30: (nit) I don't think the description is true. we no longer do debug policy

naugtur:
###  module-transforms.js issues

1.
`evadeGibson`is not what we want to do.
There's no need for us to include the whitespace between `--` and `>`, no minification is expected to occur after this transform. It's all we actually need AFAIR to do `.replace(/-->/,'-- >'))`

Also, `evadeGibson('if (item[1]--> b)') ,` does not evade.

2.
`...import()` is matching SES censorship
This would fail:
```
assert.equal(evadeImportString('"some string...import();"'), '"some string...import（);"')
```
And I don't think this is overall a good idea. It is possible for strings to contain code that ens up passed to new Function or something and we generally support that.

Classic lavamoat-node was adding underscores to all occurences of `import` in cjs modules (in case something was named import before import was a reserved word AFAIR - to keep it working thanks to consistent renaming)

Last but not least - creating an internal mechanism that takes in matching braces and turns them into a single closing brace is a potential vulnerability.

naugtur:
### power.js

if a custom fs implementation is passed that has a distinct filesystem emulated (like a memFS that's not emulating the same OS we're running in) I think it's possible that the `url` module functionality that Endo uses will produce incorrect results. i recall some of the path processing stuff was relying on information from the OS.

packages/node/src/error.js
```diff
@@ -92,9 +84,12 @@ export const InvalidPolicyError = createLavaMoatError(
...
-  'NoExecutableError',
-  ErrorCodes.NoExecutable,
+/**
+ * Error thrown if no bin script entrypoint could found
+ */
+export const NoBinScriptError = createLavaMoatError(
+  'NoBinScriptError',
+  ErrorCodes.NoBinScript,
   class extends Error {}
 )
```
naugtur, line 95: seems the file has not been formatted

packages/node/src/policy-util.js
```diff
@@ -11,21 +11,38 @@
 import { jsonStringifySortedPolicy, mergePolicy } from 'lavamoat-core'
```
naugtur, line 12: `TODO: Capabilities are inconsistent`
Care to elaborate?

packages/node/src/policy-util.js
```diff
@@ -11,21 +11,38 @@
...
 import nodeFs from 'node:fs'
 import nodePath from 'node:path'
+import { toKeypath } from 'to-keypath'
```
naugtur: nice one, but let's vendor it. No need to have a prod dependency and the risk is always there.

packages/node/src/policy-gen/generate.js
boneskull: I'm gonna push back on the colors, but we can just vendor some ANSI codes.

packages/node/src/policy-gen/inspector.js
boneskull: > Why would htere be a nested property marked as writable?

Because this is what tofu does when it finds a global nested property which is written to.  I was under the impression—at least historically—that tofu was correct and we simply didn't support this in the kernel; I'm pretty sure we have code elsewhere that works around the problem.  I don't think I should be changing this behavior in tofu as part of the PR, though.

packages/node/src/policy-gen/inspector.js
boneskull: `inspectSesCompat` mutates the `error` property of the AST in such a way that the AST is no longer structurally equivalent to a `ParseResult`.  This is bad, but that's what it does.  There's likely something in `lavamoat-node` that's relying on this behavior, so I'm loathe to try to fix it.

packages/node/src/policy-gen/load-for-policy.js
boneskull: `packagePoliciesMap` seemed more appropriate to me, so there it is.

packages/node/src/policy-gen/load-for-policy.js
boneskull: It can't; I think I missed this after some refactoring.

packages/node/src/policy-gen/load-for-policy.js
boneskull: It's the `ModuleSourceForModuleSourceHook` type, which didn't exist at one point.  Now it does and I can use it.

packages/node/src/policy-util.js
boneskull: Some of the functions accept readPowers "capabilities", and others don't. I'm pretty sure I'm using them somewhere in the tests to read data from JSON fixtures, at any rate.  Shall I elaborate more in the comment?

packages/node/src/types.ts
boneskull: Ah, I think this was leftover from before the type existed.

- [x] do this

packages/node/src/types.ts
boneskull: No, it's a type which takes most of endo's `LoadForMapOptions` and throws out the stuff we're going to explicitly provide.

packages/node/src/types.ts
boneskull: I'll see if Claude can help me with this one 😜 

- [x] do this

packages/node/src/exec/load.js
boneskull: IDK man, I'd rather chew broken glass than implement a fluent API.  And I know I'm  not as offended by the current API as yourself (though I agree it could be streamlined, I could live in bliss never touching it again).  But we should brainstorm a bit more, I think.

packages/node/src/exec/exec-compartment-class.js
boneskull: You just added this, right?

boneskull:
> module-transforms.js issues

We'll likely need to find alternatives for these.

The Gibson was one that distinctly remember encountering when trying to get webpack working.  Maybe it could be better, but it needs to happen in some fashion.

The second one was encountered in the TypeScript source (within a string). While you have mentioned `tsc` is not really important to support (at least as an entry file), certain build tools that _consume_ `tsc` _are_.  So we need to figure out some sort of replacement for this, as well.

boneskull:
> power.js

I'm content to sit on it until we end up with some sort of bug report.

packages/node/src/policy-gen/load-for-policy.js
boneskull: By "elsewhere" you mean in the parent process somewhere?  Or "elsewhere" in some separate function?  I agree it could be split up.

naugtur:
> > module-transforms.js issues
> 
> The Gibson was one that distinctly remember encountering when trying to get webpack working. Maybe it could be better, but it needs to happen in some fashion.

That's ok. We do need it. We used to have a simpler version of it somewhere that would turn `-->` into `-- >` and that should suffice. I'm asking if there's a reason it doesn't


> The second one was encountered in the TypeScript source (within a string). While you have mentioned `tsc` is not really important to support (at least as an entry file), certain build tools that _consume_ `tsc` _are_. So we need to figure out some sort of replacement for this, as well.

I'm asking why don't we use the same transform as lavamoat-node was using for this?

packages/node/src/policy-util.js
naugtur: Since it's a TODO, elaborating a little in the comment so that the intention is clear to someone else would make it easier to address the todo by someone who's not in the state of mind to know what's to be done from just looking. (eg. future self :P )

packages/node/src/policy-gen/inspector.js
naugtur: We need to discuss and explore, this TOFU behavior was earlier masked by policy merging collapsing nested fields (sometimes leading to incorrect results) 

```
"sth": true,
"sth.more": write
```
Would end up collapsed to just 
```
"sth": true
```

packages/node/src/policy-gen/inspector.js
naugtur: If you're sure it's not just types getting mixed up, let's add an issue for fixing it and move on.

packages/node/src/policy-gen/load-for-policy.js
naugtur: I mean splitting the code into separate functions/files to keep the logic and the workers code not entangled (as in: if you wanted to rip out workers, what structure of the code would make it the easiest to do without risking introducing bugs in logic?)

packages/node/src/policy-gen/generate.js
naugtur: I'm indifferent about colors, personally prefer ascii art or shapes. Just saying it's not worth having a dependency for it if that dependency introduces risk of malware.
vendoring ANSI codes as functions with template strings (so a subset of chalk I guess, but maybe lacking some compatibility fanciness) sounds better.
(I know we can't get rid of chalk from dependencies entirely yet, but in preparation for that)

packages/node/src/exec/exec-compartment-class.js
naugtur: The function was there for a long time, I added a fix to it for one particularly unexpected corner case where constructors have a prototype chain.

packages/node/src/policy-gen/generate.js
boneskull: Just to note here, a switch from `loggerr` to `@lavamoat/vog` will eliminate the dependency on `chalk`.

packages/node/src/exec/default-attenuator.js
boneskull: - [ ] do this

packages/node/src/cli.js
boneskull: - [x] do this

packages/node/src/policy-gen/inspector.js
boneskull: Created https://github.com/LavaMoat/LavaMoat/issues/1874

packages/node/src/policy-gen/inspector.js
boneskull: Right, we did that as a crappy alternative to actually implementing the behavior in the kernel so that it'd work.  ¯\\\_(ツ)\_/¯

boneskull:
@naugtur 

> I'm asking why don't we use the same transform as lavamoat-node was using for this?

I'm willing to chalk that up to ignorance.  I'll look into it.

- [x] do this

packages/node/src/policy-gen/load-for-policy.js
boneskull: OK, I've split this up into like four functions and moved the reporting business into `report.js`.

packages/node/src/types.ts
boneskull: Well, wait a minute. There is a place to view most of them all at once:

image

But we can't view ALL of them at once, because there's just a ton of them (like even 50), and the vast majority are passed thru to Endo's `loadFromMap()`.

- I think we can make this a little less awful by _nesting_ the Endo-specific options.
- I can also try to logically group these options by identifying which always appear together.

boneskull: Nesting the options is proving difficult.  Will continue plugging at it though, since I think it makes the most sense.  This would add `loadForMapOptions` & `mapNodeModulesOptions` fields to `ExecuteOptions` and a `mapNodeModulesOptions` field to `GenerateOptions`.

packages/node/src/exec/exec-compartment-class.js
naugtur: fixed

naugtur:
@boneskull 
I've made some progress on the remaining bits. 
evasive transport for imports needs more work (or moving to an AST based processing in Endo)

packages/node/src/types.ts
boneskull: I did some consolidation but I think I'd like to work on the options further in a subsequent PR.

boneskull:
@naugtur It looks like you fixed the evasions.  I think I've resolved everything I was going to resolve.  Unless you have more changes, I think we're just waiting on an Endo release.

naugtur:
> @naugtur It looks like you fixed the evasions. I think I've resolved everything I was going to resolve. Unless you have more changes, I think we're just waiting on an Endo release.

I fixed the comment evasion (we don't need the complex one) but as stated above, the import evasion fails some of the tests and I don't think we can get it to properly work without going AST.
Old evasion was easy because we knew we don't support dynamic import. Now that we do, we have to leave it alone and that's rough. 
With strings, we might be forced to do `"import("` -> `"imp"+"ort("` 
For comments, I think we'd be better off erasing them (there's a sourcemap-preserving option in endo to do that)

edit:  
https://github.com/endojs/endo/pull/3026

naugtur:

- branch: `main`
  - **[chore(node): add support for webpack #1593](https://github.com/LavaMoat/LavaMoat/pull/1593)** - branch: `boneskull/node-refactor` *(this PR)*
    - branch: `naugtur/enter-evasive-transforms`

---

packages/node/src/compartment/import-hook.js
naugtur, line 51: we must check with isBuiltin here and error out otherwise, unless we can guarantee isBuiltin is called in all paths leading to this (which I don't think is the case)

naugtur (approved):
renewing the approval

packages/node/src/compartment/import-hook.js
boneskull: So the thing is builtins go thru the exit module, but so do things like JSON files and native modules.  We can do a naive check of the filename at this point (`.json`/`.node`) as well as checking `isBuiltin()`, but that's about all we can do.

I note that Endo actually calls the exit module import hook _before_ checking policy (and I still don't really understand why) which means if we add this check we're going to hit it even before we report invalid canonical names, and invalid canonical names will _always_ be considered exit modules.

https://github.com/endojs/endo/blob/9463241e6014366ba9956cf01bd4299119a78b9d/packages/compartment-mapper/src/import-hook.js#L813-827

So I think we need to discuss further and address this post-merge.

naugtur: AFAIR the problem was we need the error that it's missing reported before it complain about policy. The memory is faded, but I remember eing frustrated that I have to call it before checking the policy for some specific reason I couldn't work around at the time.

we have parsers defined for endo for .json and .node - why are they going through exit module import hook?

We definitely need to talk about it more.

boneskull: Hmm, right. We do have parsers. 

I think I want to experiment with that code in Endo. Because IMO it's _really_ bad behavior the way it works now.

PR #1603, opened 2025-04-17
[poc] - support scuttling endo-node

weizman:
goes with https://github.com/endojs/endo/pull/2780


weizman:
current state is that scuttling scenarios are reenabled for endo-node, also for the testing infra specifically, the scuttleGlobalThis options are being passed into the scenarios, but there's a list of globals that if are not exempted break tests and it's hard to tell who's relying on them (endo-node code, end-node deps or the testing infra - most likely all three)

weizman:
ditched in favor of https://github.com/LavaMoat/LavaMoat/pull/1627

PR #1623, opened 2025-04-26
prevent exploding on install

naugtur:
when a package manager decides to install a peer dependency, allow-scripts becomes impossible to install successfully. the always-fail can only be installed _AFTER_ `allow-scripts setup` was called to configure ignoring scripts.


boneskull:
how long has this been broken?  Does this change how users are to install it?

naugtur:
> how long has this been broken? Does this change how users are to install it?

We went back and forth on it a bunch
https://github.com/LavaMoat/LavaMoat/commit/49bef05117fc579f43682d4dc5fc826c97955696
https://github.com/LavaMoat/LavaMoat/commit/c2c3dc122d4526c6c7bca89d8637b78ad628e8a1

The opinions of package managers were also going back and forth. I don't see a point in indicating it as a peer dependency as there's possibly severe disadvantages to having it and the only advantage is it being indicated as a warning when missing by some versions of some package managers.

boneskull (approved):

PR #1627, opened 2025-04-29
feat(node): add support in scuttling

weizman:
add support for scuttling in the new node (endo) pkg (awaits https://github.com/LavaMoat/LavaMoat/pull/1645 & goes with https://github.com/endojs/endo/pull/2780)

### DIFF
--- a/packages/core/src/scuttle.d.ts
+++ b/packages/core/src/scuttle.d.ts
@@ -1,0 +1,22 @@
+/**
+ * Options for scuttling global properties
+ */
+export interface ScuttleOpts {
+  enabled: boolean;
+  exceptions?: Array<string | RegExp>;
+  scuttlerName?: string;
+}
+
+/**
+ * Reference to the global object
+ */
+export interface GlobalRef {
+  globalThis?: Record<PropertyKey, unknown>;
+}
+
+/**
+ * Scuttles (disables or restricts) certain global properties
+ * @param globalRef The global object to scuttle
+ * @param [opts] Options for scuttling
+ */
+export function scuttle(globalRef: GlobalRef, opts?: ScuttleOpts | boolean): void;
--- a/packages/node/src/cli.js
+++ b/packages/node/src/cli.js
@@ -397,6 +397,12 @@ const main = async (args = hideBin(process.argv)) => {
               implies: 'generate-recklessly',
               hidden: true,
             },
+            scuttle: {
+              type: 'boolean',
+              describe: 'Enable scuttling of globalThis',
+              group: BEHAVIOR_GROUP,
+              coerce: Boolean,
+            },
           })
           /**
            * Resolve entrypoint from `root`
@@ -420,6 +426,7 @@ const main = async (args = hideBin(process.argv)) => {
           'policy-override': policyOverridePath,
           dev,
           root: projectRoot,
+          scuttle: scuttleGlobalThis,
           write,
         } = argv
 
@@ -471,6 +478,7 @@ const main = async (args = hideBin(process.argv)) => {
           trustRoot,
           dev,
           projectRoot,
+          scuttleGlobalThis,
         })
       }
     )
--- a/packages/node/src/exec/default-attenuator.js
+++ b/packages/node/src/exec/default-attenuator.js
@@ -6,6 +6,7 @@
  */
 
 import { endowmentsToolkit } from 'lavamoat-core'
+import { scuttle } from 'lavamoat-core/src/scuttle.js'
 import {
   ENDO_POLICY_ITEM_ROOT,
   GLOBAL_THIS_REFS,
@@ -14,6 +15,7 @@ import {
 import { isObjectyObject } from '../util.js'
 
 /**
+ * @import {MakeGlobalsAttenuatorOptions} from '../types.js'
  * @import {GlobalAttenuatorFn, ModuleAttenuatorFn} from '@endo/compartment-mapper'
  * @import {GlobalAttenuatorParams} from '../types.js'
  * @import {LavaMoatPolicy} from 'lavamoat-core'
@@ -45,14 +47,14 @@ export const attenuateModule = (params, originalObject) => {
  *
  * **REMEMBER: The attenuator is _not applied_ to packages without policy!**
  *
- * @param {object} options
- * @param {Partial<LavaMoatPolicy>} [options.policy]
+ * @param {MakeGlobalsAttenuatorOptions} [options]
  * @returns {GlobalAttenuatorFn<GlobalAttenuatorParams>}
  * @internal
  */
-export const makeGlobalsAttenuator = (
-  { policy: { resources } = {} } = { policy: {} }
-) => {
+export const makeGlobalsAttenuator = ({
+  policy: { resources } = {},
+  scuttleGlobalThis = { enabled: false },
+} = {}) => {
   /** @type {Set<string>} */
   const knownWritableFields = new Set()
 
@@ -67,6 +69,10 @@ export const makeGlobalsAttenuator = (
         }
       }
     }
+  }
+
+  if (typeof scuttleGlobalThis === 'boolean') {
+    scuttleGlobalThis = { enabled: scuttleGlobalThis }
   }
 
   const { getEndowmentsForConfig, copyWrappedGlobals } = endowmentsToolkit({
@@ -104,6 +110,10 @@ export const makeGlobalsAttenuator = (
         'globalThis',
         'global',
       ])
+      scuttle(originalGlobalThis, {
+        ...scuttleGlobalThis,
+        enabled: !!scuttleGlobalThis.enabled,
+      })
     } else {
       if (!rootCompartmentGlobalThis) {
         rootCompartmentGlobalThis = new Compartment().globalThis
@@ -111,6 +121,10 @@ export const makeGlobalsAttenuator = (
           'globalThis',
           'global',
         ])
+        scuttle(originalGlobalThis, {
+          ...scuttleGlobalThis,
+          enabled: !!scuttleGlobalThis.enabled,
+        })
       }
       const endowments = getEndowmentsForConfig(
         rootCompartmentGlobalThis,
--- a/packages/node/src/types.ts
+++ b/packages/node/src/types.ts
@@ -4,7 +4,7 @@
  *
  * @packageDocumentation
  */
-import type { IsBuiltinFn, LavaMoatPolicy } from 'lavamoat-core'
+import type { IsBuiltinFn, LavaMoatPolicy, Resources } from 'lavamoat-core'
 import type { Loggerr } from 'loggerr'
 import type nodeFs from 'node:fs'
 import type { SetFieldType, Simplify } from 'type-fest'
@@ -170,6 +170,16 @@ export interface WithTrustRoot {
   trustRoot?: boolean
 }
 
+export interface ScuttleGlobalThisOptions {
+  enabled?: boolean
+  exceptions?: string[]
+  scuttlerName?: string
+}
+
+export interface WithScuttleGlobalThis {
+  scuttleGlobalThis?: boolean | ScuttleGlobalThisOptions
+}
+
 /**
  * Superset of {@link FsInterface} necessary to write a policy to disk
  */
@@ -281,6 +291,7 @@ export type GeneratePolicyOptions = Simplify<
     CompartmentMapToPolicyOptions &
     WritePolicyOptions &
     WithTrustRoot &
+    WithScuttleGlobalThis &
     WithPolicyOverridePath &
     WithReadFile &
     WithProjectRoot
@@ -356,6 +367,7 @@ export type RunOptions = Simplify<
   WithRawReadPowers &
     WithDev &
     WithTrustRoot &
+    WithScuttleGlobalThis &
     WithLog &
     WithProjectRoot &
     WithPolicyOverrideOrPath &
@@ -455,3 +467,11 @@ export interface FsUtilInterface {
 export interface WithFs {
   fs?: FsUtilInterface
 }
+
+export interface WithPolicy {
+  policy?: Partial<LavaMoatPolicy<Resources>>
+}
+
+export type MakeGlobalsAttenuatorOptions = Simplify<
+  WithPolicy & WithScuttleGlobalThis
+>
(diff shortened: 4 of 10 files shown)


weizman:
This is a working state of supporting scuttling via the new node pkg including tests.
* [ ] cli should be tested to actually work
* [x] types are a mess, must be a better way to do this
* [x] added hardcoded support in scuttling exceptions that babel needs (which this pkg relies on)
  * [x] this needs to happen for the old node pkg too
* [x] enabled the core scuttling tests and made sure they work
* [x] cached globals in places we have affect on to avoid clashing with scuttling
* [ ] the invokation of scuttling within the `packages/node/src/exec/default-attenuator.js` needs a refactor

packages/core/src/scuttle.d.ts
```diff
@@ -0,0 +1,22 @@
...
+  enabled: boolean;
+  exceptions?: Array<string | RegExp>;
+  scuttlerName?: string;
+}
+
+/**
+ * Reference to the global object
+ */
+export interface GlobalRef {
+  globalThis?: Record<string, unknown>;
```
boneskull: This is probably fine for our purposes (because it should a) allow us to remove properties and b) discourage us from modifying properties otherwise due to type `unknown`), but it begs the question: what happens in scuttling if we have a global symbol? [graceful-fs](https://npm.im/graceful-fs) does this.  or a numeric key, for that matter?

packages/core/src/scuttle.d.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/**
+ * Scuttles (disables or restricts) certain global properties
+ * @param globalRef The global object to scuttle
+ * @param opts Options for scuttling
+ */
+export function scuttle(globalRef: GlobalRef, opts: ScuttleOpts | boolean): void;
```
boneskull: Is `opts` required?

packages/node/src/cli.js
```diff
@@ -285,6 +285,13 @@ const main = async (args = hideBin(process.argv)) => {
         global: true,
         group: BEHAVIOR_GROUP,
       },
+      'scuttle-global-this': {
```
boneskull: I'd prefer just `scuttle`, but feel free to use an alias if you disagree 😄

packages/node/src/cli.js
```diff
@@ -285,6 +285,13 @@ const main = async (args = hideBin(process.argv)) => {
...
+        describe: 'Enable scuttling of globalThis',
+        type: 'boolean',
+        global: true,
```
boneskull: I don't think this is correct. Scuttling only happens at runtime, therefore this option should be defined in the builder for the `run` command (and not have `global: true`)

packages/node/src/cli.js
```diff
@@ -285,6 +285,13 @@ const main = async (args = hideBin(process.argv)) => {
...
+        group: BEHAVIOR_GROUP,
+        coerce: (value) => value === true ? {enabled: true} : value,
```
boneskull: Would prefer to just keep this as a boolean for the sake of the CLI and we can coerce it later

```suggestion
        coerce: Boolean,
```

packages/node/src/exec/default-attenuator.js
```diff
@@ -6,6 +6,7 @@
  */
 
 import { endowmentsToolkit } from 'lavamoat-core'
+import { scuttle } from "lavamoat-core/src/scuttle.js"
```
boneskull: This should be exported from the root e.g.

```js
import { scuttle } from 'lavamoat-core'
```

Ideally this is a subpath export such as `lavamoat-core/scuttle` but adding an `exports` field is often fraught with peril and must be done carefully

packages/node/src/exec/default-attenuator.js
```diff
@@ -45,13 +46,12 @@ export const attenuateModule = (params, originalObject) => {
  *
  * **REMEMBER: The attenuator is _not applied_ to packages without policy!**
  *
- * @param {object} options
- * @param {Partial<LavaMoatPolicy>} [options.policy]
+ * @param {import('../types.js').MakeGlobalsAttenuatorOptions} [options]
```
boneskull: Use 

```js
/**
 * @import {MakeGlobalsAttenuatorOptions} from '../types.js'
 */
```

near the top of the file and just reference `MakeGlobalsAttenuatorOptions` here.

packages/node/src/exec/default-attenuator.js
```diff
@@ -69,6 +69,16 @@ export const makeGlobalsAttenuator = (
     }
   }
 
+  if (typeof scuttleGlobalThis === 'boolean') {
+    scuttleGlobalThis = {enabled: scuttleGlobalThis}
+  }
+  if (scuttleGlobalThis?.enabled) {
+    if (!Array.isArray(scuttleGlobalThis?.exceptions)) {
```
boneskull: probably need to destructure `isArray`

packages/node/src/exec/default-attenuator.js
```diff
@@ -69,6 +69,16 @@ export const makeGlobalsAttenuator = (
...
+      scuttleGlobalThis.exceptions = []
+    }
+    scuttleGlobalThis.exceptions.push('Map', 'parseFloat', 'WeakMap', 'WeakSet', 'Boolean', 'Number', 'Promise', 'String', 'isNaN')
```
boneskull: pls store this list in `constants.js` with an explanation of what it is and why it exists, e.g.

```js
/**
 * describe me
 */
export const SCUTTLE_EXCEPTIONS = Object.freeze(/** @type {const} */(
  'Map', 'parseFloat', 'WeakMap', 'WeakSet', 'Boolean', 'Number', 'Promise',
  'String', 'isNaN'))
```

packages/node/src/types.ts
```diff
@@ -170,6 +170,16 @@ export interface WithTrustRoot {
   trustRoot?: boolean
 }
 
+export interface ScuttleGlobalThis {
```
boneskull:
```suggestion
export interface ScuttleGlobalThisOptions {
```

I realize that this demands changes to the reference so is not easily just done thru the UI

packages/node/src/types.ts
```diff
@@ -455,3 +467,9 @@ export interface FsUtilInterface {
 export interface WithFs {
   fs?: FsUtilInterface
 }
+
+export interface WithPolicy {
+  policy?: Partial<LavaMoatPolicy<Resources>>
```
boneskull, line 472:
```suggestion
  policy?: LavaMoatPolicy
```

The options for `makeGlobalsAttenuator` as in current `main` are kind of old and busted.

Also: `makeGlobalsAttenuator` is not a public API, so this stuff should live in `internal.ts`.  `internal.ts` can pull from `types.ts`, but not vice-versa.

packages/node/src/types.ts
```diff
@@ -455,3 +467,9 @@ export interface FsUtilInterface {
...
+}
+
+export type MakeGlobalsAttenuatorOptions = WithPolicy & WithScuttleGlobalThis
```
boneskull:
```suggestion
export type MakeGlobalsAttenuatorOptions = Simplify<WithPolicy & WithScuttleGlobalThis>
```

packages/node/test/unit/scenario-util.js
```diff
@@ -210,7 +210,7 @@ export function createScenarioRunner(log = fallbackLog.error.bind(console)) {
       const worker = new Worker(RUNNER_MODULE_PATH, {
         stdout: true,
         stderr: true,
-        workerData: { entryPath, policy: lavamoatPolicy, vol: vol.toJSON() },
+        workerData: { scuttleGlobalThis:scenario?.opts?.scuttleGlobalThis, entryPath, policy: lavamoatPolicy, vol: vol.toJSON() },
```
boneskull: this should have been prettified

packages/node/src/cli.js
boneskull: (somewhere around https://github.com/LavaMoat/LavaMoat/blob/b4c9c4e981637dc007df28acb2da49c3e6680388/packages/node/src/cli.js#L368)

packages/core/src/scuttle.d.ts
weizman: We can def stumble upon Symbol/Numeric keys, and from the perspective of scuttling that changes nothing (updated 6553eeb4e)

packages/core/src/scuttle.d.ts
weizman: no, fixed d4fd0d38e

packages/node/src/cli.js
weizman: fixed aa97a196e

packages/node/src/cli.js
weizman: fixed 937e97f3b

packages/node/src/cli.js
weizman: fixed 937e97f3b

packages/node/src/exec/default-attenuator.js
weizman: agreed. should we defer to another pr? since this "changes" core pkg? your call

packages/node/src/exec/default-attenuator.js
weizman: fixed db5770c39

packages/node/src/exec/default-attenuator.js
weizman: nice catch fd7fc50f9

packages/node/src/exec/default-attenuator.js
weizman: fixed 93d27f4c7

packages/node/src/cli.js
```diff
@@ -397,6 +397,13 @@ const main = async (args = hideBin(process.argv)) => {
               implies: 'generate-recklessly',
               hidden: true,
             },
+            scuttle: {
+              type: 'boolean',
+              describe: 'Enable scuttling of globalThis',
+              group: BEHAVIOR_GROUP,
+              hidden: true,
```
naugtur: I don't think we want it hidden. Or am I missing something?

weizman: d2e8927c5

packages/node/src/types.ts
weizman: @boneskull I think I need some clarification on this, how `makeGlobalsAttenuator` being "old and busted" relate to the suggestion of turning `Partial<LavaMoatPolicy<Resources>>` into `LavaMoatPolicy`? And am I supposed to be surprised committing your suggestion as-is fails, or is it obvious and I am missing something?

packages/node/src/types.ts
weizman: like this? 6a560735a

packages/node/src/exec/default-attenuator.js
boneskull: yes, it should be in another PR (I am not asking you to do this). sorry for the confusion

packages/node/src/exec/default-attenuator.js
```diff
@@ -69,6 +73,16 @@ export const makeGlobalsAttenuator = (
...
   }
 
+  if (typeof scuttleGlobalThis === 'boolean') {
+    scuttleGlobalThis = {enabled: scuttleGlobalThis}
+  }
+  if (scuttleGlobalThis?.enabled) {
+    if (!isArray(scuttleGlobalThis?.exceptions)) {
+      scuttleGlobalThis.exceptions = []
+    }
+    scuttleGlobalThis.exceptions.push(...SCUTTLE_EXCEPTIONS)
```
boneskull: are duplicates in the `exceptions` array going to cause any problems?

packages/node/src/exec/default-attenuator.js
```diff
@@ -104,13 +118,15 @@ export const makeGlobalsAttenuator = (
         'globalThis',
         'global',
       ])
+      scuttle(originalGlobalThis, { ...scuttleGlobalThis, enabled: scuttleGlobalThis.enabled ?? false })
```
boneskull:
```suggestion
      scuttle(originalGlobalThis, { ...scuttleGlobalThis, enabled: !!scuttleGlobalThis.enabled })
```

packages/node/src/exec/default-attenuator.js
```diff
@@ -104,13 +118,15 @@ export const makeGlobalsAttenuator = (
...
     } else {
       if (!rootCompartmentGlobalThis) {
         rootCompartmentGlobalThis = new Compartment().globalThis
         copyWrappedGlobals(originalGlobalThis, rootCompartmentGlobalThis, [
           'globalThis',
           'global',
         ])
+        scuttle(originalGlobalThis, { ...scuttleGlobalThis, enabled: scuttleGlobalThis.enabled ?? false })
```
boneskull:
```suggestion
        scuttle(originalGlobalThis, { ...scuttleGlobalThis, enabled: !!scuttleGlobalThis.enabled })
```

packages/node/src/types.ts
boneskull: The point I was trying to make was that:

- We should not copy the types from the docstring (as in the `main` branch), because it is _wrong_. The `policy` option should be a `LavaMoatPolicy`; not a `Partial<LavaMoatPolicy>`.
- The docstring of `makeGlobalsAttenuator` does not follow the convention of defining a type a function's options 

...but you have already fixed this, so there's nothing more for you to do.

I just pulled your branch and didn't see any typechecking failure, fwiw.

(You may need to rebase onto `main` if CI is failing)

boneskull (approved):
No blockers, but please run `prettier` 😄

packages/node/src/exec/default-attenuator.js
weizman: no because the loop attempts a scuttle only if the property is configurable, so the 2nd time it goes through it, it won't be, so it gets skipped

weizman:
Addressed all comments (including prettify) so squashed everything, now I'd like to adjust this pr to absorb #1645

weizman:
absorbed #1645 which allowed us to drop the part in this pkg that sends scuttling exceptions to the core logic

packages/node/src/types.ts
sei358394: @sei358392

PR #1630, opened 2025-04-30
feat(webpack): enable syntax checks by default, check prior to concatenation

naugtur:
This follows advice from the Ottersec audit more closely but also solves for a situation when the plugin is running under lavamoat and `Function` is not in sloppy mode, so won't tolerate with statements. 

If running checks by default is a breaking change for someone, that's likely actually good for their project, so I'm not considering it a breaking change (while in beta)

### DIFF
--- a/packages/webpack/src/buildtime/wrapper.js
+++ b/packages/webpack/src/buildtime/wrapper.js
@@ -81,7 +81,7 @@ exports.wrapper = function wrapper({
     ','
   )}}))${optionalBinding}()`
   if (runChecks) {
-    validateSource(before + sesCompatibleSource + after)
+    validateSource(sesCompatibleSource)
   }
   return {
     before,
--- a/packages/webpack/src/plugin.js
+++ b/packages/webpack/src/plugin.js
@@ -82,6 +82,7 @@ class LavaMoatPlugin {
       policyLocation: path.join('lavamoat', 'webpack'),
       lockdown: lockdownDefaults,
       isBuiltin: () => false,
+      runChecks: true,
       ...options,
     }
 
@@ -252,7 +253,7 @@ class LavaMoatPlugin {
          * @type {import('./buildtime/aa.js').IdentifierLookup}
          */
         let identifierLookup
-        const runChecks = this.options.runChecks || diag.level > 0
+        const runChecks = this.options.runChecks
 
         // Caveat: this might be called before the lookup map is ready if a plugin is running a child compilation or alike.
         // Note that in those cases wrapped code is not meant to run and policy will be empty.
--- a/packages/webpack/test/e2e-hack.spec.js
+++ b/packages/webpack/test/e2e-hack.spec.js
@@ -55,3 +55,20 @@ test('webpack/hack/fetch - cannot pollute prototypes', (t) => {
     { message: /Cannot add property polluted, object is not extensible/ }
   )
 })
+
+test('webpack/hack/mismatched-curlies', async (t) => {
+  t.plan(1)
+  try {
+    t.context.build = await scaffold({
+      ...webpackConfigDefault,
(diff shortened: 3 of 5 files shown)


packages/webpack/src/plugin.js
```diff
@@ -252,7 +253,7 @@ class LavaMoatPlugin {
          * @type {import('./buildtime/aa.js').IdentifierLookup}
          */
         let identifierLookup
-        const runChecks = this.options.runChecks || diag.level > 0
```
naugtur, line 255: This was annoying in practice and would be much more annoying now that you might explicitly set runChecks to false for valid reasons you had.

packages/webpack/src/buildtime/wrapper.js
```diff
@@ -81,7 +81,7 @@ exports.wrapper = function wrapper({
     ','
   )}}))${optionalBinding}()`
   if (runChecks) {
-    validateSource(before + sesCompatibleSource + after)
+    validateSource(sesCompatibleSource)
```
boneskull, line 84: What does this change have to do with the `runChecks` default?

boneskull (commented):
Can we add a test that asserts default behavior?

packages/webpack/src/buildtime/wrapper.js
naugtur: This is the main change here - checking that the source after transformation is valid javascript.
As indicated in the audit too - we benefit more from checking the source before wrapping because it could contain unmatched curlies for example. Previously it was relying on earlier steps requiring correct syntax, which is not guaranteed.

naugtur:
> Can we add a test that asserts default behavior?

I'm not sure what you're requesting here. All tests that invoke bundling are running this and using the default behavior. I could make a test that asserts the check spots a broken script under default settings.

[edit]
Added a test that demonstrates the need for checks

packages/webpack/test/e2e-hack.spec.js
```diff
@@ -55,3 +55,20 @@ test('webpack/hack/fetch - cannot pollute prototypes', (t) => {
...
+    t.context.build = await scaffold({
+      ...webpackConfigDefault,
+      entry: {
+        hack4: './hack4.js',
+      },
+    })
+  } catch (err) {
+    t.assert(
+      err.compilationErrors[0].message.startsWith('E_VALIDATION'),
+      'Expected a validation error from erapper check'
```
leotm, line 71:
```suggestion
      'Expected a validation error from wrapper check'
```

leotm: (referring to `packages/webpack/src/buildtime/wrapper.js`)

packages/webpack/src/buildtime/wrapper.js
leotm: how are we checking the original (before ses-compat transforms) pkg/module source
if we're no longer include the 'before' source here to validate/check?

leotm: nvm 'before' here is just a var which is the first half of the 
<s>`with` quad backflips (would be nice if there was a way we could import [make-evaluate](https://github.com/endojs/endo/blob/master/packages/ses/src/make-evaluate.js#L33) instead)</s>
triple `with` backflips for some scope chaining (to construct an isolated strict fn)

leotm (approved):

leotm:
- https://github.com/LavaMoat/LavaMoat/pull/1665

packages/webpack/test/e2e-hack.spec.js
naugtur: I wouldn't mind if you pushed up a commit fixing that without waiting for anyone else.

PR #1645, opened 2025-05-05
feat(core): relax scuttling by skipping intrinsics

weizman:
As agreed, we plan on relaxing scuttling by avoiding scuttling of properties that are the same as can be found within compartments. This is because scuttling such basic intrinsics makes the protecting environment too limited to function, especially given the fact they can be easily retrieved by creating a new Compartment.

What happens in this PR:

* The `avoidForLavaMoatCompatibility` variable now initializes at the top, and in addition to having 3 hardcoded props LavaMoat needs, it now also dynamically generates all the intrinsics a Compartment gives (by creating one) to add to this "do not scuttle" list.
  * Note that it filters out intrinsics that are not identical to those that live outside of Compartments, because if they are not the same, it means Compartments get a tamed version of them, so since the outer ones are more powerful, it still makes sense to scuttle them
* adjusted test scenarios
* specs needed extra care, because when requiring `scuttle.js`, it crashes for not having access to a Compartment via the globalThis, which is the case under specs environment. To address that, there's a new chunk of code that:
  * Sets a fake Compartment to the globalThis that returns the list of intrinsics
  * Generate the list of intrinsics to skip (the same way `scuttle.js` does) to use for tests
  * require `scuttle.js` (which would now work having Compartment set on the globalThis)
  * restore the state of globalThis by removing the fakeCompartment property from it
  * (this is all wrapped in an anon function to express how atomic this operation must be)

### DIFF
--- a/packages/core/src/scuttle.js
+++ b/packages/core/src/scuttle.js
@@ -1,14 +1,17 @@
 /**
  * @typedef {object} ScuttleOpts
  * @property {boolean} enabled - Whether scuttling is enabled or not.
- * @property {Array<string|RegExp>} exceptions - List of properties to exclude from scuttling.
- * @property {string} scuttlerName - Name of the scuttler function to use which is expected to be found as a
- * property on the global object (e.g. if scuttlerName is 'x', scuttler function is obtained from globalThis['x']).
+ * @property {(string | RegExp)[]} exceptions - List of properties to exclude
+ *   from scuttling.
+ * @property {string} scuttlerName - Name of the scuttler function to use which
+ *   is expected to be found as a property on the global object (e.g. if
+ *   scuttlerName is 'x', scuttler function is obtained from globalThis['x']).
  */
 
 /**
  * @typedef {object} GlobalRef
- * @property {Record<string, any>} [globalThis] - Reference to the global object.
+ * @property {Record<string, any>} [globalThis] - Reference to the global
+ *   object.
  */
 
 const { Object, Array, Error, RegExp, Set, console, Proxy, Reflect } =
@@ -21,6 +24,19 @@ const {
   create,
   defineProperty,
 } = Object
+
+const preserveLanguageIntrinsics = (function (
+  globalThisCompartment,
+  globalThis
+) {
+  return (
+    Object.getOwnPropertyNames(globalThisCompartment)
+      // skip all intrinsics that a Compartment already grants
+      .filter((a) => globalThisCompartment[a] === globalThis[a])
+      // support LM,SES exported APIs and polyfills that LM counts on
+      .concat(['Compartment', 'Error', 'globalThis'])
+  )
+})(new Compartment().globalThis, globalThis)
 
 const { isArray, from } = Array
 
@@ -45,8 +61,9 @@ function generateInvokers(prop) {
 }
 
 /**
- * Applies scuttling, with the default set of options, including using Snow if passed in as scuttlerFunc.
- * Scuttle globalThis right after we used it to create the root package compartment.
+ * Applies scuttling, with the default set of options, including using Snow if
+ * passed in as scuttlerFunc. Scuttle globalThis right after we used it to
+ * create the root package compartment.
  *
  * @param {GlobalRef} globalRef - Reference to the global object.
  * @param {ScuttleOpts} opts - Scuttling options.
@@ -68,7 +85,8 @@ function scuttle(globalRef, opts) {
 
 /**
  * @param {GlobalRef} globalRef - Reference to the global object.
- * @param {ScuttleOpts|boolean} originalOpts - Scuttling options. Accepts `true` for backwards compatibility.
+ * @param {ScuttleOpts | boolean} originalOpts - Scuttling options. Accepts
+ *   `true` for backwards compatibility.
  * @returns {ScuttleOpts} - Final scuttling options.
  */
 function generateScuttleOpts(globalRef, originalOpts = create(null)) {
@@ -101,8 +119,8 @@ function generateScuttleOpts(globalRef, originalOpts = create(null)) {
   return opts
 
   /**
-   * @param {string|RegExp} except - Exception to convert to RegExp.
-   * @returns {string|RegExp} - Converted exception.
+   * @param {string | RegExp} except - Exception to convert to RegExp.
+   * @returns {string | RegExp} - Converted exception.
    */
   function toRE(except) {
     // turn scuttleGlobalThis.exceptions regexes strings to actual regexes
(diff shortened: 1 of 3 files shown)


packages/core/src/scuttle.js
```diff
@@ -22,6 +22,14 @@ const {
   defineProperty,
 } = Object
 
+const avoidForLavaMoatCompatibility = (function(cg, g) {
+  return Object.getOwnPropertyNames(cg)
+    // skip all intrinsics that a Compartment already grants
+    .filter(a => cg[a] === g[a])
+    // support LM,SES exported APIs and polyfills that LM counts on
+    .concat(['Compartment', 'Error', 'globalThis'])
+}(new Compartment().globalThis, globalThis))
```
naugtur: neat. We don't need an API for listing intrinsics after all

packages/core/src/scuttle.js
```diff
@@ -22,6 +22,14 @@ const {
   defineProperty,
 } = Object
 
+const avoidForLavaMoatCompatibility = (function(cg, g) {
```
naugtur: Don't like the name, disagree with what it says. 🫣

I'd call it `preserveLanguageIntrinsics` or something like that

packages/core/src/scuttle.js
naugtur: @boneskull let's also ponder how that relates to us giving Endo a subclassed compartment constructor

packages/core/test/scuttle.spec.js
```diff
@@ -18,9 +17,82 @@ const getPropsGroups = (g) => ({
         Object.getOwnPropertyDescriptor(g, p).writable
     )
     // remove hardcoded props scuttling ignores on purpose
-    .filter((p) => !['Compartment', 'Error', 'globalThis'].includes(p)),
+    .filter((p) => !avoidForLavaMoatCompatibility.includes(p)),
 })
 
+const {scuttle, avoidForLavaMoatCompatibility} = (function(){
+  function fakeCompartment() {
```
naugtur, line 24: Why doesn't this use a real compartment?

naugtur:
Final question I have - do we want to add a setting to affect this behavior / decide how severely to scuttle?

For the record: I don't think we'll need it, just opening the question for thought in case we come up with a reason to have it.

packages/core/src/scuttle.js
weizman: fixed

packages/core/test/scuttle.spec.js
weizman: Currently worried loading SES just for a compartment might be an overkill, plus it's just a test, so whether the list of intrinsics is even accurate or not doesn't matter that much. Perhaps we can revisit in the future

weizman:
perhaps in the future, good q https://github.com/LavaMoat/LavaMoat/pull/1645#issuecomment-2862886796

packages/core/src/scuttle.js
```diff
@@ -22,6 +22,14 @@ const {
   defineProperty,
 } = Object
 
+const preserveLanguageIntrinsics = (function(cg, g) {
+  return Object.getOwnPropertyNames(cg)
+    // skip all intrinsics that a Compartment already grants
+    .filter(a => cg[a] === g[a])
```
boneskull: please use descriptive variable names 😄

packages/core/src/scuttle.js
```diff
@@ -22,6 +22,14 @@ const {
...
+    // support LM,SES exported APIs and polyfills that LM counts on
+    .concat(['Compartment', 'Error', 'globalThis'])
```
boneskull: - what of `lockdown`/`harden`? I'm not clear on how these three global props were chosen.
- these should probably live in `src/constants.js` as a read-only array

packages/core/src/scuttle.js
```diff
@@ -22,6 +22,14 @@ const {
   defineProperty,
 } = Object
 
+const preserveLanguageIntrinsics = (function(cg, g) {
```
boneskull: if `function` is used here intentionally and cannot be substituted by an arrow expression, please explain why in a comment

packages/core/test/scuttle.spec.js
```diff
@@ -18,9 +17,82 @@ const getPropsGroups = (g) => ({
...
+        unescape: globalThis['unescape'],
+        lockdown: globalThis['lockdown'],
+        harden: globalThis['harden'],
+      }
+    }
+  }
+  // set stub to globalThis so requiring scuttle.js works
+  globalThis.Compartment = fakeCompartment
+  // generate props to skip (same code as from scuttle.js)
+  const avoidForLavaMoatCompatibility = (function(cg, g) {
```
boneskull: if this _must_ be the same operation as in `src/scuttle.js`, we should probably just create a function that does this an exports it.  otherwise divergence will cause bugs.

Can all this be removed if we use a real `Compartment`?

packages/core/test/scenarios/scuttle.js
```diff
@@ -45,14 +41,18 @@ module.exports = [
...
         scuttleGlobalThis: {
           enabled: true,
           exceptions: [
             'WebAssembly',
             'process',
-            '/[0-9]+/' /*'Set', 'Reflect', 'Object', 'console', 'Array', 'RegExp', 'Date', 'Math'*/,
+            'console',
+            'Math',
+            'Date',
+            // 'AAA',
```
boneskull, line 55:
```suggestion
```

packages/core/src/scuttle.js
boneskull: @naugtur please explain 😄 

I assume you mean the subclassed `Compartment` for execution instead of policy?

packages/core/test/scuttle.spec.js
boneskull: [opinion] if it is straightforward enough to use a real `Compartment` without awful hacks, please do so

boneskull (approved):
I don't have any blockers here, but there are suggestions / questions / opinions that I hope you can address

boneskull (approved):
I don't have any blockers here, but there are suggestions / questions / opinions that I hope you can address

packages/core/src/scuttle.js
weizman: * They were chosen a while ago for breaking LavaMoat (or other vetted shims of MetaMask) when are scuttled (can be revisited in the future)
* They can't (not easily), since this is a static file that is not being imported dynamically, but being statically injected into output bundles (for browser oriented pkgs)

packages/core/src/scuttle.js
weizman: is there a good argument for using arrow exp over function that I'm missing?

packages/core/test/scuttle.spec.js
weizman: Because `src/scuttle.js` is a static file that is not being imported dynamically, but being statically injected into output bundles (for browser oriented pkgs)

yes it can but https://github.com/LavaMoat/LavaMoat/pull/1645#discussion_r2079662702

packages/core/test/scenarios/scuttle.js
weizman: Leaving this here on purpose so it's clear how the 2 tests are different from each other

packages/core/src/scuttle.js
weizman: fixed

packages/core/test/scuttle.spec.js
naugtur: We did a thought experiment and it was awful.

packages/core/src/scuttle.js
boneskull: The only real argument for this is that if the function won't be called with a context and/or it does not reference a context, then you don't need `function` (similar argument for when to use [method definitions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions)).  

Otherwise it's just a matter of consistent style and ease of future refactors. For example, Mocha always executes callbacks with a context, so it's recommended to use `function` when defining a suite/test/hook callback even if you're not currently using `this` (though plenty disregard this advice, as it's plainly more characters to type 🙂).

Determining whether that's a _good_ argument or not is an exercise left to the reader.

PR #1666, opened 2025-05-22
fix(webpack): undo the forced overwrite of parent and top references

naugtur:
With https://github.com/LavaMoat/LavaMoat/pull/1659 needing more work because there's an expectation that `message.source === window.parent` etc. we need to stop forcefully overwriting parent and top.

In the future we'll provide a more narrow capability for postmessage. There's a few possible designs

### DIFF
--- a/packages/webpack/src/runtime/runtime.js
+++ b/packages/webpack/src/runtime/runtime.js
@@ -160,7 +160,7 @@ const installGlobalsForPolicy = (resourceId, packageCompartmentGlobal) => {
     )
     LAVAMOAT?.scuttling?.scuttle(
       theRealGlobalThis,
-      LAVAMOAT.options.scuttleGlobalThis
+      LAVAMOAT.options?.scuttleGlobalThis
     )
   } else {
     const endowments = getEndowmentsForConfig(
--- a/packages/webpack/src/runtime/runtime.js
+++ b/packages/webpack/src/runtime/runtime.js
@@ -14,11 +14,7 @@ const {
   values,
 } = Object
 
-const {
-  lockdown,
-  Proxy,
-  Math, Date,
-} = globalThis
+const { lockdown, Proxy, Math, Date } = globalThis
 
 const warn = typeof console === 'object' ? console.warn : () => {}
 
@@ -35,9 +31,7 @@ if (LOCKDOWN_ON) {
 }
 
 // harden appears on globalThis only after lockdown is called
-const {
-  harden,
-} = globalThis
+const { harden } = globalThis
 
 const knownWritableFields = new Set()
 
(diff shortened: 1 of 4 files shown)


packages/webpack/src/runtime/runtime.js
```diff
@@ -35,9 +31,7 @@ if (LOCKDOWN_ON) {
 }
 
 // harden appears on globalThis only after lockdown is called
-const {
-  harden,
-} = globalThis
+const { harden } = globalThis
 
 const knownWritableFields = new Set()
```
boneskull, line 36: do we need to pull `Set` out of `globalThis` as well?

packages/webpack/src/runtime/runtime.js
```diff
@@ -164,7 +158,10 @@ const installGlobalsForPolicy = (resourceId, packageCompartmentGlobal) => {
       rootCompartmentGlobalThis,
       globalAliases
     )
-    LAVAMOAT?.scuttling?.scuttle(theRealGlobalThis, LAVAMOAT.options.scuttleGlobalThis)
+    LAVAMOAT?.scuttling?.scuttle(
+      theRealGlobalThis,
+      LAVAMOAT.options.scuttleGlobalThis
```
```suggestion
      LAVAMOAT.options?.scuttleGlobalThis
```

boneskull (approved):
probably needs a test?

packages/webpack/src/runtime/runtime.js
naugtur: We're assuming we're working under lockdown, so theoretically we'd not need anything pulled, but Scuttling forces our hand in some places. The reason I pull harden from globalThis is so that I don't get squiggles for using it globally.

boneskull: should probably fix the root cause of the squiggle.  I can take a look after merging

packages/webpack/test/e2e-custom-globals.spec.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  // Build both versions
+  await t.notThrowsAsync(async () => {
+    t.context.buildAllowed = await scaffold(configAllowed)
+  }, 'Expected the allowed globals build to succeed')
+
+  await t.notThrowsAsync(async () => {
+    t.context.buildRestricted = await scaffold(configRestricted)
+  }, 'Expected the restricted globals build to succeed')
```
boneskull, line 60: just thinking out loud:

I note that `notThrowsAsync`'s value here is in its custom assertion message. Because this is in a hook, AVA does not necessarily expect an assertion (as it would in tests), so it's not needed to satisfy the test runner.  

Given the configuration passed into `scaffold`, it wouldn't have enough context upon failure to tell the user the "restricted" or "allowed" build failed; it could only dump the config bag, and the "type" of build which failed must be inferred by user (which, all things considered, kinda sucks and is better in a debug log message).

This assertion, then, amounts to little more than a `try` / `catch` where you throw a custom error having a `cause` of the original (AVA prints the original error & its stack). It's certainly convenient to use AVA's facilities here, and I don't necessarily see a problem in this case.

My point is that I think I/we/everybody should be intentional about doing this sort of thing; use it as a _cue_ to take a closer look at what information is provided by exceptions thrown by the "unit".

PR #1671, opened 2025-05-24
chore(ci): add a thing that updates coverage thresholds

boneskull:
For workspaces which have a `test:cover` script--presumably which uses `c8`--this automatically updates the `.c8rc` in each workspace if the coverage check passes. `.c8rc` controls the coverage threshold, which must pass on subsequent executions of `c8 --check-coverage`.

It is build as a scheduled or on-demand workflow which only works on the main branch.
I am not super enthusiastic about having commits automatically pushed to `main`. This is just one way to do it.

An alternative would be to automatically update `.c8rc` with the relevant contents of `coverage/coverage-summary.json` as a pre-commit hook; this is maybe a better idea, but we would likely want to write a custom JS script in lieu of `.github/actions/coverage-threshold/action.yml`.

The point is this: _if_ we want to maintain a high level of code coverage, we do not want it to ever drop. Developers will not remember to manually update the thresholds, so it should be automated. This is an experiment.

### DIFF
--- a/.github/actions/coverage-threshold/action.yml
+++ b/.github/actions/coverage-threshold/action.yml
@@ -1,0 +1,43 @@
+name: 'Update Coverage Threshold'
+
+description: 'Updates the coverage thresholds for all packages having a test:cover script'
+
+outputs:
+  updated:
+    description: 'Whether or not the coverage thresholds were updated'
+    value: ${{ steps.check-updated.outputs.updated }}
+
+runs:
+  steps:
+    # test:cover should be using --check-coverage anyway, but this forces the issue;
+    # without it, coverage thresholds could decrease.
+    # we allow failures here because we're not running PR checks.
+    - id: check-coverage
+      run: |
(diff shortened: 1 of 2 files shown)


boneskull:
* **#1671** <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1671?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1671?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

boneskull:
Note that _no_ workspaces currently have `test:cover` script, but I do intend to add one for `@lavamoat/node`.

My personal feelings are that code coverage is a useful metric. While getting to 100% can have diminishing returns, it also _might not_ have diminishing returns because the 1% you are missing could be critical. 

It is _our_ job to make sure that we're not in that situation, but code coverage can act as a cue to remind us to do our jobs. 😄 

There's always the escape-hatch of dropping the threshold, if necessary.  I would also entertain the ideas of:

- not using code coverage thresholds at all
- setting some arbitrary thresholds which do not often change and applying these on a per-workspace or global basis

PR #1686, opened 2025-06-04
chore(webpack): update example project

naugtur:


boneskull (approved):
I added a commit which tells Socket to ignore examples

PR #1687, opened 2025-06-04
refactor(webpack): refactoring for maintainability

naugtur:
The goal is to untangle the control flow and separate concerns better.

TODOs:
- [x] Add a diagram of flow control (separate PR)
- [x] rebase


packages/webpack/test/fixtures/main/webpack.config.js
```diff
@@ -11,7 +11,7 @@ const defaultLmOptions = {
   policyLocation: path.resolve(__dirname, 'policy'),
   readableResourceIds: true,
   runChecks: true,
-  diagnosticsVerbosity: 0,
+  diagnosticsVerbosity: 1,
```
naugtur, line 14: Changed with intention to cover some of the level 1 output in test

naugtur:
* **#1688** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1688?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1687** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1687?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1687?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/webpack/src/buildtime/diagnostics.js
```diff
@@ -17,14 +18,23 @@ module.exports = {
...
     if (level >= verbosity) {
+      let prefix = '\n  >'
+      if (level > 2) {
+        const now = Date.now()
+        if (!startTime) {
+          startTime = now
+        }
+        prefix = `\n  [${now - startTime}] >`
+      }
       // @ts-expect-error - trust me, the method exists
```
leotm, line 36: 😄

packages/webpack/test/e2e-main.spec.js
```diff
@@ -14,6 +14,12 @@ test.before(async (t) => {
 test('webpack/main - dist shape', (t) => {
   // Note: The fixture contains a few cases of resources that are emitted by default. This test is to ensure that the default behavior is not changed for the main app (hence the oddly named html file and the svg file from a dependency are emitted). resources from dependencies are prevented from getting emitted, so the other html resource is not showing up in dist. For the svg to be there a loader had to be added to the webpack config.
   t.snapshot(Object.keys(t.context.build.snapshot))
+  t.log(
+    'Warnings sample: \n',
+    t.context.build.stdout
+      .split('\n')
+      .filter((str) => str.includes('WARNING in LavaMoat'))
```
leotm: nit

```suggestion
      .filter((str) => str.includes('WARNING in LavaMoatPlugin'))
```

packages/webpack/test/fixtures/main/webpack.config.js
leotm: or could keep the default `0` and add another test for diagnosticsVerbosity `1`

leotm:
nb: control flow diagram
- https://github.com/LavaMoat/LavaMoat/pull/1688

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -25,65 +20,61 @@ const POLICY_SNAPSHOT_FILENAME = 'policy-snapshot.json'
  */
 
 module.exports = {
+  stringifyPolicyReliably: jsonStringifySortedPolicy,
+  /**
+   * @param {Object} opts
+   * @param {import('lavamoat-core').LavaMoatPolicy | undefined} opts.policyFromOptions
+   * @param {string} opts.location
+   * @returns {import('lavamoat-core').LavaMoatPolicy}
```
leotm: pushed https://github.com/LavaMoat/LavaMoat/commit/d9a69093f5959a8c2397cd7d7367fc64de7a26ea to avoid repeating `import('lavamoat-core').LavaMoatPolicy` (and in aa.js as well)

packages/webpack/README.md
```diff
@@ -78,6 +78,14 @@ Example: avoid wrapping CSS modules:
 
 See: `examples/webpack.config.js` for a complete example.
 
+### diagnosticsVerbosity
+
+| level | description                                                                                                            |
+| ----- | ---------------------------------------------------------------------------------------------------------------------- |
+| 1     | Prints more detailed warnings from compilation and wrapping of modules.                                                |
+| 2     | Enables verbose error output for all compilation errors and syntax checks. Logs the steps the plugin is going through. |
+| 3+    | Debugging logs for the plugin internals.                                                                               |
```
leotm, line 87: maybe silly question, why is this not just `3`

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -0,0 +1,157 @@
+// @ts-ignore
+const { RUNTIME_KEY } = require('../ENUM.json')
+const diag = require('../buildtime/diagnostics.js')
+const { assembleRuntime } = require('./assemble.js')
+const path = require('node:path')
+
+/** @typedef {import('../buildtime/types').LavaMoatPluginOptions} LavaMoatPluginOptions */
```
leotm:
```suggestion
/** @import {LavaMoatPluginOptions} from '../buildtime/types' */
```

packages/webpack/src/buildtime/modulesData.js
leotm, line 1: pushed https://github.com/LavaMoat/LavaMoat/pull/1687/commits/d9a69093f5959a8c2397cd7d7367fc64de7a26ea to avoid `import('webpack')` 8 times

packages/webpack/README.md
leotm: i see diag.run logic for 1 and 2, then also diag.rawDebug logic ranging from 1-5, so perhaps worth documenting 4 and 5 as well?

packages/webpack/src/buildtime/diagnostics.js
leotm: why are there no docs?

leotm: is it like when CLIs default `--help` doesn't show the hidden gems (flags)?

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -0,0 +1,157 @@
+// @ts-ignore
+const { RUNTIME_KEY } = require('../ENUM.json')
```
leotm: can fix the ts error by adding the file to `webpack/tsconfig.json`
`"include": ["src", "src/ENUM.json"],`
(or could include all json files in the project)

packages/webpack/src/plugin.js
```diff
@@ -50,6 +62,9 @@ class VirtualRuntimeModule extends RuntimeModule {
   }
 }
 
+/** @typedef {import('./buildtime/types').LavaMoatPluginOptions} LavaMoatPluginOptions */
+/** @typedef {import('./buildtime/types').ScuttlerConfig} ScuttlerConfig */
```
leotm:
```suggestion
```

packages/webpack/src/plugin.js
```diff
@@ -146,341 +196,232 @@ class LavaMoatPlugin {
...
+
+            diag.run(1, () => {
+              assertFields(STORE, ['runtimeOptimizedPolicy'])
+              const originalKeys = Object.keys(policyToApply.resources)
+              const optimizedKeys = Object.keys(
+                STORE.runtimeOptimizedPolicy?.resources
+              )
+              const policyKeyDiff = originalKeys.filter(
+                (k) => !optimizedKeys.includes(k)
+              )
```
leotm:
```suggestion
              const optimizedKeysSet = new Set(optimizedKeys)
              const policyKeyDiff = originalKeys.filter(
                (k) => !optimizedKeysSet.has(k)
              )
```

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -95,73 +90,58 @@ module.exports = {
...
+          // connections are a much better source of information than module.dependencies which contain
+          // all imported references separately along with exports and fluff
+          ...Array.from(connections).reduce((acc, dep) => {
+            // @ts-expect-error - bad types?
+            const depSpecifier = dep.resolvedModule.userRequest
+            acc[depSpecifier] = depSpecifier
+            return acc
+          }, /** @type {Record<string, string>} */ ({})),
+        },
+        //ast: module._ast, - would have to translate to babel anyway
```
leotm:
```suggestion
```

(unless it's still worth documenting)

packages/webpack/test/fixtures/main/webpack.config.js
naugtur: That would double the time for the test run - It's a valid thing to do normally, but for verifying one warning message that I care about, redoing the whole identical build would be time wasted.

packages/webpack/src/buildtime/diagnostics.js
naugtur: It's node's own internal tool for logging totally synchronously. (there's some buffering that may or may not happen when you console.log)

packages/webpack/README.md
naugtur: everything above 3 is verbose and messy and should be avoided. Over time I'm hoping to consolidate it more and have it be useful for reporting bugs - so that people can dump their level 4 output and it'd give us the context for figuring out the bug. It''s not there yet.

packages/webpack/src/buildtime/policyGenerator.js
naugtur: nah, the hope to use this is long gone.

packages/webpack/src/plugin.js
leotm: suggest removing since the type is unused in this file

leotm: (ideally should catch unused types with a lint check, similar to @typescript-eslint/no-unused-vars)

packages/webpack/src/buildtime/types.d.ts
```diff
@@ -0,0 +1,28 @@
+import type { LavaMoatPolicy } from 'lavamoat-core'
+import type { LockdownOptions } from 'ses'
+
+export interface ScuttlerObjectConfig {
+  enabled?: boolean
+  exceptions?: string[]
+  scuttlerName?: string
+}
+
+export type ScuttlerConfig = ScuttlerObjectConfig | boolean | undefined
```
leotm, line 10:
```suggestion
```

leotm: see https://github.com/LavaMoat/LavaMoat/pull/1687/files#r2157280757

this is only imported in plugin.js, where it isn't used to type anything at all

(unless this is being saved for later)

leotm: ignore me, this is used in packages/webpack/src/runtime/runtime-namespace.ts so we want to keep it

leotm (approved):
LGTM

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -95,73 +90,57 @@ module.exports = {
...
+        file: module.userRequest,
+        type: isBuiltin(module.userRequest) ? 'builtin' : 'js',
+        packageName,
+        content: module.originalSource()?.source()?.toString(),
+        importMap: {
+          // connections are a much better source of information than module.dependencies which contain
+          // all imported references separately along with exports and fluff
+          ...Array.from(connections).reduce((acc, dep) => {
+            // @ts-expect-error - bad types?
+            const depSpecifier = dep.resolvedModule.userRequest
```
leotm, line 118: perhaps worth noting this as a TODO

made initial attempt to fix this, but results in 

Found 9 errors in 6 files.

Errors  Files
     1  ../aa/src/index.js:5
     1  src/buildtime/aa.js:50
     4  src/buildtime/policyGenerator.js:6
     1  src/buildtime/wrapper.js:1
     1  src/plugin.js:128
     1  src/runtime/runtime-namespace.ts:1

😅

leotm: it was there before this refactor anyhow (so hasn't changed)

naugtur: This is a problem with the poor choice of types exported from webpack. We could create our own, but it'd just create a bunch of false sense of safety. I'd rather rely on types Webpack ship and skip rarely than create and have to keep up to date a new set of them.

packages/webpack/src/buildtime/modulesData.js
```diff
@@ -0,0 +1,242 @@
...
+ * work when reached by runtime policy enforcement. Below is the most reliable
+ * way I've found to date to identify ignored modules.
+ * @param {WebPack.Module} m
+ * @returns {boolean}
+ */
+const isIgnoredModule = (m) => {
+  return Boolean(
+    m.type === JAVASCRIPT_MODULE_TYPE_DYNAMIC &&
+      // @ts-expect-error BAD TYPES
+      m.identifierStr?.startsWith('ignored')
```
leotm, line 26: think i tried logging `m.identifierStr` but was getting `undefined` each time in our tests 🤔

leotm: ah nvm this is covered by our tests in one of our fixtures, which has m.identifierStr:

`ignored|/Users/leo/Documents/GitHub/LavaMoat/packages/webpack/test/fixtures/main/src|crypto`

naugtur: I get a lot of undefined, but then eventually - `ignored|/home/.../LavaMoat/packages/webpack/test/fixtures/main/src|crypto`

packages/webpack/README.md
```diff
@@ -105,7 +113,8 @@ This plugin will skip policy enforcement for such ignored modules.
 
 #### HMR
 
-LavaMoat is not compatible with Hot Module Replacement (HMR). Disable LavaMoat for development builds where HMR is required while keeping it enabled for production builds.
+LavaMoat is not compatible with Hot Module Replacement (HMR) and is highly unlikely to ever be - they're the antithesis of sandboxing. If you wish to use HMR for development, disable LavaMoat in your dev configuration while keeping it enabled for production builds.
```
```suggestion
LavaMoat is not compatible with Hot Module Replacement (HMR) and is highly unlikely to ever be; HMR is the antithesis of sandboxing. If you wish to use HMR for development, disable LavaMoat in your dev configuration while keeping it enabled for production builds.
```

packages/webpack/README.md
```diff
@@ -122,7 +131,7 @@ LavaMoat is not compatible with Hot Module Replacement (HMR). Disable LavaMoat f
 
 - Webpack itself is considered trusted.
 - All plugins can bypass LavaMoat protections intentionally.
-- It's unlikely _but possible_ that a plugin can bypass LavaMoat protections _unintentionally_.
+- It's unlikely _but possible_ that a plugin can bypass LavaMoat protections _unintentionally_. Some of the warnings LavaMoat Plugin reports can help detect that.
```
boneskull, line 134: (which?)

packages/webpack/README.md
```diff
@@ -131,6 +140,8 @@ LavaMoat is not compatible with Hot Module Replacement (HMR). Disable LavaMoat f
 
 Elements of the Webpack runtime (e.g., `__webpack_require__.*`) are currently mostly left intact. To avoid opening up potential bypasses, some functionality of the Webpack runtime is not available.
 
+Module Federation depends on extending runtime in ways we would not expose to modules. We're open to contributions or collaboration on carefully enabling it.
```
boneskull, line 143: Maybe this should be in a "notes" or "future" section.

packages/webpack/src/buildtime/diagnostics.js
boneskull: AFAIR `console.error` logs synchronously. If we're not logging _program output_ then IMO writing to `stderr` is appropriate (also, `console.error` invokes `util.format` which means you can use e.g., `%O` to mimic `util.inspect()`)

packages/webpack/src/buildtime/exclude.js
```diff
@@ -2,7 +2,7 @@
 /** @typedef {import('./policyGenerator').InspectableWebpackModule} InspectableWebpackModule */
 
 const path = require('node:path')
-const EXCLUDE_LOADER = path.join(__dirname, '../excludeLoader.js')
+const EXCLUDE_LOADER = path.resolve(path.join(__dirname, '../excludeLoader.js'))
```
boneskull:
```suggestion
const EXCLUDE_LOADER = path.join(__dirname, '../excludeLoader.js')
```

`__dirname` is always absolute, so this `path.resolve()` call is redundant.

packages/webpack/src/buildtime/generator.js
```diff
@@ -51,130 +51,154 @@ function processRequirements(requirements, module) {
...
 // this is to avoid wrapping the same instance twice
 const wrappedGeneratorInstances = new WeakSet()
 
 /**
  * @param {object} options
  * @param {string[]} options.excludes
- * @param {(path: string) => string | undefined} options.getIdentifierForPath
  * @param {boolean | undefined} options.runChecks
- * @param {import('./progress').ProgressAPI} options.PROGRESS
+ * @param {import('./utils.js').ProgressAPI} options.PROGRESS
```
boneskull: (recommend) use `@import`

packages/webpack/src/buildtime/generator.js
```diff
@@ -51,130 +51,154 @@ function processRequirements(requirements, module) {
...
+  }
+
+  let wrappingEnabed = false
+
+  return {
     /**
-     * @param {NormalModule} module
-     * @param {any} options - GeneratorOptions type not exported from webpack
-     * @returns {Source}
+     * @param {object} options
```
boneskull, line 79: seems like we should be able to get the shape of this object out of `webpack`

packages/webpack/src/buildtime/generator.js
```diff
@@ -51,130 +51,154 @@ function processRequirements(requirements, module) {
...
-          get: () => false,
-          set: () => {
-            // TODO: make the error more informative - explaining why the attempt to strict mode had to be skipped here but is applied anyway
-            console.warn(
-              'Attempted to set strict mode on module',
-              module.rawRequest,
-              Error().stack
-            )
-          },
+        // used to be .apply(this, arguments) but typescript complained. I feel it's worse that way
```
boneskull, line 114: `arguments` is the bad parts 😄

packages/webpack/src/buildtime/generator.js
```diff
@@ -51,130 +51,154 @@ function processRequirements(requirements, module) {
...
+        // skip doing anything if marked as excluded by the excludeLoader
+        // this may clean up the `tooEarly` list too
+        if (isExcluded(module)) {
+          excludes.push(module.userRequest)
+          diag.rawDebug(3, `skipped wrapping ${module.rawRequest}`)
+          return originalGeneratedSource
+        }
+
+        if (!wrappingEnabed) {
+          tooEarly.push(module.request || module.rawRequest)
```
boneskull, line 130: what's the significance of preferring `module.request` over `module.rawRequest`?

packages/webpack/src/buildtime/generator.js
```diff
@@ -51,130 +51,154 @@ function processRequirements(requirements, module) {
...
+        // originalGenerate adds requirements to options.runtimeRequirements. runtimeKit needs to be derived from those.
+        // We also depend on __webpack_require__ being there, so let's add it
+        options.runtimeRequirements.add('__webpack_require__')
+
+        // Turn off "use strict" being added in front of modules on final wrapping by webpack.
+        // If anything attempts to reverse it, we want to ignore it
+        if (
+          module.buildInfo && // never found an actual case in which buildInfo wasn't there, but types suggest it's possible
+          (module.buildInfo.strict === true ||
+            module.buildInfo.strict === undefined) // seems like it's possible for generation to run more than once for the same module
```
boneskull, line 147:
```suggestion
          (module.buildInfo.strict !== false) // seems like it's possible for generation to run more than once for the same module
```

I know this is not quite equivalent but may be feasible

packages/webpack/src/buildtime/generator.js
```diff
@@ -51,130 +51,154 @@ function processRequirements(requirements, module) {
...
+        ) {
+          Object.defineProperty(module.buildInfo, 'strict', {
+            get: () => false,
+            set: () => {
+              // TODO: make the error more informative - explaining why the attempt to strict mode had to be skipped here but is applied anyway
+              console.warn(
+                'Attempted to set strict mode on module',
+                module.rawRequest,
+                Error().stack
```
boneskull:
```suggestion
              console.trace(
                'Attempted to set strict mode on module',
                module.rawRequest)
```

packages/webpack/src/buildtime/generator.js
```diff
@@ -51,130 +51,154 @@ function processRequirements(requirements, module) {
...
+          options.runtimeRequirements,
+          module
+        )
+
+        let { before, after, source, sourceChanged } = wrapper({
+          // There's probably a good reason why webpack stores source in those objects instead
+          // of strings. Turning it into a string here might mean we're loosing some caching.
+          // Wrapper checks if transforms changed the source and indicates it, so that we can
+          // decide if we want to keep the original object representing it.
+          source: originalGeneratedSource.source().toString(),
```
boneskull, line 176: yes, but why are we turning it into a string?

packages/webpack/src/buildtime/generator.js
```diff
@@ -51,130 +51,154 @@ function processRequirements(requirements, module) {
...
+        if (packageId === undefined) {
+          throw Error(`Failed to find a packageId for ${module.resource}`)
+        }
+
+        const { runtimeKit, runtimeFlags } = processRequirements(
+          options.runtimeRequirements,
+          module
+        )
+
+        let { before, after, source, sourceChanged } = wrapper({
```
boneskull:
```suggestion
        const { before, after, source, sourceChanged } = wrapper({
```
??

packages/webpack/src/buildtime/modulesData.js
```diff
@@ -0,0 +1,242 @@
...
+ *   there's no superclass of all reasonable module types
+ */
+const isExternalModule = (m, moduleClass) =>
+  ['ExternalModule'].includes(moduleClass) &&
+  'externalType' in m &&
+  m.externalType !== undefined
+/**
+ * @param {WebPack.Module} m
+ * @param {string} moduleClass
+ * @returns {m is import('./policyGenerator.js').InspectableWebpackModule}
```
boneskull: `@import`

packages/webpack/src/buildtime/modulesData.js
```diff
@@ -0,0 +1,242 @@
...
+ */
+const isInspectableModule = (m, moduleClass) =>
+  'userRequest' in m ||
+  m.type?.startsWith('javascript') ||
+  isExternalModule(m, moduleClass)
+
+/**
```
boneskull, line 73: move to `types.d.ts`?

packages/webpack/src/buildtime/modulesData.js
```diff
@@ -0,0 +1,242 @@
...
+   * Array of module ids that are context modules and need to be double-wrapped.
+   *
+   * @type {{ moduleId: string | number; context: string }[]}
+   */
+  const contextModules = []
+
+  /**
+   * An array of modules deemed fit for inspecting for policy
+   *
+   * @type {import('./policyGenerator.js').InspectableWebpackModule[]}
```
boneskull: `@import`

packages/webpack/src/buildtime/modulesData.js
```diff
@@ -0,0 +1,242 @@
...
+   */
+  const inspectable = []
+
+  /**
+   * A record of module ids that are externals and need to be enforced as
+   * builtins.
+   *
+   * @type {Record<string | number, string>}
```
boneskull, line 131: `number`, really?

packages/webpack/src/buildtime/modulesData.js
```diff
@@ -0,0 +1,242 @@
...
+      )
+      diag.rawDebug(4, { module })
+      return
+    }
+
+    // ==================================================
+    // Fixing bad modules
+
+    // Fixes the issue with assets being emitted to dist without a loader
+    // TODO: refactor to move random hardening of the build somewhere it's easier to track.
```
boneskull, line 155: I'm not gonna complain about this TODO since I know you have limited time to land this

packages/webpack/src/buildtime/modulesData.js
```diff
@@ -0,0 +1,242 @@
...
+        )
+      )
+
+      // We can't use `chunkGraph.disconnectChunkAndModule` here
+      // because the require statement remains and errors out
+
+      if (module.generatorOptions) {
+        // generatorOptions was not present in testing, but types indicate it might be there
+        module.generatorOptions.emit = false
+      }
```
boneskull, line 173:
```suggestion
      // generatorOptions was not present in testing, but types indicate it might be there
      module.generatorOptions?.emit = false
```

packages/webpack/src/buildtime/modulesData.js
```diff
@@ -0,0 +1,242 @@
...
+      if (module.generator) {
+        module.generator = Object.create(module.generator, {
```
boneskull, line 175: do we care about dereferencing `Object` methods?

packages/webpack/src/buildtime/modulesData.js
```diff
@@ -0,0 +1,242 @@
...
+          emit: {
+            value: false,
+            writable: false,
+            configurable: false,
+            enumerable: true,
+          },
```
boneskull:
```suggestion
          emit: {
            value: false,
            enumerable: true,
          },
```

`writable` and `configurable` always default to `false`, fwiw

packages/webpack/src/buildtime/modulesData.js
```diff
@@ -0,0 +1,242 @@
...
+    // Note: module.context on an empty context module when no context information was guessable from code is going to point to the module that loads it.
+    if (isContextModule(module, moduleClass)) {
+      diag.rawDebug(3, {
+        contextModule: {
+          moduleId,
+          context: module.context,
+          // @ts-expect-error we want to see it if available
+          request: module?.options?.request,
+          // @ts-expect-error we want to see it if available
+          _identifier: module?._identifier,
```
boneskull, line 196: what are these and how do you know they are there if unmentioned in types? are these considered private properties?

packages/webpack/src/buildtime/modulesData.js
```diff
@@ -0,0 +1,242 @@
...
+      unenforceableModuleIds.push(moduleId)
+    } else {
+      if (isExternalModule(module, moduleClass)) {
+        externals[moduleId] = module.userRequest
+      }
+      if (isInspectableModule(module, moduleClass)) {
+        inspectable.push(module)
+      }
+
+      // typescript is complaining about the use of `resource` here, but it's actually there.
```
boneskull, line 219:
```suggestion
      // typescript is complaining about the use of `resource` here, but it's actually there.
      /* c8 ignore next */
      if (!('resource' in module)) {
        // or warn or whatever
        throw Error('Expected module to have a resource')
      } 
```

This type guard (assuming it retains the `throw`) will make `module` `IdentifiedModule & {resource: unknown}` so you'd have to change yr assertion on L225 to `/** @type {string} */(module.resource)`.

...but you could also just add `resource: string` to the `IdentifiedModule` type, since you own it.  I'd still probably add the conditional & `throw` above unless I was 100% sure there's no situation in which `module.resource` is not a `string` (if it's a private API or otherwise undocumented, I wouldn't trust it to still be there after an upgrade; this will help us notice & fix the problem if it breaks).

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -24,66 +19,66 @@ const POLICY_SNAPSHOT_FILENAME = 'policy-snapshot.json'
...
+  loadPolicy({ policyFromOptions, location }) {
+    const { policy, applyOverride } = loadPoliciesSync({
+      policyPath: path.join(location, 'policy.json'),
+      policyOverridePath: path.join(location, 'policy-override.json'),
+      debugMode: false,
+    })
+
+    /** @type {LavaMoatPolicy} */
+    let final = { resources: {} }
+    if (policyFromOptions) {
```
boneskull, line 42: this is a great place for a ternary 😄

packages/webpack/src/buildtime/modulesData.js
```diff
@@ -0,0 +1,242 @@
...
+ * @param {IdentifiedModule[]} opts.allIdentifiedModules
+ * @returns {{
+ *   inspectable: import('./policyGenerator.js').InspectableWebpackModule[]
+ *   contextModules: { moduleId: string | number; context: string }[]
+ *   knownPaths: { path: string; moduleId: string | number }[]
+ *   unenforceableModuleIds: (string | number)[]
+ *   externals: Record<string | number, string>
+ * }}
+ */
+exports.analyzeModules = ({
```
boneskull, line 96: (opinion) I would probably not define these types inline since it's rather verbose

(also `@import`)

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -95,73 +90,57 @@ module.exports = {
...
-          'utf8'
+      const packageName = getPackageNameForModulePath(
+        canonicalNameMap,
+        module.userRequest
+      )
+      const moduleRecord = new LavamoatModuleRecord({
+        // Knowing the actual specifier is not relevant here, they're used as unique identifiers that match between here and dependencies
+        specifier: module.userRequest,
+        file: module.userRequest,
+        type: isBuiltin(module.userRequest) ? 'builtin' : 'js',
```
boneskull, line 110: is this `isBuiltin` business for bundling Node.js projects, or...? why do we care if they are builtins on the Node.js side if they are replaced with shims during bundling?

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -95,73 +90,57 @@ module.exports = {
...
+            const depSpecifier = dep.resolvedModule.userRequest
+            acc[depSpecifier] = depSpecifier
+            return acc
+          }, /** @type {Record<string, string>} */ ({})),
+        },
+      })
+
+      try {
+        moduleInspector.inspectModule(moduleRecord)
+      } catch (/** @type any */ e) {
```
boneskull, line 127: I think this might accidentally work since it's invalid

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -24,66 +19,66 @@ const POLICY_SNAPSHOT_FILENAME = 'policy-snapshot.json'
...
+
+  /**
+   * @param {Object} opts
+   * @param {import('@lavamoat/aa').CanonicalNameMap} opts.canonicalNameMap
+   * @param {string} opts.location
+   * @param {IsBuiltinFn} opts.isBuiltin
+   * @param {ModuleWithConnections[]} opts.modulesToInspect
+   * @returns {LavaMoatPolicy}
+   */
+  generatePolicy({ canonicalNameMap, location, isBuiltin, modulesToInspect }) {
```
boneskull, line 73: curious about why this is explicitly synchronous

packages/webpack/src/buildtime/types.d.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface LavaMoatPluginOptions {
```
boneskull, line 12: each member here should have a description, but it doesn't have to happen in this PR

packages/webpack/src/buildtime/utils.js
```diff
@@ -107,5 +107,33 @@ module.exports = function progress({ steps }) {
...
+
+/**
+ * @template {Record<string, any>} T
+ * @template {keyof T} K
+ * @param {T} storeObj
+ * @param {readonly K[]} fields
+ * @returns {asserts storeObj is RequireFields<T, K>}
+ */
+const assertFields = (storeObj, fields) => {
+  const missingFields = fields.filter((field) => storeObj[field] === undefined)
```
boneskull, line 129: noting that this differs from 

```js
const missingFields = fields.filter((field) => !(field in storeObj))
```

packages/webpack/src/buildtime/wrapper.js
```diff
@@ -121,6 +121,6 @@ ${e}
         `
       )
     })
-    throw Error(validityFlag + 'wrapped module is not valid JS\n' + e)
+    throw Error(validityFlag + ': wrapped module is not valid JS\n' + e)
```
boneskull, line 124:
```suggestion
    throw Error(validityFlag + ': wrapped module is not valid JS\n', {cause: e})
```

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -0,0 +1,156 @@
...
+       * @param {string | undefined} params.currentChunkName - The webpack chunk
+       *   object
+       * @param {(string | number)[]} params.chunkIds - Array of chunk
+       *   identifiers
+       * @param {Object} params.policyData - LavaMoat security policy
+       *   configuration
+       * @param {Object} params.identifiers - Object containing module
+       *   identifier mappings
+       * @param {string} params.identifiers.root - Root identifier
+       * @param {Object} params.identifiers.identifiersForModuleIds - Module ID
```
boneskull:
```suggestion
       * @param {Record<string, string>} params.identifiers.identifiersForModuleIds - Module ID
```

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -0,0 +1,156 @@
...
+    return {
+      /**
+       * Generates the LavaMoat runtime source code based on chunk configuration
+       *
+       * @param {Object} params - The parameters object
+       * @param {string | undefined} params.currentChunkName - The webpack chunk
+       *   object
+       * @param {(string | number)[]} params.chunkIds - Array of chunk
+       *   identifiers
+       * @param {Object} params.policyData - LavaMoat security policy
```
boneskull:
```suggestion
       * @param {LavaMoatPolicy} params.policyData - LavaMoat security policy
```

idk if that's right tho

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -0,0 +1,156 @@
...
+       *   configuration
+       * @param {Object} params.identifiers - Object containing module
+       *   identifier mappings
+       * @param {string} params.identifiers.root - Root identifier
+       * @param {Object} params.identifiers.identifiersForModuleIds - Module ID
+       *   to identifier mappings
+       * @param {any} params.identifiers.unenforceableModuleIds - IDs of modules
+       *   that cannot be enforced
+       * @param {any} [params.identifiers.contextModuleIds] - Context module IDs
```
boneskull: booo! hiss!

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -0,0 +1,156 @@
...
+       * @param {Object} [params.identifiers.externals] - External module
```
boneskull: if we do not even access this value anywhere, use `unknown`

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -0,0 +1,156 @@
...
+const { assembleRuntime } = require('./assemble.js')
+const path = require('node:path')
+
+/** @import {LavaMoatPluginOptions} from '../buildtime/types' */
+
+module.exports = {
+  /**
+   * Builds the LavaMoat runtime configuration and generates runtime source code
+   *
+   * @param {Object} params - The parameters object
```
boneskull, line 13: 😜

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -0,0 +1,156 @@
...
+            },
+            {
+              name: 'runtime',
+              file: require.resolve('./runtimeUnlocked.js'),
+            },
+          ]
+        } else {
+          diag.rawDebug(2, `adding runtime for chunk ${currentChunkName}`)
+
+          runtimeChunks = [
```
boneskull, line 85:
```suggestion
          // TODO: document these
          runtimeChunks = [
```

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -0,0 +1,156 @@
...
+              json: true,
+            },
+            {
+              name: 'externals',
+              data: externals || null,
+              json: true,
+            },
+            { name: 'options', data: runtimeOptions, json: true },
+            (typeof runtimeOptions?.scuttleGlobalThis === 'boolean' &&
+              runtimeOptions.scuttleGlobalThis === true) ||
```
boneskull, line 126:
```suggestion
            (runtimeOptions.scuttleGlobalThis === true) ||
```

redundant

boneskull (approved):
No blockers, some questions, many mostly type-related suggestions.  I'm happy with just landing as-is with intention to make changes later given the time constraints

packages/webpack/src/buildtime/diagnostics.js
naugtur: _rawDebug does the formatting as well, but it's the only way to output to stderr that doesn't invoke async_hooks - I learned about it by creating infinite recursion the first time I tried to debug an async hook with concole, as one does :) 

Some async machinery is being triggered behind the scenes even in console.error - I never investigated what exactly tho.

packages/webpack/README.md
naugtur: Increasingly many :D 
This was left deliberately vague so that the error messages aren't and adding more doesn't need to be reflected here.

packages/webpack/README.md
naugtur: Added to DEVELOPMENT.md along with other updates

packages/webpack/src/buildtime/exclude.js
naugtur: :shaking_face:

packages/webpack/src/buildtime/generator.js
naugtur: We'd need to create a type that represents something in webpack's control and maintain it. I think enough tests depend on it working (pretty much all of them) that type safety wouldn't give us much return on investment.

packages/webpack/src/buildtime/generator.js
naugtur: not in good hands :P

packages/webpack/src/buildtime/generator.js
naugtur: `module.request` is the result of some processing by webpack, while `module.rawRequest` is the literal string accepted to be resolved. I recall it's possible for a module with rawRequest and without request to exist

packages/webpack/src/buildtime/generator.js
naugtur: It's valid, but not what I was aiming to say to the developer looking at it in the future. 
Regardless of whether it's true or unset, we want to step in. 

Now I'm thinking maybe we always want to step in...

packages/webpack/src/buildtime/generator.js
naugtur: Because we need to process it with evasive transforms.

packages/webpack/src/buildtime/modulesData.js
naugtur: yes, when webpack runs in production mode, it uses numbers from a sequence as ids for modules and chunks.

packages/webpack/src/buildtime/modulesData.js
naugtur:
```
Uncaught SyntaxError: Invalid left-hand side in assignment
```
node v22.16.0

Does it work for you?

packages/webpack/src/buildtime/modulesData.js
naugtur: not at build time where presumably we're gonna work under lockdown

packages/webpack/src/buildtime/modulesData.js
naugtur: options.request only exists on some module types, _identifier is a private field that webpack pretends isn't there but it can be helpful when looked up and it does exist for ContextModules (some modules require computation to get an identifier and there's a function to get that, but it might have side effects and it doesn't guarantee you'll get the identifier if too early in the compilation process anyway)
This is debug output for context modules only where I know it's there.

packages/webpack/src/buildtime/modulesData.js
naugtur: I'm not sure I get this. Let's talk about it tomorrow and I can fix later.

packages/webpack/src/buildtime/modulesData.js
naugtur: IT was better for the diff/history of the refactor to keep it here IMHO. Happy to move it out later.

packages/webpack/src/buildtime/modulesData.js
naugtur: did the @import, keeping them inline for the sake of the refactor history but happy to extract later.

packages/webpack/src/buildtime/policyGenerator.js
naugtur: It's part of a hook that's synchronous and all the structure it processes to generate policy is already loaded.

packages/webpack/src/buildtime/policyGenerator.js
naugtur: yes, isBuiltin is for people building bundles targeting node.js (eg. Snaps - they have a testing CLI that runs snaps in node locally)

packages/webpack/src/buildtime/policyGenerator.js
naugtur: Typescript was being annoying about `e` potentially being an unknown type instead of Error.

packages/webpack/src/buildtime/types.d.ts
naugtur: agree and agree

packages/webpack/src/buildtime/utils.js
naugtur: yup. I'm happy with the stricter meaning.

packages/webpack/src/buildtime/wrapper.js
naugtur: The way these are processed I'm not sure the cause would be surfaced to the user.

packages/webpack/src/runtime/runtimeBuilder.js
naugtur: :D I think that's a copilot leftover I didn't clean up after generating the jsdoc block

packages/webpack/src/runtime/runtimeBuilder.js
naugtur: updated to match the actual types thrown at it.

packages/webpack/src/runtime/runtimeBuilder.js
naugtur: It being an eyesore makes us more likely to eventually deprecate the old scuttling usage.

PR #1699, opened 2025-06-09
remove survey & perf

### LINKED ISSUE #1580: do something with perf and survey
- I don't believe any _currently active maintainers_ are using `packages/perf` and/or `packages/survey` for anything
- Renovate is trying to update deps in these packages, but this is unwise since nothing validates that the upgrades are functional
- They pull in more dependencies into CI and dev environs, which always carries risk.

Possible solutions:

1. Export them from this repo and into their own (archived) repos; they can be resurrected later or moved back if desired
2. Maintain them; create test suites and run them in CI
3. Disable automated upgrades of their dependencies and remove them from the list of workspaces. This would at least ensure their dependencies are not automatically installed in CI and dev environments.

Thoughts?

@naugtur @leotm @weizman

boneskull:
I moved these to https://github.com/LavaMoat/ad-hoc

If we hate it, we can 1. not merge this and 2. delete that repo.

Fixes #1580


leotm (approved):
thoughts here
- https://github.com/LavaMoat/LavaMoat/issues/1580#issuecomment-2959850245

boneskull:
Per discussion in #1580, I think it makes sense to archive `LavaMoat/ad-hoc`. I haven't done so yet because archival appears permanent; waiting for more feedback.

boneskull:
(Can also rename it something else. I—obviously—didn't know what to call it)

naugtur:
I'd like to keep perf, but adjust to cover `@lavamoat/webpack` and `@lavamoat/node`. Maybe drop some of the existing examples and bring in new ones, generalize how they're built.
We definitely need to have an answer to "how is performance affected"

boneskull:
@naugtur 

> I'd like to keep perf, but adjust to cover @lavamoat/webpack and @lavamoat/node. Maybe drop some of the existing examples and bring in new ones, generalize how they're built.

I'll happily agree if we can get this work on the schedule.  Is that feasible?  I am eager to remove them because they are currently unmaintained.

boneskull:
@naugtur Should I change this to archive just `survey`, then?

naugtur:
Let's merge this and bring back perf as a concept, but maybe focus on making it simple and easy to figure out when one gets back to it after a couple months of not looking.

PR #1704, opened 2025-06-11
fix(react-native-lockdown): add types

boneskull:
This configures `@lavamoat/react-native-lockdown` to ship type declarations. A custom ESLint config is also provided.

### DIFF
--- a/packages/react-native-lockdown/.eslintrc.js
+++ b/packages/react-native-lockdown/.eslintrc.js
@@ -1,0 +1,26 @@
+// TODO: use types when we determine what types to use
+module.exports = {
+  extends: ['../../.config/eslintrc.typed-workspace'],
+  parserOptions: {
+    sourceType: 'script',
+    ecmaVersion: 2020,
+  },
+  env: {
+    browser: false,
+    es6: false,
+    node: true,
+  },
+  overrides: [
+    {
+      files: ['./test/**/*.js'],
+      rules: {
+        // use t.log()
+        'no-console': 'error',
+
+        // of dubious value
+        'ava/use-t-well': 'off',
+      },
+    },
+  ],
+  ignorePatterns: ['./types/*', '**/*.tsbuildinfo'],
+}
--- a/packages/react-native-lockdown/package.json
+++ b/packages/react-native-lockdown/package.json
@@ -18,8 +18,14 @@
   },
   "main": "./src/index.js",
   "exports": {
-    ".": "./src/index.js",
-    "./repair": "./src/repair.js",
+    ".": {
+      "types": "./types/types.d.ts",
+      "default": "./src/index.js"
+    },
+    "./repair": {
+      "types": "./types/repair.d.ts",
+      "default": "./src/repair.js"
+    },
     "./package.json": "./package.json"
   },
   "files": [
--- a/packages/react-native-lockdown/src/index.js
+++ b/packages/react-native-lockdown/src/index.js
@@ -7,17 +7,11 @@
  */
 
 /**
- * @import {SerializerConfig} from '@react-native/metro-config'
- */
-
-/**
- * @typedef {(moduleId: number | string) => string} GetRunModuleStatement
- * @see {@link https://metrobundler.dev/docs/configuration/#getrunmodulestatement}
- */
-
-/**
- * @typedef {(options: { platform?: string }) => ReadonlyArray<string>} GetPolyfills
- * @see {@link https://metrobundler.dev/docs/configuration/#getpolyfills}
+ * @import {GetRunModuleStatementFn,
+ *  GetPolyfillsFn,
+ *  LockedDownSerializerConfig,
+ *  LockdownSerializerOptions,
+ *  LockdownSerializerConfig} from './lockdown-serializer.js'
  */
 
 const { warn } = console
@@ -28,7 +22,7 @@ const path = require('node:path')
 /**
  * Default implementation of Metro serializer option 'getRunModuleStatement'.
  *
- * @type {GetRunModuleStatement}
+ * @type {GetRunModuleStatementFn}
  */
 const defaultGetRunModuleStatement = (moduleId) => `__r(${moduleId})`
 
@@ -42,6 +36,12 @@ const ENTRY_FILE_MODULE_ID = 0
  *   call log
  */
 const warnAboutAnomalies = () => {
+  /**
+   * @type {{
+   *   calledOnceOnly: string | false
+   *   neverCalledWithEntryModule: string | false
+   * }}
+   */
   const deferredWarnings = {
     calledOnceOnly:
       'LavaMoat: getRunModuleStatement was only called once instead of at least twice in quick succession, with react-native InitializeCore first and bundle entry second. This is suspicious. Lockdown might not be in effect.',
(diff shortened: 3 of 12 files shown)


boneskull:
* **#1704** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1704?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1704?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/react-native-lockdown/package.json
boneskull, line 1: It's worth mentioning that because this is a CommonJS package, we have to do some sort of weird stuff with the exports if we want to expose top-level types.  We can't just export `index.d.ts`, because there's no way to _re-export types_ from our `index.js` (`@typedef` does something _similar_, but creates duplicate types).  

We use the same strategy here as in `lavamoat-core`:

1. Use `types.d.ts` for the `types` conditional entry point
2. `src/types.ts` re-exports all types we want to expose publicly _in addition_ to all of the types from `index.js` (in this case, those types are the type signatures of the two public API functions; if we had `@typedef`s in `index.js` they would also be exported)

To avoid a cycle (`index.js` -> `types.ts` -> `index.js`), I moved the types for `lockdownSerializer()` into their own file.

packages/react-native-lockdown/src/index.js
```diff
@@ -7,17 +7,11 @@
...
- */
-
-/**
- * @typedef {(options: { platform?: string }) => ReadonlyArray<string>} GetPolyfills
- * @see {@link https://metrobundler.dev/docs/configuration/#getpolyfills}
+ * @import {GetRunModuleStatementFn,
+ *  GetPolyfillsFn,
+ *  LockedDownSerializerConfig,
+ *  LockdownSerializerOptions,
+ *  LockdownSerializerConfig} from './lockdown-serializer.js'
```
boneskull, line 14: manual formatting 😄

packages/react-native-lockdown/src/index.js
```diff
@@ -42,6 +36,12 @@ const ENTRY_FILE_MODULE_ID = 0
  *   call log
  */
 const warnAboutAnomalies = () => {
+  /**
+   * @type {{
+   *   calledOnceOnly: string | false
+   *   neverCalledWithEntryModule: string | false
+   * }}
+   */
```
boneskull, line 44: didn't bother defining a type here because it's internal to this function

packages/react-native-lockdown/src/index.js
```diff
@@ -118,19 +106,19 @@ const lockdownSerializer = ({ hermesRuntime = true } = {}, userConfig = {}) => {
     )
   }
 
-  /** @type {SerializerConfig} */
-  const config = assign({}, userConfig)
+  /** @type {LockedDownSerializerConfig} */
+  const config = /** @type {any} */ (assign({}, userConfig))
```
boneskull, line 110: it isn't a `LockedDownSerializerConfig` _yet_, but it will be.

packages/react-native-lockdown/src/lockdown-serializer.ts
```diff
@@ -0,0 +1,59 @@
...
+  hermesRuntime?: boolean
+}
+
+/**
+ * Result of calling `lockdownSerializer()`.
+ *
+ * This will not be the same object as the provided
+ * {@link LockdownSerializerConfig}.
+ */
+export type LockedDownSerializerConfig = Omit<
```
boneskull, line 41: if it's not clear, `getRunModuleStatement` & `getPolyfills` are both _optional_ in `LockdownSerializerConfig`, so this is how we can make them _required._

packages/react-native-lockdown/src/repair.js
```diff
@@ -1,3 +1,5 @@
+/// <reference types="ses" />
```
boneskull, line 1: This exposes the SES globals. I think this can be done w/ configuration as well, but probably requires some fiddling (`compilerOptions: {types: ['ses']}` is not sufficient). Normally just an `import 'ses'` / `require('ses')` would do this, but it seems we expect it to already be loaded.

(Is there an assertion that it actually _is_ already loaded?)

packages/react-native-lockdown/test/unit/assertPolyfills.spec.js
```diff
@@ -1,4 +1,10 @@
-const test = require('ava')
+'use strict'
+
+/**
+ * @import {TestFn} from 'ava'
+ */
+
+const test = /** @type {TestFn} */ (/** @type {unknown} */ (require('ava')))
```
boneskull, line 7: I'm not sure why AVA's CJS types are so totally broken, even in the latest version.

boneskull (commented):
Added some explanations

packages/react-native-lockdown/.eslintrc.js
```diff
@@ -0,0 +1,26 @@
...
+  },
+  overrides: [
+    {
+      files: ['./test/**/*.js'],
+      rules: {
+        // use t.log()
+        'no-console': 'error',
+
+        // of dubious value
+        'ava/use-t-well': 'off',
```
leotm, line 21: NB: rules https://github.com/avajs/eslint-plugin-ava/tree/main/docs/rules

packages/react-native-lockdown/src/index.js
leotm: neat 😄 

Screenshot 2025-06-17 at 3 03 27 pm

should we update the file ext and tsconfig?

(not too bothered since VS Code resolves the path fine)

packages/react-native-lockdown/src/repair.js
leotm: i can see this triple slash ref is needed now ^ otherwise VSCode sees the `@link` tags as `any`, so LGTM

leotm: on `main` (before merging this), do you know how VSCode is able to [see](https://github.com/user-attachments/assets/b0e516a2-f164-468b-af9f-d8ca6b0682c6) the `@link` tags w/o the triple slash ref?

leotm: > (Is there an assertion that it actually is already loaded?)

good point our tests check

```js
// ...
  t.true(polyfills.some((p) => p === sesPath))
  t.true(polyfills.some((p) => p.endsWith('repair.js')))
// ...
```

but ideally the sesPath first

packages/react-native-lockdown/src/lockdown-serializer.ts
```diff
@@ -0,0 +1,59 @@
...
+ * as a production dependency.
+ * @see {@link https://github.com/facebook/metro/blob/main/packages/metro-config/types/configTypes.d.ts}
+ */
+export interface LockdownSerializerConfig {
+  getPolyfills?: GetPolyfillsFn
+
+  getRunModuleStatement?: GetRunModuleStatementFn
+
+  [key: PropertyKey]: unknown
+}
```
leotm, line 22: i was hoping we could follow convention
https://github.com/react-native-community/template/blob/main/template/metro.config.js#L7
so import `('@react-native/metro-config').MetroConfig['serializer']`
so we maintain less types

then since it's a devDependency in typical RN project
https://github.com/react-native-community/template/blob/main/template/package.json#L26
we add it as a peerDependency

unless you still reckon not worth the bother? 😜

leotm: since `@react-native/metro-config` imports types from `metro-config`, is there a reason the template doesn't directly import them from `metro-config`? perhaps extraneous dependencies are less favoured

leotm (approved):
LGTM and thank you! only few minor questions

leotm:
no blocking questions/suggestions, good to merge

PR #1712, opened 2025-06-16
create @lavamoat/vog

boneskull:
This creates `@lavamoat/vog` ~~(and subsequently adopts it in `@lavamoat/node`, though I should probably split that into a second PR)~~.

This is a general-purpose library for LavaMoat's CLI tools. It includes:

- Logger
- Progress bars
- Spinners
- Prompts
- Colors
- Other various formatting utils

Yes, [vog is a thing](https://en.wikipedia.org/wiki/Vog).

Currently, it is an ES module with TS sources, which may want to change. I would prefer to publish a dual-module package, but that would require adopting a tool like [tshy](https://npm.im/tshy) or [tsup](https://tsup.egoist.dev). Might not be a bad idea anyway, but I don't want to block this PR on that decision.

**UPDATE Sep 10 2025**: This package no longer targets #1593 and instead targets `main`. It only adds the package itself.  It uses `node:test` instead of AVA and leverages type-stripping. Changes outside the package are limited to:

- Update root `tsconfig.json` to build `vog`
- Update `clean:types` script to _always_ call `tsc -b --clean`, which should catch things that the `git` command would not (needed to clean `vog`)
- Lockfile update

### DIFF
--- a/packages/vog/package.json
+++ b/packages/vog/package.json
@@ -1,0 +1,61 @@
+{
+  "name": "@lavamoat/vog",
+  "version": "0.1.0",
+  "type": "module",
+  "description": "CLI utilities for LavaMoat",
+  "repository": {
+    "directory": "packages/vog",
+    "type": "git",
+    "url": "git+https://github.com/LavaMoat/LavaMoat.git"
+  },
+  "homepage": "https://lavamoat.github.io",
+  "bugs": "https://github.com/LavaMoat/LavaMoat/issues",
+  "author": "LavaMoat Project",
+  "license": "MIT",
+  "engines": {
+    "node": "^22.6.0 || ^24.0.0",
+    "npm": ">=9.0.0"
+  },
+  "exports": {
+    ".": {
+      "default": "./src/index.js",
+      "types": "./src/index.d.ts"
+    },
+    "./log": {
+      "default": "./src/log.js",
+      "types": "./src/log.d.ts"
+    },
+    "./package.json": "./package.json",
+    "./progress": {
+      "default": "./src/progress.js",
+      "types": "./src/progress.d.ts"
+    },
+    "./spinner": {
+      "default": "./src/spinner.js",
+      "types": "./src/spinner.d.ts"
+    }
+  },
+  "directories": {
+    "test": "test"
+  },
+  "files": [
+    "CHANGELOG.md",
+    "src/**/*",
+    "!*.tsbuildinfo",
+    "!tsconfig.json"
+  ],
+  "scripts": {
+    "lint:deps": "depcheck",
+    "test": "node --test --experimental-strip-types",
+    "test:cover": "c8 npm test"
+  },
+  "dependencies": {
+    "@opentf/cli-pbar": "0.7.2",
+    "@types/cli-spinner": "0.2.3",
+    "cli-spinner": "0.2.10",
+    "consola": "3.4.2"
+  },
+  "publishConfig": {
+    "access": "public"
+  }
+}
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -27,7 +27,39 @@ jobs:
               --workspace=packages/preinstall-always-fail
               --workspace=packages/yarn-plugin-allow-scripts
           - node-version: 18.x
+            workspaces: >-
+              --workspace=packages/aa
+              --workspace=packages/allow-scripts
+              --workspace=packages/browserify
+              --workspace=packages/core
+              --workspace=packages/git-safe-dependencies
+              --workspace=packages/lavamoat-node
+              --workspace=packages/lavapack
+              --workspace=packages/laverna
+              --workspace=packages/node
+              --workspace=packages/preinstall-always-fail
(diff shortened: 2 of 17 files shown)


boneskull:
* **#1712** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1712?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1712?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

naugtur:
I was ready to complain it's too much to maintain, but it actually isn't. Just a lightweight wrapper for a selection of packages.

boneskull:
I've striped out usages of `vog` from this PR so now it's just the package. I can stack usage onto this

naugtur (approved):

packages/vog/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "directory": "packages/vog",
+    "type": "git",
+    "url": "git+https://github.com/LavaMoat/LavaMoat.git"
+  },
+  "homepage": "https://lavamoat.github.io",
+  "bugs": "https://github.com/LavaMoat/LavaMoat/issues",
+  "author": "LavaMoat Project",
+  "license": "MIT",
+  "engines": {
+    "node": "^18.13.0 || ^20.0.0 || ^22.0.0 || ^24.0.0",
```
boneskull: `node:test`'s mocking functionality was backported to v18.13.0

packages/vog/package.json
```diff
@@ -0,0 +1,60 @@
...
+  ],
+  "scripts": {
+    "lint:deps": "depcheck",
+    "test": "node --test --experimental-strip-types",
+    "test:cover": "c8 npm test"
+  },
+  "dependencies": {
+    "consola": "3.4.2",
+    "ora": "8.2.0",
+    "progress-estimator": "0.3.1"
```
naugtur: This is all great, but I think Id rather have less dependencies in lavamoat than display spinners and progress bars.

consola is great, has no dependencies

But the other 2 are a big liability in exchange for eyecandy
```
└─┬ @lavamoat/vog@0.1.0 -> ./packages/vog
  ├── consola@3.4.2
  ├─┬ ora@8.2.0
  │ ├── chalk@5.6.0
  │ ├─┬ cli-cursor@5.0.0
  │ │ └─┬ restore-cursor@5.1.0
  │ │   ├─┬ onetime@7.0.0
  │ │   │ └── mimic-function@5.0.1
  │ │   └── signal-exit@4.1.0
  │ ├── cli-spinners@2.9.2
  │ ├── is-interactive@2.0.0
  │ ├── is-unicode-supported@2.1.0
  │ ├─┬ log-symbols@6.0.0
  │ │ ├── chalk@5.6.0 deduped
  │ │ └── is-unicode-supported@1.3.0
  │ ├── stdin-discarder@0.2.2
  │ ├─┬ string-width@7.2.0
  │ │ ├── emoji-regex@10.4.0
  │ │ ├── get-east-asian-width@1.3.0
  │ │ └── strip-ansi@7.1.0 deduped
  │ └─┬ strip-ansi@7.1.0
  │   └── ansi-regex@6.2.0
  └─┬ progress-estimator@0.3.1
    ├─┬ chalk@2.4.2
    │ ├─┬ ansi-styles@3.2.1
    │ │ └─┬ color-convert@1.9.3
    │ │   └── color-name@1.1.3
    │ ├── escape-string-regexp@1.0.5
    │ └─┬ supports-color@5.5.0
    │   └── has-flag@3.0.0
    ├── cli-spinners@1.3.1
    ├── humanize-duration@3.33.1
    └─┬ log-update@2.3.0
      ├── ansi-escapes@3.2.0
      ├─┬ cli-cursor@2.1.0
      │ └─┬ restore-cursor@2.0.0
      │   ├─┬ onetime@2.0.1
      │   │ └── mimic-fn@1.2.0
      │   └── signal-exit@3.0.7
      └─┬ wrap-ansi@3.0.1
        ├─┬ string-width@2.1.1
        │ ├── is-fullwidth-code-point@2.0.0
        │ └── strip-ansi@4.0.0 deduped
        └─┬ strip-ansi@4.0.0
          └── ansi-regex@3.0.1
```

naugtur: I noticed both have a cli-spinner package in deps, which is also a zero dependency one https://www.npmjs.com/package/cli-spinner

we can have a spinner like that. Especially if we pin dependencies.

boneskull: OK, I can live with `cli-spinner`, but progress bars will eventually be something I'll want to see when we're crunching big projects for policy generation or performing other long-running operations.

I think [cli-progress](https://npm.im/cli-progress) is probably our best bet to keep the transitives low.

boneskull:
actually, I think I want to go with [@opentf/cli-pbar](https://npm.im/@opentf/cli-pbar) because a) it is ostensibly still maintained, and b) it has even fewer dependencies.

PR #1716, opened 2025-06-17
feat(react-native-lockdown): use vendored SES version

leotm:
Use vendored SES version
- https://github.com/endojs/endo/pull/2855
  - https://github.com/endojs/endo/pull/2855/commits/eebf99c4a42fea8818c5ba6a39d2a8254db0dda8
  - `cd packages/ses && yarn run build`

Then replace with next SES release once that's out

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -144,7 +144,6 @@ module.exports = {
     '**/types/**/*',
     '**/test/**/fixture*/**/*',
     '**/examples/**/*',
-    'ses.cjs',
-    'ses-hermes.cjs',
+    '**/vendor/**/*',
   ],
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -144,5 +144,7 @@ module.exports = {
     '**/types/**/*',
     '**/test/**/fixture*/**/*',
     '**/examples/**/*',
+    'ses.cjs',
+    'ses-hermes.cjs',
   ],
 }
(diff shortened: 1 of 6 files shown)


.eslintrc.js
```diff
@@ -144,5 +144,7 @@ module.exports = {
     '**/types/**/*',
     '**/test/**/fixture*/**/*',
     '**/examples/**/*',
+    'ses.cjs',
+    'ses-hermes.cjs',
```
naugtur: I'd prefer `**/vendor/*`

packages/react-native-lockdown/vendor/ses-hermes.cjs
```diff
@@ -0,0 +1,13708 @@
+// ses@1.13.0
```
naugtur: not true. Document the specific commit ID from which this was built. If not sure, get the commit id and build it and copy over to be sure.

leotm: updated to `ses@<sha>` format, which refers to
- PR commit https://github.com/endojs/endo/pull/2855/commits/eebf99c4a42fea8818c5ba6a39d2a8254db0dda8
- standalone commit https://github.com/endojs/endo/commit/eebf99c4a42fea8818c5ba6a39d2a8254db0dda8

.eslintrc.js
leotm: updated to recursive `**/vendor/**/*` too just in case

naugtur (approved):

PR #1724, opened 2025-06-24
chore(aa): disable symlink perf test on all darwin

boneskull:
The symlink perf test was disabled in CI on darwin, but it needs to be disabled on darwin regardless of CI env.

This was preventing a release thru the usual process (test first).

### DIFF
--- a/packages/aa/test/symlink-perf.spec.js
+++ b/packages/aa/test/symlink-perf.spec.js
@@ -39,7 +39,7 @@ test.after(async (t) => {
 })
 
 // FIXME: this test fails on darwin running in GH actions; diagnose and fix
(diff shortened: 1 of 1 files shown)


boneskull:
cc @naugtur

naugtur:
We need a setup to create benchmarks for our tools alongside tests and run them rarely (nightly on new PRs and main?)
That would be a good replacement for this situation.

The change here eliminates the ability to benchmark Darwin entirely.
If this is regularly failing, do we need to look into the viability of the performance improvement now, or is it still better performance but no longer an order of magnitude?

naugtur:
It's failing MacOS anyway because of different things being slower `32 tests remained pending after a timeout`

leotm (approved):

PR #1739, opened 2025-07-16
fix(webpack): fix how ambient asset files requests are filtered

naugtur:
addresses https://github.com/MetaMask/metamask-extension/issues/34299

Next steps:

add either 
- a capability that can be granted to individual packages in the policy  
  
or
- a snapshot mechanism to fail the build if new ambient assets from dependencies were emitted


itsyoboieltr (approved):

packages/webpack/src/buildtime/modulesData.js
```diff
@@ -51,6 +51,18 @@ const isAmbientAsset = (m) =>
   Array.isArray(m.loaders) &&
   m.loaders.length === 0
 
+/**
+ * @typedef {Parser & { dataUrlCondition: boolean }} AssetParser
+ */
+
+/**
+ *
```
naugtur:
```suggestion
 * Recognize an AssetParser bu constructor name for the lack of ability to get the type from Webpack
```

boneskull (approved):
approved, but we should add a test for the change to inlined data URL assets.  please add a TODO if you're not gonna add it 😄

naugtur:
Covered data url in tests.

PR #1740, opened 2025-07-16
feat(webpack): defend webpack runtime from scuttling

naugtur:
Addresses issues with webpack runtime modules using scuttled globals.
fixes: https://github.com/MetaMask/metamask-extension/issues/34300

caveats and considerations:
- [x] consider&decide: lavamoat runtime chunk got moved later in the runtime concatenation so that it can potentially tweak other runtime modules in the future; on the other hand, if we didn't need it to be so, we could have one runtime chunk be added without closure and share the scope caching
- [x] destructuring from globalThis is less repetetive, but more problematic if the fields are not there (eg. `location` in node.js) - I think we'll have to fall back to `var NAME = globalThis.NAME`

### DIFF
--- a/packages/webpack/src/runtime/runtimeBuilder.js
+++ b/packages/webpack/src/runtime/runtimeBuilder.js
@@ -29,13 +29,16 @@ module.exports = {
        * @returns {string}
        */
       getDefensiveCodingPreamble() {
-        return `
-const {
-  setTimeout, clearTimeout,
-  setInterval, clearInterval,
-  document, self, location
-} = globalThis
-`
+        const globals = [
+          'setTimeout',
+          'clearTimeout',
+          'setInterval',
+          'clearInterval',
+          'document',
+          'self',
+          'location',
+        ]
+        return `var ${globals.map((g) => `${g} = globalThis.${g}`).join(',')};`
       },
       /**
        * Generates the LavaMoat runtime source code based on chunk configuration
--- a/packages/webpack/src/runtime/runtimeBuilder.js
+++ b/packages/webpack/src/runtime/runtimeBuilder.js
@@ -34,6 +34,7 @@ module.exports = {
           'setTimeout', // LoadScriptRuntimeModule.js
           'clearTimeout', // LoadScriptRuntimeModule.js
           'document', // LoadScriptRuntimeModule.js, AutoPublicPathRuntimeModule.js
+          'trustedTypes', // GetTrustedTypesPolicyRuntimeModule.js
           'self',
         ]
         return `var ${globals.map((g) => `${g} = globalThis.${g}`).join(',')};`
--- a/packages/webpack/src/runtime/runtimeBuilder.js
+++ b/packages/webpack/src/runtime/runtimeBuilder.js
@@ -30,13 +30,11 @@ module.exports = {
        */
       getDefensiveCodingPreamble() {
         const globals = [
-          'setTimeout',
-          'clearTimeout',
-          'setInterval',
-          'clearInterval',
-          'document',
+          'location', // AutoPublicPathRuntimeModule.js
+          'setTimeout', // LoadScriptRuntimeModule.js
+          'clearTimeout', // LoadScriptRuntimeModule.js
+          'document', // LoadScriptRuntimeModule.js, AutoPublicPathRuntimeModule.js
           'self',
-          'location',
         ]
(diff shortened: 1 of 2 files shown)


packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -21,6 +21,25 @@ module.exports = {
...
+          'setTimeout',
+          'clearTimeout',
+          'setInterval',
+          'clearInterval',
+          'document',
+          'self',
+          'location',
+        ]
+        return `var ${globals.map((g) => `${g} = globalThis.${g}`).join(',')};`
+      },
```
leotm: how do we know this list is adequate to fix the issue? and won't cause another
`Property "anotherGlobal" of globalThis is inaccessible under scuttling mode`

leotm: i note from the issue `setTimeout` ✅ and from the dupe issue `clearTimeout` ✅

leotm: prob a silly question, why not add them to [scuttleGlobalThisExceptions](https://github.com/MetaMask/metamask-extension/blob/main/development/build/index.js#L99-L153)?

naugtur: We don't want them to be exceptions as they're somewhat powerful APIs.

As for the list - it's not exhaustive but the aim is to cover a minimal set of what we expect or know to be used in webpack runtime modules instead of covering for all globals, which would not benefit the bundle size.

HMR is probably using more of these, but we will never support HMR

leotm: good point RE HMR can be more minimal, i'll add a suggestion and comments for documentation

leotm (approved):

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -21,6 +21,25 @@ module.exports = {
...
+       */
+      getDefensiveCodingPreamble() {
+        const globals = [
+          'setTimeout',
+          'clearTimeout',
+          'setInterval',
+          'clearInterval',
+          'document',
+          'self',
+          'location',
```
leotm:
```suggestion
          'location', // AutoPublicPathRuntimeModule.js
          'setTimeout', // LoadScriptRuntimeModule.js
          'clearTimeout', // LoadScriptRuntimeModule.js
          'document', // LoadScriptRuntimeModule.js, AutoPublicPathRuntimeModule.js
          'self',
```

leotm: for ref excluded the intervals
- setInterval (HMR: https://github.com/webpack/webpack/blob/main/hot/poll.js#L38)
- clearInterval (not in lib/runtime)
- probably self can be as well (not in lib/runtime), but it's used in other libs so i'm not comfortable excluding it

leotm:
was going to add [Function](https://github.com/webpack/webpack/blob/main/lib/runtime/GlobalRuntimeModule.js#L29) as well, but it's already defended in [scuttleGlobalThisExceptions](https://github.com/MetaMask/metamask-extension/blob/main/development/build/index.js#L137) ✅

edit: but perhaps should be included for all users since part of webpack lib/runtime

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -21,6 +21,23 @@ module.exports = {
...
+       *
+       * @returns {string}
+       */
+      getDefensiveCodingPreamble() {
+        const globals = [
+          'location', // AutoPublicPathRuntimeModule.js
+          'setTimeout', // LoadScriptRuntimeModule.js
+          'clearTimeout', // LoadScriptRuntimeModule.js
+          'document', // LoadScriptRuntimeModule.js, AutoPublicPathRuntimeModule.js
+          'self',
```
leotm, line 38: just realised i _think_ [trustedTypes](https://github.com/webpack/webpack/blob/main/lib/runtime/GetTrustedTypesPolicyRuntimeModule.js#L31) needs to be here as well

```suggestion
          'trustedTypes', // GetTrustedTypesPolicyRuntimeModule.js
          'self',
```

naugtur: That's very niche, but maybe not in our target audience

PR #1741, opened 2025-07-22
feat(webpack): properly expose types for the plugin

### LINKED ISSUE #34303: (Optional) Type declarations for the plugin
It'd be great to see the available configuration options without needing to jump into the source code when using the lavamoat webpack plugin.

naugtur:
- fix scuttling config type
- wire up types in package.json
- get the right types exported

Damn this was tedious to get working on the default export.
All attempts to have a main types.ts file as the types entrypoint were unsuccessful because of that.

### DIFF
--- a/packages/webpack/package.json
+++ b/packages/webpack/package.json
@@ -18,8 +18,16 @@
     "npm": ">=7.0.0"
   },
   "main": "src/plugin.js",
+  "types": "types/plugin.d.ts",
+  "exports": {
+    ".": {
+      "types": "./types/plugin.d.ts",
+      "default": "./src/plugin.js"
+    }
+  },
   "files": [
-    "src"
+    "src",
+    "types"
   ],
   "keywords": [
     "webpack",
--- a/packages/webpack/src/buildtime/types.ts
+++ b/packages/webpack/src/buildtime/types.ts
@@ -1,15 +1,9 @@
-import type { LavaMoatPolicy } from 'lavamoat-core'
+import type { LavaMoatPolicy, LavaMoatScuttleOpts } from 'lavamoat-core'
 import type { LockdownOptions } from 'ses'
 
-export interface ScuttlerObjectConfig {
-  enabled?: boolean
-  exceptions?: string[]
-  scuttlerName?: string
-}
+export type ScuttlerConfig = LavaMoatScuttleOpts | boolean | undefined
 
-export type ScuttlerConfig = ScuttlerObjectConfig | boolean | undefined
-
-export interface LavaMoatPluginOptions {
+export interface CompleteLavaMoatPluginOptions {
   generatePolicy?: boolean
   rootDir?: string
   policyLocation: string
@@ -26,3 +20,5 @@ export interface LavaMoatPluginOptions {
   debugRuntime?: boolean
   unlockedChunksUnsafe?: RegExp
 }
+
+export type LavaMoatPluginOptions = Partial<CompleteLavaMoatPluginOptions>
--- a/packages/webpack/src/plugin.js
+++ b/packages/webpack/src/plugin.js
@@ -26,8 +26,16 @@ const {
 const { loadCanonicalNameMap } = require('@lavamoat/aa')
 
 /**
+ * This is jsdoc for reexport
+ *
+ * @typedef {import('./buildtime/types').LavaMoatPluginOptions} LavaMoatPluginOptions
+ */
+
+/**
+ * Just import
+ *
  * @import {LockdownOptions} from 'ses'
- * @import {LavaMoatPluginOptions, ScuttlerConfig} from './buildtime/types'
+ * @import {CompleteLavaMoatPluginOptions} from './buildtime/types'
  * @import {CanonicalNameMap} from '@lavamoat/aa'
  * @import {LavaMoatPolicy} from 'lavamoat-core'
  */
@@ -65,12 +73,6 @@ class VirtualRuntimeModule extends RuntimeModule {
   }
 }
 
-/**
- * @typedef {LavaMoatPluginOptions} LavaMoatPluginOptions
- *
- * @typedef {ScuttlerConfig} ScuttlerConfig
- */
-
 // =================================================================
 // Plugin code
 // =================================================================
@@ -91,10 +93,10 @@ const lockdownDefaults = /** @type {const} */ ({
 
 class LavaMoatPlugin {
   /**
-   * @param {Partial<LavaMoatPluginOptions>} [options]
+   * @param {LavaMoatPluginOptions} [options]
    */
   constructor(options = {}) {
-    /** @type {LavaMoatPluginOptions} */
+    /** @type {CompleteLavaMoatPluginOptions} */
     if (options.scuttleGlobalThis === true) {
       options.scuttleGlobalThis = { enabled: true }
     } else if (typeof options.scuttleGlobalThis === 'object') {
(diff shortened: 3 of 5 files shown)


packages/webpack/src/buildtime/types.ts
```diff
@@ -1,15 +1,9 @@
-import type { LavaMoatPolicy } from 'lavamoat-core'
+import type { LavaMoatPolicy, LavaMoatScuttleOpts } from 'lavamoat-core'
 import type { LockdownOptions } from 'ses'
 
-export interface ScuttlerObjectConfig {
-  enabled?: boolean
-  exceptions?: string[]
```
naugtur, line 6: this was wrong.

packages/webpack/src/plugin.js
```diff
@@ -65,12 +73,6 @@ class VirtualRuntimeModule extends RuntimeModule {
   }
 }
 
-/**
- * @typedef {LavaMoatPluginOptions} LavaMoatPluginOptions
- *
- * @typedef {ScuttlerConfig} ScuttlerConfig
- */
```
naugtur, line 72: It seems these were breaking the exported types

packages/webpack/src/plugin.js
```diff
@@ -91,10 +93,10 @@ const lockdownDefaults = /** @type {const} */ ({
 
 class LavaMoatPlugin {
   /**
-   * @param {Partial<LavaMoatPluginOptions>} [options]
+   * @param {LavaMoatPluginOptions} [options]
```
naugtur, line 96: nicer display name for end users

packages/webpack/src/buildtime/types.ts
```diff
@@ -1,15 +1,9 @@
...
-}
-
-export type ScuttlerConfig = ScuttlerObjectConfig | boolean | undefined
+export type ScuttlerConfig = LavaMoatScuttleOpts | boolean | undefined
 
-export interface LavaMoatPluginOptions {
+export interface CompleteLavaMoatPluginOptions {
   generatePolicy?: boolean
   rootDir?: string
   policyLocation: string
```
boneskull, line 9: (navel-gazing) since this is required, maybe it shouldn't necessarily be an option

packages/webpack/src/plugin.js
```diff
@@ -26,8 +26,16 @@ const {
 const { loadCanonicalNameMap } = require('@lavamoat/aa')
 
 /**
+ * This is jsdoc for reexport
```
boneskull, line 29: (navel-gazing)

What I've done elsewhere is create a `types.ts` which re-exports types from other sources (including maybe the `index.js`) _and_ a `types.js` which is an empty module.  this becomes the `types` export in `package.json`.

I'm not sure how well it works in CJS, but it might?  this way we can avoid `@typedef` for any reason.

packages/webpack/src/plugin.js
```diff
@@ -91,10 +93,10 @@ const lockdownDefaults = /** @type {const} */ ({
 
 class LavaMoatPlugin {
   /**
-   * @param {Partial<LavaMoatPluginOptions>} [options]
+   * @param {LavaMoatPluginOptions} [options]
    */
   constructor(options = {}) {
-    /** @type {LavaMoatPluginOptions} */
+    /** @type {CompleteLavaMoatPluginOptions} */
```
boneskull: this doesn't seem to refer to any variable?

packages/webpack/src/plugin.js
```diff
@@ -122,7 +124,7 @@ class LavaMoatPlugin {
   apply(compiler) {
     /**
```
boneskull, line 125: while generally I dislike `@typedef`, it makes sense to define "scoped" types this way. with `.ts` sources you'd do the same thing

packages/webpack/src/runtime/runtime-namespace.ts
```diff
@@ -1,13 +1,18 @@
 import { EndowmentsToolkitFactory, LavaMoatPolicy } from 'lavamoat-core'
-import { LavaMoatPluginOptions, ScuttlerConfig } from '../plugin'
+import { LavaMoatPluginOptions, ScuttlerConfig } from '../buildtime/types'
```
boneskull:
```suggestion
import type { EndowmentsToolkitFactory, LavaMoatPolicy } from 'lavamoat-core'
import type { LavaMoatPluginOptions, ScuttlerConfig } from '../buildtime/types'
```

packages/webpack/src/runtime/runtime-namespace.ts
```diff
@@ -1,13 +1,18 @@
...
 }
 
 type RequiredProperty<T, K extends keyof T> = Omit<T, K> & Required<Pick<T, K>>
 export interface RuntimeNamespace {
-  scuttling: {scuttle: (globalThis: Record<string, unknown>, scuttleGlobalThis: ScuttlerConfig) => {}}
+  scuttling: {
+    scuttle: (
+      globalThis: Record<string, unknown>,
+      scuttleGlobalThis: ScuttlerConfig
+    ) => {}
```
boneskull, line 14: I don't think `{}` is the correct return type, but I don't know what it is.

packages/webpack/package.json
```diff
@@ -18,8 +18,16 @@
...
+  "exports": {
+    ".": {
+      "types": "./types/plugin.d.ts",
+      "default": "./src/plugin.js"
+    }
+  },
   "files": [
-    "src"
+    "src",
+    "types"
```
boneskull, line 30:
```suggestion
    "src",
    "types",
    "!*.tsbuildinfo"
```

packages/webpack/package.json
```diff
@@ -18,8 +18,16 @@
     "npm": ">=7.0.0"
   },
   "main": "src/plugin.js",
```
boneskull, line 20: if we are not targeting Node.js runtimes that don't support `exports`, then `main` can safely be eliminated

boneskull (approved):
no blockers, but some suggestions / comments.

packages/webpack/src/buildtime/types.ts
naugtur: All fields are optional to the user and the `Complete` type is to internally apply after defaults were set.

packages/webpack/src/plugin.js
naugtur: I took 3 separate distinct attempts at that and could not get the default export to work correctly across typescript and javascript with cjs

packages/webpack/src/plugin.js
naugtur: somehow it gets applied to options. At least I thought it did, maybe it broke. I could move it around to a better place.

packages/webpack/src/runtime/runtime-namespace.ts
naugtur: this says the function is not meant to return.
It doesn't bubble up to the plugin options, it's a typing for the runtime.js to have warnings and we want it to not rey on return value of scuttler.

last but not least, this just got reformatted because prettier changes its mind regularly

packages/webpack/src/runtime/runtime-namespace.ts
naugtur: can't apply suggestions on removals :shrug:

packages/webpack/package.json
naugtur: why? this seems wrong. files is an allowlist

packages/webpack/package.json
naugtur: There's still so much tooling out there with gaps in exports support that I don't want to risk it.

PR #1749, opened 2025-08-04
feat(core): decouple writable store from root compartment global

naugtur:
Tests contain an example of a new usage pattern. The change is backward compatible. 

Performance impact:
It's likely got a minimal negative impact on performance - one more copy is needed and writable fields are propagated through getters not directly.

Compatibility considerations:  
Old way of using it, if a field was not marked as writable and root wrote to it, the new field would be visible in new compartments created afterwards (if endowed).

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -34,7 +34,7 @@
   "scripts": {
     "lib:ses": "cp ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js",
     "lint:deps": "depcheck",
-    "test": "ava && npm run test:ses",
+    "test": "npm run test:ses && ava",
     "test:ses": "bash ./test/ses.sh"
   },
   "dependencies": {
--- a/packages/core/src/endowmentsToolkit.js
+++ b/packages/core/src/endowmentsToolkit.js
@@ -58,9 +58,9 @@ function endowmentsToolkit({
 } = {}) {
   return {
     // public API
+    createDonor,
     getEndowmentsForConfig,
     copyWrappedGlobals,
-    endowAll,
     getBuiltinForConfig,
     createFunctionWrapper,
     // internals exposed for core
@@ -69,6 +69,73 @@ function endowmentsToolkit({
     copyValueAtPath,
     applyGetSetPropDescTransforms,
     applyEndowmentPropDescTransforms,
+  }
+
+  /**
+   * Creates a donor object for the specified global reference.
+   *
+   * @param {any} globalRef
+   */
+  function createDonor(globalRef) {
+    const endowmentDescriptorsFlat = collectAllDescriptors(globalRef)
+
+    // expose all own properties of globalRef, including non-enumerable
+    const globalBackup = create(null)
+    for (const key in endowmentDescriptorsFlat) {
+      const desc = endowmentDescriptorsFlat[key]
+      if (desc.value && desc.value === globalRef) {
+        continue
+      }
+      defineProperty(globalBackup, key, desc)
+    }
+
+    return {
+      /**
+       * Creates an object populated with all properties of the global backup,
+       * but applies this-wrapping and writables by reusing the same
+       * implementation as getEndowmentsForConfig. The way to use this is:
+       *
+       * - Create a cache of globals by calling copyWrappedGlobals
+       * - Call endowAll on that cache for root compartment
+       * - Call getEndowmentsForConfig on that cache for other compartments
+       *
+       * @param {object} [unwrapTo] - For getters and setters, when the
+       *   this-value is unwrapFrom, is replaced as unwrapTo
+       * @param {object} [unwrapFrom] - For getters and setters, the this-value
+       *   to replace (default: targetRef)
+       */
+      endowAll(unwrapTo, unwrapFrom) {
+        const whitelistedReads = getOwnPropertyNames(globalBackup)
+        const allowedWriteFields = knownWritableFields
+        return makeMinimalViewOfRef(
+          globalBackup,
+          whitelistedReads,
+          unwrapTo,
+          unwrapFrom,
+          [], // nothing is explicitly banned
+          allowedWriteFields
+        )
+      },
+      /**
+       * Creates an object populated with only the deep properties specified in
+       * the packagePolicy
+       *
+       * @param {LMPolicy.PackagePolicy} packagePolicy - LavaMoat policy item
+       *   representing a package
+       * @param {object} [unwrapTo] - For getters and setters, when the
+       *   this-value is unwrapFrom, is replaced as unwrapTo
+       * @param {object} [unwrapFrom] - For getters and setters, the this-value
+       *   to replace (default: targetRef)
+       */
+      endowSpecified: (packagePolicy, unwrapTo, unwrapFrom) => {
+        return getEndowmentsForConfig(
+          globalBackup,
+          packagePolicy,
+          unwrapTo,
+          unwrapFrom
+        )
+      },
+    }
   }
 
   /**
@@ -147,42 +214,6 @@ function endowmentsToolkit({
       unwrapTo,
       unwrapFrom,
       explicitlyBanned,
-      allowedWriteFields
-    )
-  }
-
-  /**
-   * Creates an object populated with all properties of the source, but applies
-   * this-wrapping and writables by reusing the same implementation as
-   * getEndowmentsForConfig. The way to use this is:
-   *
-   * - Create a cache of globals by calling copyWrappedGlobals
-   * - Call endowAll on that cache for root compartment
-   * - Call getEndowmentsForConfig on that cache for other compartments
-   *
-   * @template {object} T Deep properties specified in the packagePolicy
-   * @param {T} sourceRef - Object from which to copy properties
-   * @param {object} [unwrapTo] - For getters and setters, when the this-value
-   *   is unwrapFrom, is replaced as unwrapTo
-   * @param {object} [unwrapFrom] - For getters and setters, the this-value to
-   *   replace (default: targetRef)
-   * @returns {Partial<T>} - The targetRef
-   */
-  function endowAll(sourceRef, unwrapTo, unwrapFrom) {
-    const proto = getPrototypeOf(sourceRef)
-    if (proto !== null && proto !== prototype) {
-      throw new Error(
-        `LavaMoat - endowAll does not support sourceRefs with custom prototype`
-      )
-    }
-    const whitelistedReads = getOwnPropertyNames(sourceRef)
-    const allowedWriteFields = knownWritableFields
-    return makeMinimalViewOfRef(
-      sourceRef,
-      whitelistedReads,
-      unwrapTo,
-      unwrapFrom,
-      [], // nothing is explicitly banned
       allowedWriteFields
     )
   }
@@ -483,6 +514,9 @@ function endowmentsToolkit({
             (thisValue) => thisValue === unwrapFromGlobalThis,
             unwrapToGlobalThis
           )
+          // NOTE: Should we censor/wrap return values?
+          // } else if (result === receiverRef) {
+          //   return unwrapToGlobalThis
         } else {
           return result
         }
@@ -571,27 +605,25 @@ function endowmentsToolkit({
   }
 
   /**
-   * @type {CopyWrappedGlobals}
+   * Extracts descriptors of own properties and properties on the prototype
+   * chain into a flat collection of descriptors
+   *
+   * @param {any} sourceGlobal
    */
-  function copyWrappedGlobals(
-    globalRef,
-    target,
-    globalThisRefs = ['globalThis']
-  ) {
+  function collectAllDescriptors(sourceGlobal) {
     // find the relevant endowment sources
-    const globalProtoChain = getPrototypeChain(globalRef)
+    const globalProtoChain = getPrototypeChain(sourceGlobal)
     // the index for the common prototypal ancestor, prototype
     // this should always be the last index, but we check just in case
     const commonPrototypeIndex = globalProtoChain.findIndex(
       (globalProtoChainEntry) => globalProtoChainEntry === prototype
     )
     if (commonPrototypeIndex === -1) {
-      // TODO: fix this error message
       throw new Error(
-        'Lavamoat - unable to find common prototype between Compartment and globalRef'
+        'Lavamoat - sourceGlobal is not a descendant of the shared Object.prototype. Might be from a different Realm.'
       )
     }
-    // we will copy endowments from all entries in the prototype chain, excluding prototype
+    // we will copy endowments from all entries in the prototype chain, excluding Object.prototype
     const endowmentSources = globalProtoChain.slice(0, commonPrototypeIndex)
 
     // call all getters, in case of behavior change (such as with FireFox lazy getters)
@@ -599,10 +631,11 @@ function endowmentsToolkit({
     endowmentSources.forEach((source) => {
       const descriptors = getOwnPropertyDescriptors(source)
       values(descriptors).forEach((desc) => {
+        // TODO: reconsider this. We have facilities to wrap and unwrap getters, maybe improving the fidelity there would be nough to keep the getters alive
         if ('get' in desc && desc.get) {
           try {
             // calling getters can potentially throw (e.g. localStorage inside a sandboxed iframe)
-            apply(desc.get, globalRef, [])
+            apply(desc.get, sourceGlobal, [])
           } catch {}
         }
       })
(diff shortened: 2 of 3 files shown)


packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -2,11 +2,12 @@ const test = require('ava')
 const endowmentsToolkit = require('../src/endowmentsToolkit.js')
 
 function prepareTest({ knownWritable } = {}) {
```
boneskull, line 9:
```suggestion
/**
 * @param {Object} [options]
 * @param {Set<string>} [options].knownWritable]
 * @returns {ReturnType<typeof endowmentsToolkit>}
 */
function prepareTest({ knownWritable } = {}) {
```

- [complaining] I note that the return type of `endowmentsToolkit` probably needs to be explicitly defined somewhere because rn it's only inferred.

packages/core/src/endowmentsToolkit.js
```diff
@@ -136,6 +137,39 @@ function endowmentsToolkit({
     )
   }
 
+  /**
+   * Creates an object populated with all properties of the source, but applies
+   * all the wrapping and writable mapping
```
boneskull: [should]

- might be nice to enumerate what exactly "all the wrapping and writable mapping" is
- also would be helpful to mention why this exists at all and how it should be used

packages/core/src/endowmentsToolkit.js
```diff
@@ -136,6 +137,39 @@ function endowmentsToolkit({
...
+   *
+   * @template {object} T Deep properties specified in the packagePolicy
+   * @param {T} sourceRef - Object from which to copy properties
+   * @param {object} unwrapTo - For getters and setters, when the this-value is
+   *   unwrapFrom, is replaced as unwrapTo
+   * @param {object} unwrapFrom - For getters and setters, the this-value to
```
```suggestion
   * @param {object} [unwrapFrom] - For getters and setters, the this-value to
```

[must] TS won't be able to infer this param is optional otherwise

packages/core/src/endowmentsToolkit.js
```diff
@@ -136,6 +137,39 @@ function endowmentsToolkit({
...
+   *   replace (default: targetRef)
+   * @returns {Partial<T>} - The targetRef
+   */
+  function endowAll(sourceRef, unwrapTo, unwrapFrom) {
+    const proto = Object.getPrototypeOf(sourceRef)
```
boneskull: Why would we choose `Object.getPrototypeOf` here, but `Reflect.getPrototypeOf` in the body of `endowmentsToolkit()`?

packages/core/src/endowmentsToolkit.js
```diff
@@ -136,6 +137,39 @@ function endowmentsToolkit({
...
+    if (proto !== null && proto !== Object.prototype) {
+      throw new Error(
```
```suggestion
      throw new TypeError(
```

[suggestion]

packages/core/src/endowmentsToolkit.js
```diff
@@ -136,6 +137,39 @@ function endowmentsToolkit({
...
+        `LavaMoat - endowAll does not support sourceRefs with custom prototype`
+      )
+    }
+    const whitelistedReads = Object.getOwnPropertyNames(sourceRef).sort(
```
boneskull: - Would it be a better idea to do this sort in `makeMinimalViewOfRef()` instead?
- Am I correct that we're doing this for sake of consistency, because presumably `Object.getOwnPropertyNames` returns prop names in the order in which they were created?

packages/core/src/endowmentsToolkit.js
```diff
@@ -136,6 +137,39 @@ function endowmentsToolkit({
...
+      (a, b) => a.length - b.length
+    )
+    const allowedWriteFields = knownWritableFields
+    return makeMinimalViewOfRef(
```
boneskull: [complaining] this function takes too many parameters

packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -144,6 +145,75 @@ test('instrumentDynamicValueAtPath puts a getter at path', (t) => {
...
+  // in lavamoat. Note that endowAll uses the same optional references for unwrapping
+
+  // Storage for globals we cache in lavamoat
+  const goldenMasterRecordGlobal = copyWrappedGlobals(originalSource, {})
+  // root compartment global
+  const rootGlobal = endowAll(goldenMasterRecordGlobal)
+  // non-root compartments
+  const endowments = getEndowmentsForConfig(goldenMasterRecordGlobal, config)
+  const endowments2 = getEndowmentsForConfig(goldenMasterRecordGlobal, config)
+  endowments.a = 42
```
boneskull, line 186: Would it be valuable to have a failure case for this?  e.g., when

```js
const config = {
  globals: {
    a: true
  }
}
```

packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -144,6 +145,75 @@ test('instrumentDynamicValueAtPath puts a getter at path', (t) => {
   t.is(target.a.b.c, 2)
 })
 
+test('copyWrappedGlobals - writable fields get updated in compartments and in root but not globally', (t) => {
+  const knownWritable = new Set(['a'])
```
boneskull, line 163: - There appears to be an implicit relationship between `config.globals..write` and `knownWritable`; am I correct that names in `knownWritable` are expected to be in `config.globals` with a policy value of `write`?
- What happens when `a` is not in `knownWritable` but is writable in config?
- What happens when `a` is not writable in config but is in `knownWritable`?
- Is there coverage for the above two cases, and if not, would they be valuable?
- Would the API be nicer to use if we eliminated `knownWritables` entirely and `endowmentsToolkit()` accepted a Policy in its place (from which it would derive the `Set`)?

packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -400,6 +470,30 @@ test('getEndowmentsForConfig - specify unwrap from, unwrap to', (t) => {
   t.is(getter.call(), globalThis)
 })
 
+test('endowAll - specify unwrap from, unwrap to', (t) => {
+  'use strict'
```
boneskull, line 575: Maybe move L493 up here so we can more easily understand that this is necessary and not just a way to make ESLint shut up.  Or is it?  From that comment, it sounds like it's needed at runtime, right? Would just like this clarified

[should]

packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -400,6 +470,30 @@ test('getEndowmentsForConfig - specify unwrap from, unwrap to', (t) => {
...
+  const unwrapTo = {}
+  const unwrapFrom = {}
+  const { endowAll } = prepareTest()
+  const sourceGlobal = {
+    get xyz() {
+      return this
+    },
+  }
+  const resultGlobal = endowAll(sourceGlobal, unwrapTo, unwrapFrom)
+  process._rawDebug(1, Object.getOwnPropertyDescriptors(resultGlobal))
```
boneskull: [should]

```suggestion
```

packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -457,3 +551,37 @@ test('copyWrappedGlobals - copy from prototype too', (t) => {
 
   t.is(Object.keys(target).sort().join(), 'onTheObj,onTheProto,window')
 })
+
+test('copyWrappedGlobals+endowAll - ends up including fields from prototype too', (t) => {
```
boneskull: [suggestion]

```suggestion
test('copyWrappedGlobals+endowAll - includes fields from source prototype', (t) => {
```

packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -457,3 +551,37 @@ test('copyWrappedGlobals - copy from prototype too', (t) => {
...
+  const sourceProto = {
+    onTheProto: function () {},
+  }
+  const source = Object.create(sourceProto)
+  source.onTheObj = function () {}
+  const target = Object.create(null)
+  copyWrappedGlobals(source, target, ['window'])
+  const endowments = endowAll(target, {}, {})
+
+  t.is(Object.keys(endowments).sort().join(), 'onTheObj,onTheProto,window')
```
boneskull: [suggestion] arguably reduces cognitive overhead and maintenance burden:

```suggestion
  t.like(Object.keys(endowments).sort(), ['onTheObj', 'onTheProto', 'window'])
```

(or `t.deepEqual()`)

packages/core/src/endowmentsToolkit.js
boneskull: - Why is this explicitly disallowed, anyway?  I _think_ it's because we want to ensure that the object passed into `endowAll()` was created by LavaMoat.   Can you answer in a comment?

  > If that's true, alternative (IMO, more defensive) implementations may:
  > - Use a _specific_ prototype we own when preparing the object for `endowAll()`
  > - Place a module-scoped `Symbol` on the prepared object, "tagging" it as our own, and removing it immediately after validation.
- Should this same check be made elsewhere in the endowments toolkit?

packages/core/package.json
```diff
@@ -34,7 +34,7 @@
   "scripts": {
     "lib:ses": "cp ../../node_modules/ses/dist/lockdown.umd.js ./lib/lockdown.umd.js",
     "lint:deps": "depcheck",
-    "test": "ava && npm run test:ses",
+    "test": "npm run test:ses && ava",
```
boneskull, line 37: Why?

boneskull (changes requested):
I only have [one quick blocker](https://github.com/LavaMoat/LavaMoat/pull/1749#discussion_r2252423628)

packages/core/test/endowmentsToolkit.spec.js
naugtur: I was told by an expert at a conference that the most common mistake people make is defining return types explicitly too often, while of you do typescript right it should infer all of them correctly.

I agree this type needs a name tho 😁

packages/core/package.json
naugtur: so that you can do  `npm test test/filename` like in every other package

packages/core/src/endowmentsToolkit.js
naugtur: but somehow it does when it's being passed directly to a function that treats it as optional. Or at least it's not complaining. Anyway, updated all to be optional

packages/core/src/endowmentsToolkit.js
naugtur: I don't know. 

making them consistently local scope instead.

packages/core/src/endowmentsToolkit.js
naugtur: The expectation is that we don't have to go down the prototype chain because copyWrappedGlobals already did that. 

1. We could add all return values of copyWrappedGlobals to a weakmap and only accept those, but that restricts future use (and testing) of this function. The idea is we won't graft any of the fields from one to the other. 

2. we could make a stricter workflow/pattern - the golden master is not really needed as a direct reference. So we could have a method that creates one (also, skips the wrapping as it'll have methods wrapped later anyway) and then returns only getEndowments and endowAll

I like the second option.

packages/core/src/endowmentsToolkit.js
naugtur: I cargo-culted this one.  It's unnecessary here. It is only necessary when we have globals from policy that have dots in them. sorting by length gets us out of a more expensive config situation.

what this also means is we might get tripped up by `globalThis['a.b]=1`

packages/core/src/endowmentsToolkit.js
naugtur: yes. It's been bothering me for a while. The thing is, these functions are called a lot and options bag is an extra allocation, so I couldn't sacrifice performance to clear the annoyance.

packages/core/test/endowmentsToolkit.spec.js
naugtur: `knownWritable` is a set created in an initial pass of the entire policy and it contains all keys everywhere that had the value of `write`

packages/core/test/endowmentsToolkit.spec.js
naugtur: I added the test and noticed a minor compatibility issue.
If a field is writable elsewhere, an attempt to write to it in a compartment that is in strict mode and had it endowed as readonly is going to cause a `Cannot set property a of #<Object> which has only a getter` error.

I'm wondering whether it's a good thing to have an error (maybe a setter that throws a LavaMoat specific one) or should we let it overwrite the local copy by creating a setter that replaces itself (and the getter) with a value when invoked.

That would be a more compatible behavior - effectively identical to `a` not being writable elsewhere - but a silent failure when you have code attempting to write to a global that's already writable elsewhere and it might help in debugging.

packages/core/test/endowmentsToolkit.spec.js
naugtur: In my experience the test output of comparing arrays is horrible. I was considering switching to a string comparison in webpack dist file snapshots because the comparison is so unreaedble when they're an array.
In this case it doesn't really matter for 3 elements.

I don't want to use `like` because it leaves room for flexibility. Might be convinced to use `deepEqual`.

packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -144,7 +158,127 @@ test('instrumentDynamicValueAtPath puts a getter at path', (t) => {
...
+  const rootGlobal = endowAll(goldenMasterRecordGlobal)
+  // non-root compartments
+  const endowments = getEndowmentsForConfig(goldenMasterRecordGlobal, config)
+  const endowments2 = getEndowmentsForConfig(goldenMasterRecordGlobal, config)
+
+  t.throws(
+    () => {
+      endowments.a = 42
+    },
+    { message: 'Cannot set property a of #<Object> which has only a getter' }
```
naugtur, line 222: We might want to change this behavior. See thread above.

packages/core/test/endowmentsToolkit.spec.js
naugtur: did more cleanup in that regard.

packages/core/test/endowmentsToolkit.spec.js
boneskull: fair enough. I want to switch to `node:test` at some point anyhow

packages/core/src/endowmentsToolkit.js
boneskull: agree on 2

packages/core/test/endowmentsToolkit.spec.js
boneskull: with the caveat that explicit type definitions are necessary for [isolated declarations](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#isolated-declarations), though we don't need that atm

packages/core/src/endowmentsToolkit.js
boneskull: iirc the tests in core aren't typechecked (unless `//@ts-check`)

packages/core/src/endowmentsToolkit.js
boneskull: we absolutely do not support `globalThis['a.b.c']` and should probably do something about it.

packages/core/test/endowmentsToolkit.spec.js
boneskull: presumably, we're avoiding duplication of work by passing it in from elsewhere because we found the caller(s) must traverse the globals anyhow—and it might as well create `knownWritables` while it's doing that.  but if such a traversal _also_ happens in `endowmentsToolkit()` at the appropriate time, it may just make sense to keep `knownWritables` within `endowmentsToolkit()` itself.  ??

packages/core/test/endowmentsToolkit.spec.js
boneskull: I'm not sure how this relates to any other thread 😄

boneskull (approved):
Blocker has been resolved, but it sounds like you want to do some more work on this.

packages/core/test/endowmentsToolkit.spec.js
naugtur: The underlying idea was that endowmentsToolkit doesn't see the policy in its entirety and it was dictated by the situation in endo, where at the time of calling getEndowmentsForConfig the entire policy doesn't exist anymore. So we're saving the knownWritable only.
I think that assumption might no longer be true, but the design of endowmentsToolkit remains. Can change it, but don't see a good reason to do that.

packages/core/src/endowmentsToolkit.js
```diff
@@ -750,19 +848,6 @@ function defaultCreateFunctionWrapper(sourceValue, unwrapTest, unwrapTo) {
  * @returns {(...args: any[]) => any}
  */
 
-/**
- * Makes a copy of all globals from the global ref to a target and wraps them
- * with the wrapper this endowmentsToolkit was configured to use. It also copies
- * all circular references to the root package compartment globalThis.
- *
- * @callback CopyWrappedGlobals
```
naugtur, line 818: It was annoying me that these are at the bottom of the file, so I moved one to see if anything fails.
@boneskull do you remember why we moved them there? I'm not seeing anything failing yet.

naugtur:
The ultimate test for this is gonna be trying it out in webpack on metamask. I just found an issue that this might be fixing or making worse :D

naugtur:
this is unlikely to reach acceptable compatibility and we've remedied the main erason we needed it. closing.

PR #1751, opened 2025-08-07
feat(webpack): repairs for globals that need more than unwrap from core

naugtur:
- fixes a silent failure in MM where `event.source` was the real `window` being compared with the compartment global
- introduces a general mechanism for including repairs into the bundle runtime only if needed (I was considering putting repairs in core, but then they would all go into the bundle)
- https://github.com/LavaMoat/LavaMoat/pull/1659 can be implemented as a repair

### DIFF
--- a/packages/webpack/src/runtime/repairs/messageevent.js
+++ b/packages/webpack/src/runtime/repairs/messageevent.js
@@ -46,8 +46,10 @@ exports.MessageEvent = (
   theRealGlobalThis,
   packageCompartmentGlobal
 ) => {
-  const originalListener = endowments.addEventListener
   const originalConstructor = endowments.MessageEvent
+  if (!originalConstructor) {
+    return
+  }
 
   /**
    * @param {string} type
@@ -62,7 +64,24 @@ exports.MessageEvent = (
   // WARNING: fidelity breaks down at MessageEvent.prototype.constructor,
   // which runs the original constructor and does not pin the resulting object
   // to a matching global in the WeakMap
-
+}
+/**
+ * Wraps the source getter on MessageEvent prototype with conversion from actual
+ * global to compartment global
+ *
+ * @param {{ [key: string]: any }} endowments
+ * @param {any} theRealGlobalThis
+ * @param {any} packageCompartmentGlobal
+ */
+exports.addEventListener = (
+  endowments,
+  theRealGlobalThis,
+  packageCompartmentGlobal
+) => {
+  const originalListener = endowments.addEventListener
+  if (!originalListener) {
+    return
+  }
   /**
    * @param {string} type
    * @param {function} listener
--- a/packages/webpack/src/runtime/repairs/messageevent.js
+++ b/packages/webpack/src/runtime/repairs/messageevent.js
@@ -1,0 +1,39 @@
+const { defineProperty } = Object
+const { call } = Function.prototype
+
+// TODO:
+// - implement wider support for messageable realms matching (parent, top, opener)
+// - once other realm reference are emulated, match and return the right references from the getter as well
+
+/**
+ * Wraps the source getter on MessageEvent prototype with conversion from actual
+ * global to compartment global
+ *
+ * @param {{ [key: string]: any }} endowments
+ * @param {any} theRealGlobalThis
+ * @param {any} packageCompartmentGlobal
+ */
+exports.MessageEvent = (
+  endowments,
+  theRealGlobalThis,
+  packageCompartmentGlobal
+) => {
+  const original = Object.getOwnPropertyDescriptor(
+    endowments['MessageEvent'].prototype,
+    'source'
+  )
+  if (original && original.get) {
+    const sourceGetter = call.bind(original.get)
+    defineProperty(endowments['MessageEvent'].prototype, 'source', {
+      ...original,
+      get() {
+        const w = sourceGetter(this)
+        if (w === theRealGlobalThis) {
+          return packageCompartmentGlobal
+        } else {
+          return w
+        }
+      },
+    })
+  }
+}
--- a/packages/webpack/src/runtime/repairs/messageevent.js
+++ b/packages/webpack/src/runtime/repairs/messageevent.js
@@ -84,12 +84,13 @@ exports.addEventListener = (
   }
   /**
    * @param {string} type
-   * @param {function} listener
+   * @param {(event: MessageEvent) => void} listener
    * @param {any} [options]
    */
   endowments.addEventListener = (type, listener, options) => {
     if (type === 'message') {
-      const wrappedListener = (/** @type {MessageEvent} */ event) => {
+      /** @type {typeof listener} */
+      const wrappedListener = (event) => {
         messageToGlobalMap.set(event, packageCompartmentGlobal)
         return listener.call(packageCompartmentGlobal, event)
       }
--- a/packages/webpack/src/runtime/repairs/messageevent.js
+++ b/packages/webpack/src/runtime/repairs/messageevent.js
@@ -1,9 +1,37 @@
 const { defineProperty } = Object
 const { call } = Function.prototype
 
-// TODO:
-// - implement wider support for messageable realms matching (parent, top, opener)
-// - once other realm reference are emulated, match and return the right references from the getter as well
+const messageToGlobalMap = new WeakMap()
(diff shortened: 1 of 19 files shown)


packages/webpack/src/runtime/repairs/messageevent.js
```diff
@@ -0,0 +1,39 @@
...
+  theRealGlobalThis,
+  packageCompartmentGlobal
+) => {
+  const original = Object.getOwnPropertyDescriptor(
+    endowments['MessageEvent'].prototype,
+    'source'
+  )
+  if (original && original.get) {
+    const sourceGetter = call.bind(original.get)
+    defineProperty(endowments['MessageEvent'].prototype, 'source', {
```
naugtur: PROBLEM: this is a shared prototype and it's only working because there's only one compartment with it in the test :facepalm: 

fixing...

naugtur:
Now I'm considering making the repair-globalname relationship not 1:1, so that if MessageEvent is not uttered but addEventListener is, we'd apply it too. 
It'd be great to let user choose, but that'll hinder adoption of this repair. Unless TOFU finds that it's a message event listener and encode that in policy. 

I'm slowly becoming a TC39 snob looking at how Web APIs mix everything together.

boneskull:
@naugtur `git-safe-dependencies`' tests are causing errors again.  Unclear if these are transient.  Maybe we should consider excluding some subset of these tests from the main test workflow and run them separately in a way that could be retried or as a check that is allowed to fail...

boneskull (commented):
I think I'll need a walkthrough of what is going on here.  Unclear if you're done changing it

naugtur:
I'll have to stabilize the tests for git safe, it's not abandoned.
I'll give you a walkthrough tomorrow.

boneskull:
@naugtur lol I didn't mean to imply it was abandoned. just that since the tests seem to have a hard dependency on network connectivity, they are prone to flake

naugtur:
I thought depending on github being up in the tests is fine. The intended use for this tool is to query github too, so if used in CI it'll have the same problems actually. Maybe it should retry or something...

naugtur:
* **#1751** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1751?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1751?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1766** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1766?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#1744](https://github.com/LavaMoat/LavaMoat/pull/1744) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1744?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1761](https://github.com/LavaMoat/LavaMoat/pull/1761) <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1761?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1767** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1767?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/webpack/src/runtime/repairs/messageevent.js
```diff
@@ -0,0 +1,112 @@
...
+// realms.set(globalThis.top || {}, 'top')
+// realms.set(globalThis.parent || {}, 'parent')
+// realms.set(globalThis.opener || {}, 'opener')
+
+const mep = MessageEvent.prototype
+// eslint-disable-next-line no-undef
+const lastResortGlobal = new Compartment().globalThis
+
+const original = Object.getOwnPropertyDescriptor(mep, 'source')
+if (original && original.get) {
```
boneskull, line 17: (suggestion)

```suggestion
if (original?.get) {
```

packages/webpack/src/runtime/repairs/messageevent.js
```diff
@@ -0,0 +1,112 @@
...
+  theRealGlobalThis,
+  packageCompartmentGlobal
+) => {
+  const originalListener = endowments.addEventListener
+  if (!originalListener) {
+    return
+  }
+  /**
+   * @param {string} type
+   * @param {function} listener
```
boneskull: (suggestion)

```suggestion
   * @param {(event: MessageEvent) => void} listener
```

packages/webpack/src/runtime/repairs/messageevent.js
```diff
@@ -0,0 +1,112 @@
...
+   * @param {any} [options]
+   */
+  endowments.addEventListener = (type, listener, options) => {
+    if (type === 'message') {
+      const wrappedListener = (/** @type {MessageEvent} */ event) => {
```
boneskull: (suggestion)

```suggestion
      /** @type {typeof listener} */
      const wrappedListener = (event) => {
```

packages/webpack/src/runtime/repairsBuilder.js
```diff
@@ -0,0 +1,38 @@
+const { readFileSync } = require('node:fs')
+const { repairs } = require('./repairs/index')
+
+/**
+ * Combines the sources of only the repairs that are needed based on the policy
+ *
+ * @param {import('lavamoat-core').LavaMoatPolicy} policy
```
boneskull: (suggestion)

```suggestion

/** @import {LavaMoatPolicy} from 'lavamoat-core' */

/**
 * Combines the sources of only the repairs that are needed based on the policy
 *
 * @param {LavaMoatPolicy} policy
```

packages/webpack/src/runtime/repairsBuilder.js
```diff
@@ -0,0 +1,38 @@
...
+exports.buildRepairs = (policy, skipRepairs) => {
+  const repairsAvailable =
+    Array.isArray(skipRepairs) && skipRepairs.length > 0
+      ? repairs.filter(
+          (repair) =>
+            !repair.target.some((target) => {
+              return skipRepairs.includes(target)
+            })
+        )
+      : repairs
```
boneskull, line 21: (suggestion) extract into a function

packages/webpack/src/runtime/repairsBuilder.js
```diff
@@ -0,0 +1,38 @@
...
+      }
+    }
+  }
+  const repairsToInclude = repairsAvailable.filter((repair) =>
+    repair.target.some((target) => {
+      return allGlobalsInvolved.has(target)
+    })
+  )
+
+  return repairsToInclude
```
boneskull, line 37: (comment) we don't have many repairs rn so this is probably quick enough

packages/webpack/test/scaffold.js
```diff
@@ -116,12 +116,18 @@ const defaultGlobals = () => ({
   URL,
 })
 
+/**
+ * @typedef {object} RunReturn
+ * @property {import('vm').Context} context - VM context object
+ * @property {any} result - Return value from the executed code
+ */
```
boneskull: (suggestion)

```suggestion

/** @import {Context} from 'node:vm' */

/**
 * @typedef {object} RunReturn
 * @property {Context} context - VM context object
 * @property {any} result - Return value from the executed code
 */
```

boneskull (approved):

packages/webpack/src/runtime/repairs/messageevent.js
naugtur: I think I might have to downgrade to older ES version in all of runtime because it doesn't get processed before landing in the bundle. planning to set up a linting rule for that.

packages/webpack/src/runtime/repairsBuilder.js
naugtur: compared to what webpack is doing reading thousands of files and concatenating them, this will always be quick enough (it happens at build time)
And we'll have at most tens of repairs.

packages/webpack/test/scaffold.js
naugtur: I intended the inline import as an indication that it's the only place it's being used.

PR #1762, opened 2025-08-12
fix(webpack): avoid scuttling the global Webpack uses for sharing chunks

naugtur:
If chunks are being loaded dynamically after scuttling happens, they will not work.
Since we're not prescribing anything about chunk structure and loading and already handle dynamic chunks, this is necessary to be the default.

I might consider ways to allow adding chunks but limit reading them from the outside of the bundle.

### DIFF
--- a/packages/webpack/src/plugin.js
+++ b/packages/webpack/src/plugin.js
@@ -260,7 +260,7 @@ class LavaMoatPlugin {
         // Adjust scuttling configuration to not scuttle webpack chunk loading facilities
         if (
           typeof STORE.options.scuttleGlobalThis === 'object' &&
-          STORE.options.scuttleGlobalThis.exceptions
+          Array.isArray(STORE.options.scuttleGlobalThis.exceptions)
         ) {
           STORE.options.scuttleGlobalThis.exceptions.push(
             compilation.outputOptions.chunkLoadingGlobal || 'webpackChunk'
--- a/packages/webpack/src/plugin.js
+++ b/packages/webpack/src/plugin.js
@@ -110,12 +110,14 @@ class LavaMoatPlugin {
    */
   constructor(options = {}) {
     if (options.scuttleGlobalThis === true) {
-      options.scuttleGlobalThis = { enabled: true }
+      options.scuttleGlobalThis = { enabled: true, exceptions: [] }
     } else if (typeof options.scuttleGlobalThis === 'object') {
       options.scuttleGlobalThis = { ...options.scuttleGlobalThis }
       if (Array.isArray(options.scuttleGlobalThis.exceptions)) {
         options.scuttleGlobalThis.exceptions =
           options.scuttleGlobalThis.exceptions.map((e) => e.toString())
+      } else {
+        options.scuttleGlobalThis.exceptions = []
       }
     }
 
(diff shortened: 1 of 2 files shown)


packages/webpack/src/plugin.js
```diff
@@ -255,6 +257,16 @@ class LavaMoatPlugin {
           )
         )
 
+        // Adjust scuttling configuration to not scuttle webpack chunk loading facilities
+        if (
+          typeof STORE.options.scuttleGlobalThis === 'object' &&
+          STORE.options.scuttleGlobalThis.exceptions
```
boneskull: [suggestion] if you're gonna check types, might as well check types

```suggestion
          Array.isArray(STORE.options.scuttleGlobalThis.exceptions)
```

packages/webpack/test/e2e-scuttle.spec.js
```diff
@@ -68,22 +76,53 @@ test(`webpack/scuttled - hosting globalThis's "Function" is scuttled`, async (t)
...
+  await scuttle(
+    t,
+    {
+      enabled: true,
+      scuttlerName,
+    },
+    {
+      [scuttlerName]: (globalRef, scuttle) => {
+        t.context.scuttler_func_called = true
+        scuttle(globalRef)
```
boneskull, line 123: does this line serve a purpose?

packages/webpack/test/e2e-scuttle.spec.js
```diff
@@ -68,22 +76,53 @@ test(`webpack/scuttled - hosting globalThis's "Function" is scuttled`, async (t)
...
+  }, 'Unexpected error in scenario')
+  t.truthy(t.context.globalThis.webpackChunkTEST)
+})
+
+test(`webpack/scuttled - webpackChunk global is transparently added to exceptions even when not specified 2`, async (t) => {
+  await scuttle(t, true)
+
+  t.notThrows(() => {
+    t.context.globalThis.webpackChunkTEST
+  }, 'Unexpected error in scenario')
```
boneskull, line 108: [suggestion] this is redundant with the next assertion

```suggestion
```

packages/webpack/test/e2e-scuttle.spec.js
```diff
@@ -36,9 +36,17 @@ async function scuttle(t, scuttleGlobalThis, globals) {
...
   }
+  // force webpack into creating chunks so that testing chunkApp globals is possible
+  webpackConfig.optimization.runtimeChunk = 'single'
+  // specify the chunk global namefor tests
+  webpackConfig.output.chunkLoadingGlobal = 'webpackChunkTEST'
+
   await t.notThrowsAsync(async () => {
     t.context.build = await scaffold(webpackConfig)
-    t.context.bundle = t.context.build.snapshot['/dist/app.js']
+    // Load both chunks effectively
```
boneskull, line 46: how does mashing them together like this work?

boneskull (approved):
nothing blocking, but does this need end-user documentation?

packages/webpack/test/e2e-scuttle.spec.js
naugtur: same as two script tags. They're prepared to run independently in a sequence and that's what they do.

packages/webpack/test/e2e-scuttle.spec.js
naugtur: It indicates that the configuration not working results in an error from scuttling. I'm not sure I like the raw error being thrown.
And the second assertion is for making sure that the first didn't succeed because there's no webpackChunkTEST field.
I'd consider changing the assertion message, but I'd like to keep the assertions.

packages/webpack/test/e2e-scuttle.spec.js
naugtur: it demonstrates how scuttling integrates with Snow.

naugtur:
> does this need end-user documentation?

It's a minute detail where we make sure scuttling won't break webpack. The user is already overwhelmed with warnings (I have testimonials) so I don't want to show this.

PR #1768, opened 2025-08-20
feat(webpack): isolate parts of the app to contain slop

naugtur:
just like that

That's one more thing to do settings for.


boneskull (approved):
I'm not sure what this does exactly, but it seems harmless.

PR #1776, opened 2025-08-25
chore: limit when renovate runs

naugtur:
Update Renovate config to schedule its work to avoid flooding actions with builds upon every push to main.


naugtur:
There's a flake in the tests. I'll stop rerunning it for now.

naugtur:
yay, npm is down

boneskull (commented):
What's the aim of this?  What's mostly happening is that there is a handful of PRs from Renovate that we can't merge yet, and it has a limit on how many can be open at once. There's a queue, which means as soon as we merge one it'll open the next in line.

naugtur:
@boneskull whenever you merge one of these (or any other PR) renovate kicks in and starves out everything else because it's updating its things.

boneskull (approved):

PR #1785, opened 2025-08-27
feat(webpack): repair for window.event deprecated leaky poperty

naugtur:
The full story was too much fun to leave untold

https://dev.to/naugtur/sometimes-the-best-thing-to-do-is-an-explicit-nothing-2i71


boneskull (approved):
obviously

PR #1786, opened 2025-08-27
feat(core): support Firefox contentscript shenanigans with cross-realm window prototypes

naugtur:
There's no impact on correctness and no breaking change. In the rare situation of finding a global that doesn't have `Object.prototype` in its prototype chain, we default to taking the whole prototype chain into account instead of panicking.

Potential improvement:
We could skip the `Object.prototype` from the other realm if we put enough effort into checking if it has all the same keys and is the last before reaching null. 
Could it be dangerous? We might be risking bringing in a fake `hasOwnProperty` on `globalThis` that someone in the other unlocked realm could modify.


packages/core/src/endowmentsToolkit.js
```diff
@@ -531,14 +531,23 @@ function endowmentsToolkit({
...
-    const commonPrototypeIndex = globalProtoChain.findIndex(
+    let commonPrototypeIndex = globalProtoChain.findIndex(
       (globalProtoChainEntry) => globalProtoChainEntry === Object.prototype
     )
-    if (commonPrototypeIndex === -1) {
-      // TODO: fix this error message
-      throw new Error(
-        'Lavamoat - unable to find common prototype between Compartment and globalRef'
+    // "Why would a global have no reference in prototype chain that matches Object.prototype?" you might ask.
+    // It's when the global is partially from a different Realm. And that's possible in extension contentscript in FireFox
```
boneskull, line 538: "global is partially from a different realm" is some acrobatics

packages/core/src/endowmentsToolkit.js
```diff
@@ -531,14 +531,23 @@ function endowmentsToolkit({
...
+    const noSharedPrototype = commonPrototypeIndex === -1
+    if (noSharedPrototype) {
+      // take the entire prototype chain
+      commonPrototypeIndex = globalProtoChain.length
+      // Skip overriding Object.prototype keys in case it's the other realm Object.prototype
+      const possiblyObjectProto = globalProtoChain.pop()
+      const replacement = Object.fromEntries(
+        Object.entries(possiblyObjectProto).filter(
+          ([key]) => key in Object.prototype
```
boneskull: This is intentional, right? We don't want `hasOwn` or w/e?

packages/core/test/endowmentsToolkit.spec.js
```diff
@@ -444,6 +444,25 @@ test('getEndowmentsForConfig - ensure setTimeout calls dont trigger illegal invo
...
+  sourceProto.onTheProto = function () {}
+  const source = Object.create(sourceProto)
+  source.onTheObj = function () {}
+  const target = Object.create(null)
+
+  // used to throw
+  // Error: Lavamoat - unable to find common prototype between Compartment and globalRef
+  copyWrappedGlobals(source, target, ['window'])
+
+  t.is(Object.keys(target).sort().join(), 'onTheObj,onTheProto,window')
```
boneskull, line 463: I hate this 😄

boneskull (approved):

packages/core/src/endowmentsToolkit.js
naugtur: this is there to make sure that we don't transfer methods from Object.ptorortype from a different realm if that's possible in the environment.
And it has a bug, it's missing a NOT. The test found it.

packages/core/src/endowmentsToolkit.js
naugtur: and also a fact of life for contentscripts

packages/core/test/endowmentsToolkit.spec.js
naugtur: Screenshot_2025-09-10_12-31-34
Screenshot_2025-09-10_12-30-50

Left is what this looks like, right is what deepequal on array looks like

PR #1787, opened 2025-08-29
fix(webpack): MessageEvent repair should not prevent further overrides

naugtur:
This fixes an unfortunate clash with Snow. The two overrides should stack correctly, because this repair is added first and it replaces window with compartmentGlobalThis if suitable, meanwhile Snow is replacing windows it recognizes with proxies that prevent setting fields. Snow will not recognize compartmentGlobalThis as a window it has a proxy for and return it as-is.

### DIFF
--- a/packages/webpack/src/runtime/repairs/messageevent.js
+++ b/packages/webpack/src/runtime/repairs/messageevent.js
@@ -29,7 +29,8 @@ if (original && original.get) {
         return w
       }
     },
-    configurable: false,
+    // left configurable because replacing it with original is impossible after and we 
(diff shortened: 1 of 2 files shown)


boneskull (approved):
LGTM, but:

1. A test to ensure it is configurable will prevent a regression
2. Maybe even make it configurable for the use-case where this is not paired with Snow.

naugtur:
>     1. A test to ensure it is configurable will prevent a regression
yup

>     2. Maybe even make it configurable for the use-case where this is not paired with Snow.
it doesn't override the default configurability at all now, so effectively it's always configurable:true

PR #1789, opened 2025-09-01
feat(webpack): add generatePolicyOnly option

naugtur:
It skips as much extra work and output as could comfortably be done without too much hacking.


packages/webpack/README.md
```diff
@@ -20,6 +20,7 @@ The LavaMoat plugin takes an options object with the following properties (all o
...
 | `policyLocation`           | Directory to store policy files in.                                                                                                                                                                                                                                                                                   | `./lavamoat/webpack`     |
 | `generatePolicy`           | Whether to generate the `policy.json` file. Generated policy is used in the build immediately. `policy-override.json` is applied before bundling, if present.                                                                                                                                                         | `false`                  |
+| `generatePolicyOnly`       | Enables `genratePolicy` and skips finishing the build (useful if you only need to regenerate policy files)                                                                                                                                                                                                            | `false`                  |
```
itsyoboieltr:
```suggestion
| `generatePolicyOnly`       | Enables `generatePolicy` and skips finishing the build (useful if you only need to regenerate policy files)                                                                                                                                                                                                            | `false`                  |
```

packages/webpack/src/plugin.js
```diff
@@ -196,21 +206,24 @@ class LavaMoatPlugin {
...
     // compiler.options.optimization.usedExports = false;
     // compiler.options.optimization.providedExports = false;
     // compiler.options.optimization.sideEffects = false;
-    // compiler.options.optimization.moduleIds = "hashed";
-    // compiler.options.optimization.chunkIds = "named";
     Object.freeze(STORE.options)
 
     // =======================================
 
+    // loadCanonicalNameMap depends on having a resolver. It'd be best to use webpack's own, but it's problematic and the discrepancies between resolvers are not on the package level, but individual exports, which has not tripped us up yet.
```
boneskull, line 196: what's problematic about `enhanced-resolve`?

packages/webpack/src/plugin.js
```diff
@@ -250,12 +263,28 @@ class LavaMoatPlugin {
...
         STORE.mainCompilationWarnings = compilation.warnings
-        assertFields(STORE, ['mainCompilationWarnings'])
-        STORE.mainCompilationWarnings.push(
-          new WebpackError(
-            'LavaMoatPlugin: Concatenation of modules disabled - not compatible with LavaMoat wrapped modules.'
+        assertFields(STORE, ['mainCompilationWarnings', 'options'])
+
+        if (STORE.options.generatePolicyOnly) {
+          compiler.options.devtool = false // source maps are expensive to make and unnecessary
+          compiler.hooks.shouldEmit.tap(PLUGIN_NAME, () => false)
```
boneskull, line 240: what does this do?

packages/webpack/src/plugin.js
```diff
@@ -250,12 +263,28 @@ class LavaMoatPlugin {
...
+          compilation.hooks.shouldGenerateChunkAssets.tap(
+            PLUGIN_NAME,
+            () => false
           )
-        )
+          // replacing generator with something returning an empty string could shave off some extra time, but is too invasive to seem worth it
+        }
+
+        if (
+          FORCED_CONCATENATEMODULES_OFF ||
```
boneskull, line 249: a screaming snake, cut in thirds?

packages/webpack/src/plugin.js
```diff
@@ -250,12 +263,28 @@ class LavaMoatPlugin {
...
+          STORE.options.diagnosticsVerbosity > 0
+        ) {
+          STORE.mainCompilationWarnings.push(
+            new WebpackError(
```
boneskull, line 253: creating an error for a warning is unintuitive to me, but that must be the webpack way. right??

packages/webpack/src/plugin.js
```diff
@@ -51,6 +51,12 @@ const JAVASCRIPT_MODULE_TYPE_AUTO = 'javascript/auto'
 const JAVASCRIPT_MODULE_TYPE_DYNAMIC = 'javascript/dynamic'
 const JAVASCRIPT_MODULE_TYPE_ESM = 'javascript/esm'
 
+const COVERED_MODULE_TYPES = [
+  JAVASCRIPT_MODULE_TYPE_AUTO,
+  JAVASCRIPT_MODULE_TYPE_DYNAMIC,
+  JAVASCRIPT_MODULE_TYPE_ESM,
+]
```
```suggestion
const COVERED_MODULE_TYPES = /** @type {const} */([
  JAVASCRIPT_MODULE_TYPE_AUTO,
  JAVASCRIPT_MODULE_TYPE_DYNAMIC,
  JAVASCRIPT_MODULE_TYPE_ESM,
])
```

boneskull (approved):
I wrote some stuff, but no big deals.

packages/webpack/src/plugin.js
naugtur: by webpack's own, I meant the exact resolution it's using at compile time. Generting canonicalNameMap from webpack graph would be great.

I need to look into using enhanced-resolve

packages/webpack/src/plugin.js
naugtur: webpack calls this hook to ask whether we want an asset to be emitted. This causes an empty dist.

packages/webpack/src/plugin.js
naugtur: visible enough? :)

packages/webpack/src/plugin.js
naugtur: yes, that's a convention they want - errors and warnings are arrays of Error

packages/webpack/src/plugin.js
naugtur: Oh, this is reaching for enhanced-resolve and has been left unused because enhanced-resolve doesn't have the sync option really - it complains about fs being async and won't let us use sync version. I couldn't find how to give it a sync fs.

PR #1792, opened 2025-09-01
feat(webpack): static shims for runtime

naugtur:
Adds an option to do vetted shims inbetween lockdown phases and including files before things start.

Useful for:
- polyfills to intrinsics / vetted shims
- including Snow with the runtime and setting it up

Could also expose:
- a way to run after lockdown 
- a way to define a per-compartment custom repair
But I'm not sure If I want those to be webpack specific.

I think none of this should be webpack specific. 

I just convinced myself to mark this feature as experimental.

### DIFF
--- a/packages/webpack/README.md
+++ b/packages/webpack/README.md
@@ -29,7 +29,7 @@ The LavaMoat plugin takes an options object with the following properties (all o
 | `diagnosticsVerbosity`     | Number property to represent diagnostics output verbosity. A larger number means more overwhelming diagnostics output.                                                                                                                                                                                                | `0`                      |
 | `debugRuntime`             | Only for local debugging use - Enables debugging tools that help detect gaps in generated policy and add missing entries to overrides                                                                                                                                                                                 | `false`                  |
 | `policy`                   | The LavaMoat policy object (if not loading from file; see `policyLocation`)                                                                                                                                                                                                                                           | `undefined`              |
-| `staticShims_experimental` | Standalone JS files to be added to the runtime chunk before lavamoat runtime starts and executes lockdown.                                                                                                                                                                                        | `undefined`              |
+| `staticShims_experimental` | Standalone JS files to be added to the runtime chunk before lavamoat runtime starts and executes lockdown.                                                                                                                                                                                                            | `undefined`              |
 
 ```js
 const LavaMoatPlugin = require('@lavamoat/webpack')
@@ -63,26 +63,20 @@ module.exports = {
   plugins: [
     new LavaMoatPlugin({
       staticShims_experimental: [
-        'package-name', 
-        path.join(__dirname, './local/file.js')
-      ]
+        'package-name', // a package whose main export is a built standalone script
+        path.join(__dirname, './local/file.js'),
+      ],
     }),
   ],
 }
 ```
 
-The static shims are executed in order before any other code runs.
-A shim can also run code between the repair and harden phases of SES lockdown. To do that, assign a synchronous function to LOCKDOWN_SHIM variable that exists in the scope of the shim.
-It's the only way to polyfill functionality on intrinsics.
-
-> [!IMPORTANT]
-> You must _assign_ to `LOCKDOWN_SHIM` without re-declaring it.
+The static shims are executed between the repair and harden phases of SES lockdown.
+It's the only way to polyfill functionality on intrinsics or run any privileged code outside of LavaMoat protections.
 
 ```js
-// lockdown-shim.js
-LOCKDOWN_SHIM = () => {
+// resolvers-shim.js
   Promise.withResolvers = ...
-}
 ```
 
 ### Excluding modules
--- a/packages/webpack/README.md
+++ b/packages/webpack/README.md
@@ -29,6 +29,7 @@ The LavaMoat plugin takes an options object with the following properties (all o
 | `diagnosticsVerbosity`     | Number property to represent diagnostics output verbosity. A larger number means more overwhelming diagnostics output.                                                                                                                                                                                                | `0`                      |
 | `debugRuntime`             | Only for local debugging use - Enables debugging tools that help detect gaps in generated policy and add missing entries to overrides                                                                                                                                                                                 | `false`                  |
 | `policy`                   | The LavaMoat policy object (if not loading from file; see `policyLocation`)                                                                                                                                                                                                                                           | `undefined`              |
+| `staticShims` | Standalone JS files to be added to the runtime chunk before lavamoat runtime starts and executes lockdown.                                                                                                                                                                                        | `undefined`              |
 
 ```js
 const LavaMoatPlugin = require('@lavamoat/webpack')
@@ -46,6 +47,40 @@ module.exports = {
 ```
 
 One important thing to note when using the LavaMoat plugin is that it disables the `concatenateModules` optimization in webpack. This is because concatenation won't work with wrapped modules.
+
+### Using static shims
+
+Static shims are a way to include additional code in the runtime chunk before LavaMoat starts.
+
+> [!WARNING]
+> Shims cannot use import or require, they must be standalone scripts.
+
+```js
+const LavaMoatPlugin = require('@lavamoat/webpack')
+const path = require('path')
+
+module.exports = {
+  plugins: [
+    new LavaMoatPlugin({
+      staticShims: [
+        'package-name', 
+        path.join(__dirname, './local/file.js')
+      ]
+    }),
+  ],
+}
+```
+
+The static shims are executed in order before any other code runs.
+A shim can also run code between the repair and harden phases of SES lockdown. To do that, assign a synchronous function to LOCKDOWN_SHIM variable that exists in the scope of the shim.
+It's the only way to polyfill functionality on intrinsics.
+
+```js
+// lockdown-shim.js
+LOCKDOWN_SHIM = () => {
+  Promise.withResolvers = ...
+}
+```
 
 ### Excluding modules
 
--- a/packages/webpack/README.md
+++ b/packages/webpack/README.md
@@ -75,6 +75,9 @@ The static shims are executed in order before any other code runs.
 A shim can also run code between the repair and harden phases of SES lockdown. To do that, assign a synchronous function to LOCKDOWN_SHIM variable that exists in the scope of the shim.
 It's the only way to polyfill functionality on intrinsics.
 
+> [!IMPORTANT]
+> You must _assign_ to `LOCKDOWN_SHIM` without re-declaring it.
+
 ```js
 // lockdown-shim.js
 LOCKDOWN_SHIM = () => {
--- a/packages/webpack/README.md
+++ b/packages/webpack/README.md
@@ -29,7 +29,7 @@ The LavaMoat plugin takes an options object with the following properties (all o
 | `diagnosticsVerbosity`     | Number property to represent diagnostics output verbosity. A larger number means more overwhelming diagnostics output.                                                                                                                                                                                                | `0`                      |
 | `debugRuntime`             | Only for local debugging use - Enables debugging tools that help detect gaps in generated policy and add missing entries to overrides                                                                                                                                                                                 | `false`                  |
 | `policy`                   | The LavaMoat policy object (if not loading from file; see `policyLocation`)                                                                                                                                                                                                                                           | `undefined`              |
-| `staticShims` | Standalone JS files to be added to the runtime chunk before lavamoat runtime starts and executes lockdown.                                                                                                                                                                                        | `undefined`              |
+| `staticShims_experimental` | Standalone JS files to be added to the runtime chunk before lavamoat runtime starts and executes lockdown.                                                                                                                                                                                        | `undefined`              |
 
 ```js
 const LavaMoatPlugin = require('@lavamoat/webpack')
@@ -62,7 +62,7 @@ const path = require('path')
 module.exports = {
   plugins: [
     new LavaMoatPlugin({
-      staticShims: [
+      staticShims_experimental: [
         'package-name', 
         path.join(__dirname, './local/file.js')
       ]
--- a/packages/webpack/src/runtime/assemble.js
+++ b/packages/webpack/src/runtime/assemble.js
@@ -70,7 +70,7 @@ function prepareSource(specifier) {
 
 /**
  * @param {string} KEY
- * @param {RuntimeFragment[]} runtimeModules
+ * @param {readonly RuntimeFragment[]} runtimeModules
  * @returns {string}
  */
 const assembleRuntime = (KEY, runtimeModules) => {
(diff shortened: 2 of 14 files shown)


packages/webpack/README.md
```diff
@@ -47,6 +48,40 @@ module.exports = {
...
+
+The static shims are executed in order before any other code runs.
+A shim can also run code between the repair and harden phases of SES lockdown. To do that, assign a synchronous function to LOCKDOWN_SHIM variable that exists in the scope of the shim.
+It's the only way to polyfill functionality on intrinsics.
+
+```js
+// lockdown-shim.js
+LOCKDOWN_SHIM = () => {
+  Promise.withResolvers = ...
+}
```
naugtur: I'd appreciate comments on the design of this.

I could handle CJS style export instead, but tried to do something else so that I don't give people an idea that the module system might be working where it's not.

I might have put too much effort into making sure that the array of shims is lexically unavailable to the shim. As a result, if you redefine `LOCKDOWN_SHIM` instead of assigning to it, you fail to register a shim.
Is that a problem? I could fix that.

packages/webpack/src/runtime/assemble.js
```diff
@@ -24,29 +75,22 @@ const assembleRuntime = (KEY, runtimeModules) => {
...
-    if ((rawSource || shimRequire) && sourceString) {
-      sourceString = removeMultilineComments(sourceString)
-      sourceString = `;(()=>{
-        const module = {exports: {}};
-        const exports = module.exports;
-          ${sourceString}
-        ;
-        LAVAMOAT['${name}'] = module.exports;
-      })();`
+    } else {
```
naugtur, line 91: With the new implementation of removing comments that actually works correctly I needed to prevent it from attempting to parse JSON, so this rearrangement of logic here was necessary.

packages/webpack/test/assemble.spec.js
```diff
@@ -0,0 +1,154 @@
+const test = /** @type {import('ava').TestFn} */ (require('ava'))
+const { removeMultilineComments } = require('../src/runtime/assemble')
+
```
naugtur, line 3: This file mostly lists reasons why `removeMultilineComments` should not be implemented with a  regexp

packages/webpack/src/runtime/assemble.js
```diff
@@ -1,18 +1,69 @@
...
-  return source.replace(/\/\*[\s\S]*?\*\//g, '')
+  try {
+    const ast = parse(source, { sourceType: 'script' })
+    return (ast.comments || [])
+      .filter((c) => c.type === 'CommentBlock')
+      .reverse()
+      .reduce(
+        (src, { start, end }) => src.slice(0, start) + src.slice(end),
+        source
+      )
```
naugtur, line 21: slicing them out is going to be much faster than regenerating the code from AST 
I'm also wondering if it would make sense to replace them with spaces, which is a fashionable choice lately.

Whatever character counting issues exist in JS it seems they align between the parser and `slice`

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -16,191 +60,335 @@ module.exports = {
...
-          } else {
-            repairs = require('./repairsBuilder.js').buildRepairs(
-              policyData,
-              options.skipRepairs
-            )
+          runtimeConfiguration = {
+            mode: 'unlocked_unsafe',
           }
+        }
+        if (options.runtimeConfigurationPerChunk_experimental) {
```
naugtur, line 326: It was such a low hanging fruit after the refactor, that I went and did it.

naugtur:
* **#1794** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1794?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1792** <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1792?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/LavaMoat/LavaMoat/1792?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/webpack/src/runtime/assemble.js
```diff
@@ -1,21 +1,72 @@
 const { readFileSync } = require('node:fs')
+const { parse } = require('@babel/parser')
 
 /**
  * @param {string} source
  * @returns {string}
  */
 function removeMultilineComments(source) {
-  return source.replace(/\/\*[\s\S]*?\*\//g, '')
+  try {
```
boneskull, line 13: you explained it to me, but it's worth mentioning in a comment why we need to remove multiline comments

packages/webpack/src/runtime/assemble.js
```diff
@@ -1,21 +1,72 @@
...
+
+/**
+ * Wraps the source code for a given module in a function that simulates the
+ * CommonJS module system.
+ *
+ * @param {string} sourceString - The source code to wrap.
+ * @param {string} [name] - The name of the module.
+ * @returns {string} The wrapped source code.
+ */
+function shimSource(sourceString, name = 'unknown') {
```
boneskull, line 51: this might be a problem if two sources lack names

packages/webpack/src/runtime/assemble.js
```diff
@@ -68,4 +112,4 @@ const assembleRuntime = (KEY, runtimeModules) => {
  * @property {string} [shimRequire]
  */
 
-module.exports = { assembleRuntime }
+module.exports = { assembleRuntime, prepareSource, removeMultilineComments }
```
boneskull, line 119: are the new exports used elsewhere or is this for testing?

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -1,11 +1,55 @@
...
-const { assembleRuntime } = require('./assemble.js')
+const { assembleRuntime, prepareSource } = require('./assemble.js')
 const path = require('node:path')
 
-/** @import {LavaMoatPluginOptions} from '../buildtime/types' */
+const { RuntimeModule } = require('webpack')
+
+class VirtualRuntimeModule extends RuntimeModule {
+  /**
+   * @param {Object} options - The options for the VirtualRuntimeModule.
```
boneskull: (style suggestion)

The extra `-` is not treated as special by VSCode's on-hover documentation panel, which means at least in _that_ context it is unhelpful.  I would probably omit these.

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -1,11 +1,55 @@
...
+  }) {
+    super(name, stage)
+    this.withoutClosure = withoutClosure
+    this.virtualSource = `;${source};`
+  }
+  shouldIsolate() {
+    return !this.withoutClosure
+  }
+
+  generate() {
```
boneskull, line 38: (suggestion)

tag methods with `@override` if they are overrides of the superclass

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -1,11 +1,55 @@
...
+/** @import {LavaMoatPluginOptions, LavaMoatChunkRuntimeConfiguration} from '../buildtime/types' */
 /** @import {LavaMoatPolicy} from 'lavamoat-core' */
-/** @import {RuntimeModule} from './assemble.js' */
+/** @import {RuntimeFragment} from './assemble.js' */
+/** @import {Chunk} from 'webpack' */
+
+/**
+ * @typedef {Object} LavaMoatRuntimeIdentifiers
+ * @property {string} root - Root identifier
+ * @property {[string, (string | number)[]][]} identifiersForModuleIds - Module
```
boneskull: if this is a data structure we own, maybe we should change it.  at the very least, since it's a tuple, we can give the members names, e.g.:

```ts
type Thing = [foo: string, bar: (string|number)[]][]
```

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -16,191 +60,335 @@ module.exports = {
...
+      const globals = [
+        'location', // AutoPublicPathRuntimeModule.js
+        'setTimeout', // LoadScriptRuntimeModule.js
+        'clearTimeout', // LoadScriptRuntimeModule.js
+        'document', // LoadScriptRuntimeModule.js, AutoPublicPathRuntimeModule.js
+        'trustedTypes', // GetTrustedTypesPolicyRuntimeModule.js
+        'self',
+      ]
+      return `var ${globals.map((g) => `${g} = globalThis.${g}`).join(',')};
+const LOCKDOWN_SHIMS = [];`
```
boneskull, line 85: This doesn't appear to capture any variables from its function context, so maybe worth pulling out?

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -16,191 +60,335 @@ module.exports = {
...
+          })();
+          if(typeof LOCKDOWN_SHIM === 'function') {
+            LOCKDOWN_SHIMS.push(LOCKDOWN_SHIM)
+          };
+        }`
+
+      const shims = dependencies.map((dep) => {
+        const source = prepareSource(dep)
+        try {
+          new Function(source)
```
boneskull, line 112: why is this check needed?

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -16,191 +60,335 @@ module.exports = {
...
+        {
+          name: 'ENUM',
+          file: require.resolve('../ENUM.json'),
+          json: true,
+        },
+        {
+          name: 'runtime',
+          file: require.resolve('./runtimeUnlocked.js'),
+        },
+      ]
```
boneskull: this surfaces the intent a bit better, imo. that `@satisfies` might be `readonly [RuntimeFragment, RuntimeFragment]` even

```suggestion
      /** @satisfies {RuntimeFragment[]} */
      const runtimeFragments = /** @type {const} */([
        {
          name: 'ENUM',
          file: require.resolve('../ENUM.json'),
          json: true,
        },
        {
          name: 'runtime',
          file: require.resolve('./runtimeUnlocked.js'),
        },
      ])
```

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -16,191 +60,335 @@ module.exports = {
...
+
+      return unlockedRuntime
+    }
+
+    /**
+     * Generates the LavaMoat runtime source code based on chunk configuration
+     *
+     * @param {Object} params - The parameters object
+     * @param {LavaMoatChunkRuntimeConfiguration['embeddedOptions']} params.embeddedOptions
+     *   - The options to embed in the
```
boneskull: ...?

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -16,191 +60,335 @@ module.exports = {
...
+      if (options.skipRepairs === true) {
+        repairs = ''
+      } else {
+        repairs = require('./repairsBuilder.js').buildRepairs(
+          policyData,
+          options.skipRepairs
+        )
+      }
+
+      /** @type {RuntimeFragment[]} */
```
boneskull: likewise, I presume this is used as a constant instead of a mutable array

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -16,191 +60,335 @@ module.exports = {
...
+        // main lavamoat runtime
+        {
+          name: 'runtime',
+          file: require.resolve('./runtime.js'),
+        },
+      ]
+
+      if (options.debugRuntime) {
+        // optional debug helpers
+        runtimeFragments.push({
```
boneskull: (you'd have to do this inline, though)

packages/webpack/README.md
boneskull: By redefine you mean assigning a second time, or using `Obj.defineProps`?

IMO this is fine to get in front of MM and see what they think. If they complain about the API, then we can change it. _It doesn't have to be perfect the first time_ is what I'm saying.  Iterate!

boneskull (approved):
No blockers, but some suggestions / questions per usual.

packages/webpack/README.md
```diff
@@ -47,6 +48,40 @@ module.exports = {
...
+      ]
+    }),
+  ],
+}
+```
+
+The static shims are executed in order before any other code runs.
+A shim can also run code between the repair and harden phases of SES lockdown. To do that, assign a synchronous function to LOCKDOWN_SHIM variable that exists in the scope of the shim.
+It's the only way to polyfill functionality on intrinsics.
+
```
rekmarks, line 76:
```suggestion

> [!IMPORTANT]
> You must _assign_ to `LOCKDOWN_SHIM` without re-declaring it.

```

packages/webpack/README.md
rekmarks: Maybe name the variable `PRE_LOCKDOWN_SHIM`?

The API is a bit strange, but it will work!

packages/webpack/src/runtime/assemble.js
naugtur: AFAICT the name is only used for a comment on top of the code.

packages/webpack/README.md
naugtur: >  get in front of MM and see what they think

They need shims before lockdown, not inbetween lockdown and are unlikely to have the need for those. We need to look at prospective users here :D 

> Maybe name the variable PRE_LOCKDOWN_SHIM?

the top-level code runs before lockdown. the function assigned runs between `repair` and `harden` stages of lockdown. PRE_LOCKDOWN sounds more like the former to me.

naugtur: oh no. I just realized if I wanted this to support people adding intrinsics polyfills from the ecosystem, they won't use anything I put in the scope :( 

Gotta redesign anyway.

naugtur: redesigned

packages/webpack/src/runtime/assemble.js
naugtur: prepareSource is used for shims. multiline comments is not used beyond tests at this point

packages/webpack/src/runtime/runtimeBuilder.js
naugtur: I considered that, but it being here the proximity between `LOCKDOWN_SHIMS` being defined and used is better and I'm anticipating the funciton will use `options`  from higher scope eventually.

packages/webpack/src/runtime/runtimeBuilder.js
naugtur: It makes sure the JS is correct before concatenation. We don't want it to contain a surplus if closing braces.

packages/webpack/src/runtime/runtimeBuilder.js
naugtur: It makes things more cognitively complex to read but somehow still nicer.

packages/webpack/src/runtime/runtimeBuilder.js
naugtur: done.

PR #1793, opened 2025-09-02
fix(webpack): fix the naive assumption that there can only be 1 module with a certain path

naugtur:
todo:
- [x] add a basic test with layers
  ~~- [ ] make it use the shared chunk properly~~
- [x] expand layers test to also cover dependencies and policy
- [x] analyze if the same naive assumption was made anywhere else

layers compatibility:
- path-moduleId relationship fixed
- more control of runtime for chunks is needed to effectively use layers for running parts of the project without lavamoat and the pairing of modules in the layer with runtime needs to be consciously made by the dev configuring webpack, there's no way to reasonably discover it.
- manual runtime config possible via: (https://github.com/LavaMoat/LavaMoat/pull/1792/files#diff-3dde54dcaa2e2055be732a506ef2d299f60bd2f6951e325304a0d04cea5562c8R326)


packages/webpack/test/scaffold.js
```diff
@@ -115,6 +115,7 @@ const defaultGlobals = () => ({
   location: { href: 'https://localhost/' }, // necessary for webpack's runtime
   URL,
 })
+exports.getDefaultGlobalsForRunScript = defaultGlobals
```
naugtur, line 118: this will be useful in other tests and I'll propagate it later

packages/webpack/src/buildtime/aa.js
```diff
@@ -79,16 +79,21 @@ exports.generateIdentifierLookup = ({
...
+        if (mapping[p.path]) {
+          mapping[p.path].moduleIds.push(p.moduleId)
+          diag.rawDebug(2, `Duplicated moduleId at path ${ p.path }, moduleIds: ${ mapping[p.path].moduleIds.join(', ') }`)
+        } else {
+          mapping[p.path] = {
+            aa: getPackageNameForModulePath(canonicalNameMap, p.path),
+            moduleIds: [p.moduleId],
+          }
         }
       }
```
boneskull: (suggestion) avoid nondescript variable name

```suggestion
  for (const {path, moduleId} of paths) {
    if (path) {
      if (mapping[path]) {
        mapping[path].moduleIds.push(moduleId)
        diag.rawDebug(2, `Duplicated moduleId at path ${ path }, moduleIds: ${ mapping[path].moduleIds.join(', ') }`)
       } else {
         mapping[path] = {
           aa: getPackageNameForModulePath(canonicalNameMap, path),
           moduleIds: [moduleId],
         }
       }
    }
```

packages/webpack/test/fixtures/main/webpack.config.js
```diff
@@ -18,6 +18,11 @@ Object.defineProperty(module.exports, 'makeConfig', {
   value: makeConfig,
 })
 
+/**
+ *
+ * @param {import('../../../src/plugin.js').LavaMoatPluginOptions} lmOptions
```
boneskull: (suggestion) `@import`

packages/webpack/test/snapshots/e2e-layers.spec.js.md
```diff
@@ -0,0 +1,18 @@
...
+The actual snapshot is saved in `e2e-layers.spec.js.snap`.
+
+Generated by [AVA](https://avajs.dev).
+
+## webpack/path-moduleId - bundle files complete
+
+> Snapshot 1
+
+    [
+      '/dist/4892327f4fdaee1a826b.js',
```
boneskull, line 12: Is this deterministic?

packages/webpack/test/e2e-layers.spec.js
```diff
@@ -0,0 +1,87 @@
+const test = /** @type {import('ava').TestFn} */ (require('ava'))
```
boneskull, line 1: (suggestion) I'm not certain, but I don't think these tests are type-checked. If adding the directive causes type errors, then they weren't. 😄 I think you might need to type-assert `require('ava')` as `unknown` before type-asserting it to `TestFn`.

also: `@import`

```suggestion
// @ts-check

const test = /** @type {import('ava').TestFn} */ (require('ava'))
```

packages/webpack/test/e2e-layers.spec.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  await t.notThrowsAsync(async () => {
+    t.context.build = await scaffold(config, { writeFS: false })
+  }, 'Expected the build to succeed')
+})
+
+test('webpack/path-moduleId - bundle files complete', (t) => {
+  // Note '/dist/4892327f4fdaee1a826b.js' is not actually a chunk, but an asset that gets emitted from the URL constructor
+  t.snapshot(Object.keys(t.context.build.snapshot))
```
boneskull, line 70: (suggestion) looking at the snapshot, I'm not sure a snapshot is necessary here.

packages/webpack/test/scaffold.js
```diff
@@ -5,6 +5,15 @@ const { toSnapshotSync } = require('memfs/lib/snapshot')
 const { createContext, runInNewContext } = require('node:vm')
 const { readFileSync } = require('node:fs')
 
+/**
+ * Concatenates chunks to pretend they have loaded successfully, avoids attempts
+ * by webpack to dynamically load
+ *
+ * @param {string[]} chunks
+ * @returns
```
boneskull:
```suggestion
 * @returns {string}
```

packages/webpack/test/scaffold.js
```diff
@@ -132,11 +140,11 @@ const defaultGlobals = () => ({
  * @param {Record<string, any>} globals
  * @returns {RunReturn}
  */
-function runScript(code, globals = defaultGlobals()) {
+function runScript(code, globals) {
   if (typeof code !== 'string' || code === '') {
     throw new Error('runScript requires a bundle string as the first argument')
   }
-  const context = createContext(globals)
+  const context = createContext(Object.assign({}, defaultGlobals(), globals))
```
boneskull: (suggestion)

```suggestion
  const context = createContext({...defaultGlobals(), ...globals})
```

packages/webpack/test/scaffold.js
```diff
@@ -145,14 +153,23 @@ function runScript(code, globals = defaultGlobals()) {
 }
 exports.runScript = runScript
 
+/**
```
boneskull, line 156: `chunks` is just an array of sources, or?

boneskull (approved):
no blockers

packages/webpack/src/buildtime/aa.js
naugtur: something is off with this suggestion diff, added the change manually.

packages/webpack/test/e2e-layers.spec.js
naugtur: It's useful as:
- a dicumentation of what the dist is
- an early warning if changing the main webpack config for tests will cause this to get split into chunks differently.

packages/webpack/test/scaffold.js
naugtur: yes. All it does is concatenate them in a way that doesn't risk messing thiungs up even if something ends with a comment or starts with a brace

packages/webpack/test/snapshots/e2e-layers.spec.js.md
naugtur: yes. it's a simple hash

packages/webpack/test/e2e-layers.spec.js
naugtur: it does change something, but changes to tsconfig would be required anyway - it starts complaining about the ava line first.

PR #1804, opened 2025-09-11
feat(webpack): null runtime

naugtur:
Opens the door to having a part of the webpack project not covered at all by specifying:
- a layer that uses exclude loader
- an entry with the layer
- a null runtime

This is part of the experimental runtime hook thing, but I have an idea how to make it really simple to configure brewing.


boneskull (approved):
this could use a test imo

naugtur:
Added a check to fail compilation if someone accidentally set up the whole project to use null runtime and exclude.

PR #1816, opened 2025-09-19
drop all global types

naugtur:
I believe these are unnecessary and potentially harmful.

When reviewing, please check if this could be solved better by putting things in @lavamoat/types


packages/core/src/endowmentsToolkit.js
```diff
@@ -1,5 +1,9 @@
 // @ts-check
 
+/**
+ * @import { LMPolicy } from '@lavamoat/types/policy'
+ */
+
 /**
  * Utilities for generating the endowments object based on a `globalRef` and a
  * {@link LMPolicy.PackagePolicy}.
```
naugtur, line 9: we could go further and delete the namespace now that we have `@import`

boneskull (approved):
Fine by me as long as it compiles.

PR #1820, opened 2025-09-22
fix(core): noSharedPrototype case in copyWrappedGlobals must include non-enumerable properties

naugtur:
- applies the filtering on the descriptor instead of sloppily making a copy of the source object instead.

### DIFF
--- a/packages/core/src/endowmentsToolkit.js
+++ b/packages/core/src/endowmentsToolkit.js
@@ -540,14 +540,6 @@ function endowmentsToolkit({
     if (noSharedPrototype) {
       // take the entire prototype chain
       commonPrototypeIndex = globalProtoChain.length
-      // Skip overriding Object.prototype keys in case it's the other realm Object.prototype
-      const possiblyObjectProto = globalProtoChain.pop()
-      const replacement = Object.fromEntries(
-        Object.entries(possiblyObjectProto).filter(
-          ([key]) => !(key in Object.prototype)
-        )
-      )
-      globalProtoChain.push(replacement)
     }
     // we will copy endowments from all entries in the prototype chain, excluding Object.prototype
     const endowmentSources = globalProtoChain.slice(0, commonPrototypeIndex)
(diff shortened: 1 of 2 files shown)


boneskull (approved):
OK, but I'm going to need you to walk me through wtf is going on here.

packages/core/src/endowmentsToolkit.js
```diff
@@ -567,8 +559,22 @@ function endowmentsToolkit({
...
+          // might be an Object.prototype from another realm. In that case we want to skip overriding Object.prototype keys
+          const filteredDescriptors = Object.fromEntries(
+            Object.entries(candidateDescriptors).filter(
+              ([key]) => !(key in Object.prototype)
+            )
+          )
+          return filteredDescriptors
+        }
+        return candidateDescriptors
+      }
```
naugtur, line 577: @boneskull 
> OK, but I'm going to need you to walk me through wtf is going on here.

What happens in this function is we're taking the global and its prototype chain and copying all of the properties as own properties onto the root compartment globalThis. It'd be weird to see methods from  `Object.prototype` as own keys on globalThis, so we're attempting to skip it.

In browser extensions, if you register a contentscript, it's running in a realm that has access to the current website's `document` but does not share the `window` and intrinsics with the page. 

The way Firefox implements that is by truncating the prototype chain of globalThis and therefore after 
```js
   let commonPrototypeIndex = globalProtoChain.findIndex(
      (globalProtoChainEntry) => globalProtoChainEntry === Object.prototype
    )
```
`commonPrototypeIndex` is `-1`

Realistically there are two possible reasons it could be happening - the Firefox case (there's no `Object.prototype` in the chain) or the prototype chain of global has an `Object.prototype` at the bottom, but from a different realm, so it's not strict equal the  `Object.prototype`  of the current realm. In that case we want to avoid grafting methods from  `Object.prototype` from other realm as own properties of the global.
Normally we avoid that by skipping `Object.prototype` when found.

It's very unlikely to ever happen - unlikely even at the scale of what we're doing, but it's here because it'd be really hard to figure out it's happening.

naugtur: The change this PR introduces is about keeping all own properties, not just the enumerable ones as the previous implementation was accidentally skipping them.

PR #1823, opened 2025-09-22
feat(allow-scripts): change the recommended use of allow-scripts to global install

naugtur:
This PR incorporates the change from https://github.com/LavaMoat/LavaMoat/pull/1822


boneskull (approved):
OK, but needs conflict resolution

PR #1827, opened 2025-09-23
feat(webpack): prevent webpack from eliminating reexports and failing policy enforcement

naugtur:
Add one more config override `sideEffects: false` - that prevents webpack from messing up policy enforcement on packages dependencies

### DIFF
--- a/packages/webpack/test/fixtures/main/webpack.config.js
+++ b/packages/webpack/test/fixtures/main/webpack.config.js
@@ -49,7 +49,9 @@ function makeConfig(lmOptions = {}) {
       }),
       new HtmlWebpackPlugin(),
     ],
-    optimization: {},
+    optimization: {
+      concatenateModules: true, // only here to trigger explicit warning about being set back to false
+    },
     resolve: {
       fallback: { crypto: false },
     },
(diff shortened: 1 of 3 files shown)


packages/webpack/test/fixtures/main/webpack.config.js
```diff
@@ -49,7 +49,9 @@ function makeConfig(lmOptions = {}) {
       }),
       new HtmlWebpackPlugin(),
     ],
-    optimization: {},
+    optimization: {
+      concatenateModules: true, // only here to trigger explicit warning about being set back to false
```
boneskull, line 53: What of `sideEffects`?

boneskull (approved):

packages/webpack/test/fixtures/main/webpack.config.js
naugtur: The flag seems to be only doing one thing - collapsing reexports. 
If a module reexports another module clearly, it will get dropped in favor of webpack putting a direct import in the module importing the reexport. That messes up policy enforcement for the negligible benefit of not including the extra module that reexports things.

It's not a huge loss. We need to turn it off.

PR #1829, opened 2025-09-24
feat(webpack): document and officially release scuttling in webpack plugin

naugtur:
types diverged again into two separate ones on conflict resolution with @lavamoat/types work. Fixing it. 
I'm also using the opportunity that it was not documented earlier to skip exposing the deprecated boolean-only scuttling.

### DIFF
--- a/packages/core/src/scuttle.d.ts
+++ b/packages/core/src/scuttle.d.ts
@@ -1,9 +1,12 @@
 /**
  * Options for scuttling global properties
  */
-export interface ScuttleOpts {
+export type LavaMoatScuttleOpts = {
+  /** Whether scuttling is enabled or not. */
   enabled: boolean;
+  /** List of properties to exclude from scuttling. */
   exceptions?: Array<string | RegExp>;
+  /** Name of the scuttler function to use which is expected to be found as a property on the global object (e.g. if scuttlerName is 'x', scuttler function is obtained from globalThis['x']). */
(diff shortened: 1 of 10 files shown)


packages/core/src/scuttle.d.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * Options for scuttling global properties
  */
-export interface ScuttleOpts {
+export type LavaMoatScuttleOpts = {
```
boneskull, line 4: Why?

packages/core/src/scuttle.js
```diff
@@ -1,5 +1,5 @@
 /**
- * @typedef {object} ScuttleOpts
+ * @typedef {object} LavaMoatScuttleOpts
```
boneskull: This seems redundant?

boneskull (commented):
Approved, but seems fishy

boneskull (approved):

packages/core/src/scuttle.d.ts
naugtur: Interface can be expanded on. I don't want that possibility to exist.

The rename is there because it's a public facing type

packages/core/src/scuttle.js
naugtur: Umm. Yes. I think I got something back in a merge. I'll look into it

PR #1832, opened 2025-09-26
secret prototype repair for FF contentscript sandbox

naugtur:
This fixes the situation where in contentscript (see: https://dev.to/naugtur/the-missing-link-3ejk) the EventTarget methods are not possible to list by copyWrappedGlobals. But the fix is generic and will apply successfully to any such sandbox.

### DIFF
--- a/packages/webpack/src/runtime/repairs/globalthis.js
+++ b/packages/webpack/src/runtime/repairs/globalthis.js
@@ -1,0 +1,71 @@
+const { getOwnPropertyNames } = Object
+
+/**
+ * Util for getting the prototype chain as an array includes the provided value
+ * in the result
+ *
+ * @param {any} value
+ * @returns {any[]}
+ */
+function getPrototypeChain(value) {
+  const protoChain = []
+  let current = value
+  while (
+    current &&
+    (typeof current === 'object' || typeof current === 'function')
+  ) {
+    protoChain.push(current)
+    current = Reflect.getPrototypeOf(current)
+  }
+  return protoChain
+}
+/**
+ * For odd sandboxing situations including Firefox contentscript environment,
+ * where the globalThis is a sandbox with a truncated prototype chain that
+ * magically inherits properties from a hidden part of the prototype chain.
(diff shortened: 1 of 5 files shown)


packages/webpack/src/runtime/repairs/globalthis.js
```diff
@@ -0,0 +1,71 @@
...
+    const considerKeys = windowProtoChain.flatMap((proto) =>
+      getOwnPropertyNames(proto)
+    )
+
+    const globalKeys = new Set(
+      globalProtoChain.flatMap((proto) => getOwnPropertyNames(proto))
+    )
+
+    for (const key of considerKeys) {
+      // this condition looks self-contradictory, but it's actually explicitly testing for fields that are reachable on globalRef by explicit lookup but cannot be discovered as own properties on the prototype chain items.
```
boneskull, line 53: 😝

packages/webpack/src/runtime/repairs/globalthis.js
```diff
@@ -0,0 +1,71 @@
...
+      if (
+        !globalKeys.has(key) &&
+        key in globalThis &&
+        typeof win[key] !== 'undefined'
+      ) {
+        try {
+          // Make a property that exists on the hidden prototype chain an own property of globalThis so that the regular logic of copying globals can find it.
+          // The try-catch exists to avoid errors if a property is a getter-setter pair that we didn't know existed.
+          // eslint-disable-next-line no-self-assign
+          globalRef[key] = globalRef[key]
```
boneskull, line 63: this might be the best line of code I've ever seen written.  at least top 5

boneskull (approved):
absurd

packages/webpack/src/runtime/repairs/globalthis.js
naugtur: I think I'll report it to FF for a chance to remove this code. Not sure why they did what they did. There's nothing special in EventTarget.prototype AFAIK

PR #1833, opened 2025-09-29
fix(webpack): use the final module info in connections to handle the optimizations webpack sideEffect option does

naugtur:
This PR effectively adds support for the `sideEffects` optimization in webpack.

Policy generation goes through module and their "connections" as webpack calls them.
Turns out `resolvedModule` is not the final one, but the first module resolved for this connection. When optimizations happen, they replace the `module` on the `connection`. It happens eg. when treeshaking removes an intermediate dependency that was reexporting what the current module actually depends on.

The changes webpack makes to the dependency graph may be confusing (a policy is generated indicating a module can import something it never imports in its source code) so we've added `meta` field to the generated policy with explanations.

Additionally, a runtime change is added that makes webpack `fallback` modules work (when `Buffer` is provided by `buffer` package explicitly as a fallback, it's through a module type webpack calls"external" and these were only supported as builtins to date, but are not related to working under node.)


packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -113,9 +113,49 @@ module.exports = {
...
+                '>>>reexport collapsed>>>',
+                // @ts-expect-error - bad types?
+                connection.module?.userRequest,
+                // @ts-expect-error - bad types?
+                connection?.resolvedModule?.userRequest,
+              ])
+            }
             // @ts-expect-error - bad types?
-            const depSpecifier = dep.resolvedModule.userRequest
+            const depSpecifier = connection.module?.userRequest
```
naugtur, line 184: the change from `resolvedModule` to `module` is the change this was all about.

Turns out `resolvedModule` is not the final one, but the first module resolved for this connection. When optimizations happen, they replace the `module` on the connection. It happens when treeshaking removes an intermediate dependency that was reexporting what the current module actually depends on.

packages/webpack/src/plugin.js
```diff
@@ -187,13 +189,10 @@ class LavaMoatPlugin {
...
     }
-    if (compiler.options.optimization.sideEffects) {
-      FORCED_CONFIG.push('sideEffects=true->false')
-    }
     // Concatenation won't work with wrapped modules. Have to disable it.
     compiler.options.optimization.concatenateModules = false
-    // package policy enforcement won't work if webpack takes shortcuts
-    compiler.options.optimization.sideEffects = false
+    // This setting creates a bit of confusion when it removes a reexport between packages and collapses the dependency tree, but it's got a potential to optimize some bundles a lot. We can support it by making sure we use the optimized connections when generating policy.
+    // compiler.options.optimization.sideEffects;
```
naugtur, line 195: We no longer need to force `sideEffects` off. And that's good for bundle size.

The change in size was negligible in MM Extension, but 4X in Snaps.

packages/webpack/src/runtime/runtime.js
```diff
@@ -105,24 +105,33 @@ const enforcePolicy = (specifier, referrerResourceId, wrappedRequire) => {
...
+          referrerPolicy.builtin
+        )
+      }
+    } 
+    if (referrerPolicy.packages) {
+      // if an external was automatically generate by webpack and is not recognized as builtin, but allowed as a package, we still need to pass it in.
+      const requestedResourceId = findResourceId(externalName)
+      if (requestedResourceId && referrerPolicy.packages[requestedResourceId]) {
+        return wrappedRequire()
+      }
```
naugtur, line 137: This is a different fix for an issue that I found when diagnosing this sideEffects thing. If I wanted to recreate some polyfilling action webpack does by default with sideEffects, I got policy generated with a module that should be treated as a builtin listed in `"packages"` because it was the browser implementation (of buffer in that case) 

This change is a compatibility improvement. 

Need to think some more about security implications.

packages/types/src/policy-schema/lavamoat-policy.v0-0-1.schema.ts
```diff
@@ -135,6 +135,10 @@ export interface ResourcePolicy {
    * Allow native modules
    */
   native?: boolean
+  /**
+   * Metadata about the resource
+   */
+  meta?: Record<string, string|string[]>
```
naugtur: I considered `meta` and `notes`
The field is keyed so that it merges well.

Useful for:
- leaving explanations of unusual situations in policy
- leaving notes in policy-override eg. `meta: { reason: "have to be manually added because it's only imported dynamically" }`

CC @boneskull @kumavis - policy design

packages/webpack/src/plugin.js
```diff
@@ -256,9 +255,7 @@ class LavaMoatPlugin {
           // replacing generator with something returning an empty string could shave off some extra time, but is too invasive to seem worth it
         }
 
-        if (
-          FORCED_CONFIG.length > 0
-        ) {
+        if (FORCED_CONFIG.length > 0) {
```
naugtur, line 258: Prettier can't make up its mind.

packages/webpack/test/fixtures/main/policy/policy.json
```diff
@@ -26,6 +26,20 @@
...
         "define": true
       }
+    },
+    "uses-reexport-package": {
+      "globals": {
+        "console.log": true
+      },
+      "meta": {
+        "webpack-optimization": [
+          "Dependency 'uses-reexport-package>reexport-package' reexports from 'es6-module-package' and webpack collapsed that to a direct import."
```
naugtur, line 36: neat, huh?

packages/types/src/policy-schema/lavamoat-policy.v0-0-1.schema.ts
boneskull: Seems reasonable, but also we could support JSONC if we want comments.

boneskull: (suggestion) define a type for this field

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -113,9 +153,32 @@ module.exports = {
         importMap: {
           // connections are a much better source of information than module.dependencies which contain
           // all imported references separately along with exports and fluff
-          ...Array.from(connections).reduce((acc, dep) => {
+          ...Array.from(connections).reduce((acc, connection) => {
```
boneskull, line 159: (opinion)
```suggestion
          ...[connections].reduce((acc, connection) => {
```

packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -133,6 +196,12 @@ module.exports = {
     }
 
     const policy = moduleInspector.generatePolicy({})
+    Object.entries(meta).forEach(([packageName, packageMeta]) => {
+      if (!policy.resources[packageName]) {
+        return
+      }
+      policy.resources[packageName].meta = packageMeta
+    })
```
boneskull, line 207: What's this doing?  I'm not aware of any other place we're actually mutating a policy

boneskull (approved):

packages/types/src/policy-schema/lavamoat-policy.v0-0-1.schema.ts
naugtur: 👍I had a type for it when it was more complex but then I simplified to a point of not feeling like I need one, but I see where that would go.

packages/webpack/src/buildtime/policyGenerator.js
naugtur: You mean `...[... `?

That's weird 😅

packages/webpack/src/buildtime/policyGenerator.js
naugtur: It's adding the meta info to the policy. Meta is collected earlier and needs to be merged in before policy is written

I should check if it ends up in the bundle... 🤔

naugtur:
@boneskull The JSONC suggestion was very tempting. It'd fix the problem with stripping the metadata too. I was very close to going with it, but I see value in keeping the policy JSON.parsea-ble and having the metadata in one place. We could start adding things helpful for visualization in there.

PR #1835, opened 2025-10-07
fix!(core): stop overwriting the policy file while loading.

naugtur:
The only reason we wanted it was for there to exist a single file where the policy and overrides are merged. That should happen in policy generation. We now only warn if we detect some overrides were created after the policy.json was last generated.

### DIFF
--- a/packages/core/src/loadPolicy.js
+++ b/packages/core/src/loadPolicy.js
@@ -1,9 +1,7 @@
 // @ts-check
 
-const fs = require('node:fs/promises')
 const { readFileSync } = require('node:fs')
 const { mergePolicy } = require('./mergePolicy')
-const { jsonStringifySortedPolicy } = require('./stringifyPolicy')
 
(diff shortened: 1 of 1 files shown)


packages/core/src/loadPolicy.js
```diff
@@ -74,10 +72,30 @@ async function loadPolicy({ debugMode, policyPath }) {
   return policy
 }
 
+/**
+ * Checks if all resources from overrides exist in policy
+ *
+ * @param {any} policy
+ * @param {any} override
```
boneskull:
```suggestion
 * @param {import('@lavamoat/types').LavaMoatPolicy} policy
 * @param {import('@lavamoat/types').LavaMoatPolicy} override
```

boneskull (approved):
LGTM, but use types plox

boneskull:
you want `fix(core)!`

packages/core/src/loadPolicy.js
naugtur: weird. copilot broke types. LavaMoatPolicy is in scope.

PR #1838, opened 2025-10-20
fix(webpack): add j to the list of harmless fields on __webpack_require__

naugtur:
as documented here: https://github.com/webpack/webpack/blob/00a162b01901c9db8c1284d9b1541d04f953e8f2/lib/RuntimeGlobals.js#L363

I missed it the first time around. 

It seems to be used to figure out whether a chunk is loading under the runtime it was expecting to be loading when there's multiple distinct runtimes in a compilation.


boneskull (approved):
Yes, it is good that you have added "j" to this array.  ???

PR #1839, opened 2025-10-20
fix(webpack): adapt scuttling config for runtime regardless of where it comes from

naugtur:
There was a bug that caused scuttling config to not be processed if it came from `runtimeConfigurationPerChunk` callback instead of the main plugin options. 
Refactored scuttling config processing to happen after it has been obtained.

### DIFF
--- a/packages/webpack/src/plugin.js
+++ b/packages/webpack/src/plugin.js
@@ -87,20 +87,6 @@ class LavaMoatPlugin {
   constructor(options = {}) {
     if (typeof options.scuttleGlobalThis === 'object') {
       options.scuttleGlobalThis = { ...options.scuttleGlobalThis }
-      if (Array.isArray(options.scuttleGlobalThis.exceptions)) {
-        options.scuttleGlobalThis.exceptions =
-          options.scuttleGlobalThis.exceptions.map(
-            /**
-             * Convert exception to string
-             *
-             * @param {string | RegExp} e
-             * @returns {string}
-             */
-            (e) => e.toString()
-          )
-      } else {
-        options.scuttleGlobalThis.exceptions = []
-      }
     } else {
       options.scuttleGlobalThis = { enabled: false }
     }
@@ -261,16 +247,6 @@ class LavaMoatPlugin {
               'LavaMoatPlugin: Following options had to be overriden for security: ' +
                 FORCED_CONFIG.join(', ')
             )
-          )
-        }
-
-        // Adjust scuttling configuration to not scuttle webpack chunk loading facilities
-        if (
-          typeof STORE.options.scuttleGlobalThis === 'object' &&
-          Array.isArray(STORE.options.scuttleGlobalThis.exceptions)
-        ) {
-          STORE.options.scuttleGlobalThis.exceptions.push(
-            compilation.outputOptions.chunkLoadingGlobal || 'webpackChunk'
           )
         }
 
@@ -564,6 +540,9 @@ class LavaMoatPlugin {
                     contextModuleIds: STORE.contextModuleIds,
                     externals: STORE.externals,
                   },
+                  chunkLoaderName:
+                    compilation.outputOptions.chunkLoadingGlobal ||
+                    'webpackChunk',
                 })
 
                 // Add the runtime modules to the chunk, which handles
--- a/packages/webpack/src/plugin.js
+++ b/packages/webpack/src/plugin.js
@@ -541,7 +541,7 @@ class LavaMoatPlugin {
                     externals: STORE.externals,
                   },
                   chunkLoaderName:
-                    compilation.outputOptions.chunkLoadingGlobal ||
+                    compilation.outputOptions.chunkLoadingGlobal ??
                     'webpackChunk',
                 })
 
--- a/packages/webpack/src/runtime/scuttlingConf.js
+++ b/packages/webpack/src/runtime/scuttlingConf.js
@@ -1,0 +1,29 @@
+/** @import {LavaMoatScuttleOpts} from 'lavamoat-core' */
+
+/**
+ * @param {LavaMoatScuttleOpts} scuttleGlobalThis
+ * @param {Object} [options]
+ * @param {string} [options.chunkLoaderName='webpackChunk'] The name of the global that loads chunks
+ * @returns {LavaMoatScuttleOpts}
+ */
+exports.adjustScuttleConfig = (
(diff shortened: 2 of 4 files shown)


packages/webpack/src/runtime/scuttlingConf.js
```diff
@@ -0,0 +1,29 @@
...
+  scuttleGlobalThis = { ...scuttleGlobalThis }
+  if (Array.isArray(scuttleGlobalThis.exceptions)) {
+    scuttleGlobalThis.exceptions = scuttleGlobalThis.exceptions.map(
+      /**
+       * Convert regex to string
+       *
+       * @param {string | RegExp} e
+       * @returns {string}
+       */
+      (e) => e.toString()
```
boneskull, line 22: I'm not sure this is entirely safe.  From [escape-string-regexp](https://npm.im/escape-string-regexp):

```js
	// Escape characters with special meaning either inside or outside character sets.
	// Use a simple backslash escape when it’s always valid, and a `\xnn` escape when the simpler form would be disallowed by Unicode patterns’ stricter grammar.
	return string
		.replace(/[|\\{}()[\]^$+*?.]/g, '\\$&')
		.replace(/-/g, '\\x2d');
```

packages/webpack/src/plugin.js
```diff
@@ -564,6 +540,9 @@ class LavaMoatPlugin {
                     contextModuleIds: STORE.contextModuleIds,
                     externals: STORE.externals,
                   },
+                  chunkLoaderName:
+                    compilation.outputOptions.chunkLoadingGlobal ||
```
```suggestion
                    compilation.outputOptions.chunkLoadingGlobal ??
```

packages/webpack/test/e2e-scuttle.spec.js
```diff
@@ -50,6 +50,40 @@ async function scuttle(t, scuttleGlobalThis, globals) {
     t.context.globalThis = runScriptWithSES(t.context.bundle, globals).context
   }, 'Expected the build to succeed')
 }
+async function scuttleViaRuntimeConf(t, scuttleGlobalThis, globals) {
+  const webpackConfigDefault = makeConfig({
+    generatePolicy: true,
+    diagnosticsVerbosity: 1,
+    policyLocation: path.resolve(__dirname, 'fixtures/main/policy-scuttling'),
+    runtimeConfigurationPerChunk_experimental: (/*chunk*/) => {
```
boneskull, line 58:
```suggestion
    runtimeConfigurationPerChunk_experimental: (_chunk) => {
```

packages/webpack/test/e2e-scuttle.spec.js
```diff
@@ -91,6 +125,15 @@ test(`webpack/scuttled - webpackChunk global is transparently added to exception
   t.truthy(t.context.globalThis.webpackChunkTEST)
 })
 
+test(`webpack/scuttled - webpackChunk global is transparently added to exceptions when configuration comes from runtimeConfigurationPerChunk`, async (t) => {
+  await scuttleViaRuntimeConf(t, { enabled: true, exceptions: ['Function'] })
+
+  t.notThrows(() => {
+    t.context.globalThis.webpackChunkTEST
```
boneskull, line 133: Is this a getter? Was this supposed to be a function call?

boneskull (approved):
OK, but concerns about the regexp and the test.

packages/webpack/src/runtime/scuttlingConf.js
boneskull: Wait, we're not converting from a string to a regexp, it's the other way around.  If this gets passed thru then to webpack, I assume they handle the escaping.  But I'm confused why we do this at all

naugtur: scuttling is recreating these from strings. The implementation relies on being able to pass the config around to other windows/frames where snow runs scuttling too. So the whole thing is made serializable. 
These are not regexes from untrusted sources, so the failure mode here is: the develoepr tried to unscuttle a thing and failed, has to try again.

packages/webpack/src/plugin.js
naugtur: I'm not sold on the benefit :P

packages/webpack/test/e2e-scuttle.spec.js
naugtur: We only have that rule in Endo.

packages/webpack/test/e2e-scuttle.spec.js
naugtur: If scuttled, it's a getter that throws

packages/webpack/src/runtime/scuttlingConf.js
naugtur: Also, I don't think you can stringify a regex in a way that would mess up the constructor. 

And it goes into the bundle after JSON.strigify, not as naive concatenation, so I don't see room for failure there.

boneskull: I'm not sure if you can "mess up" a constructor but you can certainly break the regex upon construction _if the original string was not a regex_.  Recall that if you use the constructor, you need to escape stuff, whereas with regex literals you don't.  But if we're _roundtripping_ the regexes (they _start_ as regexes) this shouldn't be a problem.

boneskull: i.e. the use-case for `escape-string-regexp` is not our use-case.

packages/webpack/test/e2e-scuttle.spec.js
boneskull: really? huh.  I want it.  I think I may have added it in my overhaul of `@lavamoat/node`'s eslint config.

packages/webpack/test/e2e-scuttle.spec.js
boneskull: I'd probably add a comment to that effect because at first glance it is fishy

packages/webpack/test/e2e-scuttle.spec.js
naugtur: do share ;)

packages/webpack/test/e2e-scuttle.spec.js
naugtur, line 133:
```suggestion
    // accessing the field, if scuttled, will call the getter that throws
    t.context.globalThis.webpackChunkTEST
```

PR #1843, opened 2025-10-29
fix(core): fix policy narrowing logic in the policy merge 

naugtur:
starting with a test

### DIFF
--- a/packages/core/src/mergePolicy.js
+++ b/packages/core/src/mergePolicy.js
@@ -7,31 +7,39 @@ const {
 } = require('lavamoat-tofu/src/util')
 const mergeDeep = require('merge-deep')
 
-const { values, hasOwn } = Object
+const { entries, hasOwn } = Object
 
 /**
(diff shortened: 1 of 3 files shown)


packages/core/src/mergePolicy.js
```diff
@@ -41,14 +49,26 @@ function mergePolicy(policyA, policyB) {
 
 /**
  * @template {BuiltinPolicy | GlobalPolicy} T
- * @param {T} packagePolicy
+ * @param {T} policyItems
+ * @param {T} [overrideItems]
  * @returns {T}
```
boneskull: This whole thing is wrong and it's my fault. It needs function overloads instead of generics. I'll deal with it later.

boneskull (approved):
:shipit:

packages/core/src/mergePolicy.js
naugtur: @copilot add the proposed fix

naugtur: well, that didn't work

PR #1845, opened 2025-11-03
feat(tofu): enable deep dependency analysis of ES modules

boneskull:
This makes `lavamoat-tofu` generate exports like `fs.readFile` when encountering `import {readFile} from 'fs'`; previously it would only generate `fs`.

### DIFF
--- a/packages/tofu/src/inspectSource.js
+++ b/packages/tofu/src/inspectSource.js
@@ -183,37 +183,122 @@ function inspectGlobals(
  * @param {import('@babel/types').Node} ast
  * @param {string[]} [packagesToInspect] - List of module IDs to look for; if
  *   omitted, will inspect all imports
+ * @param {boolean} [deep] Return named exports and imports
  * @returns {string[]} - List of imported identifiers
  */
-function inspectEsmImports(ast, packagesToInspect) {
+function inspectEsmImports(ast, packagesToInspect, deep = true) {
   const shouldInspectAll = !packagesToInspect
   const pkgsToInspect = new Set(packagesToInspect)
 
   /** @type {Set<string>} */
-  const esmImports = new Set()
-
-  /**
-   * @param {import('@babel/traverse').NodePath<
-   *   | import('@babel/types').ImportDeclaration
-   *   | import('@babel/types').ExportNamedDeclaration
-   * >} path
-   */
-  const handleNodePath = (path) => {
-    const importSource = path.node.source?.value
-
-    if (importSource && (shouldInspectAll || pkgsToInspect.has(importSource))) {
-      esmImports.add(importSource)
-    }
-  }
+  const result = new Set()
 
   traverse(ast, {
-    ExportNamedDeclaration: (path) => {
-      handleNodePath(path)
+    ImportDeclaration: (path) => {
+      const importSource = path.node.source.value
+      if (!shouldInspectAll && !pkgsToInspect.has(importSource)) {
+        return
+      }
+
+      if (!deep) {
+        result.add(importSource)
+        return
+      }
+
+      const { specifiers } = path.node
+
+      const hasNamespaceImport = specifiers.some(
+        (spec) => spec.type === 'ImportNamespaceSpecifier'
+      )
+
+      const hasOnlyDefaultImport = specifiers.every(
+        (spec) => spec.type === 'ImportDefaultSpecifier'
+      )
+
+      if (
+        hasNamespaceImport ||
+        hasOnlyDefaultImport ||
+        specifiers.length === 0
+      ) {
+        // import * as foo from 'bar', import foo from 'baz', import 'quux'
+        result.add(importSource)
+      } else {
+        // Mixed or named imports
+        specifiers.forEach((spec) => {
+          if (spec.type === 'ImportSpecifier') {
+            // it can be an identifier or a string literal, to my amazement
+            const importedName =
+              spec.imported.type === 'Identifier'
+                ? spec.imported.name
+                : spec.imported.value
+            result.add(`${importSource}.${importedName}`)
+          } else if (spec.type === 'ImportDefaultSpecifier') {
+            // import foo, {bar} from 'baz'
+            result.add(importSource)
+          }
+        })
+      }
     },
-    ImportDeclaration: handleNodePath,
+
+    ExportNamedDeclaration: (path) => {
+      const exportSource = path.node.source?.value
+      if (
+        !exportSource ||
+        (!shouldInspectAll && !pkgsToInspect.has(exportSource))
+      ) {
+        return
+      }
+
+      if (!deep) {
+        result.add(exportSource)
+        return
+      }
+
+      const { specifiers } = path.node
+
+      const hasNamespaceExport = specifiers.some(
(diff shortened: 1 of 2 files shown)


boneskull:
* **#1845** <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1845?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1845?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1846** <a href="https://app.graphite.com/github/pr/LavaMoat/LavaMoat/1846?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/tofu/test/inspectEsmImports.spec.js
```diff
@@ -29,22 +29,45 @@ testInspect(
...
+    'package08', // default import - no keypath
+    'package09', // default + namespace import - no keypath
+    'package10', // side-effect import - no keypath
+    'package12', // namespace re-export - no keypath
+    'package13.name4', // named re-export - keypath
+    'package14', // default re-export - no keypath
+    'package15', // mixed default + named import - module for default
+    'package15.namedExport', // mixed default + named import - keypath for named
+    'package16.kebab-case', // string literal named import - keypath with string literal
+    'package17.with-hyphen', // string literal named re-export - keypath with string literal
```
naugtur, line 58: oh God 🙈 
I wouldn't have guessed this is valid.

packages/tofu/src/inspectSource.js
```diff
@@ -183,37 +183,110 @@ function inspectGlobals(
  * @param {import('@babel/types').Node} ast
  * @param {string[]} [packagesToInspect] - List of module IDs to look for; if
  *   omitted, will inspect all imports
+ * @param {boolean} [deep] Return named exports and imports
  * @returns {string[]} - List of imported identifiers
  */
-function inspectEsmImports(ast, packagesToInspect) {
+function inspectEsmImports(ast, packagesToInspect, deep = true) {
```
naugtur, line 189: appreciate the opt-in for backwards cmopatibility, but I think we should turn it on in lavamoat-node classic too if circumstances permit.

packages/tofu/src/inspectSource.js
```diff
@@ -183,37 +183,110 @@ function inspectGlobals(
...
 
-  /**
-   * @param {import('@babel/traverse').NodePath<
-   *   | import('@babel/types').ImportDeclaration
-   *   | import('@babel/types').ExportNamedDeclaration
-   * >} path
-   */
-  const handleNodePath = (path) => {
-    const importSource = path.node.source?.value
+  traverse(ast, {
```
naugtur, line 196: There are a few other node types that ModuleSource in SES is looking at
https://github.com/endojs/endo/blob/naugtur/e2e-test-conditional-conditions/packages/module-source/src/babel-plugin.js#L313

We don't need all of them because we're not interested in finding exports. I'll try to make a comparison to see if this is not missing anything tho.

packages/tofu/src/inspectSource.js
```diff
@@ -183,37 +183,110 @@ function inspectGlobals(
...
+        // Mixed or named imports
+        specifiers.forEach((spec) => {
+          if (spec.type === 'ImportSpecifier') {
+            // it can be an identifier or a string literal, to my amazement
+            const importedName =
+              spec.imported.type === 'Identifier'
+                ? spec.imported.name
+                : spec.imported.value
+            result.add(`${importSource}.${importedName}`)
+          } else if (spec.type === 'ImportDefaultSpecifier') {
```
naugtur, line 235: doesn't seem to handle `ImportNamespaceSpecifier`
Could be just `else` without a further condition.

naugtur: Also, makes me wonder if, for a `*` namespace import we could get the name and use the code from cjs side of things to figure out what fields on that thing are being accessed. The fidelity loss is possible, but the code on the cjs side was being tested for years and seems good enough, so reuse is tempting.

packages/tofu/src/inspectSource.js
```diff
@@ -183,37 +183,110 @@ function inspectGlobals(
...
+            // import foo, {bar} from 'baz'
+            result.add(importSource)
+          }
+        })
+      }
+    },
 
-  traverse(ast, {
     ExportNamedDeclaration: (path) => {
```
naugtur, line 243: `ExportAllDeclaration` also exists

packages/tofu/test/inspectEsmImports.spec.js
boneskull: It was news to me!

packages/tofu/src/inspectSource.js
boneskull: This is _opt-out_ since it defaults to `true` (just aping `inspectRequires`)

packages/tofu/src/inspectSource.js
boneskull: Are you sure that it doesn't handle importnamespace specifier? The tests seem to suggest it does, do they not?

packages/tofu/src/inspectSource.js
boneskull: I'm not sure what this is.  This handler works for `export * as name from 'pkg'`, as that's an `ExportNamedDeclaration`.

boneskull: Oh, I see, it's `export * from 'pkg'`

boneskull:
@naugtur I've `ExportAllDeclaration` support.  For packages which want to re-export everything from `fs` 🙄

packages/tofu/src/inspectSource.js
naugtur: Oh, I see, it's handled before the foreach.

packages/tofu/src/inspectSource.js
```diff
@@ -183,37 +183,122 @@ function inspectGlobals(
...
+      } else {
+        specifiers.forEach((spec) => {
+          if (spec.type === 'ExportSpecifier') {
+            // Handle both Identifier and StringLiteral (for arbitrary module namespace identifier names)
+            const exportedName =
+              // @ts-expect-error - @babel/types is wrong, but the parser handles string literals
+              spec.local.type === 'StringLiteral'
+                ? // @ts-expect-error - @babel/types is wrong, but the parser handles string literals
+                  spec.local.value
+                : spec.local.name
```
naugtur, line 275: I cross-checked this with what babel produces via astexplorer.net

packages/tofu/test/inspectEsmImports.spec.js
```diff
@@ -29,22 +29,47 @@ testInspect(
   export * as name3 from "package12";
   export { name4 } from "package13";
```
naugtur, line 30: one combination not listed here is 
export { name as default } from "package";
but it works and is redundant. I don't think it needs to be added.

naugtur (approved):
I've read the hell out of this and it's correct.

PR #1848, opened 2025-11-05
feat(core): add informative errors when the result of policy merge contains invalid entries

naugtur:
As a result of fixing the bug that was erasing all nested fields below one already existing we've opened the possibility of some existing policies being invalid, which manifested at runtime if a global property was a getter. 

```
Error: unable to copy on to targetRef, targetRef has a getter at "location"
```

This change will make redundant policy overrides fail early.


boneskull (approved):
LGTM, but I wonder if tofu is the right place for this validation.  If you think it is, then I agree.  If you don't think it is, then I agree with that, too.

naugtur:
> LGTM, but I wonder if tofu is the right place for this validation. If you think it is, then I agree. If you don't think it is, then I agree with that, too.

The code for finding duplicates is a mutation of code that removes them, they belong next to each other because they'll be updated together.

PR #1856, opened 2025-11-25
fix(lavamoat-node): fix missing cjs extension in the list of extensions to resolve

naugtur:
This fix ends up providing a tiny improvement in ecosystem compatibility where conventionally built packages have an index.cjs and index.mjs and `resolve.sync` doesn't support looking at package.json->exports but finds the index.cjs by convention. 

It seems like a regression introduced in https://github.com/LavaMoat/LavaMoat/commit/84303100bc250dd564b5769d41fdcd3f11f0f58c


packages/lavamoat-node/src/parseForPolicy.js
```diff
@@ -362,6 +364,8 @@ function parseModule(moduleSrc, filename = '<unknown file>') {
       sourceType: 'unambiguous',
       // someone must have been doing this
       allowReturnOutsideFunction: true,
+      // allow typescript syntax
+      plugins: ['typescript'],
```
naugtur: This experiment doesn't belong here.
```suggestion
```

packages/lavamoat-node/src/parseForPolicy.js
```diff
@@ -235,6 +236,7 @@ function makeImportHook({
     // parse
     const ast = parseModule(content, specifier)
     // get imports
+    // TODO: ts support means this would need to understand esm syntax too?
```
naugtur:
```suggestion
```

FrederikBolding (approved):

PR #1857, opened 2025-11-26
fix(lavamoat-node): cli entry args

### LINKED ISSUE #344: `lavamoat`: Scrub CLI options from `process.argv` before running script
The LavaMoat node CLI includes its own CLI options, but it leaves them on `process.argv` to be picked up by the script being run by LavaMoat. This could result in them overriding options of the same name intended for the script being run, and it prevents using Yargs "strict mode" where unexpected CLI options are disallowed.

This could be solved by using a CLI delimiter argument like `--`. Treat everything before the `--` as intended for LavaMoat, and everything after as intended for the script being run. This is the same solution that `npm` uses.

See here for the context where this came up: https://github.com/MetaMask/metamask-extension/pull/14836#discussion_r889261864

itsyoboieltr:
fixes: https://github.com/LavaMoat/LavaMoat/issues/344

### DIFF
--- a/packages/lavamoat-node/src/cli.js
+++ b/packages/lavamoat-node/src/cli.js
@@ -24,15 +24,16 @@ function parseArgs() {
       })
       yargsFlags(yargs, defaults)
     })
+    .parserConfiguration({
+      'populate--': true,
+    })
     .help()
 
   const parsedArgs = argsParser.parse()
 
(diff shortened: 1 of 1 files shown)


naugtur:
merged into a local branch to continue work. 
This needs a test and assessment whether it's a breaking change (likely)

PR #1859, opened 2025-11-27
fix(lavamoat-node): cli entry args (#1857)

### LINKED ISSUE #344: `lavamoat`: Scrub CLI options from `process.argv` before running script

naugtur:
fix #344

continuation of a contribution from @itsyoboieltr 

todo:
- [x] validate it's needed and correct
- [x] figure out if it's a breaking change and whether we can keep backward-compatibility
- [x] tests
- [x] docs
- [ ] compare with `@lavamoat/node`

### DIFF
--- a/packages/lavamoat-node/src/cli.js
+++ b/packages/lavamoat-node/src/cli.js
@@ -40,6 +40,7 @@ function parseArgs() {
   // so we just remove the node path
   process.argv.shift()
   } else {
+    // Rebuild process.argv to pass only entry args
     process.argv = [process.argv[0], parsedArgs.entryPath, ...entryArgs]
   }
 
--- a/packages/lavamoat-node/src/cli.js
+++ b/packages/lavamoat-node/src/cli.js
@@ -24,15 +24,16 @@ function parseArgs() {
       })
       yargsFlags(yargs, defaults)
     })
+    .parserConfiguration({
+      'populate--': true,
+    })
     .help()
 
   const parsedArgs = argsParser.parse()
 
-  // patch process.argv so it matches the normal pattern
-  // e.g. [runtime path, entrypoint, ...args]
-  // we'll use the LavaMoat path as the runtime
-  // so we just remove the node path
-  process.argv.shift()
+  // Rebuild process.argv to pass only entry args
+  const entryArgs = parsedArgs['--'] || []
+  process.argv = [process.argv[0], parsedArgs.entryPath, ...entryArgs]
 
   return parsedArgs
 }
--- a/packages/lavamoat-node/src/cli.js
+++ b/packages/lavamoat-node/src/cli.js
@@ -32,8 +32,16 @@ function parseArgs() {
   const parsedArgs = argsParser.parse()
 
   // Rebuild process.argv to pass only entry args
-  const entryArgs = parsedArgs['--'] || []
-  process.argv = [process.argv[0], parsedArgs.entryPath, ...entryArgs]
+  const entryArgs = parsedArgs['--']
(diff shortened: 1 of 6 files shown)


packages/lavamoat-node/src/run-command.js
```diff
@@ -20,18 +20,30 @@ function parseArgs() {
     .usage(
       '$0',
       'lavamoat-run-command [flags for lavamoat] -- command [args for the command]',
-      (yarn) => yargsFlags(yarn, defaults)
+      (yargs) => yargsFlags(yargs, defaults)
```
naugtur, line 23: lol

packages/lavamoat-node/src/run-command.js
```diff
@@ -20,18 +20,30 @@ function parseArgs() {
...
     )
+    .parserConfiguration({
```
naugtur, line 25: the old way worked and was doing the same, but this is cleaner and consistent across both files.

packages/lavamoat-node/src/cli.js
```diff
@@ -24,15 +24,24 @@ function parseArgs() {
...
 
-  // patch process.argv so it matches the normal pattern
+  // Rebuild process.argv to pass only entry args
+  const entryArgs = parsedArgs['--']
+  if (!entryArgs) {
+    // patch process.argv so it matches the normal pattern
   // e.g. [runtime path, entrypoint, ...args]
   // we'll use the LavaMoat path as the runtime
   // so we just remove the node path
   process.argv.shift()
```
naugtur: I don't remember why we're doing a shift here instead of dropping the second arg, but leaving this in for back comp.

packages/lavamoat-node/src/cli.js
```diff
@@ -24,15 +24,24 @@ function parseArgs() {
...
+  } else {
+    process.argv = [process.argv[0], parsedArgs.entryPath, ...entryArgs]
```
FrederikBolding, line 44:
```suggestion
    // Rebuild process.argv to pass only entry args
    process.argv = [process.argv[0], parsedArgs.entryPath, ...entryArgs]
```
Nit: More readable to place this inline IMO

packages/lavamoat-node/src/run-command.js
```diff
@@ -20,18 +20,30 @@ function parseArgs() {
...
+      'populate--': true,
+    })
     .help()
 
   const parsedArgs = argsParser.parse()
-  const commandName = parsedArgs._[0]
+  const passDownArgs = parsedArgs['--'] || []
```
FrederikBolding, line 31: Is this not breaking? I don't entirely follow how this is equivalent to the old way

naugtur: I think the difference is - in the old version, if you provided some arguments before the -- that were not consumed by lavamoat, they might have been passed on. So this might differ on misuse, but the correct usecase is unaffected.

(also, this cli is somewhat new in comparison and I'm not sure it's been used)

FrederikBolding (approved):

PR #1860, opened 2025-11-27
fix(lavamoat-node): return entry module exports namespace from programmatic use of lavamoat runLava function

naugtur:
trivial but useful. I needed it to set up a benchmark. Others will need it too.


boneskull (commented):
I guess I was expecting a change to `runLava`?  I'm confused.

naugtur:
You've added the return in another PR (linked above) but without the test update. We can merge just the test update and/or squash with that other one.

boneskull (approved):

PR #1863, opened 2025-12-01
fix: typescript got confused into creating a cyclic reference in types for scuttling in core

naugtur:
hopefully this is the last time we're fixing scuttling types.

Resulting types/src/scuttling.d.ts

**before**

```ts
/**
 * Applies scuttling, with the default set of options, including using Snow if
 * passed in as scuttlerFunc. Scuttle globalThis right after we used it to
 * create the root package compartment.
 *
 * @param {GlobalRef} globalRef - Reference to the global object.
 * @param {LavaMoatScuttleOpts} opts - Scuttling options.
 */
export function scuttle(globalRef: GlobalRef, opts: LavaMoatScuttleOpts): void;
import type { GlobalRef } from './scuttle.d.ts';
import type { LavaMoatScuttleOpts } from './scuttle.d.ts';
//# sourceMappingURL=scuttle.d.ts.map
```

**after**

```ts
/**
 * Options for scuttling `globalThis`.
 */
export type LavaMoatScuttleOpts = {
    /**
     * Whether scuttling is enabled or not.
     */
    enabled: boolean;
    /**
     * List of properties to exclude from scuttling.
     */
    exceptions?: (string | RegExp)[] | undefined;
    /**
     * Name of the scuttler function to use which is expected to be found as a property on the global object (e.g. if scuttlerName is 'x', scuttler function is obtained from globalThis['x']).
     */
    scuttlerName?: string | undefined;
};
/**
 * Reference to the global object
 */
export type GlobalRef = {
    /**
     * Reference to the global object.
     */
    globalThis?: Record<PropertyKey, unknown> | undefined;
};
/**
 * Applies scuttling, with the default set of options, including using Snow if
 * passed in as scuttlerFunc. Scuttle globalThis right after we used it to
 * create the root package compartment.
 *
 * @param {GlobalRef} globalRef - Reference to the global object.
 * @param {LavaMoatScuttleOpts} opts - Scuttling options.
 */
export function scuttle(globalRef: GlobalRef, opts: LavaMoatScuttleOpts): void;
//# sourceMappingURL=scuttle.d.ts.map
```

### DIFF
--- a/packages/core/src/scuttle.d.ts
+++ b/packages/core/src/scuttle.d.ts
@@ -1,25 +1,11 @@
-/**
- * Options for scuttling global properties
- */
-export type LavaMoatScuttleOpts = {
-  /** Whether scuttling is enabled or not. */
-  enabled: boolean;
-  /** List of properties to exclude from scuttling. */
-  exceptions?: Array<string | RegExp>;
-  /** Name of the scuttler function to use which is expected to be found as a property on the global object (e.g. if scuttlerName is 'x', scuttler function is obtained from globalThis['x']). */
-  scuttlerName?: string;
-}
+// Declaring types here and importing them in scuttle.js 
+// leads to typescript getting confused into generating 
+// a self-referential cycle in the generated .d.ts files:
+// types/src/scuttle.d.ts
+// import type { LavaMoatScuttleOpts } from './scuttle.d.ts';
 
-/**
- * Reference to the global object
- */
-export interface GlobalRef {
-  globalThis?: Record<PropertyKey, unknown>;
-}
+export type {
+    LavaMoatScuttleOpts
+} from './scuttle.js'
 
-/**
- * Scuttles (disables or restricts) certain global properties
- * @param globalRef The global object to scuttle
- * @param [opts] Options for scuttling
- */
-export function scuttle(globalRef: GlobalRef, opts?: LavaMoatScuttleOpts | boolean): void;
+export { scuttle } from './scuttle.js'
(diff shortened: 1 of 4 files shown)


FrederikBolding (approved):

boneskull:
I've changed this PR to just rename `scuttle.d.ts` to `scuttle.ts`, which results in a `types/src/scuttle.d.ts` of:

```ts
/**
 * Options for scuttling global properties
 */
export type LavaMoatScuttleOpts = {
    /** Whether scuttling is enabled or not. */
    enabled: boolean;
    /** List of properties to exclude from scuttling. */
    exceptions?: Array<string | RegExp>;
    /** Name of the scuttler function to use which is expected to be found as a property o
n the global object (e.g. if scuttlerName is 'x', scuttler function is obtained from globa
lThis['x']). */
    scuttlerName?: string;
};
/**
 * Reference to the global object
 */
export interface GlobalRef {
    globalThis?: Record<PropertyKey, unknown>;
}
/**
 * Scuttles (disables or restricts) certain global properties
 * @param globalRef The global object to scuttle
 * @param [opts] Options for scuttling
 */
export declare function scuttle(globalRef: GlobalRef, opts?: LavaMoatScuttleOpts | boolean
): void;
//# sourceMappingURL=scuttle.d.ts.map
```

I had to change `export function` to `export declare function` which signals to TS the function exists _but is not implemented "here"_ (in the `.ts` file).  I'm fairly certain `export declare function` and `export function` are functionally identical in a `d.ts` file. Either way, this should fix the cycle.

boneskull:
Meh, I'm gonna just rebase my changes away.  It confuses tseslint and I don't want to deal with that.

boneskull (approved):

PR #1867, opened 2025-12-02
fix(core): untangle another tsc confusion in generating .d.ts files

naugtur:
I found one more place where .d.ts files were broken (`LavaMoatScuttleOpts` was not getting imported into options.d.ts)


packages/core/test/scenario.ts
```diff
@@ -8,7 +8,7 @@ import {
 } from '@lavamoat/types'
 import type { ExecutionContext } from 'ava'
 import { type SetRequired } from 'type-fest'
-import type { LavaMoatOpts } from '../src/options'
+import type { LavaMoatOpts } from '../src/types'
```
naugtur: This was the only place where `LavaMoatOpts` were not imported from main types export

boneskull (commented):
This opens us up for future cycles if we don't have them now.  `types.ts` is an entry point.  other modules shouldn't depend on it, since it grabs so much other stuff from elsewhere.  Maybe better to rename `options.js` to `options.ts`?  What else did you try here?

naugtur:
Changing extensions was a whack-a-mole game for me - one thing worked, something else broke. The worst thing was a working IDE ts server but silently messed up generated types with missing stuff that ends up resolving some of the things we typed to `any` in outside use.

I tried many ways and this is the only one that works for me.

naugtur:
Tested the latest commit and it seems to work fine. Could go into a release now. (can't approve as an author)

PR #1878, opened 2026-01-14
fix: performant implementation of getPackageDirForModulePath in aa and minor memoization in lavamoat-node

naugtur:
the assumption that splitting by path separator works is valid.

there's a possibility that `path.sep` is not the right separator if a memfs is used. 
should I add a fallback to detect a separator?
should this be an expensive to construct prefix tree for an algorithm that is agnostic to the concept of a path? (very meh imho)

The build I tested this on goes from 90sec to 64sec with this fix.

### DIFF
--- a/packages/aa/src/index.js
+++ b/packages/aa/src/index.js
@@ -275,33 +275,22 @@ function getPackageNameForModulePath(canonicalNameMap, modulePath) {
   return packageName
 }
 
+const pathSeparator = path.sep;
+
 /**
  * @param {CanonicalNameMap} canonicalNameMap
  * @param {string} modulePath
  * @returns {string | undefined}
  */
 function getPackageDirForModulePath(canonicalNameMap, modulePath) {
-  // find which of these directories the module is in
-  const matchingPackageDirs = Array.from(canonicalNameMap.keys()).filter(
-    (packageDir) => modulePath.startsWith(packageDir)
-  )
-  if (matchingPackageDirs.length === 0) {
-    return undefined
-    // throw new Error(`Could not find package for module path: "${modulePath}" out of these package dirs:\n${Array.from(canonicalNameMap.keys()).join('\n')}`)
+  let subPath = modulePath
+  while (subPath.length > 0) {
+    if (canonicalNameMap.has(subPath)) {
+      return subPath
+    }
+    subPath = subPath.substring(0, subPath.lastIndexOf(pathSeparator))
   }
-  const longestMatch = matchingPackageDirs.reduce(takeLongest)
-  return longestMatch
-}
-
-/**
- * For comparing string lengths
- *
- * @param {string} a
- * @param {string} b
- * @returns {string}
- */
-function takeLongest(a, b) {
-  return a.length > b.length ? a : b
+  return undefined
 }
 
 /**
--- a/packages/aa/src/index.js
+++ b/packages/aa/src/index.js
@@ -275,7 +275,7 @@ function getPackageNameForModulePath(canonicalNameMap, modulePath) {
   return packageName
 }
 
-const pathSeparator = path.sep;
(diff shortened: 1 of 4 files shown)


packages/lavamoat-node/src/kernel.js
```diff
@@ -200,14 +204,16 @@ function createModuleLoader({ canonicalNameMap }) {
...
+      const moduleData = {
 
-      return {
         type: 'js',
         file: absolutePath,
         package: packageName,
         source: wrappedContent,
         id: absolutePath,
       }
+      memo.set(absolutePath, moduleData)
```
naugtur, line 215: I'm memoizing moduleData because it seems to me it will never be GCd ayway, so this might be less memory overall.

I'm not 100% sure tho - it's possible that all data from here gets copied elsewhere and the moduleData object gets cleaned up.

packages/aa/src/index.js
```diff
@@ -275,33 +275,22 @@ function getPackageNameForModulePath(canonicalNameMap, modulePath) {
   return packageName
 }
 
+const pathSeparator = path.sep;
```
naugtur: This would be incorrect for webpack with memfs on windows.

boneskull:
I think memfs always uses POSIX-style paths?

naugtur:
> I think memfs always uses POSIX-style paths? 

Yep. And there's no good path from Webpack to this, so I'm considering a fallback. Based on slash popularity in keys measured once the map is created at first or something.

The question is tho - can we merge as is to give us time to think about the best fallback?

packages/webpack/test/aa.spec.js
```diff
@@ -64,11 +64,11 @@ const lookupFixture = {
...
     Object.entries({
       '/': '$root$',
-      '/node_modules/leftpad/': 'leftpad',
+      '/node_modules/leftpad': 'leftpad',
       '/node_modules/fast-json-patch': 'fast-json-patch',
       '/node_modules/fast-json-patch/module': 'fast-json-patch',
-      '/node_modules/@ethereumjs/util/dist/': '@ethereumjs/util',
-      '/node_modules/@ethereumjs/util/node_modules/ethereum-cryptography/dist/':
+      '/node_modules/@ethereumjs/util/dist': '@ethereumjs/util',
+      '/node_modules/@ethereumjs/util/node_modules/ethereum-cryptography/dist':
```
naugtur, line 71: fixture was incorrect 🙈

packages/aa/src/index.js
```diff
@@ -275,33 +275,28 @@ function getPackageNameForModulePath(canonicalNameMap, modulePath) {
...
- * @returns {string}
- */
-function takeLongest(a, b) {
-  return a.length > b.length ? a : b
+  if (
+    canonicalNameMap.has(pathSeparator) &&
+    modulePath.startsWith(pathSeparator)
+  ) {
+    return pathSeparator
+  }
```
FrederikBolding, line 298: In what case is the separator simply a valid module path?

packages/aa/src/index.js
FrederikBolding: How often is memfs used with Webpack? Just wondering what the chances of running into this are.

Can we detect running in memfs and fail?

naugtur: At this point, because we're using the aa package, LavaMoat webpack plugin will not work with memfs anyway. There's no way to find out about it from here. It could be surfaced by the plugin maybe.

If memfs contents match the situation on disk, we could try to adapt here, but that's extremely unlikely. memfs is mostly used in tests

packages/aa/src/index.js
naugtur: If your build environment starts at root.

Currently it only happens in tests, but it is theoretically possible.

FrederikBolding (approved):

PR #1882, opened 2026-02-03
fix(aa): normalize root path to not contain a trailing slash regardless of what is provided in input

naugtur:
Earlier the test missed one possibility. It occurred in the wild.


boneskull (approved):
:shipit:

PR #1772, opened 2025-08-22
chore!: drop Node.js v16 support across the board

boneskull:
BREAKING CHANGE: This drops support for Node.js v16 in `@lavamoat/aa`, `@lavamoat/allow-scripts`, `@lavamoat/preinstall-always-fail` and `@lavamoat/yarn-plugin-allow-scripts`.

Issue #1221, opened 2024-07-09
[webpack] TypeError: Cannot read properties of undefined (reading 'keys')


Jack-Works:
```
node_modules/.pnpm/@lavamoat+aa@4.2.0/node_modules/@lavamoat/aa/src/index.js:285
  const matchingPackageDirs = Array.from(canonicalNameMap.keys()).filter(
                                                          ^

TypeError: Cannot read properties of undefined (reading 'keys')
    at getPackageDirForModulePath (node_modules/.pnpm/@lavamoat+aa@4.2.0/node_modules/@lavamoat/aa/src/index.js:285:59)
    at getPackageNameForModulePath (node_modules/.pnpm/@lavamoat+aa@4.2.0/node_modules/@lavamoat/aa/src/index.js:262:22)
    at Object.inspectWebpackModule (node_modules/.pnpm/@lavamoat+webpack@0.3.0-beta.0_webpack@5.88.2/node_modules/@lavamoat/webpack/src/buildtime/policyGenerator.js:96:24)
    at node_modules/.pnpm/@lavamoat+webpack@0.3.0-beta.0_webpack@5.88.2/node_modules/@lavamoat/webpack/src/plugin.js:255:33
    at Array.forEach (<anonymous>)
    at node_modules/.pnpm/@lavamoat+webpack@0.3.0-beta.0_webpack@5.88.2/node_modules/@lavamoat/webpack/src/plugin.js:249:47
    at Array.forEach (<anonymous>)
    at node_modules/.pnpm/@lavamoat+webpack@0.3.0-beta.0_webpack@5.88.2/node_modules/@lavamoat/webpack/src/plugin.js:248:30
    at Hook.eval [as call] (eval at create (node_modules/.pnpm/tapable@2.2.1/node_modules/tapable/lib/HookCodeFactory.js:19:10), <anonymous>:7:1)
    at Hook.CALL_DELEGATE [as _call] (node_modules/.pnpm/tapable@2.2.1/node_modules/tapable/lib/Hook.js:14:14)
```

Jack-Works:
sorry this still happens in @lavamoat/aa@4.3.0 for me (@lavamoat/webpack@0.4.0-beta.0). please reopen this thanks!

legobeat:
@Jack-Works Which Node.js version are you on?

Our current understanding:

- v18 was never affected
- Issue is fixed in v22
[...]

PR #1886, opened 2026-02-13
fix!: drop Node.js v18 support across the board

boneskull:
Minimum Node.js version is now v20.19.0. The v22 range starts at v22.5.1 to avoid a `closeSync` regression in v22.5.0 that broke LavaMoat (see nodejs/node#53902, LavaMoat/LavaMoat#1221).

`engines.node` is set to `^20.19.0 || ^22.5.1 || ^24.0.0` for all packages (except `@lavamoat/vog`, which already requires v22.6.0+).

The Node.js 18.x CI matrix entry has been removed accordingly.

BREAKING CHANGE: Node.js v18 is no longer supported.

* * *

_Note:_ This PR contains the changes from #1772 (and thus makes it redundant).


boneskull:

- branch: `main`
  - **[fix!: drop Node.js v18 support across the board #1886](https://github.com/LavaMoat/LavaMoat/pull/1886)** - branch: `boneskull/drop-v18` *(this PR)*
    - branch: `boneskull/upgrade-eslint`

---

.github/workflows/nodejs.yml
```diff
@@ -24,29 +24,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - node-version: 16.x
-            workspaces: >-
-              --workspace=packages/aa
```
naugtur, line 29: I was tempted to keep these working for older versions, but aa is in for a few bigger changes, so maybe it's time to drop the lag.

.github/workflows/nodejs.yml
naugtur, line 43: The only one not listed here is `vog`

naugtur (approved):

naugtur:
Should we incorporate this? https://github.com/LavaMoat/LavaMoat/pull/1879 ?
I think we should. (or just merge both at once)

PR #1894, opened 2026-03-05
TS support via type stripping

naugtur:
It should probably go behind a flag, but I didn't like the idea of prop-drilling it everywhere. 
Maybe an ambient singleton handling config or something?


packages/node/src/constants.js
```diff
@@ -166,6 +166,16 @@ export const NATIVE_PARSER_FILE_EXT = 'node'
...
 
+/**
+ * Extension for TypeScript modules
+ */
+export const MTS_PARSER_FILE_EXT = 'ts'
+
+/**
+ * Name of the TypeScript (type-stripped) parser language
+ */
+export const MTS_PARSER_NAME = 'mts'
```
boneskull, line 177: why not `.cts` too?

packages/node/src/compartment/with-type-stripping.js
```diff
@@ -0,0 +1,67 @@
+/**
+ * Provides {@link withTypeStripping}, a decorator that strips TypeScript type
+ * annotations from source code before delegating to an underlying parser.
+ *
+ * @packageDocumentation
+ * @internal
+ */
+
+import * as modulemodule from 'node:module'
```
boneskull, line 9: per convention I've been using
```suggestion
import * as nodeModule from 'node:module'
```

packages/node/src/compartment/with-type-stripping.js
```diff
@@ -0,0 +1,67 @@
...
+const decoder = new TextDecoder()
+const encoder = new TextEncoder()
+
+const stripTypeScriptTypes =
+  'stripTypeScriptTypes' in modulemodule
+    ? /** @type {(source: string) => string} */ (
+        modulemodule.stripTypeScriptTypes
+      )
+    : () => {
+        throw new Error(
```
boneskull, line 25: - I'd probably make a custom error for this, e.g., `NodeVersionError` or `NodeCapabilitiesError`
- ...and catch it later, wrapping in an appropriate `GenerationError` or `ExecutionError`, depending on where it bubbles up

packages/node/src/compartment/with-type-stripping.js
```diff
@@ -0,0 +1,67 @@
...
+ */
+export const withTypeStripping = (parserImpl) => {
+  // Note: we'd need to let parsers in node optionally accept strings and
+  // that would let us avoid re-encoding the string to bytes.
+  /** @type {ParserImplementation['parse']} */
+  const parse = (bytes, specifier, location, packageLocation, options) => {
+    // Only strip types in the root compartment, which is where the application
+    // code lives. This is determined by the compartment descriptor's label, if
+    // provided. If no label is provided, we conservatively assume it's not the
+    // root compartment and do not strip types.
```
boneskull, line 46: yes, but I don't understand _why_?

boneskull (commented):
some preliminary feedback

boneskull:
Note that the tests for this will need skipping under Node.js v20.x

packages/node/src/constants.js
naugtur: 1. I don't think node supports it, not sure babel would accept it
2. I don't think people use (or should use) it 

If you're using TypeScript anyway, esm works better with it.

packages/node/src/compartment/with-type-stripping.js
naugtur: Didn't know we had a convention for that. Lgtm

packages/node/src/compartment/with-type-stripping.js
naugtur: The value of wrapping/rethrowing as opposed to just complaining about node.js and let it fly all the way up is lost on me, but I appreciate the care and don't mind introducing it.

packages/node/src/compartment/with-type-stripping.js
naugtur: Is this about the comment not explaining why or you actually not knowing why?

The way people use typescript (and the way it works in node.js) is only the project code is transpiled/stripped and node_modules must ship JavaScript dist

Should I add something about that to the comment?

boneskull: Yeah, something referencing the fact that Node won't type-strip anything in `node_modules`.

PR #1897, opened 2026-03-11
feat(node): use endowmentsToolkit for module attenuation properly

naugtur:


naugtur:
CI failures seem to be from upstream.

PR #1899, opened 2026-03-12
fix(git-safe): adapt to github API change

naugtur:


boneskull (approved):

boneskull:
Doesn't seem like it works

boneskull:
I updated snapshots

naugtur:
> I updated snapshots

We explicitly don't want to update snapshots. they were correct. 
I left a console.log that was tripping up the snapshot :(

boneskull:
This is fine.  The other error in the build is fixed by #1900

PR #1905, opened 2026-03-19
fix(node): fix corner case where policy override path was guessed incorrectly

boneskull:
I could only reproduce this when using `--generate-recklessly`, but a similar programmatic use of `run()` is not beyond possibility.

- Removes some cruft from `policy-util.js` and types (policy debug stuff, unknown canonical name crunching)
- Adds more tests for `policy-util.js`
- Fix `ComposeOptions` type to allow identical properties for ease of use
- Update Wallaby config

### DIFF
--- a/packages/node/src/policy-gen/load-for-policy.js
+++ b/packages/node/src/policy-gen/load-for-policy.js
@@ -293,6 +293,8 @@ export const loadAndGeneratePolicy = async (
     )
     return { policy, packageJsonMap }
   } finally {
+    // we only check this if we have a policy override, because we're assuming
+    // the policy we just generated is valid, which may be a bad idea.
     if (policyOverride) {
       reportInvalidCanonicalNames(unknownCanonicalNames, knownCanonicalNames, {
         policy: policyOverride,
--- a/packages/node/src/policy-util.js
+++ b/packages/node/src/policy-util.js
@@ -14,12 +14,9 @@ import nodeFs from 'node:fs'
 import nodePath from 'node:path'
 import * as constants from './constants.js'
 import {
-  ATTENUATORS_COMPARTMENT,
-  DEFAULT_POLICY_DEBUG_FILENAME,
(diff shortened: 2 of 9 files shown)


boneskull:

- branch: `main`
  - **[fix(node): fix corner case where policy override path was guessed incorrectly #1905](https://github.com/LavaMoat/LavaMoat/pull/1905)** - branch: `boneskull/policy-override-path-guessing` *(this PR)*

---

packages/node/src/policy-gen/load-for-policy.js
```diff
@@ -293,6 +293,8 @@ export const loadAndGeneratePolicy = async (
     )
     return { policy, packageJsonMap }
   } finally {
+    // we only check this if we have a policy override, because we're assuming
+    // the policy we just generated is valid, which may be a bad idea.
```
naugtur, line 297: Policy generation has a less strict performance budget so I guess we could run it always

packages/node/src/policy-util.js
```diff
@@ -394,111 +390,6 @@ export const isPolicy = (value) => {
...
- * @template {CompartmentDescriptor} T
- * @template {string} K
- * @param {LavaMoatPolicy} policy
- * @param {Record<K, T>} compartmentDescriptors
- * @returns {{
- *   getInvalidCanonicalNames: () => PolicyCanonicalNameInfo[]
- *   policyCanonicalNameInfo: PolicyCanonicalNameInfo[]
- * }}
- */
-export const createGetInvalidCanonicalNamesFn = (
```
naugtur, line 470: Was this unused?

naugtur (approved):

naugtur:
(should) make lint happy before merging

packages/node/src/policy-util.js
boneskull: yes

PR #1909, opened 2026-03-21
fix(core): replace merge-deep with bespoke policy merge implementation

boneskull:
This replaces the `merge-deep` dependency (pulling in 15 _other_ dependencies) as well as `@types/merge-deep` with a bespoke merge algorithm.

It provides support for the deprecated `resolutions` field as well as retaining the `debugInfo` prop from `LavaMoatPolicyDebug` if present (does not attempt merge `debugInfo` from policy override, since policy override should never contain it). Code coverage for `mergePolicy.js` is now at or near 100%.

- Fixes some types in `dedupePolicyPaths`
- Refactor `testMerge` helper function (`mergePolicy.spec.js`) into AVA macro
- Update Wallaby config

## Questions

- Do we need to retain backwards-compat for `debugInfo`?
  - preferably yes 
- Do we need to retain backwards-compat for `resolutions`?
  -  yes. we'll get rid of them on the next major when we switch to config file
- Do we need to retain backwards-compat for the `env` field in `ResourcePolicy`? [A test suite in `lavamoat-node`](https://github.com/LavaMoat/LavaMoat/blob/9fa58f1b9d270272ed9d3bd3b25b240b9ce103b3/packages/lavamoat-node/test/envConfig.spec.js) expects this field.
  - It's not been used since MM moved off sass compiled to js from dart, but it's a feature that it had and we might eventually be pushed to reimplement it. this needs to stay.
- Why does `test/util.js` merge the policies without deduplicating?  If it just used `mergePolicy`, we wouldn't need to export `mergePolicyObjects` at all.
  - let's not export it


boneskull:

- branch: `main`
  - **[fix(core): replace merge-deep with ts-deepmerge #1909](https://github.com/LavaMoat/LavaMoat/pull/1909)** - branch: `boneskull/destroy-merge-deep` *(this PR)*

---

scutuatua-crypto (approved):

packages/core/src/mergePolicy.js
```diff
@@ -27,66 +248,76 @@ const { entries, hasOwn } = Object
...
-            policyItems: packagePolicy.builtin,
-            overrideItems: currentOverride.builtin,
-            resourceId,
-          })
-        }
+    const mergedPolicy = mergePolicyObjects(policyA, policyOverride)
+    entries(mergedPolicy.resources).forEach(([resourceId, packagePolicy]) => {
+      const currentOverride = policyOverride.resources?.[resourceId] ?? {}
+      if (hasOwn(packagePolicy, 'globals') && packagePolicy.globals) {
+        packagePolicy.globals = dedupePolicyPaths({
```
naugtur: If we didn't export the mergePolicyObjects function - we could put the globals and builtins dedupes in the specific functions that merge them and avoid the second loop. 
Unless we have negative testcases that rely on producing an undeduplicated policy, there's no reason not to. And if we do, we could feed them merged policy hardcoded. I'd go for it.

naugtur:
`merge-deep` might have been creating copies instead of what spread operator is doing, but I don't see how that would cause us any harm.

naugtur (approved):
LGTM. Hide `mergePolicyObjects` and :shipit:

packages/core/src/mergePolicy.js
boneskull: OK, I've made this refactor.

It affected a couple tests in `globals.spec.js`. The tests did not seem valuable, as they were testing a code path that cannot happen in production, and now _really_ can't happen in production, since there's no longer an option to not dedupe a policy upon merging.

I also did an overhaul of the `mergePolicy()` tests, which should now be generally be easier to understand at-a-glance.

> merge-deep might have been creating copies instead of what spread operator is doing, but I don't see how that would cause us any harm.

I've added a test to assert we're not retaining any object references in the merged policy from the original policy objects.

Please take another look and tell me what you think.

naugtur (approved):

PR #1911, opened 2026-03-23
feat(node): implement import.meta.resolve

naugtur:
I was too happy discovering I could, that I didn't think whether I should.

TODO: 
- [ ] Should there be limitations in what can be resolved/looked up?
- [x] tests


scutuatua-crypto (approved):

packages/node/src/exec/exec-compartment-class.js
```diff
@@ -6,13 +6,33 @@
...
 const wrapFunctionConstructor = endowmentsToolkit.defaultCreateFunctionWrapper
 
 /**
  * @import {SomeGlobalThis, ContextTestFn} from '../internal.js'
  * @import {CompartmentOptions} from 'ses'
  */
 
+/**
+ *
+ *
```
boneskull, line 19: pls add description

packages/node/src/exec/exec-compartment-class.js
```diff
@@ -6,13 +6,33 @@
...
+ */
+const importMetaHook = (_parentModuleSpecifier, meta) => {
+  assert(meta.url, 'meta.url should be provided by endo, but missing')
+  /**
+   * @param {string} specifier
+   * @param {string} [parent]
+   * @returns {string}
+   */
+  meta.resolve = (specifier, parent) => {
+    const require = createRequire(parent || meta.url)
```
boneskull, line 31: So it sounds like, per the test, that `parent` is always falsy (and I can confirm the API in the Node.js docs).  Why even bother with `parent`, then?

I don't actually see the second parameter appearing [on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta/resolve), either.  Where is this coming from?  Is this an Endo-ism?

packages/node/src/exec/exec-compartment-class.js
```diff
@@ -6,13 +6,33 @@
...
  */
 
+/**
+ *
+ *
+ * @param {string} _parentModuleSpecifier
+ * @param {Record<string, any>} meta
+ */
+const importMetaHook = (_parentModuleSpecifier, meta) => {
+  assert(meta.url, 'meta.url should be provided by endo, but missing')
```
boneskull, line 24: I probably wouldn't leave this here. If you want to keep it, would it be better in the body of `meta.resolve` when `parent` is falsy?

packages/node/test/unit/exec/execute.spec.js
```diff
@@ -169,4 +169,18 @@ test(
...
+test(
+  'import.meta.resolve - builtin and meta.url',
+  testExec,
+  new URL('../fixture/import-meta-resolve/app.js', import.meta.url),
+  {
+    resolvedBuiltin: true,
+    resolvedSelf: true,
+    nestedMetaUrlCorrect: true,
+    // TODO: nested relative resolution is broken because Node.js
+    // import.meta.resolve() ignores the second (parent) argument
```
boneskull, line 181: What is the TODO here, exactly?  Is this something to fix in Endo?

packages/node/src/exec/exec-compartment-class.js
naugtur: AFAIR it's meant to be added there eventually. But now that I think about it, it's too powerful anyway. Better to drop it.

packages/node/test/unit/exec/execute.spec.js
naugtur: oops, that's a comment about previous implementation

boneskull:
Looks like this just needs some updates and will be ready.  Bumping priority.

boneskull:
Regarding your question

> Should there be limitations in what can be resolved/looked up?

I'm not sure it's in our threat model, but a bad actor could probe files this way.  Of course, those files won't be able to be loaded if they aren't in the compartment map, but it still feels like it may expose too much.

Unfortunately I don't see any way to do this safely without extracting all modules from the compartment map and providing a bespoke resolution implementation (bypassing `require` altogether).

PR #1914, opened 2026-03-24
feat(core): endow the ability to write to global symbols

naugtur:


packages/core/src/endowmentsToolkit.js
```diff
@@ -117,6 +117,11 @@ function endowmentsToolkit({
           whitelistedReads.push(path)
           return
         }
+        if (packagePolicyValue === 'write-symbol') {
+          const symbolKey = Symbol.for(path)
```
boneskull, line 121: Do we need to dereference `for` here from the global `Symbol`?

naugtur: yes. my bad

PR #1923, opened 2026-03-26
chore(node): update static native tests to run under more platform/arch

boneskull:
Previously, the "static native" test was only running under `linux-x64`; now it works for `linux-arm64`, `darwin-arm64`, and `win32`.

Rebuilt all binaries and updated JSON fixture.


boneskull:
OK, never f'ing mind.

PR #1924, opened 2026-03-26
chore(node): run execution unit tests in worker thread

boneskull:
This changes unit tests of `run()` to execute in worker threads. This gives us two things:

1. Better control over stdout/stderr
2. Better isolation

The new `exec-runner.js` supports both on-disk fixtures and JSON fixtures.

While the strategy is similar for how we run scenarios, it's not quite the same—scenarios make assertions about the _output_ of the program; exec unit tests make assertions about the _exports_ of the program.


boneskull:

- branch: `main`
  - **[chore(node): run execution unit tests in worker thread #1924](https://github.com/LavaMoat/LavaMoat/pull/1924)** - branch: `boneskull/exec-worker-tests` *(this PR)*

---

PR #1928, opened 2026-03-27
feat(lavamoat-node): modernize resolve in lavamoat classic, opt-in only

naugtur:
This change makes lavamoat-node handle package.json->exports if you opt in.

- centralize resolver usage
- conditionally use the native `require.resolve` implementation 
- fall back to old implementation if it resolved symlinks outside node_modules or found incompatible files

Tests pass with combinations of flags, not sure if and how much I want to cover in CI. open to opinions.


packages/lavamoat-node/src/resolve.js
```diff
@@ -0,0 +1,101 @@
...
+const pathIsWithin = (rootPath, targetPath) => {
+  const relativePath = path.relative(rootPath, targetPath)
+  return (
+    relativePath === '' ||
+    (!relativePath.startsWith('..') && !path.isAbsolute(relativePath))
+  )
+}
+
+const NATIVE_IGNORES_SYMLINKS = !!process.env.NODE_PRESERVE_SYMLINKS
+const NATIVE_WITH_FALLBACK_ENABLED = !!process.env.LAVAMOAT_RESOLVE_NATIVE
```
boneskull, line 59: is this the opt-in?

we should probably document this.

packages/lavamoat-node/src/resolve.js
```diff
@@ -0,0 +1,101 @@
...
+
+let trace
+if (process.env.LAVAMOAT_DEBUG_RESOLVE) {
+  const traces = []
+  const num = Date.now()
+  trace = (report) => {
+    traces.push(report)
+    require('node:fs').writeFileSync(
+      `./trace-${num}.json`,
```
boneskull: probably better to go in e.g. `$XDG_STATE_HOME` imo

packages/lavamoat-node/src/resolve.js
```diff
@@ -0,0 +1,101 @@
...
+  return (
+    relativePath === '' ||
+    (!relativePath.startsWith('..') && !path.isAbsolute(relativePath))
+  )
+}
+
+const NATIVE_IGNORES_SYMLINKS = !!process.env.NODE_PRESERVE_SYMLINKS
+const NATIVE_WITH_FALLBACK_ENABLED = !!process.env.LAVAMOAT_RESOLVE_NATIVE
+
+let trace
```
boneskull, line 74: - I'd probably just leave this `undefined` and check if it exists before calling it.
- It should be typed

packages/lavamoat-node/src/resolve.js
```diff
@@ -0,0 +1,101 @@
...
+    }
+    let resolved
+    let report = {
+      via: 'unknown',
+      requestedName,
+      referrer,
+    }
+
+    if (NATIVE_IGNORES_SYMLINKS || NATIVE_WITH_FALLBACK_ENABLED) {
+      const refUrl = new URL(`file://${referrer}`)
```
boneskull, line 146: Is there any reason to use `url.pathToFileUrl()` over this?

packages/lavamoat-node/src/resolve.js
```diff
@@ -0,0 +1,101 @@
...
+      resolved = createRequire(refUrl).resolve(requestedName)
+      report.via = 'native'
+      if (
+        // resolved to a location outside the project's node_modules
+        (path.isAbsolute(resolved) &&
+          !pathIsWithin(projectNodeModules, resolved)) ||
+        !resolutionOmittedExtensions.includes(path.extname(resolved))
```
boneskull, line 153: would it be helpful to note which of these conditions caused the fallback?  different `report.via`?

packages/lavamoat-node/src/resolve.js
```diff
@@ -0,0 +1,101 @@
...
+  return (requestedName, referrer) => {
+    if (
+      isBuiltin(requestedName) ||
+      (path.isAbsolute(requestedName) &&
+        pathIsWithin(projectRoot, requestedName))
+    ) {
+      return requestedName
+    }
+    let resolved
+    let report = {
```
boneskull, line 139: pls type

packages/lavamoat-node/src/resolve.js
```diff
@@ -0,0 +1,101 @@
...
+  const projectNodeModules = path.join(projectRoot, 'node_modules')
+  return (requestedName, referrer) => {
+    if (
+      isBuiltin(requestedName) ||
+      (path.isAbsolute(requestedName) &&
+        pathIsWithin(projectRoot, requestedName))
+    ) {
+      return requestedName
+    }
+    let resolved
```
boneskull, line 137: pls type

packages/lavamoat-node/src/resolve.js
```diff
@@ -0,0 +1,101 @@
...
+          throw err
+        }
+      }
+      // return resolved path
+      return result
+    },
+  }
+}
+
+const pathIsWithin = (rootPath, targetPath) => {
```
boneskull, line 50: do we care about symlinks here?

packages/lavamoat-node/src/resolve.js
```diff
@@ -0,0 +1,101 @@
+const path = require('node:path')
+const { isBuiltin, createRequire } = require('node:module')
+const resolve = require('resolve') // file extension omitted can be omitted, eg https://npmfs.com/package/yargs/17.0.1/yargs
```
boneskull, line 4: would it be better to require `resolve` on-the-fly when we need it?

boneskull (changes requested):
looks ok. blocking on tests

a fixture setup wanting to load a package using `exports` would be fine.

packages/lavamoat-node/src/resolve.js
naugtur: yup. I wanted to keep it quiet for now so I can make breaking changes :sweat_smile:  but I can document it as experimental

packages/lavamoat-node/src/resolve.js
naugtur: That's not OS independent. I'd rather remove the `./` (which I forgot to do)

packages/lavamoat-node/src/resolve.js
naugtur: re1 I've done both many times and always hated what the code looked like when it was an undefined instead of a noop.

packages/lavamoat-node/src/resolve.js
naugtur: I took what was already there and didn't think about it. The reason to not use it is minimizing change I guess.

packages/lavamoat-node/src/resolve.js
naugtur: We explicitly care to _not_ consider symlinks here.
if symlinks are resolved in a way that make these mismatch - that's exactly what we want to consider not within. Policy relies on the node_modules structure and symlinks are one of the two problems that cause us to fall back to old implementation

packages/lavamoat-node/src/resolve.js
naugtur: I'd consider that premature optimization

packages/lavamoat-node/src/resolve.js
naugtur: done

packages/lavamoat-node/src/resolve.js
naugtur: working on this - it'll need to change anyway - it's not tracing what happened if there's a not-found error

packages/lavamoat-node/src/resolve.js
naugtur: It's visible in the paths listed, Won't help us with debugging any further and won't mean anything to end-users.

naugtur:
@boneskull I consider it shippable now

packages/lavamoat-node/package.json
```diff
@@ -32,8 +32,9 @@
   ],
   "scripts": {
     "lint:deps": "depcheck",
-    "test": "npm run test:npm && npm run test:yarn1",
+    "test": "npm run test:npm && npm run test:yarn1 && npm run test:resolve",
```
boneskull: (could)

```suggestion
    "test": "npm run test:npm; npm run test:yarn1; npm run test:resolve",
```

packages/lavamoat-node/package.json
```diff
@@ -32,8 +32,9 @@
...
     "test:npm": "npm run test:prepare && ava",
+    "test:resolve": "export LAVAMOAT_RESOLVE_NATIVE=1; npm run test:npm",
```
boneskull: (must)

```suggestion
    "test:resolve": "cross-env LAVAMOAT_RESOLVE_NATIVE=1 npm run test:npm",
```

boneskull (approved):
Just gotta fix the `test:resolve` script. Once that's done, merge

PR #1930, opened 2026-03-31
feat(allow-scripts): install scripts pinned to versions

naugtur:
TODO:
- [x] verify that skipping work
- [x] consider automated migration
- [x] document in readme
- [x] mark the feature as breaking


packages/allow-scripts/src/config.js
```diff
@@ -10,6 +10,81 @@ const normalizeBin = require('npm-normalize-package-bin')
...
 const bannedBins = new Set(['corepack', 'node', 'npm', 'pnpm', 'yarn'])
+
+/**
+ * Checks if versionToCheck is lower than or equal to versionAllowed
+ *
+ * @param {string} versionToCheck
+ * @param {string} versionAllowed
+ * @returns {boolean}
+ */
+const versionBelowOrEqual = (versionToCheck, versionAllowed) => {
```
boneskull, line 21: I'd probably just use `semver` here as we're bound to miss something.

naugtur:
failing tests are fixed in https://github.com/LavaMoat/LavaMoat/pull/1934 - need to rebase this when merged

boneskull (approved):

packages/allow-scripts/src/config.js
naugtur: semver supports way too many features for matching. but I thought about it some more and I want to get rid of this matching entirely

PR #1932, opened 2026-04-01
fix(webpack): source maps

itsyoboieltr:
sesPrefixFiles prepends SES lockdown to the runtime chunk using ConcatSource. It currently runs at stage 700 (OPTIMIZE_INLINE), which is after webpack generates source maps at stage 500 (DEV_TOOLING). The .map file gets frozen at stage 500, then the JS is modified at stage 700, so they get out of sync.

Moving to stage 499 (DEV_TOOLING - 1) means lockdown is prepended before source maps are generated. ConcatSource already correctly offsets the original source map by the number of prepended lines, so the .map file is correct with no other changes needed. Stage 499 is still after Terser (stage 400), so the lockdown code won't be minified.

### DIFF
--- a/packages/webpack/src/plugin.js
+++ b/packages/webpack/src/plugin.js
@@ -564,7 +564,7 @@ class LavaMoatPlugin {
           compilation.hooks.processAssets.tap(
             {
               name: PLUGIN_NAME,
-              stage: Compilation.PROCESS_ASSETS_STAGE_OPTIMIZE_INLINE,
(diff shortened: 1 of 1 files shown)


davidmurdoch (commented):
wonderful find!

naugtur:
This change would expose SES to be wrapped in `eval`  if `devtool` is set to one of the eval-based options. 
These are bad anyway, so I'll throw an error if they're used, for good measure. Otherwise this change is safe.

naugtur (approved):

PR #1933, opened 2026-04-02
feat(webpack): override the devtool setting that is likely to break lavamoat

naugtur:
Should have done it before. Now with the change in SES inlining stage I got down to it.


naugtur:
failing tests are fixed in https://github.com/LavaMoat/LavaMoat/pull/1934

boneskull (approved):

packages/webpack/src/plugin.js
```diff
@@ -180,6 +180,13 @@ class LavaMoatPlugin {
     // This setting creates a bit of confusion when it removes a reexport between packages and collapses the dependency tree, but it's got a potential to optimize some bundles a lot. We can support it by making sure we use the optimized connections when generating policy.
     // compiler.options.optimization.sideEffects;
 
+    if (compiler.options.devtool?.toString().includes('eval')) {
+      FORCED_CONFIG.push(
+        'devtool="eval*"->false (use any non-eval devtool instead)'
+      )
+    }
+    compiler.options.devtool = false
```
davidmurdoch, line 188: 👀 why you no let us have devtools?

PR #1941, opened 2026-04-10
fix(allow-scripts): usability improvements/fixes

naugtur:
- skip writing version for `$root$`
- fix `list` and improve its output to be more informative and helpful when versions don't match
- remove the older version matching as allowed


packages/allow-scripts/README.md
```diff
@@ -75,7 +77,6 @@ Configuration goes in `package.json`
 
 If allowScripts has an item set to true, the lifecycle scripts of that package will be executed in the run phase if the current installed version is still matching.
 
-
 > **Note** allow-scripts is pinning versions of the dependencies with lifecycle since v5. Items set to false will get automatically updated to the current version detected. To avoid churn in the list, you can remove the `#version` part from a disallowed item and it will be denied gardless of version. Versions are mandatory for allowed packages in case of maintainer compromise. You can opt out of versions with the `--skip-versions` flag.
 
 > **Note** While you can configure all install scripts that you've been running to date as allowed, it's best to limit the number of them in case a package with pre-existing install script gets exploited. To figure out which packages' scripts can be ignored, try [can-i-ignore-scripts](https://www.npmjs.com/package/can-i-ignore-scripts)
```
boneskull:
```suggestion
> [!NOTE]
>
> - `allow-scripts` is pinning versions of the dependencies with lifecycle since v5. 
> - Items set to `false` will be updated via `allow-scripts auto` to the current  detected version.
> - To avoid churn in the list, you can remove the `#version` suffix from a disallowed item and it will be denied regardless of version. 
> - Versions are mandatory for allowed packages in case of maintainer compromise. > - You can opt out of this behavior with the `--skip-versions` flag.

> [!NOTE]
>
> While you can configure all install scripts that you've been running to date as allowed, it's best to limit the number of them in case a package with pre-existing install script gets exploited. To determine which packages' scripts can be safely ignored, try [can-i-ignore-scripts](https://www.npmjs.com/package/can-i-ignore-scripts).
```

packages/allow-scripts/README.md
```diff
@@ -90,9 +91,9 @@ allow-scripts run
 
 It will fail if it detects dependencies which haven't been set up during [configuration](#Configure) of the package. You will be advised to run `yarn allow-scripts auto`.
 
-### Debug
+### List / debug
```
boneskull: So was `debug` the old command?  Or was `list` the old command? 

```suggestion
### List

> _Alias: **debug**_
```

packages/allow-scripts/README.md
```diff
@@ -90,9 +91,9 @@ allow-scripts run
...
 
-Prints comprehension of configuration and dependencies with lifecycle scripts, specifying _allowed_ and _disallowed_ packages.
+Prints the complete set of information that allow-scripts uses to populate and run the allowlist content.
```
boneskull:
```suggestion
Prints all information `allow-scripts` uses to populate and run the allowed scripts.
```

packages/allow-scripts/README.md
```diff
@@ -47,7 +49,7 @@ npx --no-install allow-scripts setup
 allow-scripts setup
 ```
 
-Adds a `.yarnrc` or `.npmrc` (the latter if `package-lock.json` is present) to the package, populates this file with the line `ignore-scripts true`. 
+Adds a `.yarnrc` or `.npmrc` (the latter if `package-lock.json` is present) to the package, populates this file with the line `ignore-scripts true`.
```
boneskull:
```suggestion
Adds (or updates) a `.yarnrc` or `.npmrc` (the latter if `package-lock.json` is present) file in the current package and appends `ignore-scripts=true`.
```

packages/allow-scripts/README.md
```diff
@@ -75,7 +77,6 @@ Configuration goes in `package.json`
 
 If allowScripts has an item set to true, the lifecycle scripts of that package will be executed in the run phase if the current installed version is still matching.
```
boneskull:
```suggestion
If `allowScripts` has an item set to true, the lifecycle scripts of that package will be executed in the run phase if the currently installed version matches.
```

packages/allow-scripts/README.md
```diff
@@ -35,10 +35,12 @@ pnpm add -D @lavamoat/allow-scripts
 In the following steps you will continue using `yarn` or `npx --no-install` if you've installed locally. Local installation is also directly callable in your project's `"scripts"` in `package.json`
 
 Example
+
 ```sh
 yarn allow-scripts setup
 npx --no-install allow-scripts setup
 ```
+
 > **Warning** if @lavamoat/allow-scripts was not installed prior, npx will try to download and run allow-scripts (note no namespace prefix) which is a different package. We suggest adding --no-install to prevent accidents.
```
boneskull:
```suggestion
> [!WARNING]
>
> If `@lavamoat/allow-scripts` is not already installed, `npx` will prompt to download and run package `allow-scripts` (missing the `@lavamoat` scope) **which is a different package**. We suggest using `npx --no-install` to prevent accidents.
```

packages/allow-scripts/src/report.js
```diff
@@ -69,42 +67,56 @@ function printPackagesByScriptConfiguration({
...
+    }
+    return pattern
+  }
+  console.log('\n# allowed packages')
   if (allowedPatterns.length) {
-    allowedPatterns.forEach((pattern) => {
-      const collection = packagesWithScripts.get(pattern) || []
-      console.log(`- ${pattern} [${collection.length} location(s)]`)
-    })
+    console.log('- ' + allowedPatterns.map(markUnused).join('\n- '), '\n')
```
boneskull: (could) use template literals consistently

packages/allow-scripts/src/config.js
```diff
@@ -11,76 +11,42 @@ const { loadCanonicalNameMap } = require('@lavamoat/aa')
...
   return {
     /**
-     * Checks if the pattern is blanketDisallowed regardless of version
+     * Checks if the pattern is known regardless of version
      *
      * @param {string} patternToCheck
      * @returns {boolean}
      */
-    disallowedExplicitlyOrAll: (patternToCheck) => {
+    known: (patternToCheck) => {
```
boneskull: (could) rename to `isKnown`

boneskull (approved):
No blockers, but I like my suggestions in `README.md`. 😄

PR #1949, opened 2026-04-13
feat(node): Read pnpm bin wrapper scripts

### LINKED ISSUE #1948: pnpm scripts are incompatible with `--bin`
Running `lavamoat --bin tsc -p lavamoat/node/tsc/policy.json -o lavamoat/node/tsc/policy-override.json -- -b` produces the following:

```
error   › Could not find executable tsc from ./
NoBinScriptError: Could not find executable tsc from ./
at new NoBinScriptError … 
  - code: 'LMN_NO_BIN_SCRIPT'
```

But the file does exist:

```
$ ls -lA node_modules/.bin/tsc
-rwxr-xr-x node_modules/.bin/tsc
```
```
$ cat !$
cat node_modules/.bin/tsc
#!/bin/sh
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")

case `uname` in
    *CYGWIN*|*MINGW*|*MSYS*)
        if command -v cygpath > /dev/null 2>&1; then
            basedir=`cygpath -w "$basedir"`
        fi
    ;;
esac

if [ -z "$NODE_PATH" ]; then
  export NODE_PATH="…"
else
  export NODE_PATH="…"
fi
if [ -x "$basedir/node" ]; then
  exec "$basedir/node"  "$basedir/../.pnpm/typescript@5.9.2/node_modules/typescript/bin/tsc" "$@"
else
  exec node  "$basedir/../.pnpm/typescript@5.9.2/node_modules/typescript/bin/tsc" "$@"
fi
```

I'm using pnpm and instead of symlinks into the package it is creating executable scripts.

whymarrh:
Closes #1948

This change updates the resolution of the `--bin` argument, now reading from the executable file where it is a pnpm wrapper script. `read-cmd-shim` will return the path to the underlying script, and we now resolve that full path.

### DIFF
--- a/packages/node/src/resolve.js
+++ b/packages/node/src/resolve.js
@@ -10,9 +10,15 @@ import path from 'node:path'
 import { PACKAGE_JSON } from './constants.js'
 import { NoBinScriptError, NoWorkspaceError } from './error.js'
 import { hrLabel, hrPath } from './format.js'
-import { isExecutableSymlink, isReadableFileSync, realpathSync } from './fs.js'
+import {
+  isExecutablePathSync,
+  isExecutableSymlink,
+  isReadableFileSync,
+  realpathSync,
+} from './fs.js'
 import { log } from './log.js'
 import { toPath } from './util.js'
+import readCmdShim from 'read-cmd-shim'
 
 /**
  * @import {ResolveBinScriptOptions, ResolveEntrypointOptions, ResolveWorkspaceOptions} from './internal.js'
@@ -103,7 +109,7 @@ export const resolveBinScript = (
     )
   }
   let current = workspace
-   
+
   while (true) {
     const maybeBinDir = path.join(current, 'node_modules', '.bin')
     const niceBinDir = hrPath(maybeBinDir)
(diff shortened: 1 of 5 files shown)


packages/node/src/resolve.js
```diff
@@ -116,6 +122,19 @@ export const resolveBinScript = (
...
+        const binPath = readCmdShim.sync(maybeBinPath)
+        const realBinPath = realpathSync(path.join(maybeBinDir, binPath), {
+          fs,
+        })
+        log.debug(
+          `Found executable ${hrPath(realBinPath)} linked from ${hrPath(maybeBinPath)}`
+        )
+        return realBinPath
+      } catch {
+        // Not a pnpm shim
```
naugtur:
```suggestion
        // Not a pnpm shim
        // readCmdShim throws 'ENOTASHIM' if a file exists but is not a recognizable shim
```

naugtur (approved):
Looks great!
Gonna have to make typescript happy tho.

TIL that the scripts are stable enough to get parsed for paths. Not sure if I like the fact that they're effectively public API surface now, but getting used to it. :)

whymarrh:
Added a type definition file for the package with the docs [copied from the README](https://www.npmjs.com/package/read-cmd-shim). TypeScript seems happy locally!

naugtur:
I noticed one more thing - I think you intended to add this to `packages/lavamoat-node` not `packages/node` 
Can you doublecheck which package you wanted it in?

whymarrh:
> I noticed one more thing - I think you intended to add this to `packages/lavamoat-node` not `packages/node` Can you doublecheck which package you wanted it in?

I did mean to add this to `@lavamoat/node`, that's what I've been working with locally.

Whether I should be doing that is maybe a separate issue. 😅

boneskull:
@whymarrh It's in heavy development, I'll say. 😄

whymarrh:
Yeah it's all good, this is part of that development!

whymarrh:
Ah, I was using `lavamoat-node` in the commit and PR title, my bad.

I fixed both of those and rebased on all the latest stuff. I think this should be good now.

PR #1956, opened 2026-04-15
fix(webpack): preserve sourcemaps when SES source transforms modify module code

itsyoboieltr:
## Problem

When `applySourceTransforms` modifies a module's source (to evade `<!--`, `-->`, `import()`, `eval()`), the generator discards the original sourcemap — it returns a plain string inside `ConcatSource` instead of a webpack `Source` object.

For most chunks this goes unnoticed because other unmodified modules still carry sourcemap data. But chunks containing **only transformed modules** end up with zero mapping data, and webpack skips `.map` file generation entirely. In MetaMask's build, this caused 2 out of 45 UI chunks to have no sourcemaps at all.

## Fix

`wrapper.js` now accepts a webpack `Source` object instead of a string. It:

1. Flattens the source via `sourceAndMap()` into a `SourceMapSource` (needed because the incoming source may be a compound `ReplaceSource` whose internal positions don't match the flattened output)
2. Runs `applySourceTransforms` on the string to get the SES-compatible result (used for validation)
3. Diffs the original and transformed strings to compute exact positional replacements
4. Applies those replacements via webpack's `ReplaceSource`, which automatically adjusts the sourcemap

`generator.js` is simplified — it passes the `Source` object directly to `wrapper` and uses the result in `ConcatSource(before, source, after)` with no branching.

## Test plan

- [x] MetaMask webpack build passes cleanly (no SES validation errors)
- [x] All 45 UI chunks produce `.map` files (previously 43/45)
- [x] Sentry stack traces resolve to original source with correct file paths, line numbers, and variable names

### DIFF
--- a/packages/webpack/src/buildtime/generator.js
+++ b/packages/webpack/src/buildtime/generator.js
@@ -3,7 +3,7 @@
 /** @typedef {import('webpack').sources.Source} Source */
 
 const {
-  sources: { ConcatSource },
+  sources: { ConcatSource, SourceMapSource },
 } = require('webpack')
 const { wrapper } = require('./wrapper.js')
 const diag = require('./diagnostics.js')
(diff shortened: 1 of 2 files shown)


naugtur:
closing in favor of #1965

PR #1964, opened 2026-04-20
feat(allow-scripts): add a check command to warn an update is needed in CI

naugtur:
Implemented in reaction to a user request :)

### DIFF
--- a/packages/allow-scripts/README.md
+++ b/packages/allow-scripts/README.md
@@ -6,13 +6,14 @@ A tool for running only the dependency lifecycle hooks specified in an _allowlis
 
 ## Usage
 
-`@lavamoat/allow-scripts` increases your project's security posture by first _denying_ all lifecycle scripts from running, then _allowing_ the ones you actually need.  The workflow looks like this:
+`@lavamoat/allow-scripts` increases your project's security posture by first _denying_ all lifecycle scripts from running, then _allowing_ the ones you actually need. The workflow looks like this:
 
 1. Run `allow-scripts setup` **once** per project to opt-out of running scripts
 2. Run `allow-scripts auto` to populate or update the allowlist (configured in `package.json`)
 3. Run `allow-scripts run` after every install to run the lifecycle scripts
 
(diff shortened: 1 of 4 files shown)


packages/allow-scripts/src/report.js
```diff
@@ -18,6 +19,27 @@ async function printPackagesList({ rootDir }) {
...
+    configs: { lifecycle },
+  } = await loadAllPackageConfigurations({ rootDir, skipVersions })
+
+  if (isScriptConfigurationClean(lifecycle, skipVersions)) {
+    console.log('allow-scripts - allowlist OK')
+  } else {
+    console.log(
+      'allow-scripts - allowlist needs update. Run "allow-scripts auto" to automatically update the allowlist configuration or "allow-scripts list" for details.'
+    )
+    process.exitCode = 1
```
boneskull, line 40: 👍

packages/allow-scripts/src/report.js
```diff
@@ -141,7 +163,46 @@ function printPackagesByScriptConfiguration({
   }
 }
 
+/**
+ * @param {ScriptsConfig} param0
+ */
```
boneskull, line 170:
```suggestion
/**
 * @param {ScriptsConfig} params
 * @param {boolean} [skipVersions]
 */
```

packages/allow-scripts/README.md
```diff
@@ -123,6 +124,14 @@ Prints all information `allow-scripts` uses to populate and run the allowed scri
 allow-scripts list
 ```
 
+### Check
+
+Exits 1 if the allowlist needs maintenance. Useful for CI to avoid running all tests on an installation that's missing results of lifecycle scripts.
```
boneskull:
```suggestion
`allow-scripts check` will exit with code `1` if the allowlist needs maintenance. This is useful to short-circuit CI on an installation that's missing results of lifecycle scripts.
```

boneskull (approved):
No blockers, a suggestion or two.

PR #1965, opened 2026-04-20
feat(webpack): sourcemap-preserving evasive transforms

naugtur:
currently rebuilt with webpack primitives.

Tried out the new babel-based evasive transforms instead for better meaning preservation. It's not that much slower thanks to how rare the need for a transform is. The main problem for now is we can't use them in webpack, because lavamoat-node doesn't support ESM so it won't work under lavamoat. Has to be postponed till we create a cjs export from `@endo/evasive-transform` or switch to lavamoat/node for good.

Babel experiments in branch: `naugtur/webpack-evasive-transform-babel`

this replaces #1956

### DIFF
--- a/packages/webpack/test/e2e-sourcemap.spec.js
+++ b/packages/webpack/test/e2e-sourcemap.spec.js
@@ -1,0 +1,131 @@
+const test = /** @type {import('ava').TestFn} */ (require('ava'))
+const path = require('node:path')
+const fs = require('node:fs')
+const { SourceMapConsumer } = require('source-map-js')
+const { scaffold } = require('./scaffold.js')
+const { makeConfig } = require('./fixtures/main/webpack.config.js')
+
+test.before(async (t) => {
+  const webpackConfig = makeConfig({
+    generatePolicy: false,
+  })
+  // Toggle to include and exclude minification
+  webpackConfig.mode = 'production'
+  webpackConfig.entry = { app: './sourcemap.js' }
+  webpackConfig.devtool = 'source-map'
+
+  await t.notThrowsAsync(async () => {
+    t.context.build = await scaffold(webpackConfig)
+  }, 'Expected the build to succeed')
+  t.context.mode = webpackConfig.mode
+  t.context.bundle = t.context.build.snapshot['/dist/app.js']
+  t.context.mapJson = JSON.parse(t.context.build.snapshot['/dist/app.js.map'])
+})
+
+test('webpack/sourcemap - .map file exists and is valid', (t) => {
+  t.truthy(t.context.build.snapshot['/dist/app.js.map'])
+  t.truthy(t.context.mapJson.version)
+  t.truthy(t.context.mapJson.mappings)
+  t.truthy(t.context.mapJson.sources.length > 0)
+})
+
+test('webpack/sourcemap - bundle references the sourcemap', (t) => {
+  t.regex(t.context.bundle, /\/\/# sourceMappingURL=app\.js\.map/)
+})
+
+test('webpack/sourcemap - SES transforms were applied in the bundle', (t) => {
+  // eval( → (0,eval)(
+  t.regex(t.context.bundle, /\(0,eval\)/)
+})
+
+test('webpack/sourcemap - production mangled identifiers away from bundle but map has them', (t) => {
+  if (t.context.mode === 'production') {
+    // These identifiers should not survive production minification
+    t.notRegex(t.context.bundle, /secretEvalResult/)
+    t.notRegex(t.context.bundle, /computeSomethingUnique/)
+  }
+  const allContent = t.context.mapJson.sourcesContent.join('\n')
+
+  t.regex(allContent, /secretEvalResult/)
+  t.regex(allContent, /computeSomethingUnique/)
+})
+
+test('webpack/sourcemap - sourcesContent has pre-transform code', (t) => {
+  const allContent = t.context.mapJson.sourcesContent.join('\n')
+
+  // The original eval( should appear in sourcesContent, not the transformed (0,eval)(
+  t.regex(allContent, /const secretEvalResult = eval\(/)
+})
+
+test('webpack/sourcemap - eval transform maps back to original position', (t) => {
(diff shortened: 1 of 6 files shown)


packages/webpack/test/e2e.manual.js
```diff
@@ -31,5 +34,6 @@ function writeFromSnapshot(snapshot) {
 scaffold(webpackConfig).then(({ stdout, snapshot }) => {
   console.log(stdout)
   writeFromSnapshot(snapshot)
+  console.log('\nBuild written to:', path.resolve('.', 'tmp', now))
```
boneskull:
```suggestion
  console.log('\nBuild written to:', path.resolve('tmp', now))
```

packages/webpack/test/e2e.manual.js
```diff
@@ -2,12 +2,15 @@
...
-}
+
+const { makeConfig } = require('./fixtures/main/webpack.config.js')
+
+const webpackConfig = makeConfig({
+  generatePolicy: false,
+})
+// webpackConfig.entry = { app: './sourcemap.js' }
+// webpackConfig.devtool = 'source-map'
+// webpackConfig.mode = 'production'
```
boneskull, line 13:
```suggestion
```

packages/webpack/test/e2e-sourcemap.spec.js
```diff
@@ -0,0 +1,131 @@
...
+  t.regex(
+    restOfLine,
+    /eval\(/,
+    'from the mapped column onwards, the source should contain eval('
+  )
+})
+
+/**
+ * Find the first occurrence of a substring in an array of lines. Returns
+ * 1-based line and 0-based column, or {line: -1, column: -1} if not found.
```
boneskull, line 121: 1-based line and 0-based column is unexpected

packages/webpack/test/e2e-sourcemap.spec.js
```diff
@@ -0,0 +1,131 @@
...
+  })
+  // Toggle to include and exclude minification
+  webpackConfig.mode = 'production'
+  webpackConfig.entry = { app: './sourcemap.js' }
+  webpackConfig.devtool = 'source-map'
+
+  await t.notThrowsAsync(async () => {
+    t.context.build = await scaffold(webpackConfig)
+  }, 'Expected the build to succeed')
+  t.context.mode = webpackConfig.mode
```
boneskull, line 20: what's the use of setting the mode in `context` if it's static?

packages/webpack/test/e2e-sourcemap.spec.js
```diff
@@ -0,0 +1,131 @@
...
+  // eval( → (0,eval)(
+  t.regex(t.context.bundle, /\(0,eval\)/)
+})
+
+test('webpack/sourcemap - production mangled identifiers away from bundle but map has them', (t) => {
+  if (t.context.mode === 'production') {
+    // These identifiers should not survive production minification
+    t.notRegex(t.context.bundle, /secretEvalResult/)
+    t.notRegex(t.context.bundle, /computeSomethingUnique/)
+  }
```
boneskull, line 46:
```suggestion
  if (t.context.mode === 'production') {
    // These identifiers should not survive production minification
    t.notRegex(t.context.bundle, /secretEvalResult/)
    t.notRegex(t.context.bundle, /computeSomethingUnique/)
  } else {
    t.regex(t.context.bundle, /secretEvalResult/)
    t.regex(t.context.bundle, /computeSomethingUnique/)
  }
```

packages/webpack/test/e2e-sourcemap.spec.js
```diff
@@ -0,0 +1,131 @@
...
+    generatePolicy: false,
+  })
+  // Toggle to include and exclude minification
+  webpackConfig.mode = 'production'
+  webpackConfig.entry = { app: './sourcemap.js' }
+  webpackConfig.devtool = 'source-map'
+
+  await t.notThrowsAsync(async () => {
+    t.context.build = await scaffold(webpackConfig)
+  }, 'Expected the build to succeed')
```
boneskull, line 19:
```suggestion
  t.context.build = await scaffold(webpackConfig)
```

rationale: this suite tests the behavior of sourcemaps, not of the test harness. let it fail if it fails.

boneskull (approved):
Some suggestions here, but nothing more important than a nit. A question or two as well. No blockers.

packages/webpack/test/e2e.manual.js
naugtur: This is a file for manual testing. These lines are useful to toggle on and off.

packages/webpack/test/e2e-sourcemap.spec.js
naugtur: It parametrizes a test later and it is useful to me to switch prod/dev for exploratory testing

packages/webpack/test/e2e-sourcemap.spec.js
naugtur: When mode is not production, it doesn't matter. For prod mode the assertions validate that the build configuration was not broken by something else. I hesitated if I even want to keep these two. No need to bloat the test IMO

packages/webpack/test/e2e-sourcemap.spec.js
naugtur: This wrapping exists to pinpoint the error to what happened. AFAIR it was confusing when the build failed. 

Note that it's happening in the before section and is reported differently. All other tests follow this convention too.

packages/webpack/test/e2e-sourcemap.spec.js
naugtur: I was initially surprised by it, but the implementation works correctly

I know line numbers start at 1.
Col numbers are not entirely correct in Webpack and beyond in general, sometimes off by 2digit numbers even without our intervention.

I'll investigate it further to make sure, but starting from -1 seems safe.

naugtur: column starts with 0 https://tc39.es/ecma426/#sec-terms-and-definitions-colun

PR #1970, opened 2026-05-04
feat(allow-scripts): add blocking git dependencies by default

naugtur:
This would probably be a breaking change

to be released with the breaking change to the bins feature.


boneskull:
Needs a `feat(allow-scripts)!` and a `BREAKING CHANGE:` in the commit msg description

boneskull (changes requested):
please add test for something. anything

PR #1971, opened 2026-05-04
fix(lavamoat-node): fix old resolution bug where paths containing # were misinterpreted

naugtur:
Took me a while to build a minimal repro, clanker failed miserably at that.


packages/lavamoat-node/src/resolve.js
```diff
@@ -55,8 +56,8 @@ const pathIsWithin = (rootPath, targetPath) => {
   )
 }
 
-const NATIVE_IGNORES_SYMLINKS = !!process.env.NODE_PRESERVE_SYMLINKS
-const NATIVE_WITH_FALLBACK_ENABLED = !!process.env.LAVAMOAT_RESOLVE_NATIVE
+const nativeIgnoresSymlinks = () => !!process.env.NODE_PRESERVE_SYMLINKS
+const nativeWithFallbackEnabled = () => !!process.env.LAVAMOAT_RESOLVE_NATIVE
```
naugtur, line 60: delayed read to allow instrumentation in testing. crude, but effective.

packages/lavamoat-node/test/projects/5/node-modules/a/package.json
```diff
@@ -1,4 +1,5 @@
 {
   "name": "a",
-  "version": "1.0.0"
-}
\ No newline at end of file
+  "version": "1.0.0",
+  "main": "./main.js"
```
naugtur, line 4: necessary to avoid the invalid resolve falling back to index.js and silencing the error. When index.js exists, we don't get `failed to load './x'` on incorrect resolution. 

A test testing the test to prevent regression is tempting...

packages/lavamoat-node/test/index_env.spec.js
```diff
@@ -0,0 +1,31 @@
+const path = require('node:path')
+const test = require('ava')
+const { parseForPolicy } = require('../src/parseForPolicy')
+const { runLavamoat } = require('./util')
+
+test.beforeEach(() => {
+  process.env.LAVAMOAT_RESOLVE_NATIVE = '1'
```
naugtur, line 7: These would mess up other tests if not in a separate file.

boneskull (approved):

PR #1972, opened 2026-05-05
feat(node)!: use @endo/parser-pipeline

boneskull:
BREAKING CHANGE: The API of `generatePolicy()` has changed; instead of resolving with a `LavaMoatPolicy`, it now resolves with an object having properties `policy: LavaMoatPolicy` and `hasWarnings: boolean`. This only affects programmatic consumption of `@lavamoat/node`. To migrate, destructure `policy` from the resolved object. `generatePolicy()` no longer accepts `write`/`writableFs`  options and does not write policy to disk; to write a policy, use `writePolicy `imported from `@lavamoat/node`.

This change adopts `@endo/parser-pipeline` for faster parsing in both execution and policy-gen workflows, as well as addition of support for `import()` in CommonJS sources. In a nutshell, this reduces the count of times a module's AST is parsed to one (1) without incurring extra memory pressure by caching.

Additional user-facing changes include:

- Addition of `--concurrency` CLI flag. This flag controls the maximum number of threads to use when generating policy.
- Addition of `--warnings` CLI flag. This flag defaults to `true` for the `generate` command and `false` for the `run` command. For the `run` command, this flag conflicts with `--quiet`. To disable warnings with `generate`, supply `--no-warnings`.
- Addition of `--treat-warnings-as-errors` CLI flag to the `generate` command.  If _any_ warnings would be emitted, the process will fail with exit code `1`.
- Aforementioned changes to the `generatePolicy` API.
- Policy generation now emits deltas after various operations. New spinner 🤪
- SES violation warnings now emit the absolute path to the problem file for easier navigation from terminal.
- Eliminate false-positives in entrypoint path resolution warning.


boneskull:

- branch: `main`
  - **[feat(node)!: use @endo/parser-pipeline #1972](https://github.com/LavaMoat/LavaMoat/pull/1972)** - branch: `boneskull/parser-pipeline` *(this PR)*
    - [feat(node)!: expose location-based includes #2018](https://github.com/LavaMoat/LavaMoat/pull/2018) - branch: `boneskull/better-bins`
      - branch: `boneskull/dogfood`
      - branch: `boneskull/e2e-dogfooding-node`
    - [fix(node)!: introduce new policy source API convention #1976](https://github.com/LavaMoat/LavaMoat/pull/1976) - branch: `boneskull/policy-api`
      - [chore(node): replace loggerr with @lavamoat/vog #1991](https://github.com/LavaMoat/LavaMoat/pull/1991) - branch: `boneskull/node-vog2`

---

packages/core/src/compact.js
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+/**
+ * Returns a compacted copy of `policyOverride` with any entries in `resources`
+ * that are already fully covered by `policy` removed, together with a flag
+ * indicating whether anything was actually removed.
+ *
+ * Only the `resources` field is examined for redundancy. All other fields
+ * (`include`, `root`, `resolutions`, etc.) are passed through unchanged because
+ * policy generation never produces them independently of overrides.
```
naugtur, line 34: User feedback is clear on this topic - they want the policy.json file to get overrides merged into for easier policy diff review. It's hard to comprehend and review the change to both at the same time otherwise.

boneskull: I recall you mentioning this on a call, but I can't see how this is here nor there.

The overrides are merged into `policy.json`.  The `Resources` that would be in `policy.json` _without merging from overrides_ are candidates for compaction.

boneskull:
This will fail CI until we cut a new Endo release, at which point I can upgrade to the latest versions of `@endo/compartment-mapper`, `ses`, `@endo/evasive-transform` and `@endo/parser-pipeline`.

packages/core/src/compact.js
naugtur: this is too handwavy a conversation to do async. Let's talk and put notes here.

packages/node/src/exec/exec-visitor.js
```diff
@@ -0,0 +1,267 @@
...
+ * @see {@link https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_EVAL_REJECTED.md}
+ */
+const SES_DIRECT_EVAL_RE = /(^|[^.])\beval(\s*\()/g
+
+/**
+ * Non-consuming variant of {@link SES_DIRECT_EVAL_RE} used to locate eval
+ * occurrences inside a string literal's value without swallowing the trailing
+ * `(`. The lookahead ensures `lastIndex` advances only past the `eval` token,
+ * which keeps multi-match scans simple.
+ */
```
naugtur, line 56: 👀

packages/core/src/compact.js
naugtur: ok, this is fine.

packages/core/src/compact.js
```diff
@@ -0,0 +1,107 @@
...
+ * @returns {CompactResult} Compacted policy override and whether any changes
+ *   were made
+ * @public
+ */
+exports.compactPolicyOverride = (policyOverride, policy) => {
+  /** @type {Resources} */
+  const compactedResources = {}
+  let compacted = false
+
+  for (const [name, overrideResource] of Object.entries(
```
boneskull: - [x] consistency

packages/node/src/policy-gen/policy-gen-parsers.js
```diff
@@ -0,0 +1,242 @@
...
+    workerData,
+    log: /** @type {LogFn} */ (log.debug.bind(log)),
+    onModuleStart,
+    onModuleComplete,
+  })
+
+  return {
+    mjs: asyncParsers[LANGUAGE_MJS],
+    cjs: asyncParsers[LANGUAGE_CJS],
+  }
```
boneskull: - [x] return the whole `asyncParsers`?

packages/webpack/package.json
```diff
@@ -51,7 +51,7 @@
     "browser-resolve": "2.0.0",
     "json-stable-stringify": "1.3.0",
     "lavamoat-core": "^18.0.5",
-    "ses": "1.15.0"
```
boneskull, line 54: - [x] pull this crap out, and in react-native-lockdown as well

packages/node/src/policy-gen/policy-gen-visitor.js
```diff
@@ -0,0 +1,140 @@
...
+      )
+      const builtins = /** @type {BuiltinsAnalyzerResults} */ (
+        new Set([...esmBuiltins, ...cjsBuiltins])
+      )
+
+      const violations = serializeViolations(
+        inspectSesCompat(/** @type {any} */ (capturedAst))
+      )
+
+      return { globals, builtins, violations }
```
boneskull, line 135: - [x] capture `globals`, `builtins`, `violations` instead; run inspection within `Program` visitor

boneskull: - [x] profile memory usage

packages/node/src/policy-gen/parse-worker.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+if (!parentPort) {
+  throw new Error('This module must be run as a worker thread')
+}
+
+/** @type {PolicyGenWorkerData} */
+const { globalRefs, globalsOptionsByLanguage = {}, builtinModules } = workerData
+
+runPipelineInWorker(
```
boneskull:
```suggestion
void runPipelineInWorker(
```

packages/node/src/policy-gen/policy-gen-visitor.js
boneskull: The results are in: the way I was doing it before was using a bit more memory than if I didn't.  Didn't seem like a whole lot, but it's better if I don't capture the AST.

packages/node/src/exec/exec-visitor.js
```diff
@@ -0,0 +1,267 @@
...
+       * direct identifier reference) makes it operate in global scope instead.
+       */
+      CallExpression(path) {
+        const { callee } = path.node
+        if (
+          callee.type === 'Identifier' &&
+          callee.name === 'eval' &&
+          !evadedEvals.has(callee)
+        ) {
+          // Mark the new callee so we don't re-visit it.
```
naugtur: why would we revisit it? is the visitor getting fed the part it creates right away? Wouldn't that cause infinite loops in many visitors?

packages/node/src/exec/exec-visitor.js
```diff
@@ -0,0 +1,267 @@
...
+        if (path.parent.type === 'TaggedTemplateExpression') {
+          return
+        }
+        for (const quasi of path.node.quasis) {
+          const escaped = escapeEvalInTemplateRaw(quasi.value.raw)
+          if (escaped !== quasi.value.raw) {
+            quasi.value.raw = escaped
+          }
+        }
+      },
```
naugtur: Worth noting: the case where `'eval'` is the name of a method on an object or a class is covered in evasive transforms already: https://github.com/endojs/endo/blob/bb4b14cbb3dea52d8209ffbead890d1b9459eafe/packages/evasive-transform/src/transform-code.js#L238

replacing `e` with an escape sequence for character e is a neat idea. Maybe it could be used to improve this: https://github.com/endojs/endo/blob/bb4b14cbb3dea52d8209ffbead890d1b9459eafe/packages/evasive-transform/src/transform-code.js#L95
It's not inserting new items into the array of strings, so maybe we don't have to bail out from tagged templates. 
The only case we'd need to bail out of would be if the function of the tagged template used `.raw` and passed it around to something that escapes characters further. Makes me wonder if it's a risk we'd want to take. I'm pretty sure "yes" in LavaMoat and expect pushback in Endo.
The likelihood of compatibility trouble because we don't do it is IMO higher than trouble because we do do it.

also, adding `eval\(` to the evadeRegexp in evasive-transform would potentially cover all this in one pass. I don't remember why we didn't do it back then, but it waas incremental work, so maybe I stopped before getting there.

packages/core/test/compact.spec.js
```diff
@@ -0,0 +1,268 @@
...
+
+test('compactPolicyOverride - does not mutate inputs', (t) => {
+  const policy = { resources: { foo: { globals: { process: true } } } }
+  const override = { resources: { foo: { globals: { process: true } } } }
+  const overrideCopy = JSON.parse(JSON.stringify(override))
+  const policyCopy = JSON.parse(JSON.stringify(policy))
+  compactPolicyOverride(override, policy)
+  t.deepEqual(override, overrideCopy)
+  t.deepEqual(policy, policyCopy)
+})
```
naugtur: maybe indicate that it was in fact changing something. Wasn't obvious at first glance.
```suggestion
  const result = compactPolicyOverride(override, policy)
  t.deepEqual(override, overrideCopy)
  t.deepEqual(policy, policyCopy)
  t.notDeepEqual(result.policy, override)
})
```

packages/node/src/exec/execute.js
```diff
@@ -32,9 +37,13 @@ import { load } from './load.js'
  */
 
 export const execute = async (entrypointPath, options = {}) => {
+  const { log = defaultLog } = options
+  const startTime = Date.now()
   const application = /** @type {ApplicationLoader<T>} */ (
     await load(entrypointPath, options)
   )
+  const duration = (Date.now() - startTime) / 1000
+  log.debug(`Loaded application in ${seconds(duration)}s`)
```
naugtur: (nit) This should not take seconds but miliseconds in most cases (I hope)
Why format it as seconds?

packages/node/src/compartment/import-hook.js
naugtur, line 67: (shower thoughts) I'm not sure if it's right or if it matters, but I just started wondering if calling `require(specifier)` inside of execute would be more correct. 
And I think if importNowHook was used in initial linking, it probably would, but in dynamic require it doesn't matter. 
What matters is that without calling it first, we don't have much of a way to figure out exports.
So not running it earlier is not really an option here.

packages/node/src/compartment/node-compartment-map.js
```diff
@@ -201,11 +202,22 @@ export const makeNodeCompartmentMap = async (
     log
   )
 
+  // if we have no policy but we do have policy overrides (policy generation
+  // use-case), we must gather the unknown canonical names after-the-fact using
+  // the overrides
+  if (policyOverride && !endoPolicy) {
+    unknownCanonicalNames = findUnknownCanonicalNames(
+      knownCanonicalNames,
+      policyOverride
```
naugtur, line 211: How does a function from `@endo/compartment-mapper` accept a LavaMoat policy override type? That feels wrong. Should it be passed through a policy translation function first?

If there's a way to avoid importing findUnknownCanonicalNames from a nested path in the package and calling it on a (accidentally matching type) policy override, but use a hook for it too, I'd consider that instead. The semver contract should not prevent findUnknownCanonicalNames changes from breaking this.

packages/node/src/policy-gen/policy-gen-parsers.js
```diff
@@ -0,0 +1,241 @@
+/**
+ * Provides {@link createPolicyGenParsers}, which creates async
+ * `ParserImplementation` instances backed by a worker pool for policy
+ * generation.
+ *
+ * Each worker runs the full pipeline (parse + tofu analysis + evasive
+ * transforms + module-source transforms + generate + record building) per
```
naugtur: I don't think there's a need to run all the transforms in policy generation, especially _after_ tofu. 

Also, from the first pass of reading, it doesn't seem like they're being included

packages/node/src/policy-gen/policy-gen-visitor.js
```diff
@@ -0,0 +1,140 @@
+/**
+ * Provides a single analyzer pass factory for policy-generation parsing.
+ *
+ * The pass captures the AST once and delegates to tofu's analysis primitives in
+ * `done()`, returning a combined `{ globals, builtins, violations }` result.
+ * Running all three analyses in a single pass avoids multiple Babel traversals
+ * and eliminates positional-tuple ordering fragility in `onModuleComplete`.
```
naugtur: I can't figure out what positional-tuple refers to.

packages/node/src/policy-gen/pipeline-config.js
```diff
@@ -0,0 +1,66 @@
...
+  globalsOptionsByLanguage = {},
+  builtinModules,
+}) => {
+  const mjsGlobalsOptions = {
+    globalRefs,
+    ...globalsOptionsByLanguage[LANGUAGE_MJS],
+  }
+  const cjsGlobalsOptions = {
+    globalRefs,
+    ...globalsOptionsByLanguage[LANGUAGE_CJS],
```
naugtur, line 50: (could) find a way to be more explicit about what this is and were it's coming from.

I'm struggling to track down where this is coming from. 
I mean, I will if I have to but this dependency injection has too many hops IMHO

packages/node/test/unit/exec/exec-visitor.spec.js
```diff
@@ -0,0 +1,283 @@
...
+  t.true(result.includes('console.log("hello")'))
+})
+
+test('rewrites direct eval to indirect eval', (t) => {
+  const source = 'const result = eval("1 + 1");'
+  const result = applyTransform(source)
+  t.false(result.includes('eval("1 + 1")'), 'should not have direct eval')
+  // Should contain (0, eval) pattern
+  t.true(
+    result.includes('(0,') && result.includes('eval'),
```
naugtur: this is a weird assertion. why not assert (0,eval) to make it clearer? If babel has a whitespace issue, at least we'll know when it does

packages/node/test/unit/exec/exec-visitor.spec.js
```diff
@@ -0,0 +1,283 @@
...
+    'live eval should still be rewritten to indirect eval'
+  )
+})
+
+test('escapes eval() inside bare template literals', (t) => {
+  const source = 'var x = `eval()`;'
+  const result = applyTransform(source)
+  t.notRegex(
+    result,
+    /(^|[^.])\beval\s*\(/,
```
naugtur: (nit) repetitions of this pattern with different descriptions are just one way this is needlessly elaborate. 
I'd prefer an array of: 
```js
{
  title: 'escapes eval() inside bare template literals',
  case: 'var x = `eval()`;',
  assertions: (result) => {} // optional
}
```
Where each case is fed through applyTransform and asserted that passing it to SES doesn't throw. The result of evaluation in a compartment is then passed to `assertions(result)` if it's defined for custom assertions like checking the string value is correct. (like 
```js
const recover = new Function(`${result}\nreturn x;`)
t.is(recover(), 'eval(foo) and eval(bar)')
```

packages/node/test/unit/policy/policy-gen-visitor.spec.js
```diff
@@ -0,0 +1,94 @@
...
+test('globals is a Map', (t) => {
+  const result = analyzeSource('export const x = 1;')
+  t.true(result.globals instanceof Map)
+})
+
+test('builtins is a Set', (t) => {
+  const result = analyzeSource('export const x = 1;')
+  t.true(result.builtins instanceof Set)
+})
+
```
naugtur: Why are these tests needed?

packages/node/README.md
```diff
@@ -144,7 +144,7 @@ npx exec lavamoat generate --help
 
 The following issues (or missing features) are _intended to be resolved_:
 
-- `await import()` is not yet supported in CommonJS scripts
+- ~~`await import()` is not yet supported in CommonJS scripts~~ (supported in v2.0.0)
 - _Scuttling_ (i.e. "deletion" of unused objects from the execution environment)
```
naugtur, line 148: I saw scuttling somewhere in config types, is this still up to date?
If so, let's make it outdated soon, so that we don't make decisions that don't go well with scuttling.

naugtur (approved):
approved, with comments to discuss.

packages/node/src/compartment/node-compartment-map.js
boneskull: This relies on the implicit similarities between our policy and Endo policy.  I _could_ pass `policyOverride` thru `toEndoPolicy`. It wouldn't make a difference, but perhaps would future-proof things.

packages/node/src/exec/execute.js
boneskull: good catch

packages/node/src/policy-gen/policy-gen-visitor.js
boneskull: - [x] clarify comment

it has to do with the previous iteration of this API

packages/node/src/policy-gen/policy-gen-parsers.js
boneskull: - [x] clarify comment

packages/node/src/exec/execute.js
boneskull: - [x] fix

packages/core/test/compact.spec.js
boneskull: - [x] fix

packages/node/test/unit/exec/exec-visitor.spec.js
boneskull: - [x] fix

packages/node/test/unit/exec/exec-visitor.spec.js
boneskull: - [x] refactor

packages/node/test/unit/policy/policy-gen-visitor.spec.js
boneskull: - [x] remove

packages/node/README.md
boneskull: I'm unclear on what you want to happen here.

packages/node/src/compartment/node-compartment-map.js
naugtur: If we don't get to use finding unknown canonical names through a hook and have to stick to the function, I'd argue it's necessary to pass through toEndoPolicy or create a simpler overridesToEndoPolicy that only copies `"pacakges"` fields

packages/node/src/policy-gen/pipeline-config.js
boneskull: I added a description to what the parameter is (a `PolicyGenWorkerData` object).  When we call `createPolicyGenParsers()` in `load-for-policy.js`, we bootstrap the workers with some configuration data which is passed thru `@endo/parser-pipeline` in `policy-gen-parsers.js`.  Searching for `PolicyGenWorkerData` usages would show this

packages/node/src/compartment/node-compartment-map.js
boneskull: Recall this: https://github.com/LavaMoat/LavaMoat/blob/d94d176d9e386e5be57cbba95ee25f951159668e/packages/node/src/exec/default-attenuator.js#L149-L166

Above, we're taking a subset of an Endo policy and masquerading as a LavaMoat policy; that's the reverse of what I'm doing here (though admittedly a smaller subset). My point is that it is not unprecedented.

`findUnknownCanonicalNames` searches `packages`, but it also searches the keys of `resources`; limiting it to the former would be insufficient.  It will ignore any fields it does not know about. Because this shape is "the same" between LM and Endo, we can get away with it. In terms of types, `findUnknownCanonicalNames` accepts an `@endo/compartment-mapper` `SomePolicy` and we're providing it our `LavaMoatPolicy`; the latter structurally satisfies the former.

So I'm saying that while I _could_ pass it thru `toEndoPolicy`, it doesn't currently buy us anything.

Theoretically, we would run into trouble if we broke this policy shape in _either_ LM or Endo.

packages/node/src/exec/exec-visitor.js
boneskull: Yeah I don't see how this could happen.

- [x] fix

boneskull:
This will need a squash, then I'll merge.

PR #1976, opened 2026-05-20
fix(node)!: introduce new policy source API convention

boneskull:
BREAKING CHANGE: removed `loadPolicies`, `isMergedPolicy`, and `constants.MERGED_POLICY_FIELD`; changed `toEndoPolicy` to accept only `Merged<LavaMoatPolicy>` (was polymorphic); changed `RunOptions` (affecting `run()`) and `GeneratePolicyOptions` (affecting `generatePolicy()`) to accept `policyInput: PolicyInput` instead of `policy`, `policyPath`, `policyOverride`, `policyOverridePath`; changed `GeneratePolicyResult.policy` type from `MergedLavaMoatPolicy` to `Merged<LavaMoatPolicy>`; removed types `MergedLavaMoatPolicy`, `UnmergedLavaMoatPolicy`, `LoadPoliciesOptions`, `WithPolicyOverridePath`, `WithPolicyOverrideOrPath`, `ToEndoPolicyOptions`, `ToEndoPolicyOptionsWithoutPolicyOverride`; see `MIGRATION.md`

This changes how public APIs accept information about policies.  Previously, we had many polymorphic and mutually exclusive options scattered around, including a notion of a `MergedPolicy` containing a brand `Symbol`.  Instead, when providing information about policies (whether on-disk or in-memory), programmatic users now provide an explicit `PolicySource` of various shapes, and unifies the internal processing (read: application of defaults).  Final path resolution (discovery/"guessing") is now decoupled; everything goes through the same internal execution flow of `loadPolicy` which results in a final `Merged<LavaMoatPolicy>`.

To assist, we export new helper functions:

### Before

```js
import { run } from '@lavamoat/node'

await run('./app.js', {
  policyPath: '/app/lavamoat/node/policy.json',
  policyOverridePath: '/app/lavamoat/node/policy-override.json',
})
```

### After

```js
import {
  run,
  policyInput,
  policySourceFromFile,
  policyOverrideSourceFromFile,
} from '@lavamoat/node'

await run('./app.js', {
  policies: policyInput({
    policy: policySourceFromFile('/app/lavamoat/node/policy.json'),
    override: policyOverrideSourceFromFile(
      '/app/lavamoat/node/policy-override.json'
    ),
  }),
})

// guess policy override path

await run('./app.js', {
  policies: policyInput({
    policy: policySourceFromFile('/app/lavamoat/node/policy.json'),
    projectRoot: '/app/'
  }),
})

// inline policy override object

await run('./app.js', {
  policies: policyInput({
    policy: policySourceFromFile('/app/lavamoat/node/policy.json'),
    override: policySourceFromInline({ resources: { /* .. */ } }),
  }),
})

```

Many permutations are possible.  While this is obviously more verbose, it normalizes the input across APIs and offers additional flexibility _for the end-user_ over what policies are loaded and how.

See the [ADR on the wiki](https://github.com/LavaMoat/LavaMoat/wiki/ADR:-@lavamoat-node-policy-input-abstraction) for background.


boneskull:

- branch: `main`
  - **[fix(node)!: introduce new policy source API convention #1976](https://github.com/LavaMoat/LavaMoat/pull/1976)** - branch: `boneskull/policy-api` *(this PR)*
    - [chore(node): replace loggerr with @lavamoat/vog #1991](https://github.com/LavaMoat/LavaMoat/pull/1991) - branch: `boneskull/node-vog2`
      - branch: `boneskull/vog-spinner`

---

packages/node/MIGRATION.md
```diff
@@ -0,0 +1,102 @@
...
+```js
+import {
+  run,
+  policyInput,
+  policySourceFromFile,
+  policyOverrideSourceFromFile,
+} from '@lavamoat/node'
+
+await run('./app.js', {
+  policyInput: policyInput({
```
naugtur, line 50: (could) 
I'm not fond of the name policyInput. It is an input like any argument to a function is. 
And the function policyInput seems unnecessarily exposed as public. Might as well run on the inside by default if we're expecting it to be called on the inputs.

As an end user I feel it's a hurdle that should not be necessary.

packages/node/MIGRATION.md
```diff
@@ -0,0 +1,102 @@
...
+
+const merged = await loadPolicy(
+  policyInput({
+    policy: policySourceFromFile('/app/lavamoat/node/policy.json'),
+    override: policyOverrideSourceFromFile(
+      '/app/lavamoat/node/policy-override.json'
+    ),
+  })
+)
+const endoPolicy = await toEndoPolicy(merged)
```
naugtur, line 90: why is this a public API? Do we need to expose any notion of Endo at all?

packages/node/src/cli.js
```diff
@@ -471,40 +478,46 @@ const main = async (args = hideBin(process.argv)) => {
           )
         }
 
-        /**
-         * This will be the policy merged with overrides, if present
-         *
-         * @type {LavaMoatPolicy | undefined}
-         */
-        let policy
+        const input = buildPolicyInput({
```
naugtur: What's the reason we're using a different policyInput function on the inside than we're recommending API users use?

packages/node/src/policy-input.js
```diff
@@ -0,0 +1,319 @@
...
+ * @public
+ */
+export const policyInput = ({
+  policy,
+  override,
+  projectRoot = process.cwd(),
+} = {}) =>
+  freeze({
+    policy: policy ?? policySourceFromDefault(projectRoot),
+    override: override ?? policyOverrideAuto(projectRoot),
```
naugtur, line 197: policyInput having defaults in case it's called with no arguments is all the more reason to call it on the inside not expose it IMO

packages/node/src/policy-input.js
```diff
@@ -0,0 +1,319 @@
...
+      policyPath = input.policy.path
+      break
+    case 'inline':
+      policy = input.policy.policy
+      break
+    default:
+      throw new InvalidArgumentsError(
+        `Unknown PolicySource kind: ${/** @type {any} */ (input.policy).kind}`
+      )
+  }
```
naugtur, line 233: I get the idea of avoiding polymorphism and that's a keeper. The layers here are too many tho. If we're making the API user wrap inputs in function calls anyway, there's no reason to have different function for each kind and separate handling for each kind on the inside. 

The different functions on the outside could already unify it and probably even read the actual file to return a policy object for validation.

resolvePolicySources is translating from a nicer API to an uglier one where policy and policyPath are separate things that the next layer has to handle anyway.

Some of the input format transitions can be eliminated and we'll get less code to maintain and follow in debugging as a result.

packages/node/src/policy-util.js
```diff
@@ -508,13 +356,69 @@ export const makeDefaultPolicyPath = (projectRoot = process.cwd()) => {
...
+ * {@link unwrapMerged}.
+ *
+ * @template {LavaMoatPolicy} [P=LavaMoatPolicy] The policy type. Default is
+ *   `LavaMoatPolicy`
+ * @param {P} policy A merged policy (typically produced by `mergePolicies`)
+ * @returns {Merged<P>}
+ * @public
+ */
+
+export const wrapMerged = (policy) => freeze({ policy, merged: true })
```
naugtur, line 410: is the reason behind this something to do with passing policy to/from workers? I'm not yet sure what serialization this is refering to.

packages/node/MIGRATION.md
boneskull: I mean, it's always been a public API, because it seemed like it'd have utility.

packages/node/MIGRATION.md
boneskull: hmm.  I'll think about this stuff more.  `policyInput` applies defaults, but it seems like we could move that up a level so that it's not directly exposed to the user.

packages/node/src/policy-util.js
boneskull: "Serialization" in this case means writing `policy.json` to disk.  We only want to write a _merged_ (`Merged`) policy to disk.  Likewise, some functions accept only merged policy; other operations should never be applied to a merged policy, etc.

boneskull:
I force-pushed, but it was only to get CI passing.

packages/node/src/policy-util.js
naugtur: I can't confidently say I follow. Let's talk this through

PR #1992, opened 2026-05-26
fix: ensure internal sequence of specifier handling does not leak con…

naugtur:
…trol flow from kernel code

It's a flaw in input processing, but not exploitable for a policy evasion because of how some comparisons were made attempts to use the object not its string value.
Better to plug it tho.


naugtur:
re test failing: I'm not sure why the stderr is still getting the output, but I'll just remove the log then.

boneskull (approved):

PR #1993, opened 2026-05-26
staged publishing with laverna

naugtur:
This is a breaking change, but over time it'll become the only acceptable default, so might as well rip the bandaid.
- yarn v4.15.0 is not out yet
- defaults to staging with `--local` for fallback

I'd be willing to switch to opt-in via `--staged` under pressure, but this is my preference and the sooner we have it out, the better for adoption ;)


naugtur:
not sure why tests didn't get triggered here at all...

boneskull:
@naugtur b/c github was broken, probably.  amend an empty commit and force-push?

boneskull (commented):
I'm good with this in principle, but we should either gracefully handle old `npm` versions (and test it!) _or_ just drop support for old `npm` versions entirely.

package.json
```diff
@@ -27,9 +27,10 @@
     "rebuild": "npm run --workspaces --if-present rebuild && npm run rebuild:types",
     "rebuild:types": "npm run clean:types && npm run build:types",
     "release": "npm run release:rebuild && npm run release:publish --newPkg=${npm_config_newpkg}",
-    "release:dry-run": "laverna --dryRun --newPkg=${npm_config_newpkg}",
-    "release:publish": "laverna --newPkg=${npm_config_newpkg}",
+    "release:dry-run": "laverna --local --dryRun --newPkg=${npm_config_newpkg}",
+    "release:publish": "laverna --local --newPkg=${npm_config_newpkg}",
     "release:rebuild": "npm run rebuild && npm run test:prep && npm run test:workspaces",
+    "release:stage": "npm run release:rebuild && laverna",
```
naugtur, line 33:
```suggestion
    "release:stage": "npm run rebuild && laverna",
```

.github/workflows/publish-stage.yml
```diff
@@ -0,0 +1,27 @@
+# this workflow stages versions to NPM with laverna after a release was merged
```
naugtur, line 1: this workflow might be unnecessary

PR #1994, opened 2026-05-26
chore(ci): use staged publishing w/ release-please

boneskull:
This modifies the `release-please` workflow to automatically publish a release to the staged registry.  Inspiration was taken from https://github.com/danielroe/uppt.

In a nutshell:

1. If the `release-please` job creates releases, they are stuffed into the job output. If it doesn't, the workflow ends.
2. If those releases are present and `release-please` succeeded, the `pack` job prepares the repo, builds, and packs all workspaces which were released.
3. The `pack` job then uploads the tarballs as an artifact. It outputs a list of the tarball filenames to be read by the `publish` job.
4. If `pack` succeeds, then the `publish` job runs, receiving the list of tarballs. It does _not_ install dependencies (though it must upgrade `npm`), but rather fetches the artifact uploaded by `pack`.
5. `publish` iterates through the list of tarballs and invokes a staged publish for each (with provenance).

Permissions are granted on a per-job basis. `release-please` needs to be able to write GitHub Releases and create PRs. `pack` doesn't need to do much of anything. `publish` needs `id-token` to enable a publish w/ provenance, as well as `environment: npm`.

Modified `prepare/action.yml` to omit `corepack`, since it no longer ships by default in latest Node.js. Replaced `--foreground-scripts` with `--ignore-scripts` just to be really damn sure.

### DIFF
--- a/.github/actions/prepare/action.yml
+++ b/.github/actions/prepare/action.yml
@@ -25,11 +25,6 @@ runs:
       with:
         node-version-file: .nvmrc
         cache: npm
-    - name: Set npm version
-      run: |
-        corepack enable
-        npm --version
-      shell: ${{ inputs.shell }}
     - name: Install dependencies
       run: npm ci --foreground-scripts
       shell: ${{ inputs.shell }}
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -6,14 +6,93 @@ on:
     branches:
       - main
 
-permissions:
-  contents: write
-  pull-requests: write
-
-name: release-please
+name: Release & Publish
 
 jobs:
+  # Creates release PRs, and creates GitHub Releases once those PRs are merged.
   release-please:
+    name: Release Please
+    concurrency:
+      group: release-please-${{ github.ref }}
+      cancel-in-progress: false
     runs-on: ubuntu-latest
+    outputs:
+      # boolean indicating whether any releases were created
+      releases_created: ${{ steps.release.outputs.releases_created }}
+      # JSON array of workspace paths that were released
+      paths_released: ${{ steps.release.outputs.paths_released }}
+    permissions:
+      contents: write
+      pull-requests: write
     steps:
       - uses: googleapis/release-please-action@5c625bfb5d1ff62eadeeb3772007f7f66fdcf071 # v4
+        id: release
+
+  # If any releases were created in release-please, this job packs the released
+  # workspaces into tarballs and uploads them as artifacts.
+  pack:
+    name: Pack Released Workspaces
+    runs-on: ubuntu-latest
+    needs: release-please
+    concurrency:
+      group: pack-${{ github.ref }}
+      cancel-in-progress: false
+    outputs:
+      # JSON array of tarball filenames that were packed (relative to outdir)
+      files: ${{ steps.pack.outputs.files }}
+    if: needs.release-please.result == 'success' && needs.release-please.outputs.releases_created == 'true'
+    permissions: {}
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/prepare
+        with:
+          node-version: 24.x
+      - run: npm run build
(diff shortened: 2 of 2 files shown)


boneskull:

- branch: `main`
  - **[chore(ci): use staged publishing w/ release-please #1994](https://github.com/LavaMoat/LavaMoat/pull/1994)** - branch: `boneskull/staged-publish` *(this PR)*

---

.github/workflows/release-please.yml
```diff
@@ -6,14 +6,87 @@ on:
...
+      - run: npm run build
+      - id: pack
+        run: |
+          paths='${{ needs.release-please.outputs.paths_released }}'
+          outdir='${{ runner.temp }}/lavamoat-pack'
+          readarray -t workspace_args < <(jq -r '.[] | "--workspace=" + .' <<< "$paths")
+          echo "Packing workspaces: ${workspace_args[*]}"
+          pack_output=$(npm pack "${workspace_args[@]}" --pack-destination="$outdir" --json)
+          files=$(jq -c '[.[].filename]' <<< "$pack_output")
+          echo "files=$files" >> "$GITHUB_OUTPUT"
```
boneskull, line 61: `danielroe/uppt` instead just publishes all tarballs in the uploaded artifact.  I felt it was _potentially_ safer to only publish the tarballs we expect.  Even if it isn't any safer, this gives us a better audit trail.

boneskull:
Unfortunately unless we want to set up verdaccio (assuming they support staged packages already), we won't know if this really works until we know if it really works.

.github/actions/prepare/action.yml
```diff
@@ -25,12 +25,7 @@ runs:
...
         node-version-file: .nvmrc
         cache: npm
-    - name: Set npm version
-      run: |
-        corepack enable
-        npm --version
-      shell: ${{ inputs.shell }}
     - name: Install dependencies
-      run: npm ci --foreground-scripts
+      run: npm ci --ignore-scripts
```
naugtur: (minor) we have .npmrc for that. should suffice. 
`--foreground-scripts` is to see outputs from the scripts that we did allow in the ogs

.github/workflows/release-please.yml
```diff
@@ -6,14 +6,87 @@ on:
...
+    permissions: {}
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - uses: ./.github/actions/prepare
+        with:
+          node-version: 24.x
+      - run: npm run build
+      - id: pack
+        run: |
+          paths='${{ needs.release-please.outputs.paths_released }}'
```
naugtur: (sec nit) if this was an action accepting inputs, it could be parsed in node.js. It presents a template injection risk if release-please job is tricked into producing user-controlled output

.github/workflows/release-please.yml
```diff
@@ -6,14 +6,87 @@ on:
...
+      cancel-in-progress: false
+    if: needs.pack.result == 'success' && needs.pack.outputs.files != '[]'
+    steps:
+      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: lavamoat-pack
+          path: ${{ runner.temp }}/lavamoat-tarballs
+      - run: npm install -g 'npm@^11.15.0'
+      - run: |
+          files='${{ needs.pack.outputs.files }}'
```
naugtur: (sec nit) - same as above - if this was an action accepting inputs, it could be parsed in node.js. It presents a template injection risk, but not exploitable as-is

docs on how to fix: https://docs.zizmor.sh/audits/#template-injection

(I'll be fixing all zizmor warinigs tomorrow)

naugtur (approved):
LGTM, would prefer to fix template injection just in case

.github/actions/prepare/action.yml
boneskull: I'll revert this

.github/workflows/release-please.yml
boneskull: https://github.com/LavaMoat/LavaMoat/pull/1995

PR #2012, opened 2026-06-03
fix(webpack): stabilize runtime lookup data

davidmurdoch:
## Summary

Stabilizes the `@lavamoat/webpack` runtime data that is serialized into emitted bundles for module/resource lookup and chunk validation.

The plugin currently embeds `idmap` and `kch` in the runtime using the order supplied by webpack/build-time traversal. Those values are runtime lookup tables, so their order does not affect behavior, but it does affect emitted bundle bytes. This sorts only those two lookup structures at the serialization boundary:

- `idmap`, including each resource's module ID list
- `kch` known chunk IDs

The patch intentionally leaves `unenforceable`, `ctxm`, and `externals` unchanged because they are not part of the observed local nondeterminism this fix targets.

## Validation

- `npm run build`
- `npm run --workspace=packages/webpack test` (96 tests passed)
- `npm run lint` (passed with existing warnings)

### DIFF
--- a/packages/webpack/src/runtime/runtimeBuilder.js
+++ b/packages/webpack/src/runtime/runtimeBuilder.js
@@ -41,6 +41,57 @@ class VirtualRuntimeModule extends RuntimeModule {
   }
 }
 
+/**
+ * Compare runtime IDs in a stable way without changing their types.
+ *
+ * @param {string | number} a
+ * @param {string | number} b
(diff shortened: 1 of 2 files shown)


naugtur:
closing in favor of #2014 
Thanks for the proposal. This was a good direction, but I preferred more generic sorting to be more future proof

PR #2013, opened 2026-06-05
fix(webpack): make runtime emission deterministic for reproducible builds

gauthierpetetin:
## Problem

Rebuilding the same source with `@lavamoat/webpack` produces a **different `runtime.js` (and the HTML/content-scripts that reference it) intermittently** — roughly a coin-flip in our testing (we observed several distinct `runtime.js` hashes for the same input). Every real code chunk is byte-identical every build; only LavaMoat's runtime metadata reshuffles.

This breaks **byte-for-byte build verification**. Concretely, it blocks Firefox Add-ons (AMO) reviewers (and our own release verification) from reproducing the published MetaMask extension from source, since reproduction requires an exact match.

## Root cause

The id lists emitted into the LavaMoat runtime are collected by iterating webpack's `compilation.chunks` — a `Set` whose iteration order is **not stable between builds** — and are serialized **without sorting**:

- `idmap` (`identifiersForModuleIds`) — `packages/webpack/src/buildtime/aa.js`. Both the entry order and the module-id order within each entry follow the chunk-Set order.
- `chunkIds` (`kch`), `unenforceableModuleIds`, `contextModuleIds` (`ctxm`) — `packages/webpack/src/plugin.js`, collected in the same `afterOptimizeChunkIds` pass.

## Fix

Sort these id lists before emitting them:

- **`aa.js`** — sort `identifiersForModuleIds` entries by resource key, and module ids within each entry.
- **`plugin.js`** — sort `chunkIds`, `unenforceableModuleIds`, and `contextModuleIds`.

Ids may be numbers or strings, so they're compared numerically (`localeCompare(..., { numeric: true })`). Runtime lookups against these structures are membership tests, so reordering is functionally safe — it only normalizes serialization order.

## Tests

- Added regression tests in `packages/webpack/test/aa.spec.js`:
  - `identifiersForModuleIds` is sorted (entries + within-entry).
  - `identifiersForModuleIds` is **independent of input order** (reversing the input paths yields identical output) — this is the key guard and fails without the fix.
- Full `@lavamoat/webpack` suite passes locally (97/97), with no snapshot changes.

## Note

This makes builds deterministic going forward. It does **not** retroactively match artifacts already published with the previous non-deterministic ordering — reproducibility applies once a release ships with this change.

### DIFF
--- a/packages/webpack/src/buildtime/aa.js
+++ b/packages/webpack/src/buildtime/aa.js
@@ -143,16 +143,36 @@ exports.generateIdentifierLookup = ({
 
   crossReference(identifiersWithKnownPaths, usedIdentifiers)
 
-  const identifiersForModuleIds = Object.entries(
-    Object.entries(pathLookup).reduce((acc, [, { aa, moduleIds }]) => {
+  // Module/chunk ids may be numbers or strings; compare them numerically so
+  // the resulting order is stable and independent of insertion order.
+  const byId = (a, b) =>
+    String(a).localeCompare(String(b), 'en', { numeric: true })
+
+  const grouped = Object.entries(pathLookup).reduce(
+    (acc, [, { aa, moduleIds }]) => {
       const key = translate(aa)
       if (acc[key] === undefined) {
         acc[key] = []
       }
       acc[key].push(...moduleIds)
       return acc
-    }, /** @type {Record<string, (string | number)[]>} */ ({}))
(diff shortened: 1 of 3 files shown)


naugtur:
To avoid chasing each unstable input, I'd rather stabilize the output, which is what we want to achieve in the first place.
Closing in favor of #2014

PR #2014, opened 2026-06-08
fix(webpack): reproducible builds require runtime to be stable on unstable input ordering

naugtur:
## Problem

Rebuilding the same source with `@lavamoat/webpack` produces a **different `runtime.js` (and the HTML/content-scripts that reference it) intermittently** — roughly a coin-flip in our testing (we observed several distinct `runtime.js` hashes for the same input). Every real code chunk is byte-identical every build; only LavaMoat's runtime metadata reshuffles.

This breaks **byte-for-byte build verification**. Concretely, it blocks Firefox Add-ons (AMO) reviewers (and our own release verification) from reproducing the published MetaMask extension from source, since reproduction requires an exact match.

## Root cause

The id lists emitted into the LavaMoat runtime are collected by iterating webpack's `compilation.chunks` — a `Set` whose iteration order is **not stable between builds** — and are serialized **without sorting**:

- `idmap` (`identifiersForModuleIds`) — `packages/webpack/src/buildtime/aa.js`. Both the entry order and the module-id order within each entry follow the chunk-Set order.
- `chunkIds` (`kch`), `unenforceableModuleIds`, `contextModuleIds` (`ctxm`) — `packages/webpack/src/plugin.js`, collected in the same `afterOptimizeChunkIds` pass.

## Fix

Stabilize runtime generation output

closes #2013

### DIFF
--- a/packages/webpack/src/runtime/assemble.js
+++ b/packages/webpack/src/runtime/assemble.js
@@ -1,5 +1,8 @@
 const { readFileSync } = require('node:fs')
 const { parse } = require('@babel/parser')
+const stableStringify = require('json-stable-stringify')
+
+const { isArray } = Array
 
 /**
  * Webpack adds a multiline comment in front of every line of the runtime code.
@@ -84,7 +87,12 @@ const assembleRuntime = (KEY, runtimeModules) => {
       sourceString = rawSource
     }
     if (data) {
-      sourceString = JSON.stringify(data)
+      // shallow sort is not breaking any mappings. nested sorting needs to be done with more caution
+      if (isArray(data)) {
+        data.sort()
+      }
+      // This was not necessary for the bugfix, but seems prudent
+      sourceString = stableStringify(data)
     }
     if (json) {
       sourceString = `LAVAMOAT['${name}'] = (${sourceString});`
--- a/packages/webpack/src/runtime/runtimeBuilder.js
+++ b/packages/webpack/src/runtime/runtimeBuilder.js
@@ -4,6 +4,8 @@ const { assembleRuntime, prepareSource } = require('./assemble.js')
 const { adjustScuttleConfig } = require('./scuttlingConf.js')
 const path = require('node:path')
 
+const { isArray } = Array
+
 const { RuntimeModule } = require('webpack')
 
 class VirtualRuntimeModule extends RuntimeModule {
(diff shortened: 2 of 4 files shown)


naugtur:
Problem: some of the arrays are order-significant afterall.

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -41,11 +43,16 @@ class VirtualRuntimeModule extends RuntimeModule {
...
+/**
+ * @import {
+ *   LavaMoatChunkRuntimeConfiguration,
+ *   LavaMoatPluginOptions
+ * } from "../buildtime/types"
+ */
+/** @import {LavaMoatPolicy} from "@lavamoat/types" */
+/** @import {RuntimeFragment} from "./assemble.js" */
+/** @import {Chunk} from "webpack" */
+/** @import {ProgressAPI} from "../buildtime/utils.js" */
```
naugtur, line 55: noting to see here, formatter changed its mind

packages/webpack/src/runtime/assemble.js
```diff
@@ -84,7 +87,12 @@ const assembleRuntime = (KEY, runtimeModules) => {
       sourceString = rawSource
     }
     if (data) {
-      sourceString = JSON.stringify(data)
+      // shallow sort is not breaking any mappings. nested sorting needs to be done with more caution
+      if (isArray(data)) {
+        data.sort()
```
naugtur, line 92: One thing to dislike that I will not fix in the current PR is that I'm using generic `.sort()` and relying on it stringifying the possible nested arrays for comparison instead of the fancy implementation that's aware of the data shape. What I get in return is the ability to future-proof by making all arrays in LM runtime sorted. I'd rather refactor to Set and `Array.from().sort()` every set while leaving arrays ordered where it's meaningful than make sorting aware of the data structures and therefore applicable only to well know unstable collections.

packages/webpack/test/runtime-stability.spec.js
```diff
@@ -0,0 +1,249 @@
...
+        ['package-a', [3, 1, 2]],
+        ['package-b', [5, 4]],
+      ],
+      unenforceableModuleIds: [30, 10, 20],
+      contextModuleIds: [300, 100, 200],
+      externals: { 60: 'crypto', 40: 'fs', 50: 'path' },
+    },
+  })
+
+  t.is(head(outputA), head(outputB), 'first 500 chars must match')
```
naugtur, line 99: this assertion is redundant, until you see how far you need to scroll back to see the difference

itsyoboieltr (approved):

PR #2018, opened 2026-06-13
feat(node)!: expose location-based includes

boneskull:
BREAKING CHANGES: The shape of `LavaMoatPolicy.include` has changed; the `entry: string` field is now `modules?: string[]`, representing an array of relative paths. Instead of the `name` field, each object can have a `location` field which is a path to the package directory relative to the project root (or CWD). `location` should be used for packages which do not appear anywhere in the dependency graph. The `name` field still references a canonical name.

Fixes
-----

- Fix issues with reporting unknown canonical names incorrectly

Types
-----

- `WithLoadForMapOptions` type removed in lieu of `ConsumerCaptureFromMapOptions`
- `WithRawPowers` renamed to `PowerBuilders`; `WithRawReadPowers` renamed `WithReadPowersXorBuilders`
- Removed type `WithReadPowersAndTrust`
- Re-export some types from `type-fest`
- Remove some unhelpful internal option-bag interfaces; simplify others
- Add missing docstrings to many types
- Fix some duplication of types between `types.ts` and `internal.ts`

Error Handling
--------------

- Better error handling of unexpected errors during policy gen
- Reformat a few messages

Internal
--------

- Rename `isObjectyObject` to `isNonArrayObject`
- Includes refactors of `load-for-policy` module; splits hooks out into another module
- Remove cruft from `makeNodeCompartmentMap`, `load`
- Fix test harness problems
- Create `isIncludeEntryByName` & `isIncludeEntryByLocation` type guards; update `isPolicy` type guard to use
- Update scenario test harness to allow exceptions to expected failures (found one!); do not just _skip_ expected failures
- Create `readEntryPackageDescriptor`
- Add some missing unit tests

* * *
 
### feat(core)!: support expanded LavaMoatPolicy.include

* * *

### feat(types)!: expand LavaMoatPolicy.include


boneskull:

- branch: `main`
  - **[feat(node)!: expose location-based includes #2018](https://github.com/LavaMoat/LavaMoat/pull/2018)** - branch: `boneskull/better-bins` *(this PR)*
    - branch: `boneskull/dogfood`
    - branch: `boneskull/e2e-dogfooding-node`

---

boneskull:
I'd recommend landing this alongside #1972.  It makes webpack work in my `__e2e-dogfooding__` branch via `lavamoat --bin webpack`.  

For the curious, the policy overrides look like this:

```json
{
  "include": [
    "webpack-cli",
    "webpack-cli>import-local",
    "file-loader>webpack",
    {
      "name": "@lavamoat/__e2e-dogfooding__",
      "entry": "./webpack.config.js"
    },
    "@lavamoat/__e2e-dogfooding__>html-webpack-plugin",
    "babel-loader",
    "@lavamoat/__e2e-dogfooding__>@babel/preset-env",
    "@lavamoat/__e2e-dogfooding__>@babel/preset-env>@babel/plugin-transform-modules-umd",
    "@lavamoat/__e2e-dogfooding__>@babel/preset-env>@babel/plugin-transform-regenerator",
    "@lavamoat/__e2e-dogfooding__>@babel/preset-env>babel-plugin-polyfill-corejs2>@babel/helper-define-polyfill-provider",
    "@babel/preset-react>@babel/plugin-transform-react-display-name"
  ],
  "resources": {
    "graceful-fs": {
      "globals": {
        "___graceful-fs_previous": "write",
        "___graceful-fs_queue": "write"
      }
    },
    "webpack": {
      "packages": {
        "webpack-cli": true
      }
    },
    "webpack-cli": {
      "packages": {
        "webpack-cli>import-local": true,
        "webpack": true,
        "@lavamoat/__e2e-dogfooding__": true
      }
    },
    "webpack-cli>import-local": {
      "packages": {
        "webpack-cli": true
      }
    },
    "loader-runner": {
      "packages": {
        "@lavamoat/__e2e-dogfooding__>html-webpack-plugin": true,
        "babel-loader": true
      }
    },
    "@lavamoat/__e2e-dogfooding__>html-webpack-plugin": {
      "builtin": {
        "node:url": true
      }
    },
    "terser-webpack-plugin>jest-worker": {
      "packages": {
        "terser-webpack-plugin": true
      }
    },
    "@babel/core": {
      "packages": {
        "@lavamoat/__e2e-dogfooding__>@babel/preset-env": true,
        "@lavamoat/__e2e-dogfooding__>@babel/preset-typescript": true,
        "@babel/preset-react": true
      }
    },
    "@lavamoat/__e2e-dogfooding__>@babel/preset-env>babel-plugin-polyfill-corejs2>@babel/helper-define-polyfill-provider": {
      "builtin": {
        "path": true,
        "module": true
      }
    }
  },
  "root": {
    "usePolicy": "webpack"
  },
  "additionalLocations": [
    {
      "location": ".",
      "modules": ["./webpack.config.js"]
    }
  ]
}
```

packages/node/src/compartment/includes.js
boneskull, line 1: - [ ] consider filtering functions which pull out the appropriate `includes` subtype, then make these functions accept those specific subtypes

packages/core/src/mergePolicy.js
```diff
@@ -206,26 +208,47 @@ const mergeInclude = (policy, policyOverride, mergedPolicy) => {
...
-          }
-          acc.push({ ...item })
-        }
+        return `s:${item}`
+      }
+      const modules = item.modules ? [...item.modules].sort().join('\0') : ''
+      return 'name' in item
+        ? `n:${item.name}\0${modules}`
+        : `l:${item.location}\0${modules}`
+    }
```
boneskull, line 231: - [x] ~~deslop~~

boneskull: after looking at this closely, it's actually not a bad idea.  `\0` is a disallowed character in filenames in both windows and linux (though they have different meanings), afaict. It's a reasonable delimiter.

naugtur:
> ```json
> "@lavamoat/__e2e-dogfooding__>*
> ```
Items starting with that should theoretically not be necessary for policy generation after the top one is provided. Policy generation would follow the graph down from there. Are they necessary for runtime?

packages/core/src/mergePolicy.js
```diff
@@ -206,26 +209,46 @@ const mergeInclude = (policy, policyOverride, mergedPolicy) => {
...
+
+    /**
+     * Deep-clones an include item so the merged result shares no object
+     * references with the inputs.
+     *
+     * @param {IncludePolicy} item
+     * @returns {IncludePolicy}
+     */
+    const clone = (item) =>
```
naugtur, line 240: should we default to cloning the whole merged policy via JSON or structuredClone?

packages/core/src/mergePolicy.js
```diff
@@ -206,26 +209,46 @@ const mergeInclude = (policy, policyOverride, mergedPolicy) => {
...
-    const seenStrings = new Set()
-    /** @type {Map<string, Set<string>>} */
-    const seenObjects = new Map()
+    const seen = new Set()
 
-    mergedPolicy.include = combined.reduce((acc, item) => {
+    /**
+     * Computes a canonical dedupe key for an include item.
+     *
+     * We hse `\0` as a separator because it's disallowed in filenames in both
```
naugtur, line 217:
```suggestion
     * We use `\0` as a separator because it's disallowed in filenames in both
```


Why is this so complex? Can't we put all includes in an array and remove duplicates by stable stringify match?

packages/core/src/mergePolicy.js
naugtur: (nit) it doesn't seem this function needs a new instance for every call to mergeInclude()

packages/node/src/compartment/includes.js
```diff
@@ -0,0 +1,141 @@
...
+
+    return include.modules?.length
+      ? include.modules.map((module) => ({
+          compartment: compartment.label,
+          entry: module,
+        }))
+      : [{ compartment: compartment.label, entry: '.' }]
+  }
+
+  // this could have been a beautiful nested ternary
```
naugtur, line 128: If it's not just because typescript couldn't grok it, maybe we should have it be nice for us and force-cast or ignore something.

packages/node/src/compartment/includes.js
```diff
@@ -0,0 +1,141 @@
...
+      return [...acc, include]
+    }
+    if (isIncludeEntryByName(include)) {
+      return [...acc, ...handleIncludeEntryByName(include)]
+    }
+    if (isIncludeEntryByLocation(include)) {
+      return [...acc, ...handleIncludeEntryByLocation(include)]
+    }
+    return acc
+  }, /** @type {NonNullable<PreloadOption['_preload']>} */ ([]))
```
naugtur, line 140: I'm not enjoying the complexity of how includes are processed here, but can't see much redundancy beyond using stable stringify for keys

packages/node/src/compartment/includes.js
```diff
@@ -0,0 +1,141 @@
...
+ * @param {string} [options.projectRoot] Project root
+ * @returns {NonNullable<MapNodeModulesOptions['additionalLocations']>}
+ */
+export const buildAdditionalLocations = (
+  includes = [],
+  { projectRoot = process.cwd() } = {}
+) =>
+  includes.reduce((acc, include) => {
+    if (isIncludeEntryByLocation(include)) {
+      if (include.location.startsWith('..')) {
```
naugtur, line 46: (should) If we're relying on it for any security or integrity, this is not enough. `./../` will pass through.
Unless this is just to prevent users from doing something nonsensical, in which case it might be good enough.

packages/node/src/policy-gen/load-for-policy.js
```diff
@@ -228,10 +257,10 @@ export const loadAndGeneratePolicy = async (
...
   })
 
-  const _preload = policyOverride?.include?.map((include) =>
-    typeof include === 'string'
-      ? { compartment: include, entry: '.' }
-      : { compartment: include.name, entry: include.entry }
+  const _preload = buildPreloads(
+    packageCompartmentMap,
+    policyOverride?.include,
+    { projectRoot }
```
naugtur, line 263: the `_` in here suggests we should have a plan to replace preload with something else and this is to bridge the gap until that happens. 

What's the situation here?

packages/core/src/mergePolicy.js
boneskull: habit, I guess. I tend to avoid stringifying for anything that isn't serialization.

- [x] use stable stringify to compare

packages/core/src/mergePolicy.js
boneskull: > should we default to cloning the whole merged policy via JSON or structuredClone?

Seems like it might simplify things if we did this up front.

- [x] investigate eager structured cloning

boneskull:
> > ```json
> > "@lavamoat/__e2e-dogfooding__>*
> > ```
> 
> Items starting with that should theoretically not be necessary for policy generation after the top one is provided. Policy generation would follow the graph down from there. Are they necessary for runtime?

Wait...what is this in reference to?

packages/node/src/compartment/includes.js
boneskull: No, it's not a type issue.  It's just that I've learned to avoid nested ternaries because people tend to react with revulsion.

packages/node/src/compartment/includes.js
boneskull: Honestly, I hadn't thought about it too hard.  If I'm being diligent, I'm going to figure out what exactly happens when someone tries this.  It seems like ascending above the project root (where relative paths are computed from) _is_ nonsensical, but I'm unaware of any other implications.

packages/node/src/policy-gen/load-for-policy.js
boneskull: variable `_preload` matches the `captureFromMap` `_preload` option. The variable could be called `preload`, but then I couldn't use object shorthand 😞

packages/node/src/compartment/includes.js
```diff
@@ -0,0 +1,141 @@
...
+        throw new InvalidPolicyError(
+          `Include location cannot ascend above the project root (${hrPath(projectRoot)}): ${hrPath(include.location)}`
+        )
+      }
+      const location = toFileURLString(
+        path.join(projectRoot, include.location, '/')
```
boneskull, line 50: - [x] `/` seems like it should do nothing??

packages/core/src/mergePolicy.js
boneskull: re: cloning; I don't think we need to make this change in this PR. needs another look tho

packages/node/src/compartment/includes.js
boneskull: No; I was confusing with `./`; `path.join('./', 'foo', 'bar')` becomes `foo/bar`

naugtur (approved):
everything addressed. 🚢

PR #2025, opened 2026-06-18
feat(run): add @lavamoat/run, a secure npx replacement

kumavis:
@lavamoat/run ships the `lavax` bin: an npx-style command that fetches and runs a package's bin while applying LavaMoat's security principles.

- install scripts are disabled by default (override with --allow-scripts), matching @lavamoat/allow-scripts
- the bin runs under @lavamoat/node: SES lockdown plus a generated, enforced per-package policy, treating the fetched package as an untrusted root
- packages install into an isolated per-spec sandbox cache; the policy is generated on first use (trust-on-first-use) and reused thereafter
- everything after <spec> is forwarded to the bin, like npx

Wires the package into the monorepo (tsconfig refs, eslint typed config, release-please) and documents it in the root README and AGENTS.md.

Claude-Session: https://claude.ai/code/session_01NSridaT397M7CiBua4Jxpc

### DIFF
--- a/packages/run/src/args.js
+++ b/packages/run/src/args.js
@@ -1,0 +1,85 @@
+/**
+ * Splits a raw argument vector into `lavax`'s own options, the package spec,
+ * and the arguments to forward to the package's bin.
+ *
+ * `lavax` follows the same contract as `npx`: every token after the `<spec>`
+ * positional belongs to the package's bin, _not_ to `lavax`. This cannot be
+ * expressed with yargs' option parser alone (separate-token option values
+ * conflict with `halt-at-non-option`), so we split first and hand yargs only
+ * the leading options plus the spec.
+ *
+ * **All exports in this module are considered part of the public API.**
+ *
+ * @packageDocumentation
+ */
+
+/**
+ * `lavax` options which consume a following token as their value when written
+ * in separate-token form (e.g. `--cache /tmp/x`).
+ */
+const VALUE_OPTIONS = new Set([
+  '-c',
+  '--call',
+  '--registry',
+  '--cache',
+  '-p',
+  '--policy',
+])
+
+/**
+ * The result of {@link splitArgs}.
+ *
+ * @typedef SplitArgs
+ * @property {string[]} optionTokens Tokens that are `lavax`'s own options
+ * @property {string | undefined} spec The package spec, if present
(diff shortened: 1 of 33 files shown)


packages/run/src/args.js
```diff
@@ -0,0 +1,85 @@
...
+
+    // First non-option token is the spec; everything after is forwarded.
+    let forwardedArgs = argv.slice(i + 1)
+    if (forwardedArgs[0] === '--') {
+      forwardedArgs = forwardedArgs.slice(1)
+    }
+    return { optionTokens, spec: token, forwardedArgs }
+  }
+
+  return { optionTokens, spec: undefined, forwardedArgs: [] }
```
naugtur, line 84: node.js has a built-in args parser now. If we narrow down the version support a little, this could use it.

wait a second... I see yargs used below. this can be done with yargs AFAIR

packages/run/src/install.js
```diff
@@ -0,0 +1,155 @@
...
+ * @param {Omit<InstallOptions, 'cwd' | 'spawn'>} [options] Options
+ * @returns {string[]}
+ */
+export const buildNpmArgs = (
+  spec,
+  { allowScripts = false, registry, quiet = false } = {}
+) => {
+  const args = ['install', '--no-audit', '--no-fund']
+  if (!allowScripts) {
+    args.push('--ignore-scripts')
```
naugtur, line 102: Latest npm 11 and upcoming npm 12 introduce a bunch of controls that are more granular than this. lavax should generate .npmrc in the location it installs to.

packages/run/src/resolve-bin.js
```diff
@@ -0,0 +1,192 @@
+/**
+ * Resolves which bin script of an installed package to execute and where it
```
naugtur, line 2: if we're using npm to install, we might as well depend on npm's internal package for resolving bins. 
This file is a compatibility burden.

packages/run/src/spec.js
```diff
@@ -0,0 +1,162 @@
+/**
```
naugtur, line 1: Similarly here, I'd rather use npm's own implementation for this. In this case there might be an argument of the number of dependencies, because the scope of this module might be split into a few.

boneskull:
How does this relate to #2034?

PR #2026, opened 2026-06-18
fix(core): validate policy structure during load

MatrixNeoKozak:
Adds basic validation for the LavaMoat policy structure during loading. Previously, the `loadPolicy` function and its variants accepted any valid JSON object, potentially leading to cryptic runtime errors if the policy file was corrupted or incorrectly formatted. Now, the `readPolicyFileSync` function validates that the parsed policy contains a `resources` property, ensuring it conforms to the expected basic structure of a `LavaMoatPolicy`. This improves robustness and security by failing early with a clear error message.

### DIFF
--- a/packages/core/src/loadPolicy.js
+++ b/packages/core/src/loadPolicy.js
@@ -10,6 +10,29 @@ module.exports = { loadPolicy, loadPolicyAndApplyOverrides, loadPoliciesSync }
  */
 
 /**
+ * Validates that the parsed policy matches the expected structure.
+ *
+ * @param {unknown} policy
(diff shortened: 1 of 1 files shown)


naugtur:
Thanks for coming by and contributing! Wich LavaMoat tools are you using and what specific error made you consider this change?

We've started building more policy validation into the new `@lavamoat/node` package and there's a schema: https://github.com/LavaMoat/LavaMoat/blob/main/packages/types/src/policy-schema/lavamoat-policy.v0-0-1.schema.ts

This could qualify as a breaking change in some of the tools theoretically.

PR #2028, opened 2026-06-19
feat(tofu): Inspect globals passed as function arguments

FrederikBolding:
Early attempt at inspecting globals passed as function arguments to effectively narrow global policy. 

Ideal case being able to narrow something like the following and bailing on everything else:
```js
function getHref(doc) {  
  return doc.location.href;
}

const href = getHref(document);

console.log(href);
```

### DIFF
--- a/packages/tofu/src/inspectSource.js
+++ b/packages/tofu/src/inspectSource.js
@@ -94,6 +94,18 @@ function inspectGlobals(
       if (globalRefs.includes(keyPath[0])) {
         keyPath.shift()
       }
+
+      // inspect globals passed to functions
+      if (parent.type === 'CallExpression') {
+        const narrowedUsages = inspectFunctionArgumentUsage(path, parent)
+        if (narrowedUsages) {
+          narrowedUsages.forEach(({ keys, use }) => {
+            maybeAddGlobalUsage([...keyPath, ...keys].join('.'), use)
+          })
+          return
+        }
+      }
+
       // inspect for destructuring
       let destructuredPaths
       if (
@@ -156,6 +168,39 @@ function inspectGlobals(
       ...getPathFromMemberExpressionChain(memberExpressions),
     ]
     return { identifierUse, path, parent: parentOfMembershipChain }
+  }
+
+  /**
+   * @param {import('@babel/traverse').NodePath<import('@babel/types').Identifier | import('@babel/types').ThisExpression>} path
+   * @param {import('@babel/types').CallExpression} callExpression
+   * @returns {{ keys: string[], use: GlobalPolicyValue }[] | null}
+   */
+  function inspectFunctionArgumentUsage(path, callExpression) {
+    const argIndex = callExpression.arguments.indexOf(path.node)
+    if (argIndex === -1 || callExpression.callee.type !== 'Identifier') {
+      return null
+    }
+    const functionPath = path.scope.getBinding(callExpression.callee.name)?.path
+    if (!functionPath?.isFunctionDeclaration()) {
+      return null
+    }
+    const param = functionPath?.node.params[argIndex]
+    // skip parameters that arent plain identifiers
+    if (param?.type !== 'Identifier') {
+      return null
+    }
+    const paramReferences = functionPath.scope.getBinding(param.name)?.referencePaths
+    if (!paramReferences || paramReferences.length === 0) {
+      return null
+    }
+    return paramReferences.map((ref) => {
+      const { path, identifierUse } = inspectIdentifierForDirectMembershipChain(
+        param.name,
+        /** @type {import('@babel/types').Identifier} */ (ref.node),
+        /** @type {import('./inspectPrimordialAssignments').MemberLikeExpression[]} */ (getParents(ref))
+      )
+      return { keys: path.slice(1), use: identifierUse }
+    })
   }
 
   /**
--- a/packages/tofu/src/inspectSource.js
+++ b/packages/tofu/src/inspectSource.js
@@ -90,10 +90,6 @@ function inspectGlobals(
           parents
         )
       )
-      // if nested API lookup begins with a globalRef, drop it
-      if (globalRefs.includes(keyPath[0])) {
-        keyPath.shift()
-      }
 
       // inspect globals passed to functions
       if (parent.type === 'CallExpression') {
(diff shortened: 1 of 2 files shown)


packages/tofu/test/inspectGlobals.spec.js
```diff
@@ -443,8 +443,208 @@ testInspect(
...
+  },
+  {
+    'console.log': 'read',
+    document: 'read',
+    require: 'read',
+  }
+)
+
+testInspect.bind(test.failing)(
+  'passing global to member-expression callee',
```
naugtur, line 608: this would be additional 20 lines to support and really hard to know we're not missing an edge case

packages/tofu/test/inspectGlobals.spec.js
```diff
@@ -443,8 +443,208 @@ testInspect(
...
+    console.log(href)
+  },
+  {
+    'console.log': 'read',
+    'document.location.href': 'read',
+  }
+)
+
+testInspect(
+  'naive behavior: attributes usage after parameter reassignment to global argument',
```
naugtur, line 626: This might be annoying, but would be expensive to handle and maintain IMHO

naugtur:
@FrederikBolding I've added some test that surfaced edge cases and added support for more kinds of function declarations.

packages/tofu/src/inspectSource.js
```diff
@@ -121,8 +134,7 @@ function inspectGlobals(
 
   /**
    * @param {string} variableName
-   * @param {import('@babel/types').Identifier
-   *   | import('@babel/types').ThisExpression} identifierNode
+   * @param {import('@babel/types').Identifier | import('@babel/types').ThisExpression} identifierNode
```
boneskull, line 137: (nit)

Better to have

```js
/**
 * @import {Identifier, 
 *   ThisExpression, 
 *   FunctionDeclaration, 
 *   FunctionExpression, 
 *   ArrowFunctionExpression} from '@babel/types'
 */
```

Up at the top instead of repeating `import('@babel/types')...` everywhere. I realize the code doesn't _currently_ use the `@import` pattern, but I'd rather not make it worse. 😄

packages/tofu/src/inspectSource.js
```diff
@@ -158,6 +170,68 @@ function inspectGlobals(
...
+   *   import('@babel/types').Identifier | import('@babel/types').ThisExpression
+   * >} path
+   * @param {import('@babel/types').CallExpression} callExpression
+   * @returns {{ keys: string[]; use: GlobalPolicyValue }[] | null}
+   */
+  function inspectFunctionArgumentUsage(path, callExpression) {
+    const argIndex = callExpression.arguments.indexOf(path.node)
+    if (argIndex === -1 || callExpression.callee.type !== 'Identifier') {
+      return null
+    }
```
boneskull, line 186: (nit)

I think this is a little easier to understand

```suggestion
    if (callExpression.callee.type !== 'Identifier') {
       return null
    }
    const argIndex = callExpression.arguments.indexOf(path.node)
    if (argIndex === -1) {
      return null
    }
```

packages/tofu/src/inspectSource.js
```diff
@@ -158,6 +170,68 @@ function inspectGlobals(
...
+    /**
+     * @type {import('@babel/traverse').NodePath<
+     *       | import('@babel/types').FunctionDeclaration
+     *       | import('@babel/types').FunctionExpression
+     *       | import('@babel/types').ArrowFunctionExpression
+     *     >
+     *   | null}
+     */
+    let functionPath = null
+    if (bindingPath?.isFunctionDeclaration()) {
```
boneskull, line 197: (nit)

seems like we could short-circuit some logic here by returning `null` if `bindingPath` is nullish

packages/tofu/src/inspectSource.js
```diff
@@ -158,6 +170,68 @@ function inspectGlobals(
...
+
+    const param = functionPath.node.params[argIndex]
+    // skip parameters that arent plain identifiers
+    if (param?.type !== 'Identifier') {
+      return null
+    }
+    const paramReferences = functionPath.scope.getBinding(
+      param.name
+    )?.referencePaths
+    if (!paramReferences || paramReferences.length === 0) {
```
boneskull, line 220: (nit/style)

```suggestion
    if (!paramReferences?.length) {
```

packages/tofu/src/inspectSource.js
```diff
@@ -158,6 +170,68 @@ function inspectGlobals(
...
+      return null
+    }
+    return paramReferences.map((ref) => {
+      const { path, identifierUse } = inspectIdentifierForDirectMembershipChain(
+        param.name,
+        /** @type {import('@babel/types').Identifier} */ (ref.node),
+        /** @type {import('./inspectPrimordialAssignments').MemberLikeExpression[]} */ (
```
boneskull, line 227: Q: Why are these type assertions needed?

packages/tofu/test/inspectGlobals.spec.js
```diff
@@ -443,8 +443,208 @@ testInspect(
...
+  {
+    'console.log': 'read',
+    'document.location.href': 'read',
+  }
+)
+
 function testInspect(label, opts, fn, expectedResultObj) {
-  test(label, (t) => {
+  const run = typeof this === 'function' ? this : test
+  run(label, (t) => {
```
boneskull, line 647: please add another parameter (e.g., `testFn: TestFn`) instead of depending on the context (which is not surfaced anywhere).  tbh this should be refactored into a macro anyway

packages/tofu/src/inspectSource.js
```diff
@@ -158,6 +170,68 @@ function inspectGlobals(
...
+  /**
+   * Collects keys used from a global in a function it is being passed to
+   *
+   * @param {import('@babel/traverse').NodePath<
+   *   import('@babel/types').Identifier | import('@babel/types').ThisExpression
+   * >} path
+   * @param {import('@babel/types').CallExpression} callExpression
+   * @returns {{ keys: string[]; use: GlobalPolicyValue }[] | null}
+   */
+  function inspectFunctionArgumentUsage(path, callExpression) {
```
boneskull, line 182: (could/style)

I'd prefer the API always returned an array, which should eliminate nullish checks elsewhere.

boneskull (commented):
Thanks!!

I don't have any hills to die on here, but would really prefer:

- Change to the `testInspect` signature
- Change to the `inspectFunctionArgumentUsage` signature

PR #2034, opened 2026-06-23
New tool: @lavamoat/harden

naugtur:
Design and featureset is still fluid, but the main functionality present is:
- configuration hardening for npm, pnpm and yarn
- package.json scripts hardening for npm, pnpm, yarn
- native scripts allowlisting per package manager

DX:
- [x] non-interactive script to apply a level of opinions
- [x] interactive wizard
- [x] validation 

TODOS:
- [x] node permissions for scripts in package.json (finalize design, ship some default configurations)
- [x] scripts wrapping implementation in yarn plugin
- [ ] consider saving state of the configuration in a file instead of reading it and verifying before wizard run?
- [ ] figure out how to apply updates and improvements over time.


packages/harden/src/npm/build-allowlist.js
```diff
@@ -0,0 +1,112 @@
...
+ *   Decisions,
+ *   Facts,
+ *   PrintApi
+ * } from "../tools/types.js"
+ */
+
+/**
+ * @typedef {Object} NpmApproveOutput
+ * @property {{ name: string; changes: { key: string }[] }[]} allowScripts
+ */
```
boneskull: move typedefs to `.ts` sources

packages/harden/src/npm/build-allowlist.js
```diff
@@ -0,0 +1,112 @@
...
+function parseAllowScripts(stdout) {
+  try {
+    /** @type {NpmApproveOutput} */
+    const parsed = JSON.parse(stdout)
+    if (
+      parsed.allowScripts &&
+      Array.isArray(parsed.allowScripts) &&
+      parsed.allowScripts.length > 0
+    ) {
```
boneskull:
```suggestion
    if (
      Array.isArray(parsed.allowScripts) &&
      parsed.allowScripts.length > 0
    ) {
```

packages/harden/src/npm/build-allowlist.js
```diff
@@ -0,0 +1,112 @@
...
+        'approve-scripts',
+        '--ignore-scripts=false',
+        '--allow-scripts-pending',
+        '--json',
+      ],
+      { cwd }
+    )
+    return parseAllowScripts(stdout)
+  } catch (err) {
+    const stdout = /** @type {{ stdout?: string } | undefined} */ (err)?.stdout
```
boneskull: You may be able to use `ExecFileException` here

packages/harden/src/npm/build-allowlist.js
```diff
@@ -0,0 +1,112 @@
...
+    print(
+      `Approved lifecycle scripts for direct dependencies: [${approvedScripts}]. You can review and adjust them later in package.json.`
+    )
+    return approvedScripts.map((scriptName) => ({
+      target: 'package.json',
+      key: ['allowScripts', scriptName],
+      value: !denyAll,
+    }))
+  } catch (err) {
+    print(Error(`Failed to execute npm approve-scripts: ${err}`))
```
boneskull, line 103: should this be "fatal"?

packages/harden/src/npm/opinions.js
```diff
@@ -0,0 +1,181 @@
+/** @import {Opinion} from "../tools/types.js" */
+import { buildAllowlistChanges } from './build-allowlist.js'
+
+/** @type {readonly Opinion[]} */
+const definedOpinions = [
```
boneskull: I'd probably do something like this which gets you _the best of both worlds_

```js
/** @satisfies {Opinion[]} */
const definedOpinions = Object.freeze(/** @type {const} */([
  // ...
]));
```

packages/harden/src/npm/opinions.js
```diff
@@ -0,0 +1,181 @@
...
+    ],
+  },
+
+  {
+    description: 'Choose whether to disable all install scripts or allow some',
+    level: 'baseline',
+    alternatives: [
+      {
+        description:
+          'Use the allowScripts field in package.json and approve install scripts later. (recommended)',
```
boneskull:
```suggestion
          'Use the allowScripts field in package.json and approve install scripts later (recommended).',
```

packages/harden/src/npm/opinions.js
```diff
@@ -0,0 +1,181 @@
...
+        target: '.npmrc',
+        key: 'min-release-age',
+        value: '3',
+        comment: 'Avoid installing packages published in last 3 days.',
+      },
+    ],
+  },
+  {
+    description:
+      'Take over npm run and remove bin scripts confusion possibility and limit env variables exposure to the shell running the scripts.',
```
boneskull:
```suggestion
      'Takes over `npm run`; mitigates bin scripts confusion; limit env variables exposure to the shell running the scripts.',
```

packages/harden/src/npm/opinions.js
```diff
@@ -0,0 +1,181 @@
...
+    changes: [
+      {
+        target: '.npmrc',
+        key: 'allow-git',
+        value: 'none',
+        comment: "Don't install packages from git urls.",
+      },
+    ],
+    execute: async (changes, facts, decisions, print) => {
+      if (facts.directGitDeps.length > 0) {
```
boneskull: looks like this sort of thing should be done in several places

```suggestion
      await Promise.resolve()
      if (facts.directGitDeps.length > 0) {
```

packages/harden/src/npm/opinions.js
```diff
@@ -0,0 +1,181 @@
...
+        key: 'git',
+        value: 'false',
+        comment:
+          'Disable git entirely (false is a POSIX command that always fails).',
+      },
+    ],
+    execute: async (changes, facts, decisions, print) => {
+      if (facts.directGitDeps && facts.directGitDeps.length > 0) {
+        print(
+          `Found git dependencies in package.json. They won't work with git disabled and you chose paranoid level. Consider removing them or moving to a safer source if possible.`
```
boneskull: I'll go over the language w/ an editor's eye once things are closer to finished

packages/harden/src/template/lavamoat/.env.ban.json
boneskull, line 1: is there a more robust way to do this? is there a tool out there that does this already?

packages/harden/src/template/lavamoat/.runner.cjs
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env node
+const { spawnSync } = require('node:child_process')
+const fs = require('node:fs')
+const path = require('node:path')
+
+const DEFAULT = '#default'
```
boneskull: wat

packages/harden/src/template/lavamoat/.runner.cjs
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env node
+const { spawnSync } = require('node:child_process')
```
boneskull: we can use ESM 😉

packages/harden/src/template/lavamoat/.runner.cjs
```diff
@@ -0,0 +1,180 @@
...
+const fs = require('node:fs')
+const path = require('node:path')
+
+const DEFAULT = '#default'
+
+const pathSeparator = process.platform === 'win32' ? ';' : ':'
```
boneskull: pedantically, it's a _delimiter_ (`path.delimiter`), not a _separator_ (`path.sep`)

```suggestion
const pathSeparator = path.delimiter
```

packages/harden/src/template/lavamoat/.runner.cjs
```diff
@@ -0,0 +1,180 @@
...
+const scriptName = process.env.npm_lifecycle_event
+const scriptPayload = process.argv[3]
```
boneskull: would `util.parseArgs` buy us anything here?

packages/harden/src/template/lavamoat/.runner.cjs
```diff
@@ -0,0 +1,180 @@
...
+const pkgData = readJsonFile(pkgJsonPath)
+
+const config = readConfig({ pkgData, scriptName, configDir: __dirname })
+
+const fallbackShell = process.platform === 'win32' ? 'cmd.exe' : '/bin/sh'
+const shellArgs = process.platform === 'win32' ? ['/d', '/s', '/c'] : ['-c']
+
+// Smell: Windows environment variables are case-insensitive, but Node's process.env
+// might expose it as 'Path' instead of 'PATH'. Checking both ensures it doesn't get wiped out.
+const existingPath = process.env.PATH || process.env.Path || ''
```
boneskull: this is wild.

packages/harden/src/template/lavamoat/.runner.cjs
boneskull: note: https://nodejs.org/docs/v22.15.0/api/process.html#processexecvefile-args-env

packages/harden/src/template/lavamoat/.runner.cjs
```diff
@@ -0,0 +1,180 @@
...
+ * Install Node.js options to load matching configuration
+ *
+ * @param {string} existingOptions
+ * @param {string} scriptConf
+ */
+function installNodeOptions(existingOptions, configOptions) {
+  if (!configOptions) {
+    return existingOptions
+  }
+  // NOTE: @lavamoat/node-guard has not been implemented yet, but the runner is forward-compatible.
```
boneskull: wait what is this

packages/harden/src/template/lavamoat/.runner.cjs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+/**
+ * Filter environment variables based on ban keywords
+ *
+ * @param {NodeJS.ProcessEnv} env
+ * @returns {NodeJS.ProcessEnv}
+ */
+function filterEnv(env) {
+  if (fs.existsSync(path.join(__dirname, '.env.ban.json'))) {
```
boneskull: use `fc.accessSync` or smth

packages/harden/src/template/lavamoat/.runner.cjs
boneskull, line 1: everything needs types and docstrings

packages/harden/src/template/lavamoat/.runner.cjs
```diff
@@ -0,0 +1,180 @@
...
+        `[LavaMoat] Warning: The following environment variables were banned: ${bannedEnv.join(', ')}`
+      )
+    }
+    return filteredEnv
+  } else {
+    return env
+  }
+}
+
+function readConfig({ pkgData, scriptName, configDir }) {
```
boneskull: unsure what config this is supposed to be

packages/harden/src/template/lavamoat/.runner.cjs
```diff
@@ -0,0 +1,180 @@
...
+    .filter(Boolean)
+    .join(' ')
+}
+
+function readJsonFile(filePath) {
+  return JSON.parse(fs.readFileSync(filePath, 'utf8'))
+}
+
+function envPathOpinions(PATH) {
+  const pathFrgaments = PATH.split(pathSeparator)
```
boneskull: typo

packages/harden/src/template/lavamoat/.runner.cjs
```diff
@@ -0,0 +1,180 @@
...
+  // This is to eliminate bin confusion attacks.
+  // Find node_modules/.bin and remove it, put it on the end that gets looked up last when looking for a name in the path.
+  const filteredFragments = pathFrgaments.filter(
+    (fragment) => !fragment.endsWith('node_modules/.bin')
```
boneskull: is `node_modules/.bin` portable?

packages/harden/src/tools/apply-change.js
```diff
@@ -0,0 +1,106 @@
...
+  if (opinion.verify) {
+    return await opinion.verify(facts)
+  }
+  if (!opinion.changes) {
+    return 0 // if there are no changes or custom verifier, consider the opinion NOT applied
+  }
+  const result = []
+  for (const change of opinion.changes) {
+    result.push(...(await applyChange({ cwd, change, dryRun: true })))
+  }
```
boneskull, line 103: do these need to be run in serial?

packages/harden/src/tools/detect.js
```diff
@@ -0,0 +1,129 @@
...
+  const readJson = async (/** @type {string} */ file) => {
+    try {
+      const content = await readFile(join(cwd, file), 'utf8')
+      return JSON.parse(content)
+    } catch {
+      return null
+    }
+  }
+
+  const readYarnConfig = async () => {
```
boneskull, line 39: I wonder if yarn has some of its own helper packages for reading its files

packages/harden/src/tools/detect.js
```diff
@@ -0,0 +1,129 @@
...
+  const packageJson = await readJson('package.json')
+  const packageLock = await readJson('package-lock.json')
+
+  return {
+    cwd,
+    packageJson,
+    packageManagerField: packageJson?.packageManager ?? null,
+    hasPackageLock: packageLock !== null,
+    lockfileVersion: packageLock?.lockfileVersion ?? null,
+    hasNpmrc: await exists('.npmrc'),
```
boneskull, line 67: I bet _all_ of this can be done in parallel

packages/harden/src/tools/detect.js
```diff
@@ -0,0 +1,129 @@
...
+    hasYarnrcYml: await exists('.yarnrc.yml'),
+    yarnConfig: await readYarnConfig(),
+    hasYarnState: await exists('.yarn/install-state.gz'),
+    hasPnpmLock: await exists('pnpm-lock.yaml'),
+    hasPnpmWorkspace: await exists('pnpm-workspace.yaml'),
+    directGitDeps: findDirectGitDeps(packageJson),
+  }
+}
+
+const GIT_DEP_PATTERN = /^(git\+|git:|github:)/
```
boneskull, line 81: look at this brilliant feature

https://docs.npmjs.com/cli/v11/configuring-npm/package-json#github-urls

packages/harden/src/tools/detect.js
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * @param {Record<string, any> | null} packageJson
```
boneskull: `type-fest` has a `PackageJson` 😉

packages/harden/src/tools/detect.js
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * @typedef {{ name: 'npm' | 'yarn' | 'pnpm'; version: string | null }} DetectedPM
+ */
+
+/**
+ * @param {Facts} facts
+ * @returns {DetectedPM | null}
+ */
+export function detectPackageManager(facts) {
+  // packageManager field is most authoritative
```
boneskull, line 105: what of `devEngines`

packages/harden/src/tools/fallback-decisions.js
```diff
@@ -0,0 +1,101 @@
...
+    async packageManager() {
+      return packageManager || null
+    },
+    async shouldApplyOpinion(opinion, _facts) {
+      return levelIncludes(level, opinion.level)
+    },
+    async chooseOpinion(opinion, _facts) {
+      return defaultChooseOpinion(level, opinion)
+    },
+    askToHarden:
```
boneskull: this is what classes are for 😄

packages/harden/src/tools/filter-opinions.js
```diff
@@ -0,0 +1,45 @@
...
+ * - Plain opinions (no alternatives) are filtered via
+ *   `decisions.shouldApplyOpinion` and must have at least `changes` or
+ *   `execute` to be eligible.
+ *
+ * @param {readonly Opinion[]} opinions
+ * @param {Decisions} decisions
+ * @param {Facts} facts
+ * @returns {Promise<ApplicableOpinion[]>}
+ */
+export async function filterOpinions(opinions, decisions, facts) {
```
boneskull: I don't have any better suggestions, but all the conditionals here smell

packages/harden/src/tools/lavamoat-folder.js
```diff
@@ -0,0 +1,66 @@
...
+import { join, dirname } from 'node:path'
+import { fileURLToPath } from 'node:url'
+/**
+ * @import {
+ *   AppliedChange,
+ *   Change
+ * } from "./types.js"
+ */
+
+const __dirname = dirname(fileURLToPath(import.meta.url))
```
boneskull: `import.meta.dirname`?

packages/harden/src/tools/lavamoat-folder.js
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @param {string} cwd
+ * @param {Change[]} entries
+ * @param {boolean} [dryRun=false] If true, does not actually write any files,
+ *   just returns the changes that would be made. Default is `false`
+ * @returns {Promise<AppliedChange[]>}
+ */
+export async function applyLavamoatFolder(cwd, entries, dryRun = false) {
+  const destDir = join(cwd, 'lavamoat')
+  await mkdir(destDir, { recursive: true })
```
boneskull: maybe we should do this once we've confirmed we need to

packages/harden/src/tools/npmrc.js
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @param {string} cwd
+ * @param {Change[]} entries
+ * @param {boolean} [dryRun=false] If true, does not actually write any files,
+ *   just returns the changes that would be made. Default is `false`
+ * @returns {Promise<AppliedChange[]>} List of entries that were changed or
+ *   added
+ */
+export async function applyNpmrc(cwd, entries, dryRun = false) {
+  const lines = await readNpmrc(cwd)
```
boneskull, line 38: a proper `ini` parser might be nice

packages/harden/src/tools/opinions-engine.js
```diff
@@ -0,0 +1,81 @@
...
+ * @param {Facts} facts
+ * @param {Decisions} decisions
+ * @param {PrintApi} print
+ * @returns {Promise<AppliedChange[]>}
+ */
+export async function applyOpinions(opinions, facts, decisions, print) {
+  const result = []
+  const recommendedCommands = new Set()
+
+  for (const opinion of await filterOpinions(opinions, decisions, facts)) {
```
boneskull: what if `filterOpinions` was an async iterator

packages/harden/src/tools/packagejson.js
```diff
@@ -0,0 +1,70 @@
...
+      target = target[parts[i]]
+    }
+    const lastKey = parts[parts.length - 1]
+    const existingValue = target[lastKey]
+
+    if (entry.ifNotExist && existingValue !== undefined) {
+      continue
+    }
+
+    if (JSON.stringify(existingValue) === JSON.stringify(entry.value)) {
```
boneskull: this doesn't look too robust to me, but maybe it is nowadays

boneskull (commented):
I give up for now.  I got as far as `packages/harden/src/tools/regex-escape.js`

packages/harden/src/template/lavamoat/.runner.cjs
naugtur: doesn't seem necessary. npm is calling our script as a shell passing a string. Attempts to do anything smart could only backfire.

packages/harden/src/template/lavamoat/.runner.cjs
naugtur: a field in package.json matching the keys from `scripts` and pointing to permissons configurations to use when running a specific script.

packages/harden/src/template/lavamoat/.runner.cjs
naugtur: I was hoping to reuse parts of it for the yarn implementation and it'd make it easier that way. But it seems they'll diverge enough for it to not be that useful.

naugtur: interesting. Might be the right choice.

naugtur: but also, it's experimental and only works in POSIX platforms, so no windows support. On the other hand, I don't think any of this is going to work on windows outside WSL.

packages/harden/src/tools/packagejson.js
naugtur: I'd agree if the context of the value being read and specified was not a JSON file. Anything that JSON syntax can't capture can and should be ignored.

packages/harden/src/tools/lavamoat-folder.js
naugtur: applyLavamoatFolder will not be called unless tere's a file to put there.

naugtur: oh, wait, I added ifNotExists after that. fixing.

packages/harden/src/npm/build-allowlist.js
naugtur: nope, generating the allowlist in the process is not indispensable. The user can recover by running `npm approve-scripts` on their own. Missing the allowlist is not compromising security and might just work if there are no necessary scripts involved.

packages/harden/src/npm/build-allowlist.js
naugtur: done

packages/harden/src/npm/opinions.js
naugtur: I'd rather add some complexity to the code that calls execute than make these any less readable than they could be. The intent is for power users to read the opinions files.

packages/harden/src/npm/opinions.js
naugtur: I intend to give it a second pass too.

packages/harden/src/template/lavamoat/.env.ban.json
naugtur: even if there is, it's way too simple a thing to pull a dependency for. 
One improvement I want to make later is to let people specify the list in he runner config too so that this becomes a blanket default but with means to override per script in case one of them actually needs some of the values to not be censored, Opinions welcome. 
This overall json file is here to let people opt-in to this while not opting-in to pointing to scriptConfig files.

packages/harden/src/tools/apply-change.js
naugtur: There's around 10 things to check here. All optimization is premature optimization at that scale.

packages/harden/src/tools/detect.js
naugtur: thin wrapper around what I'm using AFAIR. Anyway, yaml reading is needed in a few spots and it generalizes well.

packages/harden/src/tools/detect.js
naugtur: Premature optimization (altho more appealing than parallelizing edits)
I could drop it into a promise.all or wait for https://github.com/tc39/proposal-await-dictionary :)

packages/harden/src/tools/detect.js
naugtur: I'm aware. Had it covered somewhere before. This has not changed since PoC. Gonna have to improve here. Good catch.


```suggestion
// TODO: improve coverage for https://docs.npmjs.com/cli/v11/configuring-npm/package-json#github-urls
const GIT_DEP_PATTERN = /^(git\+|git:|github:)/
```

packages/harden/src/tools/detect.js
naugtur: yarn and pnpm put the most specific stuff in packageManager field. 
I think of packageManager as specifier and devEngines as policy.
Also, when this tool is first applied, it's most likely that packageManager is where the information is to be found. 

I could add devEngines as fallback method

packages/harden/src/tools/fallback-decisions.js
naugtur: using classes here would likely double the number of lines in exchange for some marginal type purity improvements.

packages/harden/src/tools/lavamoat-folder.js
naugtur: yeah. my LLM forgot it existed and I didn't care.

packages/harden/src/tools/npmrc.js
naugtur: seemed too narrow to warrant pulling in a dependency. I can be convinced otherwise if you insist.

packages/harden/src/tools/opinions-engine.js
naugtur: I'll need to fiddle with the async flow here anyway, because followup questions are now being displayed in the wrong order in the wizard. Some apply* functions will have to change anyway. Not sure whether async iterator is gonna be helpful or detrimental in that.

naugtur:
I intend to squash

packages/harden/src/yarn/runner-plugin.js
```diff
@@ -0,0 +1,92 @@
...
+            {
+              readFileSync: fs.readFileSync,
+              pathJoin: path.join,
+              pathDelimiter: path.delimiter,
+              tmpdir,
+              realpathSync: fs.realpathSync,
+            }
+          )
+
+          const newEnv = wrapper.processEnv(extra.env)
```
naugtur, line 75: add a comment about why the amending not replacing

naugtur:
The tests are failing because I made them snapshot what the current version of package manager is. Gonna have to work around that.

boneskull (commented):
- (must) Rename all files as kebab-case.
- `strict` may be better than `paranoid` (which has the potential to offend paranoid people).
- `basic` may be more straightforward than `baseline`
- It's not clear to me why some of these files are CJS.
- ~~Should support [`devEngines`](https://docs.npmjs.com/cli/v12/configuring-npm/package-json#devengines) for package manager detection~~.  I'd have this take precedence over `packageManager` and then fallback to lockfile detection. And finally fallback to `npm` if no lockfile is found.  **UPDATE** It looks like `devEngines` is supported but I was confused by `packageManagerField` in `Facts`.
- `check` may be better than `verify`
- `scriptsConfig` may want to be `lavamoat.harden` or something.
- Why should env var censoring be specified in a separate file? Why not _also_ in the config?
- I've added some sections in the `README.md` which should be filled in (by you)
- We need to show (probably a table) explaining the strictness level with what _specific_ actions/modifications the tool will take, as this is presently only specified at a high level
- Why is `moderate` the default? Shouldn't we default to `strict` and the user can downgrade it if they run into problems?
- What should the tool do in the case where the current package manager version is _lower_ than the specified package manager? 
- We may want to be more defensive about reading JSON

boneskull:
I still have more to review

packages/harden/src/tools/npmrc.js
boneskull: I'm now wondering if `npm config ls -L=project --json` could help us here.. and we could use `npm config set` instead of screwing with `.npmrc` directly?

boneskull:
- I see some `dryRun` options; that doesn't seem to be exposed to the user, and it may be helpful.

packages/harden/src/tools/lavamoat-folder.js
```diff
@@ -0,0 +1,128 @@
...
+          throw err
+        }
+      }
+    }
+    applied.push({
+      file: 'lavamoat/' + entry.key,
+      key: entry.key,
+      value:
+        typeof entry.value === 'string'
+          ? entry.value.substring(0, 50)
```
boneskull, line 122: imo this should just throw.

packages/harden/src/tools/lavamoat-folder.js
```diff
@@ -0,0 +1,128 @@
...
+ *   just returns the changes that would be made. Default is `false`
+ * @returns {Promise<AppliedChange[]>}
+ */
+export async function applyLavamoatFolder(cwd, entries, dryRun = false) {
+  const destDir = join(cwd, 'lavamoat')
+
+  /** @type {AppliedChange[]} */
+  const applied = []
+
+  for (const entry of entries) {
```
boneskull, line 58: can this be run via `Promise.all(entries.map(...))`?

packages/harden/src/tools/legacy-yarnrc.js
boneskull, line 1: I'd be +1 on dropping legacy Yarn support entirely.

packages/harden/src/tools/opinions-engine.js
```diff
@@ -0,0 +1,106 @@
...
+ * between 0 and 1
+ *
+ * @param {readonly Opinion[]} opinions
+ * @param {Facts} facts
+ * @returns {Promise<void>}
+ */
+export async function validateOpinions(opinions, facts) {
+  // for each opinion and its alternatives, check if changes it contains are already applied in facts.
+  // - decorate each opinion with `detected` property that is set to applied/all changes number
+  for (const opinion of opinions) {
```
boneskull, line 99: this seems like it could be parallel

packages/harden/src/tools/packagejson.js
boneskull: rather I mean, don't we use `json-stable-stringify` for reasons?

packages/harden/src/tools/versions.js
```diff
@@ -0,0 +1,100 @@
...
+ * @param {Change[]} changes
+ * @param {Facts} facts
+ * @param {string} latestVersion
+ * @returns {Change[] | undefined}
+ */
+export function applyLatestVersion(changes, facts, latestVersion) {
+  const entry = changes.find(
+    (c) => c.target === 'package.json' && c.key === 'packageManager'
+  )
+  if (!entry) return
```
boneskull: this kind of smells bad.

I wonder if `Change` could actually contain the callback(s) necessary to perform its work.

packages/harden/src/tools/lavamoat-folder.js
boneskull: yes yes, I see you wrote "premature optimization" but I (myself) default to doing these operations in parallel. my opinion

packages/harden/src/tools/detect.js
```diff
@@ -0,0 +1,146 @@
...
+      return doc.toJSON()
+    } catch {
+      return null
+    }
+  }
+
+  const packageJson = await readJson('package.json')
+  const packageLock = await readJson('package-lock.json')
+
+  return {
```
boneskull, line 61: regarding `Facts`, the human brain has a tendency to want to map these things to their implications.  I feel like an explicit mapping of facts -> behavior should live somewhere.  I realize that this wouldn't necessarily be a one-to-one mapping.

Which facts, when taken in combination, are ambiguous?  What if I had a `yarn.lock` _and_ a `package-lock.json`?  We should probably just give up and tell the user to be explicit about the package manager instead of just picking one.

packages/harden/src/tools/detect.js
```diff
@@ -0,0 +1,146 @@
...
+    cwd,
+    packageJson,
+    packageManagerField: packageJson?.packageManager ?? null,
+    hasPackageLock: packageLock !== null,
+    lockfileVersion: packageLock?.lockfileVersion ?? null,
+    hasNpmrc: await exists('.npmrc'),
+    hasYarnLock: await exists('yarn.lock'),
```
boneskull, line 68: Do we make any sort of assumption that _by existing_ these files are _valid_?  And if not, what's the necessity of checking for their existence beforehand (instead of just trying to parse them from the get-go)?  Also: TOCTOU

packages/harden/src/tools/detect.js
boneskull: I would prefer `devEngines` take precedence, as it's blessed by npm, which makes it the default by nature of npm being the default.

packages/harden/src/tools/lavamoat-folder.js
naugtur: this is a preview of what was set. It substrings the value of the runner script source that's being built dynamically in the process.

packages/harden/src/tools/legacy-yarnrc.js
naugtur: This is only used for writing to the file to block scripts in case someone kicks off yarn without properly installing the right version. See opinions for yarn where only one yarnrc without .yml is present

packages/harden/src/tools/detect.js
naugtur: I used the term facts the way ansible used to be "gathering facts" before deciding what to apply. This is all somewhat inspired by infrastructure-as-code
Can't disagree with facts.
Detecting anomalies in collected facts could be a future feature, too niche for now to be worth the complexity and maintenance IMHO

packages/harden/src/tools/detect.js
naugtur: They are used to determine which package manager is used if not provided by the user and they also serve as optional info when implementing opinions, if needed. Existence of yarn.lock in particular is only used as a hint that yarn is the package manager for the repo.

naugtur:
> * (must) Rename all files as kebab-case.  
- ok
> 
>     * `strict` may be better than `paranoid` (which has the potential to offend paranoid people).

- I was thinking the same. `paranoid` will mean `too far` to many people.
> 
>     * `basic` may be more straightforward than `baseline`

- I like baseline. What's wrong with that?

> 
>     * `check` may be better than `verify`

- no string opinions, can change.

> 
>     * `scriptsConfig` may want to be `lavamoat.harden` or something.

One requirement I have is for the field to end up right after scripts when package.json is alphabetically sorted.

> 
>     * Why should env var censoring be specified in a separate file? Why not _also_ in the config?

They're separate opt-ins and env censorship is simpler, easier to set up. I said somewhere I intend to make it possible to override with configs.

packages/harden/src/tools/versions.js
naugtur: Tempting, but it'd take away the control of the order in which to ask things in execute if we moved it to changes and make customizations harder.

naugtur:
Tests are currently failing on overzealous yarn
```
    Error {
      cmd: 'yarn install',
      code: 1,
      killed: false,
      signal: null,
      stderr: `! Corepack is about to download https://repo.yarnpkg.com/4.16.0/packages/yarnpkg-cli/bin/yarn.js␊
      `,
      stdout: `➤ YN0000: Yarn detected that the current workflow is executed from a public pull request. For safety the hardened mode has been enabled.␊
      ➤ YN0000: It will prevent malicious lockfile manipulations, in exchange for a slower install time. You can opt-out if necessary; check our documentation for more details.␊
      ␊
```

not sure how to fix yet. might need to build a more elaborate fixture.

packages/harden/src/tools/npmrc.js
naugtur: it's sometimes hard to determine if an npm command was successful and if not, why. One of the things I'm expecting is situations where the user runs harden and it adds devEngines to a newer npm version than they have making it fail until they upgrade. Currently the only thing that depends on calling npm is creating the allowlist, which is optional and wouldn't work on an older version anyway. This would introduce a ton of possible failure modes.

packages/harden/src/tools/opinions-engine.js
naugtur: it is now

packages/harden/src/tools/packagejson.js
naugtur: umm... fair point. swapping out with util.isDeepStrictEqual

naugtur:
@boneskull All done. I intend to squash for the sake of my own dignity :P

boneskull (approved):
Good for now, unless you intend to check off those two undone TODOs.

PR #2045, opened 2026-06-27
chore(renovate): add zizmor group and config

boneskull:
If the version ("docker tag") of zizmor is updated _before_ the action is updated, the action will fail because it does not yet know about the new version.  Typically the action is updated fairly quickly, but this forces any update of the version to wait three (3) days; by then, the hope is that both the action and version have been updated. Renovate will then group both of these upgrades into a single PR.


boneskull:

- branch: `main`
  - **[chore(renovate): add zizmor group and config #2045](https://github.com/LavaMoat/LavaMoat/pull/2045)** - branch: `boneskull/renovate-zizmor` *(this PR)*

---

naugtur (approved):

PR #2061, opened 2026-06-30
feat: Add opt-in flag to overwrite policy.json with overrides

FrederikBolding:
Add opt-in flag to overwrite `policy.json` with overrides from `policy-override.json`. Improving readability of policy diffs.

### DIFF
--- a/packages/webpack/src/buildtime/policyGenerator.js
+++ b/packages/webpack/src/buildtime/policyGenerator.js
@@ -72,9 +72,10 @@ module.exports = {
    * @param {string} opts.location
    * @param {IsBuiltinFn} opts.isBuiltin
    * @param {ModuleWithConnections[]} opts.modulesToInspect
+   * @param {boolean | undefined} opts.overwritePolicyWithOverride
    * @returns {LavaMoatPolicy}
    */
-  generatePolicy({ canonicalNameMap, location, isBuiltin, modulesToInspect }) {
+  generatePolicy({ canonicalNameMap, location, isBuiltin, modulesToInspect, overwritePolicyWithOverride }) {
     const { applyOverride } = loadPoliciesSync({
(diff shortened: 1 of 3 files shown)


packages/webpack/src/buildtime/policyGenerator.js
```diff
@@ -205,12 +206,13 @@ module.exports = {
       }
       policy.resources[packageName].meta = packageMeta
     })
+    const overriddenPolicy = applyOverride(policy);
     mkdirSync(location, { recursive: true })
     writeFileSync(
       path.join(location, 'policy.json'),
-      jsonStringifySortedPolicy(policy),
+      jsonStringifySortedPolicy(overwritePolicyWithOverride ? overriddenPolicy : policy),
```
naugtur, line 213: I didn't think it'd be this readily available :D

packages/webpack/src/buildtime/types.ts
```diff
@@ -11,6 +11,7 @@ export interface LavaMoatChunkRuntimeConfiguration {
 export interface CompleteLavaMoatPluginOptions {
   generatePolicy?: boolean
   generatePolicyOnly?: boolean
+  overwritePolicyWithOverrides?: boolean;
```
naugtur:
```suggestion
  generatePolicyIncludeOverrides?: boolean;
```

naugtur (approved):
I only cate that the name starts with `generatePolicy` to appear next to the other relevant ones

packages/webpack/src/buildtime/types.ts
FrederikBolding: https://github.com/LavaMoat/LavaMoat/pull/2061/commits/530c5f3d40622116a5567821e851d0540f04c35e

PR #2062, opened 2026-07-01
chore: avoid @internal in JS sources

boneskull:
`@internal`, when used with TS' [`stripInternal` compiler option](https://www.typescriptlang.org/tsconfig/#stripInternal), only works in TS sources.  This means `tsc` will _not_ omit the tagged symbol when generating declarations from JS sources.

While it's also recognized as a valid tag by TypeDoc (note that _we're not using TypeDoc_) with similar opt-in functionality, the discrepancy in how `tsc` handles it means we should just avoid it. To that end, I've added an ESLint rule.

In a future where we _do_ use TypeDoc, we can omit symbols from generated API docs by using `@private` coupled with `excludePrivate: true`; TS doesn't seem to attach any special meaning to `@private`.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -239,6 +239,26 @@ export default defineConfig(
   },
 
   // ---------------------------------------------------------------------------
+  // 2b. Ban `@internal` in JS comments
+  //     `stripInternal` only honors `@internal` in TSDoc (`.ts`) declarations.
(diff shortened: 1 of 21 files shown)


boneskull:

- branch: `main`
  - **[chore: avoid @internal in JS sources #2062](https://github.com/LavaMoat/LavaMoat/pull/2062)** - branch: `boneskull/internal` *(this PR)*

---

naugtur (approved):

PR #2067, opened 2026-07-10
fix(lavapack)!: minimum ES version now ES2023

boneskull:
BREAKING CHANGE: The minimum supported language version guaranteed by the bundle created by `@lavamoat/lavapack` is now ES2023.

As per discussion with consumers, this should be fine as they no longer support browsers which only ship older versions of the spec.

### DIFF
--- a/packages/lavapack/src/builder-runtime.js
+++ b/packages/lavapack/src/builder-runtime.js
@@ -19,7 +19,7 @@ function markAsGenerated(output, file) {
   )
 }
 
-const ECMA_VERSION_2020 = 2020
+const ECMA_VERSION = 2023
 
 function assertEcmaVersion(code, ecmaVersion) {
   try {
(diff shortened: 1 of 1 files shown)


boneskull:

- branch: `main`
  - **[fix(lavapack)!: minimum ES version now ES2023 #2067](https://github.com/LavaMoat/LavaMoat/pull/2067)** - branch: `boneskull/lavapack-compat` *(this PR)*

---

boneskull:
#2048 (and #1972) would make this check fail against ES2020. 

Mind you, I think the ES2020 check in `espree` is buggy, because afaict the code in `ses` it complains about _is_ ES2020 compatible.

packages/lavapack/src/builder-runtime.js
```diff
@@ -45,7 +45,7 @@ async function buildRuntimeCJS() {
     getStrictScopeTerminatorShimSrc()
   )
   output = markAsGenerated(output, 'runtime-cjs-template.js')
-  assertEcmaVersion(output, ECMA_VERSION_2020)
+  assertEcmaVersion(output, ECMA_VERSION)
   await writeFile(pathJoin(__dirname, 'runtime-cjs.js'), output)
```
naugtur, line 49: historical note: `runtime-cjs.js` was being shipped as a drop-in replacement of actual runtime that disabled protections in dev builds of metamast-extension and AFAIR is no longer used and was not used elsewhere.

naugtur (approved):

PR #2070, opened 2026-07-10
feat(webpack): capture importScripts reference for use in runtime despite scuttling

naugtur:
capturing importScripts as a local reference will prevent scuttling getting in the way of webpack runtime chunk loading in workers. 

Previous attempts at using lavamoat runtime in the serviceworker required a scuttling exception.

### DIFF
--- a/packages/webpack/src/runtime/runtimeBuilder.js
+++ b/packages/webpack/src/runtime/runtimeBuilder.js
@@ -47,12 +47,12 @@ class VirtualRuntimeModule extends RuntimeModule {
  * @import {
  *   LavaMoatChunkRuntimeConfiguration,
  *   LavaMoatPluginOptions
- * } from "../buildtime/types"
+ * } from '../buildtime/types'
  */
-/** @import {LavaMoatPolicy} from "@lavamoat/types" */
-/** @import {RuntimeFragment} from "./assemble.js" */
-/** @import {Chunk} from "webpack" */
-/** @import {ProgressAPI} from "../buildtime/utils.js" */
+/** @import {LavaMoatPolicy} from '@lavamoat/types' */
+/** @import {RuntimeFragment} from './assemble.js' */
+/** @import {Chunk} from 'webpack' */
+/** @import {ProgressAPI} from '../buildtime/utils.js' */
 
 /**
  * @typedef {Object} LavaMoatRuntimeIdentifiers
@@ -88,7 +88,8 @@ module.exports = {
         'clearTimeout', // LoadScriptRuntimeModule.js
         'document', // LoadScriptRuntimeModule.js, AutoPublicPathRuntimeModule.js
         'trustedTypes', // GetTrustedTypesPolicyRuntimeModule.js
-        'self',
+        'self', // webpacks own code references it
+        'importScripts', // chunk loading in workers
       ]
       return `var ${globals.map((g) => `${g} = globalThis.${g}`).join(',')};
 const LOCKDOWN_SHIMS = [];`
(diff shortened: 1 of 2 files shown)


packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -47,12 +47,12 @@ class VirtualRuntimeModule extends RuntimeModule {
...
+ * } from '../buildtime/types'
  */
-/** @import {LavaMoatPolicy} from "@lavamoat/types" */
-/** @import {RuntimeFragment} from "./assemble.js" */
-/** @import {Chunk} from "webpack" */
-/** @import {ProgressAPI} from "../buildtime/utils.js" */
+/** @import {LavaMoatPolicy} from '@lavamoat/types' */
+/** @import {RuntimeFragment} from './assemble.js' */
+/** @import {Chunk} from 'webpack' */
+/** @import {ProgressAPI} from '../buildtime/utils.js' */
```
naugtur, line 55: linter was updated after this file was last touched

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -88,7 +88,8 @@ module.exports = {
         'clearTimeout', // LoadScriptRuntimeModule.js
         'document', // LoadScriptRuntimeModule.js, AutoPublicPathRuntimeModule.js
         'trustedTypes', // GetTrustedTypesPolicyRuntimeModule.js
-        'self',
+        'self', // webpacks own code references it
+        'importScripts', // chunk loading in workers
```
naugtur, line 92: the actual change - now if any part of the webpack runtime references `importScripts` lexical lookup will take it from a local variable inside the main closure

boneskull (approved):
Needs a test of some sort, but otherwise LGTM.

packages/webpack/src/runtime/runtimeBuilder.js
```diff
@@ -319,7 +320,19 @@ const LOCKDOWN_SHIMS = [];`
...
+            // before any module that is not trying equally hard to be first.
+            // While the defensive preamble should not be required before
+            // scuttling occurs, it defines variables that will be hoisted and
+            // if it's not first, it'll overshadow the globals with undefined.
+            // In practice it was never problematic with webpack internal plugins,
+            // but if it happened to interfere with a different plugin or an
+            // unlucky key was added to the defensive preamble list, it'd
+            // be really hard to debug. So we use the shady -1 trick to avoid
+            // possible but unlikely issues.
+            stage: RuntimeModule.STAGE_NORMAL - 1,
```
naugtur, line 335: Maybe a snippet from the resulting bundle will help explain this

```
      /******/ (() => { // webpackBootstrap
...
      /************************************************************************/
      /******/ 	/* webpack/runtime/LavaMoat/defensive */
      /******/ 	;var location = globalThis.location,setTimeout = globalThis.setTimeout,clearTimeout = globalThis.clearTimeout,document = globalThis.document,trustedTypes = globalThis.trustedTypes,self = globalThis.self,importScripts = globalThis.importScripts;
      /******/ 	const LOCKDOWN_SHIMS = [];;
      /******/ 	
      /******/ 	/* webpack/runtime/DefensivePreambleTest */
      /******/ 	(() => {
      /******/ 		;
      /******/ 		    globalThis['importScripts'] = 44;
      /******/ 		    if(typeof importScripts !== 'number' || importScripts !== 42) {
      /******/ 		      throw new Error('"importScripts" is not the expected value: ' + importScripts)
      /******/ 		    }
      /******/ 		  ;
      /******/ 	})();
      /******/ 	
      /******/ 	/* webpack/runtime/LavaMoat/runtime */
      /******/ 	(() => {
      /******/ 		;const LAVAMOAT = Object.create(null);
...
```

Runtime modules are concatenated into a closure that webpack creates for its bootstrap. Defensive preamble makes copies of certain important globals inside that closure.
In most cases the globals we protect from scuttling with defensive preamble are not synchronously used and even if they are, the vast majority of plugins don't go into the first stage of the payload. 
That being said, anything that attempts to reach for a name that we cover with the preamble will get an undefined from hoisting if preamble isn't above it.

PR #2081, opened 2026-08-04
fix(core): prevent the endowment logic from skipping writable fields

naugtur:
The logic would cause writable fields to not be included in iteration and not made accessible to the writer in some cases.

### DIFF
--- a/packages/core/src/endowmentsToolkit.js
+++ b/packages/core/src/endowmentsToolkit.js
@@ -168,7 +168,14 @@ function endowmentsToolkit({
             unwrapFrom || targetRef
           )
         } else {
-          instrumentDynamicValueAtPath(pathParts, sourceRef, targetRef)
+          instrumentDynamicValueAtPath(
+            pathParts,
+            sourceRef,
+            targetRef,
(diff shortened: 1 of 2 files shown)


packages/core/src/endowmentsToolkit.js
```diff
@@ -103,26 +103,21 @@ function endowmentsToolkit({
           explicitlyBanned.push(path)
           return
         }
-        // write access handled elsewhere
+        whitelistedReads.push(path)
```
naugtur, line 106: if policy was not set to false, it must be listed in paths because that's what's being iterated.

packages/core/src/endowmentsToolkit.js
naugtur, line 161: `allowedWriteFields` check is not reachable if the item is not in `paths`

boneskull (approved):

naugtur:
@boneskull I'll need a new review because new changes.

Meanwhile Opus tried to introduce 2 distinct vulnerabilities and I made a few iterations of getting the logic right after that. Soo... this is not an at-a-glance thing

PR #2088, opened 2026-08-19
A collection of fixes from real world testing of @lavamoat/harden

naugtur:
See individual commits. 
Intend not to squash.


naugtur:
tests failing, looks like github ships different actions machine config after recent downtime. seems to be missing corepack?

packages/harden/src/template/lavamoat/scripts.debug.json
```diff
@@ -0,0 +1,7 @@
+{
+  "notes": "Prints all permission checks without enforcing anything. Useful for discovering what is being blocked in programs that swallow errors.",
+  "nodeOptions": {
+    "--permission-audit": true,
+    "--import": "\"data:text/javascript,import%20dc%20from%20'node%3Adiagnostics_channel'%3B%5B'fs'%2C'net'%2C'child'%2C'worker'%2C'inspector'%2C'wasi'%2C'addon'%2C'ffi'%5D.forEach(s%3D%3Edc.subscribe('node%3Apermission-model%3A'%2Bs%2C(msg)%3D%3Econsole.error('%5BLM%20DEBUG%5D'%2Cmsg.permission%2B'%20on%20'%2Bmsg.resource)))%3B\""
```
naugtur: `--permissions-audit` causes diagnostics to be reported instead of policies being enforced.
the urlencoded blob is a trick to load code into node.js through `NODE_OPTIONS` because `-e` is not allowed there :D  
It logs content from diagnostic channels for permissions

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -141,22 +167,38 @@ function makeRunScriptWrapper(
...
+    }
 
     customizePermissionsConfig(configOptions, env)
 
     permissionsModelCompatibilityExtensions(configOptions, env)
 
     const confOption = makeFlagsFromConfig(configOptions)
 
-    return `${existingOptions || ''} ${confOption.trim()}`.trim()
+    return confOption.trim()
```
naugtur, line 201: load-bearing deletion ;) 

If a script calls another script through something like `"lint": "npm run lint:types"` we don't want to accumulate NODE_OPTIONS

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -23,6 +23,27 @@ function makeRunScriptWrapper(
...
+  /**
+   * @param {Record<string, string>} configs
+   * @param {string} scriptName
+   * @returns {string | undefined}
+   */
+  function prefixMatch(configs, scriptName) {
+    const keys = Object.keys(configs)
+    const matchingKeys = keys
+      .filter((k) => k.endsWith('*'))
+      .filter((k) => scriptName.startsWith(k.slice(0, -1)))
```
boneskull, line 35: consider normalizing `keys` (somewhere; not necessarily here) or otherwise rejecting invalid patterns

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -23,6 +23,27 @@ function makeRunScriptWrapper(
...
+
+    if (matchingKeys.length === 0) {
+      return undefined
+    }
+    // longest prefix match wins
+    const bestMatch = matchingKeys.reduce((a, b) =>
+      a.length > b.length ? a : b
+    )
```
boneskull, line 43: could be

```suggestion
    const bestMatch = matchingKeys.sort((a, b) => b.length - a.length)[0]
```

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -38,11 +59,13 @@ function makeRunScriptWrapper(
       return {}
     }
     const configName =
-      scriptsConfig[scriptName] || scriptsConfig[DEFAULT_PERMISSION_KEY]
+      scriptsConfig[scriptName] ||
+      prefixMatch(scriptsConfig, scriptName) ||
+      scriptsConfig[DEFAULT_PERMISSION_KEY]
```
boneskull, line 64: consider implication of using `??` instead of `||` here

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -23,6 +23,27 @@ function makeRunScriptWrapper(
     return JSON.parse(readFileSync(filePath, 'utf8'))
   }
 
+  /**
+   * @param {Record<string, string>} configs
+   * @param {string} scriptName
+   * @returns {string | undefined}
+   */
+  function prefixMatch(configs, scriptName) {
```
boneskull, line 31: should `configs` be plural? makes more sense as singular to me

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -141,22 +167,38 @@ function makeRunScriptWrapper(
...
    * Checks the config obtained from package.json and puts it in as NODE_OPTIONS
    *
-   * @param {string | undefined} existingOptions
    * @param {ConfigOptions} configOptions
    * @param {NodeJS.ProcessEnv} env
    */
-  function installNodeOptions(existingOptions, configOptions, env) {
+  function installNodeOptions(configOptions, env) {
+    const existingOptions = env.NODE_OPTIONS
+    const inLavaMoatEnvAlready = !!env.LAVAMOAT_RUN_CONF
```
boneskull, line 175: a `LAVAMOAT_RUN_CONF=0` would cause this to be `true`, in case that matters.

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -141,22 +167,38 @@ function makeRunScriptWrapper(
...
+    // If this becomes an issue, we could use a separate env var to distinguish between options we set and options that were there already.
+
+    // If we're not nested, but there are existing options, it's safe to warn the user without being annoying.
+    if (
+      !inLavaMoatEnvAlready &&
+      existingOptions &&
+      existingOptions.length > 0
+    ) {
+      console.error(
+        `[LavaMoat] Warning: Replacing existing NODE_OPTIONS. Move them to the relevant JSON file in lavamoat/ if they were intentional.`
```
boneskull, line 191:
```suggestion
        `[LavaMoat] Warning: Replacing user-defined NODE_OPTIONS. Move them to the relevant JSON file in lavamoat/ if they were intentional.`
```

I think this makes it more clear whose responsibility it was for setting `NODE_OPTIONS` (assuming my suggestion is actually true)

packages/harden/src/runner/run-script-wrapper.cjs
boneskull: Wait, I don't see what we're doing with `confOption`?

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -234,15 +276,23 @@ function makeRunScriptWrapper(
         filteredFragments.push(fragment)
       }
     }
-    // Why would there be multiple bin fragments? In a npm workspace, local bin and workspace root bin is added
-    filteredFragments.push(...nodeModulesBinFragments)
+    // This should preserve order of items.
```
boneskull, line 279: how so?  I didn't know `Set` made any guarantees about order, unless that's unspecified engine convention

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -234,15 +276,23 @@ function makeRunScriptWrapper(
...
+    const uniqueNodeModulesBinFragments = Array.from(
+      new Set(nodeModulesBinFragments)
+    )
+
+    // Why would there even be multiple bin fragments? ecause
```
boneskull, line 284:
```suggestion
    // Why would there even be multiple bin fragments? because
```

boneskull: Also: there's no definition of "bin fragment" anywhere here that I can see, so I'm not sure what a "bin fragment" is.

and the comment below, you mean it adds `node_modules/.bin` for each parent to the filesystem root, or...?

packages/harden/src/template/lavamoat/scripts.debug.json
boneskull: barbaric

packages/harden/src/yarn/runner-plugin.js
```diff
@@ -44,15 +44,20 @@ module.exports = {
           }
 
           const pkgJson = workspace.manifest.raw
-          const binFolder =
-            extra.env.BERRY_BIN_FOLDER || `node_modules${path.sep}.bin`
+          // Abundance of caution. It seems yarn is not adding
+          // any node_modules/.bin locations to the PATH by default.
+          const binFolder = `node_modules${path.sep}.bin`
```
boneskull, line 49: seems like `path.join` would be more idiomatic here

boneskull: also: "by default" in what context?  Certainly when I execute `yarn lint` it runs the `lint` script which invokes `eslint` successfully.

packages/harden/test/runner.spec.js
```diff
@@ -1,36 +1,29 @@
...
 
-/**
- * Copies a fixture project to a temp dir and returns the temp path.
- *
- * @param {string} name - Project folder name
- */
-async function copyProject(t, name) {
-  if (DEBUG) {
-    return join(PROJECTS_DIR, name)
+function cleanupPathAfterNpm(PATH) {
```
boneskull, line 35: I assume this scrubs data for a snapshot? Should it live with the other utils?

packages/harden/test/runner.spec.js
```diff
@@ -62,3 +60,25 @@ for (const pm of PKGMGR_LIST) {
     )
   })
 }
+test(`runner wildcard support in scriptsConfig exits zero`, async (t) => {
+  t.plan(1)
+  const cwd = await copyProject(t, 'runner-features')
```
boneskull, line 83: this _looks like_ we want to clean up something after test execution?

packages/harden/test/runner.spec.js
```diff
@@ -62,3 +60,25 @@ for (const pm of PKGMGR_LIST) {
...
+    print: () => {},
+  })
+
+  const result = await execFileAsync('npm', ['test'], {
+    cwd,
+    env: { ...process.env },
+  })
+  t.log(result.stdout)
+
+  t.pass()
```
boneskull, line 101: I'm not sure this works, but I'd prefer we made _some_ sort of actual assertion in this test

```suggestion
  const result = await t.notThrowsAsync(execFileAsync('npm', ['test'], {
    cwd,
    env: { ...process.env },
  }));
  t.log(result.stdout)
```

packages/harden/test/utils.js
boneskull, line 1: I note that a) scrubbing data for E2E snapshot tests seems like a thing that we must do often, and b) it sucks and always devolves into a swamp of regex.

Even using a flag for machine-readable output (e.g., the `--json` you've proposed) won't help if the output contains shit like absolute paths.

I'm not proposing any solutions, but merely noting the problem is a bit thorny.

packages/harden/README.md
```diff
@@ -160,7 +160,26 @@ One of the more advanced capabilities `@lavamoat/harden` brings to the project i
...
+- Exact script name match has highest priority.
+- Prefix wildcard keys ending in `*` are supported (for example, `lint:*` matches `lint:eslint` and `lint:types`).
+- If multiple wildcard prefixes match, the longest prefix wins.
+- If there is no exact or wildcard match, `#default` is used when present.
+
+Example:
+
+```json
+{
+  "scriptsConfig": {
```
boneskull, line 177: I look at this and wonder why I need to reference other JSON files in my JSON file.

naugtur: because they're too big to be comfortably inlined and are expected to be heavily reusable.

packages/harden/src/runner/run-script-wrapper.cjs
naugtur: The comment will make it so :D

It is in fact the observed result and this code will only ever run on V8 so I was reluctant to explode it into array-iteration based implementation. MAybe I should? Or maybe I should go to TC39 with Array.prototype.uniq() poposal?

packages/harden/src/runner/run-script-wrapper.cjs
naugtur: it ends up in NODE_OPTIONS

packages/harden/src/runner/run-script-wrapper.cjs
naugtur: helps me remember that it's a collection

packages/harden/src/runner/run-script-wrapper.cjs
naugtur: maybe I could validate them in `harden check` or something like that, but I want to keep the runner minimal so that it can be permanently frozen soon and hopefully never change again. There's no good mechanism to deliver updates to it as the runner implementation has to exist in the repo and work potentially before dependencies are installed.

packages/harden/src/runner/run-script-wrapper.cjs
naugtur: I don't see the need to replace O(N) with O(Nlog(N))
The performance doesn't matter here tho, so is this change relieving mental load when reading? I'd say that's what the comment is intended to relieve

packages/harden/src/runner/run-script-wrapper.cjs
naugtur: A falsy value will cause a fallback to kick in. `false` is a ridiculous case, but for an empty string I prefer to fall back on defaults than returning empty. People would use that to implicitly allow all if it worked.

packages/harden/src/runner/run-script-wrapper.cjs
naugtur: LAVAMOAT_RUN_CONF is set to the string representing the filename with config, so if it's set to `0` and a file named `0` exists, it's correctly a truthy value. if the file `0` doesn't exist, this line is unreachable.

packages/harden/src/runner/run-script-wrapper.cjs
naugtur: we can't know if they're user-defined or defined by a non-node tool that subsequently calls node. 
might also be an attack attempt where someone polluted the env with some NODE_OPTIONS before the scripts were invoked.

packages/harden/src/runner/run-script-wrapper.cjs
naugtur: the word fragment here is unfortunate. these are paths from $PATH, hard to name them something not-silly. they're `PATHPaths` :D 

I'll think about it...

packages/harden/src/template/lavamoat/scripts.debug.json
naugtur: In the future we might ship some modules that can be used in config via `"--require": "@lavamoat/harden/something` and then the debug could go in there. 

I'd be willing to remove this in favor of shipping and documenting `@lavamoat/harden/permission-debug` 

BTW, `--permission` without `-audit` also causes these diagnostics to be emitted. Should we print them out by default? If permissions are on and banning something, I don't see a scenario where the user would not want to know and error propagation is unreliable. On the other hand, they might rely on clean stderr from the script that deliberately swallows the error. Unlikely but possible.

packages/harden/src/yarn/runner-plugin.js
naugtur: `BERRY_BIN_FOLDER` is a tmpdir location that yarn sets up with bins for each script run. A clanker was claiming local node_modules/.bin could also be added to PATH, but I didn't see that happening in my local testing. 

agree about path.join - didn't come up in my head as a way to use the correct path separator in the moment 😅

packages/harden/test/runner.spec.js
naugtur: it's a tmpdir location, do we need to?

It's been super useful to have the folders remain after tests to be able to inspect them manually and try things (especially when a test failed)

packages/harden/test/runner.spec.js
naugtur: nice

packages/harden/test/runner.spec.js
naugtur: nope, it's even more hairy than that. It cleans up PATH from what the fact of this being a process spawned by `npm test` causes to be there, so we don't get unrelated mess when we invoke other package managers that make less mess in PATH.

packages/harden/package.json
```diff
@@ -36,6 +36,7 @@
   "scripts": {
     "lint:deps": "depcheck",
     "test": "ava",
+    "test:prep": "corepack enable",
```
naugtur:
```suggestion
    "test:prepare": "corepack enable",
```

naugtur:
lots to squash there, but CI is finally green. 
It was a mix of corepack setup being actually fragile and my PATH cleanup being too aggressive.

Mrtenz (approved):

PR #2089, opened 2026-08-19
feat(harden): verify lavamoat files integrity in case write to project root was allowed

naugtur:
- not sure if necessary or overkill
- should it be recursive? risks processing tens of KB of policy.json files ;)

### DIFF
--- a/packages/harden/src/runner/make-run-script-wrapper.global.d.ts
+++ b/packages/harden/src/runner/make-run-script-wrapper.global.d.ts
@@ -1,4 +1,16 @@
 declare global {
+  type LMFolderIntegrityCheckOptions = {
+    projectRoot: string
+    pathJoin: (...segments: string[]) => string
+    createHash: typeof import('node:crypto').createHash
+    readdirSync: typeof import('node:fs').readdirSync
+    readFileSync: typeof import('node:fs').readFileSync
+  }
+
+  type LMFolderIntegrityCheck = {
+    verifyAfter: () => void
+  }
+
   type MakeRunScriptWrapperConfigOptions = Record<
     string,
     boolean | string | string[]
@@ -36,6 +48,10 @@ declare global {
     options: MakeRunScriptWrapperOptions,
     io: MakeRunScriptWrapperIO
   ): MakeRunScriptWrapper
+
+  function LMFolderIntegrityCheck(
+    options: LMFolderIntegrityCheckOptions
+  ): LMFolderIntegrityCheck
 }
 
 export {}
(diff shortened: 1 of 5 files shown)


packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -1,10 +1,54 @@
...
 
+/**
+ * @param {object} opts
+ * @param {string} opts.projectRoot
+ * @param {typeof import('node:path').join} opts.pathJoin
+ * @param {typeof import('node:crypto').createHash} opts.createHash
+ * @param {typeof import('node:fs').readdirSync} opts.readdirSync
+ * @param {typeof import('node:fs').readFileSync} opts.readFileSync
+ */
+// eslint-disable-next-line no-unused-vars
```
boneskull, line 11: would a `/* global */` be better?

packages/harden/src/runner/make-run-script-wrapper.global.d.ts
```diff
@@ -1,4 +1,16 @@
 declare global {
+  type LMFolderIntegrityCheckOptions = {
+    projectRoot: string
+    pathJoin: (...segments: string[]) => string
```
boneskull, line 4: why not `typeof import('node:path').join`?

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -1,10 +1,54 @@
...
+function LMFolderIntegrityCheck({
```
boneskull, line 12: wait... remind me why we need globals?

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -1,10 +1,54 @@
...
+  }
+
+  const hashZero = hashFilesInDir()
+
+  return {
+    verifyAfter: () => {
+      const hashNow = hashFilesInDir()
+      if (hashNow !== hashZero) {
+        throw new Error(
+          `[LavaMoat] lavamoat folder integrity check failed. It was modified during the script run.`
```
boneskull, line 39: We should probably note this behavior in `README.md` and suggest how to find the root cause

boneskull: also: should we be verifying the integrity of `node_modules/@lavamoat/harden` as well?  unsure of the threat model here.

packages/harden/src/yarn/runner-plugin.js
```diff
@@ -69,6 +71,14 @@ module.exports = {
             }
           )
 
+          const integrity = LMFolderIntegrityCheck({
+            projectRoot: extra.cwd,
+            pathJoin: path.join,
+            createHash,
+            readdirSync: fs.readdirSync,
```
boneskull, line 78: would it be safer to dereference `readdirSync`, `join`, etc.? I suppose I'm not clear on why we're passing capabilities this way. also unclear on the interaction between this an script-specific node.js permissions

PR #2090, opened 2026-08-20
fix(harden): use the project root for `projectRoot` in the Yarn runner plugin

Mrtenz:
The Yarn 4 runner plugin passed `extra.cwd` as `projectRoot` to `makeRunScriptWrapper`. `extra.cwd` is the directory the script is executed in, which for a workspace script is the workspace directory rather than the root of the project.

This switches it to `project.cwd`, so `projectRoot` points at the actual project root regardless of which workspace the script belongs to.

### DIFF
--- a/packages/harden/src/yarn/runner-plugin.js
+++ b/packages/harden/src/yarn/runner-plugin.js
@@ -51,7 +51,7 @@ module.exports = {
             {
               scriptName,
               scriptPayload: extra.script,
-              projectRoot: extra.cwd,
+              projectRoot: project.cwd,
               pathBinMatcher: (fragment) => {
                 return fragment.endsWith(binFolder)
               },


naugtur:
I think we should look at both. Not sure about priority.

Gonna upgrade this to "support workspaces" for all pkgmgrs

If we look at this file relative to workspace root, we'd also need to be consistent with the configs.

Or we might document that lavamoat folder should live in workspace root and support only that.

One drawback in that case - paths to json config are no longer actual relative paths while looking nice.


Food for thought

naugtur:
Plan to address it more widely.
note: I think the only reasonable way to set it up is to have a shared `lavamoat` folder for the whole workspace (because configs are almost certainly reusable across all projects in a workspace) and individual scriptConfig fields in package.json locally. I don't think there should be a lavamoat folder per workspace folder separately for the sake of this tool. This introduces a design tension, because when using `@lavamoat/node` or `@lavamoat/webpack` whey will work with local lavamoat folders in workspaces and that on the other hand is a good choice. 

This PR addresses the workspace problem for yarn with no design tension whatsoever. 

npm and pnpm are resolving the script shell field from CWD of the `(p)npm run` call, so they fail to do it when someone calls it from a workspace directory.   
I see two ways we could address it:
1. install lavamoat runner as a dependency with a bin and use it by name (I have a PoC implementation where it installs from the local folder with a randomized name so it's impossible for a non-targeted package to declare the same bin and potentially override. But it could also be a regular npm package we put in dependencies and it'd solve the version management problem for free)
 - pros: no need to process workspaces, just works from any folder
 - cons: weird setup, depends on dependencies being installed to run any script

2. put symlinks (or .npmrc override) in every workspace folder
 - pros: no need to npm install
 - cons: messy!

naugtur:
closing in favor of https://github.com/LavaMoat/LavaMoat/pull/2091

(pulled your commit there)

naugtur:
flipping that back around now - gonna merge this instead.

naugtur (approved):

PR #2091, opened 2026-08-27
feat(harden): workspaces support

naugtur:
- trivial in yarn
- only possible through symlinking in (p)npm

Also tried:
 creating a package with a `bin` and using its name  
Doesn't work because it creates a chicken-and-egg problem: scripts-shell is used for running install scripts, so if the package is not installed, install will fail with ENOENT if any scripts are allowed.  
The abandoned implementation exists in naugtur/harden-workspaces-support_failed

todo:
- [ ] check what happens if symlink exists
- [ ] check if the `script-shell` and linking trick works on windows :see_no_evil:


naugtur:
I stacked this PR using github's own functionality to see how it works.

naugtur:
@SocketSecurity ignore npm/@pnpm/network.ca-file@1.0.2 not obfuscated

packages/harden/src/npm/opinions.js
```diff
@@ -1,6 +1,10 @@
 /** @import {Opinion} from '../tools/types.js' */
 import { buildAllowlistChanges } from './build-allowlist.js'
 import { bundleRunner } from '../runner/runner-bundler.js'
+// @ts-expect-error ships without types
```
boneskull, line 4: (should)

Pull in `@types/npmcli__map-workspaces` as a dev dep

packages/harden/src/npm/opinions.js
```diff
@@ -224,6 +228,25 @@ const definedOpinions = Object.freeze(
...
+          })
+          for (const workspace of workspaces.values()) {
+            // compare workspace string to facts.cwd and create relative path
+            const projectPathRelativeToWorkspace = relative(
+              workspace,
+              facts.cwd
+            )
+            symlinkSync(
+              join(projectPathRelativeToWorkspace, 'lavamoat'),
+              join(workspace, 'lavamoat'),
```
boneskull, line 244: (could)

```js
const DEFAULT_DIR = 'lavamoat'
```

packages/harden/src/pnpm/opinions.js
```diff
@@ -175,6 +179,30 @@ const definedOpinions = Object.freeze(
...
+          const projectPathRelativeToWorkspace = relative(
+            workspacePath,
+            facts.cwd
+          )
+
+          // pnpm includes the root folder in its results.
+          // If the relative path is empty, we are at the root. Skip it.
+          if (projectPathRelativeToWorkspace === '') {
+            continue
+          }
```
boneskull, line 197: (could)

refactor by filtering the array instead of issuing a `continue`

packages/harden/src/runner/run-script-wrapper.cjs
boneskull, line 43: I feel like I complained about this somewhere else

boneskull, line 48: (should)

give `opts` an actual type

boneskull, line 68: Why not:

```suggestion
      return { config: {} }
```

boneskull, line 171: (could)

```js
   * @returns {string}
```

boneskull, line 184: (could)

wrap long comment lines

boneskull, line 188:
```suggestion
      existingOptions?.length
```

boneskull, line 191: (could)

1. Seems like we should know which file exactly needs updating
2. Might want to print the `NODE_OPTIONS` that we're replacing?
3. This might become tedious and the user may want to suppress the warning; e.g., the user's shell rc sets `NODE_OPTIONS` but they don't actually care it's getting blown away
4. Alternatively, consider throwing if `NODE_OPTIONS` is present

boneskull, line 201: (could)

`makeFlagsFromConfig` should probably do the trimming, eh?

boneskull, line 282: (could)

```suggestion
    const uniqueNodeModulesBinFragments = [...new Set(nodeModulesBinFragments)]
```

boneskull, line 310: (could)

rename to `LAVAMOAT_HARDEN_CONF` or something that distinguishes `@lavamoat/harden` specifically

packages/harden/src/tools/verifier.js
boneskull, line 29: (should)

describe options, return value

boneskull, line 30: (could)

refactor signature so the options are optional:

```suggestion
export function createVerifier(level, print, { json = false, packageManager } = {}) {
```

boneskull, line 74: (could)

rewrite as a `for..of` loop which may be easier to understand.  though, I probably wouldn't.

boneskull, line 75:
```suggestion
          [...scores].map(([name, [applied, total]]) => [
```

boneskull, line 103:
```suggestion
${[...scores]
```

boneskull, line 106: Would `Number.prototype.toFixed` be more appropriate here?

packages/harden/src/yarn/runner-plugin.js
boneskull, line 49: I totally _did_ complain about this.

packages/harden/test/index.spec.js
boneskull, line 77: (should)

string constants for these levels

boneskull, line 92:
```suggestion
    t.deepEqual(result, [], 'Expected no changes to be reported with verifier')

    const { exitCode } = await decisions.showSummary(summary)
```

boneskull, line 100: Is this necessary?  I worry about flake

packages/harden/test/runner.spec.js
```diff
@@ -1,49 +1,42 @@
...
 import { copyProject } from './utils.js'
+import { delimiter, join } from 'node:path'
 import { hardenDefaults } from '../src/index.js'
 import { createFallbackDecisions } from '../src/tools/default-decisions.js'
 
 const execFileAsync = promisify(execFile)
 
 const PKGMGR_LIST = ['npm', 'pnpm', 'yarn']
 
+function cleanupPathAfterNpm(PATH) {
```
boneskull, line 13: (could)

rename `PATH` to `filepath` or something. non-screaming variable name

packages/harden/test/runner.spec.js
```diff
@@ -1,49 +1,42 @@
...
+  const filteredFragments = []
+  for (const fragment of pathFragments) {
+    if (!fragment.includes('node_modules')) {
+      filteredFragments.push(fragment)
+    }
+  }
+  return filteredFragments.join(delimiter)
+}
+
 const runnerSetupPerPm = {
```
boneskull, line 24: I feel like this needs a comment explaining wtf

packages/harden/test/runner.spec.js
```diff
@@ -77,7 +74,55 @@ for (const pm of PKGMGR_LIST) {
       'Expected no secret leakage, but SECRET is present'
     )
   })
+
+  test(`runner loads default config in root and nested workspace package for ${pm}`, async (t) => {
+    const cwd = await copyProject(t, `runner-workspace-${pm}`)
```
boneskull, line 79: (should)

these tests don't seem to cleanup after themselves.

packages/harden/test/runner.spec.js
boneskull, line 135: (could)

make a noop fn the default value of the `print` option

boneskull, line 146:
```suggestion
  await t.notThrowsAsync(execFileAsync('npm', ['test'], {
    cwd,
    env: { ...process.env }, // does this really need a shallow copy?
  }), 'npm test exited with non-zero exit code')
  //t.log(result.stdout)
```

packages/harden/test/utils.js
boneskull, line 87: (should)

comment about what the length check is doing (bc I have no idea)

packages/harden/package.json
```diff
@@ -36,10 +36,12 @@
   "scripts": {
     "lint:deps": "depcheck",
     "test": "ava",
-    "test:prep": "corepack enable --install-directory=$(dirname $(which node))",
+    "test:prep": "corepack enable --install-directory \"$(dirname \"$(command -v node)\")\" npm pnpm yarn && corepack prepare pnpm@latest --activate && corepack prepare yarn@stable --activate",
```
boneskull, line 39: 😢 

I know it's lower priority rn, but if we can make this package portable then we'll have more success with it.

packages/harden/README.md
boneskull, line 166: should this be its own section?

boneskull, line 169:
```suggestion
- Keys ending in suffix `*` are supported (for example, `lint:*` matches `lint:eslint` and `lint:types`).
```

boneskull, line 170: (should)

I think we're using "prefix" here incorrectly

boneskull, line 171:
```suggestion
- If there is no exact or wildcard match, `#default` is used.
- If there is no `#default`, shit happens
```

boneskull (approved):
I don't think I have any blockers, but there are "should" changes I'd like to see.

boneskull:
My suggestions seem to have been hosed. Wonder if it's GH-stacked-PR-related??? ahem

boneskull:
I threw this in the PR queue but didn't assign it a priority, fwiw

packages/harden/src/runner/run-script-wrapper.cjs
naugtur: Don't recall that one. What's the complaint? 
As-is it allows intuitive behavior for
`lint:*` and `lint:eslint:*` to narrow down for multiple items.

naugtur: Why would I do that? It's never reused, it's an options bag for the function.

naugtur: Don't know. meat-proxy.

The important fix was adding `config:{}`

naugtur: typescript influencers keep telling me not to specify return types for functions and let ts infer them instead.

naugtur:
With 
- https://github.com/LavaMoat/LavaMoat/pull/2109
- https://github.com/LavaMoat/LavaMoat/pull/2090
- https://github.com/npm/cli/pull/9972
- https://github.com/pnpm/pnpm/issues/14422
this should not be necessary.

I should port some tests over

PR #2102, opened 2026-09-05
adopt changesets

boneskull:
This adopts [Changesets](https://changesets.dev) for better control over release notes.

> [!CAUTION]
> Changesets' tag names _differ_ from those created by Release Please.  For that reason, when `changests git-tag` runs, it will _not_ find the tags created by Release Please, so will tag all workspaces regardless. Changesets' format looks like `@scope/name@version` while Release Please uses `name-version`.
> After merging this PR, we'll need to manually create Git tags using Changeset's format based on the latest tag for each workspace.  For example, Changesets wants `@lavamoat/aa@5.0.1`, but there only exists a `aa-v5.0.1` as created by Release Please; we'll need to create a new tag on `aa-v5.0.1` named `@lavamoat/aa@5.0.1`.  Tedious, but we only need to do so once.

* * *

- Some random CI lint fixes from zizmor
- Updated `RELEASE.md`


boneskull:

- branch: `main`
  - **[adopt changesets #2102](https://github.com/LavaMoat/LavaMoat/pull/2102)** - branch: `boneskull/changesets` *(this PR)*

---

naugtur:
We need to only copy the latest tag in each repo to new format, right?

.github/actions/prepare/action.yml
```diff
@@ -10,6 +10,10 @@ inputs:
     description: 'Shell to use'
     required: false
     default: 'bash'
+  install-dependencies:
+    description: 'Install dependencies'
+    required: false
+    default: true
```
naugtur, line 16: This sounds like it runs npm install.
How much do we know about how?

.github/workflows/release-please.yml
```diff
@@ -17,70 +24,106 @@ permissions: {}
...
-      group: release-please-${{ github.ref }}
+      group: release-${{ github.ref }}
       cancel-in-progress: false
     runs-on: ubuntu-latest
     outputs:
-      # boolean indicating whether any releases were created
-      releases_created: ${{ steps.release.outputs.releases_created || steps.manual.outputs.releases_created }}
-      # JSON array of workspace paths that were released
-      paths_released: ${{ steps.release.outputs.paths_released || steps.manual.outputs.paths_released }}
+      # boolean string indicating whether any releases were created
```
naugtur, line 40: What's a boolean string? 😅
Is it `"true"`?

.github/workflows/release-please.yml
```diff
@@ -17,70 +24,106 @@ permissions: {}
...
     permissions:
-      contents: write
-      pull-requests: write
+      contents: write # changesets/action pushes the version branch and release tags
+      pull-requests: write # changesets/action creates/updates the version PR
     steps:
-      - uses: googleapis/release-please-action@5c625bfb5d1ff62eadeeb3772007f7f66fdcf071 # v4.4.1
-        id: release
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         if: github.event_name != 'workflow_dispatch'
```
naugtur, line 49: Why is manual dispatch different?
Are these conditionals here to handle the first run?

package.json
```diff
@@ -11,11 +11,14 @@
   ],
   "engines": {
     "node": "^22.22.2 || ^24.15.0",
-    "npm": ">=7.0.0"
+    "npm": ">=10.9.0"
```
naugtur, line 14: I'll bump it to 12 soon anyway

naugtur:
Will renovate know how to do changesets?

.github/actions/prepare/action.yml
boneskull: uhhh... what?

.github/workflows/release-please.yml
boneskull: I mean, it's strings all the way down, but yeah it's either `true` or `false`

.github/workflows/release-please.yml
boneskull: Manual dispatch is meant to be used when publishing fails for whatever reason. It assumes the "new version" PR has been merged, and we just need to invoke pack & publish.  So, currently, manual dispatch skips Release Please.  In this PR, manual dispatch skips Changesets.

package.json
boneskull: - [x] update `engines.npm`

boneskull:
@naugtur 

> Will renovate know how to do changesets?

What would it need to know?  It doesn't know anything about Release Please other than there's a GitHub Action it needs to update. Same with Changesets.

naugtur:
@boneskull 
> What would it need to know? It doesn't know anything about Release Please other than there's a GitHub Action it needs to update. Same with Changesets.

Doesn't it need to generate a changesets file? Are they falling back to conventional commits if absent or will we have to write them if we want renovate PRs to go show up in the changelog?

.github/actions/prepare/action.yml
naugtur: brainfart. I missed the fact that it's in inputs and tried to figure out what it meant

PR #2103, opened 2026-09-07
fix(harden): avoid test flakiness by relying less on one run per process

naugtur:
Running `hardenDefaults` many times in one process had some problems for tests:
- yaml cache is permanent, per folder
- options object was getting `detected` updates from multiple sources

fixed options issue by making a shallow copy

keeping the cache and addressing the issue in test setup

### DIFF
--- a/packages/harden/src/index.js
+++ b/packages/harden/src/index.js
@@ -9,8 +9,9 @@ import { opinions as pnpmOpinions } from './pnpm/opinions.js'
  * @import {
  *   HardenDefaultsOptions,
  *   HardenResult,
+ *   Opinion,
  *   PrintApi
- * } from "./tools/types.js"
+ * } from './tools/types.js'
  */
 
 /**
@@ -56,13 +57,13 @@ export async function hardenDefaults(options) {
   let opinions
   switch (pmName) {
     case 'npm':
-      opinions = npmOpinions
+      opinions = shallowCopy(npmOpinions)
       break
     case 'yarn':
-      opinions = yarnOpinions
+      opinions = shallowCopy(yarnOpinions)
       break
     case 'pnpm':
-      opinions = pnpmOpinions
+      opinions = shallowCopy(pnpmOpinions)
       break
     default:
       throw new Error(`Unsupported package manager: ${pmName}`)
@@ -122,4 +123,12 @@ export async function hardenDefaults(options) {
           .join('\n')
 
   return { result, summary }
+}
+
+/**
+ * @param {readonly Opinion[]} from
+ * @returns Opinion[]
+ */
+function shallowCopy(from) {
+  return from.map((item) => ({ ...item }))
 }
--- a/packages/harden/src/index.js
+++ b/packages/harden/src/index.js
@@ -127,7 +127,7 @@ export async function hardenDefaults(options) {
 
 /**
  * @param {readonly Opinion[]} from
- * @returns Opinion[]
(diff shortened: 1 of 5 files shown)


packages/harden/src/index.js
```diff
@@ -123,3 +124,11 @@ export async function hardenDefaults(options) {
 
   return { result, summary }
 }
+
+/**
+ * @param {readonly Opinion[]} from
+ * @returns Opinion[]
```
boneskull: (must)

```suggestion
 * @returns {Opinion[]}
```

packages/harden/test/with-scripts.spec.js
```diff
@@ -69,14 +69,18 @@ for (const pm of PKGMGR_LIST) {
 
     t.assert(result.length > 0, 'Expected some changes for strict level')
 
+    // cache busting - calling hardenDefaults multilpe times reuses dryRun cache otherwise
```
boneskull:
```suggestion
    // cache busting - calling hardenDefaults multiple times reuses dryRun cache otherwise
```

That being said, would it not be more robust as something like a timestamp? It looks like it's just appending the `-m` suffix?

packages/harden/test/with-scripts.spec.js
```diff
@@ -69,14 +69,18 @@ for (const pm of PKGMGR_LIST) {
...
+    const cwd2 = cwd + '-m'
+    await execFileAsync('mv', [cwd, cwd2])
```
boneskull: (should)

did you write this just before bed? 😆 

```suggestion
    await fsPromises.rename(cwd, cwd2)
```

packages/harden/src/tools/apply-change.js
```diff
@@ -102,9 +101,11 @@ export async function verifyOpinion(cwd, opinion, facts) {
     result.push(...(await applyChange({ cwd, change, dryRun: true })))
   }
 
+  let score = 1 - result.length / changes.length
```
boneskull, line 105: (must, or explain what I'm missing)
 
1. Per L95 (and presumably the type of `opinion.changes`), `changes` is _always_ an array
2. Thus, the `structuredClone` call will _never_ return a falsy value
3. Therefore the conditional on L96 will _never_ be truthy
4. Unless there's a side-effect here, `changes` can be an empty array
5. `changes.length` can therefore be `0` which could cause a divide-by-zero error

boneskull (approved):
This is fine to merge after requested fixes

packages/harden/src/tools/apply-change.js
naugtur: the conditional on L96 should have been about changes.length. not sure what happened there.

It didn't make sense from the start. When changes are empty, the verify function is the fallback. 

It worked because the NaN was being ignored and replaced with score from verify. 

I made it explicit - somewhat unnecessary, but we're not ignoring a potential div by zero.

packages/harden/test/with-scripts.spec.js
naugtur: It does it once per test and the test is already in a folder with randomly generated name.

packages/harden/test/with-scripts.spec.js
naugtur: It was after hours of very annoying search for the bug that I reproduced by running `watch -e` on the test run with stuff instrumented to print various states. I was braindead, yes.

PR #2104, opened 2026-09-09
fix(harden): --allow-fs-tmp needs to explicitly set read and write

naugtur:
despite my best assumprions, write doesn't grant read

### DIFF
--- a/packages/harden/src/runner/run-script-wrapper.cjs
+++ b/packages/harden/src/runner/run-script-wrapper.cjs
@@ -125,25 +125,27 @@ function makeRunScriptWrapper(
     // 2. tmp write - crossplatform
     if (configOptions['--allow-fs-tmp'] === true) {
       delete configOptions['--allow-fs-tmp']
-      if (configOptions['--allow-fs-write']) {
-        if (typeof configOptions['--allow-fs-write'] === 'string') {
-          configOptions['--allow-fs-write'] = [
-            configOptions['--allow-fs-write'],
-          ]
-        }
-        if (configOptions['--allow-fs-write'] === true) {
-          return // none of this matters
-        }
-      } else {
-        // do this for both undefined and false
-        configOptions['--allow-fs-write'] = []
-      }
       const tmp = tmpdir()
-      configOptions['--allow-fs-write'].push(tmp)
-      // because macos is being weird
       const tmpRealPath = realpathSync(tmp)
-      if (tmpRealPath !== tmp) {
-        configOptions['--allow-fs-write'].push(tmpRealPath)
+      // write doesn't grant read, sadly
+      for (const perm of ['write', 'read']) {
+        const allowOption = `--allow-fs-${perm}`
+        if (configOptions[allowOption]) {
+          if (typeof configOptions[allowOption] === 'string') {
+            configOptions[allowOption] = [configOptions[allowOption]]
+          }
+          if (configOptions[allowOption] === true) {
+            continue // none of this matters
+          }
+        } else {
+          // do this for both undefined and false
+          configOptions[allowOption] = []
+        }
+        configOptions[allowOption].push(tmp)
+        // because macos is being weird
+        if (tmpRealPath !== tmp) {
+          configOptions[allowOption].push(tmpRealPath)
+        }
       }
     }
   }
--- a/packages/harden/test/projects/runner-features/lavamoat/scripts.permissions.json
+++ b/packages/harden/test/projects/runner-features/lavamoat/scripts.permissions.json
@@ -1,5 +1,5 @@
 {
   "nodeOptions": {
-    "--permissions": true
+    "--permission": true
   }
 }
--- a/packages/harden/test/projects/runner-features/lavamoat/scripts.tmp.json
+++ b/packages/harden/test/projects/runner-features/lavamoat/scripts.tmp.json
@@ -1,0 +1,6 @@
+{
+  "nodeOptions": {
+    "--permission": true,
+    "--allow-fs-tmp": true
+  }
+}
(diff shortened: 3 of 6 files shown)


naugtur:
macos test fails with
```
      Error: Access to this API has been restricted. Use --allow-fs-read to manage permissions.␊
          at Object.realpathSync (node:fs:2815:17)␊
          at Object.<anonymous> (/private/var/folders/d8/hvxvltxn0fl4rmnd52sncbth0000gn/T/harden-test-runner-features-7nJWoA/tmp-rw.js:5:24)␊
          at Module._compile (node:internal/modules/cjs/loader:1781:14)␊
          at Object..js (node:internal/modules/cjs/loader:1913:10)␊
          at Module.load (node:internal/modules/cjs/loader:1505:32)␊
          at Function._load (node:internal/modules/cjs/loader:1309:12)␊
          at wrapModuleLoad (node:internal/modules/cjs/loader:254:19)␊
          at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:171:5)␊
          at node:internal/main/run_main_module:36:49 {␊
        code: 'ERR_ACCESS_DENIED',␊
        permission: 'FileSystemRead',␊
        resource: '/var'␊
```

seems realPathSync on a mac is meddling in `/var` or doing some silly iterative listing of dirs for links.  
Gonna need help fiddling with it.  
It's dumb

boneskull:
@naugtur I can reproduce this.  

So it looks to me like on macOS anyway, you must have read/write access to `/var` because presumably something about the realpath lookup.

If you give Node.js `--allow-fs-read="/foo/bar" --allow-fs-read="/foo"`—at least on macOS in this specific instance—you will be able to read `/foo/bar` but not `/foo`.  If you _instead_ give it `--allow-fs-read="/foo" --allow-fs-read="/foo/bar"`, then it will work as expected.

Immediately after this loop: https://github.com/LavaMoat/LavaMoat/blob/41013fa154e08a411196348de10cedbd00f845ea/packages/harden/src/runner/run-script-wrapper.cjs#L131-L149

I added this:

```js
      if (tmpRealPath !== tmp) {
        configOptions['--allow-fs-read'].unshift('/var', '/private/var')
        configOptions['--allow-fs-write'].unshift('/var', '/private/var')
      }
```

which caused `npm run tmp` in the temp project to succeed.  I haven't thought about how to avoid hardcoding these paths and generalize the solution, but it may behoove us to special-case macOS here instead of depending on `tmpRealPath !== tmp`.

> TIL about `NODE_DEBUG_NATIVE=permission_model` which prints a lovely tree!

packages/harden/src/runner/make-run-script-wrapper.global.d.ts
```diff
@@ -26,6 +26,8 @@ declare global {
     pathDelimiter: string
     tmpdir: () => string
     realpathSync: (path: string) => string
+    lstatSync: (path: string) => import('fs').Stats
```
boneskull: (should)

no inline imports; use `import { type Stats } from 'fs'` at top of file.

I'm pretty sure this is enforced in `.ts` sources, but it may be that eslint is ignoring `.d.ts` files wholesale.

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -125,25 +133,40 @@ function makeRunScriptWrapper(
...
-            configOptions['--allow-fs-write'],
-          ]
-        }
-        if (configOptions['--allow-fs-write'] === true) {
-          return // none of this matters
+      const tmp = tmpdir()
+      let tmpRealPath = tmp
+      try {
+        tmpRealPath = realpathSync(tmp)
+      } catch (_) {
```
boneskull: (should)

```suggestion
      } catch {
```

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -125,25 +133,40 @@ function makeRunScriptWrapper(
...
+        // if realpathSync fails, it's been restricted by permissions
+        try {
+          const stats = lstatSync(tmp)
+          if (stats.isSymbolicLink()) {
+            tmpRealPath = readlinkSync(tmp)
+          }
+        } catch (_) {
```
boneskull: (should)

```suggestion
        } catch {
```

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -125,25 +133,40 @@ function makeRunScriptWrapper(
...
-      if (tmpRealPath !== tmp) {
-        configOptions['--allow-fs-write'].push(tmpRealPath)
+      // write doesn't grant read, sadly
+      for (const perm of ['write', 'read']) {
+        const allowOption = `--allow-fs-${perm}`
+        if (configOptions[allowOption]) {
+          if (typeof configOptions[allowOption] === 'string') {
+            configOptions[allowOption] = [configOptions[allowOption]]
+          }
+          if (configOptions[allowOption] === true) {
```
boneskull: (should)

if the first conditional is hit, the second conditional is impossible, so:


```suggestion
          } else if (configOptions[allowOption] === true) {
```

boneskull: (could)

handle unexpected type of `configOptions[allowOption]`

packages/harden/src/runner/run-script-wrapper.cjs
```diff
@@ -125,25 +133,40 @@ function makeRunScriptWrapper(
...
+            continue // none of this matters
+          }
+        } else {
+          // do this for both undefined and false
+          configOptions[allowOption] = []
+        }
+        configOptions[allowOption].push(tmp)
+        // because macos is being weird
+        if (tmpRealPath !== tmp) {
```
boneskull, line 173: I'm still undecided whether or not it's better to rely on this behavior or to just detect macOS.  We know how macOS behaves, but if this situation arises on another platform, is it safe to assume it's safe?  Is there any attack vector here?

boneskull (approved):
approved pending changes

packages/harden/src/runner/run-script-wrapper.cjs
naugtur: You can do anything with containers I suppose, so having a made up linux environment that uses a symlinked path for tmpdir is not impossible.

packages/harden/src/runner/run-script-wrapper.cjs
naugtur: rearranged them for early exit and added a guard against going forward with non-array

PR #2109, opened 2026-09-17
fix(harden): improve shell-script location resolution in npm

naugtur:
- turns out PWD is resolved early enough that most relative path problems disappear
- JS files cannot be shells in CMD, so wWindows users always had to use bash. It's better for security anyway I suppose
- the snapshots being removed were redundant before (from a test rename)


packages/harden/src/npm/opinions.js
```diff
@@ -201,7 +201,7 @@ const definedOpinions = Object.freeze(
     {
       id: 'n_runner',
       description:
-        'Take over npm run and remove bin scripts confusion possibility and configure other limitations.',
+        'Take over npm run and remove bin scripts confusion possibility and configure other limitations. [!No Windows cmd support]',
```
boneskull:
```suggestion
        'Take over "npm run" and mitigate bin-scripts confusion; configure other limitations. [!No Windows cmd support]',
```

packages/harden/README.md
```diff
@@ -157,6 +157,9 @@ Depending on your strictness choice, for yarn we can use the built-in `dependenc
 
 ### Script Execution Environment
 
+> Warning:
+> The script-shell override this depends on is not supported in Windows' cmd.exe, you need to be running bash or a compatible shell.
+
```
boneskull:
```suggestion
> [!WARNING]
> 
> The script-shell override `@lavamoat/harden` depends on is **unsupported in Windows' `cmd.exe`**.  Only `bash`, `zsh`, and like POSIX shells are supported.
```

boneskull (approved):
No blockers, but a couple suggestions for verbiage

